PR #39, opened 2018-05-04
fix print() and NY=int(nvertex/NX)

daisy20170101:


daisy20170101:
There is error about print "..." which can be fixed by replacing print ("...").
Also NY=nvertex/NX is float instead of integer, which can be fixed by NY=int(nvertex/NX)

Thomas-Ulrich:
Hi,
SeisSol has currently 2 main branches (master and easi, in my opinion easi is the one you should prefer). Actually, most of the changes you proposed have been already implemented in easi, I just forgot the cherry-pick the commit to the master branch also.  The scripts are now updated.
(but not the int, as it was not complaining on my python 3.4).
Maybe the best would be to move these scripts in an independent repository to avoid the double commit (one in each branch).

Thomas.

daisy20170101:
Hi, Thomas,

Thanks for the reply. 
It maybe easier for us to sit together and you can show me how to get start SeisSol. I was trying to double check all the workflow to make it more clearly for future users.
Do you have time today or tomorrow? We can do it either in your office or mine.

Thanks,

Best
Duo

——————————————————————
Dr. Duo Li
Department of Earth and Environmental Sciences
Geophysics
Ludwig-Maximilians-Universität München
Theresienstr. 41
80333 Munich 
Germany
liduoduo07@gmail.com

PR #60, opened 2019-04-15
Documentation has been added i.e. the installation process

ravil-mobile:
The installation process has been corrected and checked with both GNU and Intel compilers.

### DIFF
--- a/Documentation/compilation.rst
+++ b/Documentation/compilation.rst
@@ -16,24 +16,51 @@ In order to run SeisSol, you need to first install:
 Inital Adjustments to .bashrc
 -----------------------------
 
-Add the following lines to your .bashrc (vi ~/.bashrc).
+Create a new text file in your $HOME directory
 
 .. code-block:: bash
 
-  source /opt/intel/compiler/VERSION/bin/compilervars.sh intel64
+  cd $HOME
+  touch seissol_env.sh
+
+Open the file with your favorite text editor (vi ./seissol_env.sh) and add following lines:
+
+.. code-block:: bash
 
   # User specific aliases and functions
-  export PATH=$HOME/bin:$PATH
-  export LD_LIBRARY_PATH=$HOME/lib:$HOME/lib64:$LD_LIBRARY_PATH
+  export PATH=$S3_HOME/bin:$PATH
+  export LD_LIBRARY_PATH=$S3_HOME/lib:$S3_HOME/lib64:$LD_LIBRARY_PATH  # lookup path for executables
+  export LIBRARY_PATH=$LD_LIBRARY_PATH:$LIBRARY_PATH  # lookup path for compilers
   export EDITOR=vi
-  export SCONS_LIB_DIR=$HOME/lib64/scons-2.2.0/
-  export PATH=$HOME/../libxsmm/bin:$PATH
-  export C_INCLUDE_PATH=$HOME/include:$C_INCLUDE_PATH 
-  export CPLUS_INCLUDE_PATH=$HOME/include:$CPLUS_INCLUDE_PATH
+  export SCONS_LIB_DIR=$S3_HOME/lib64/scons-2.2.0/
+  export C_INCLUDE_PATH=$S3_HOME/include:$C_INCLUDE_PATH
+  export CPLUS_INCLUDE_PATH=$S3_HOME/include:$CPLUS_INCLUDE_PATH
+  export PKG_CONFIG_PATH=$S3_HOME/lib/pkgconfig
+
+  export FC=mpiifort  # or mpifort
+  export CXX=mpiicpc  # or mpicxx
+  export CC=mpiicc    # or mpicc
 
   ######  ParMetis library necessary (Optional) ##############
-  export PARMETIS_BASE='path_to_parmetis'
-  export PARMETIS_LIBDIR='path_to_parmetis/lib'
+  export PARMETIS_BASE=$S3_HOME
+  export PARMETIS_LIBDIR=$S3_HOME/lib
+
+
+Add the following to your .bashrc (vi ~/.bashrc).
+
+.. code-block:: bash
+
+  source /opt/intel/compiler/VERSION/bin/compilervars.sh intel64 # if you use intel compilers
+  source $HOME/seissol_env.sh
+
+
+.. code-block:: bash
+
+  source ./seissol_env.sh  # update your enviroment variable 
+
+  mkdir s3_software
+  cd ./s3_software
+
 
 Installing SCons
 ----------------
@@ -41,9 +68,9 @@ Installing SCons
 .. code-block:: bash
 
   wget http://prdownloads.sourceforge.net/scons/scons-2.2.0.tar.gz
-  tar -xaf scons-2.2.0.tar.gz
+  tar -xvf ./scons-2.2.0.tar.gz
   cd scons-2.2.0
-  python setup.py install --prefix=$HOME
+  python setup.py install --prefix=$S3_HOME
   cd ..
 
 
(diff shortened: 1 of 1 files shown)


Thomas-Ulrich:
Hi,
Thx for this pull request!
I think you are not defining $S3_HOME in your script, am I right? Is it $HOME/s3_software?
Can you explain why you call it S3 by the way?
Concerning
`ln -s $PWD/build/SeisSol_<version> $S3_HOME/bin/Seissol`
I think the information contained in the name of the executable are useful, and I would not rename against something more simple.

ravil-mobile:
Hi, Thomas

You are right I forgot to add the definition of S3_HOME environment variable. Regarding to your second question, I was looking for an abbreviation for Seissol. I noticed there were 3 letters 'S' in the name and it determined my decision to name it as "S3". Maybe Im wrong. If you can suggest an appropriate abbreviation  for it I'll appreciate it.

According to the third suggestion, I don't know. As for me it is cumbersome to always copy an executable from one directory to another, or always go to Seissol build directory. However, if you thing that the soft link looks ugly I can remove it

Documentation/compilation.rst
```diff
@@ -16,34 +16,63 @@ In order to run SeisSol, you need to first install:
...
-  export PARMETIS_LIBDIR='path_to_parmetis/lib'
+  export PARMETIS_BASE=$S3_HOME
+  export PARMETIS_LIBDIR=$S3_HOME/lib
+
+
+Add the following to your .bashrc (vi ~/.bashrc).
+
+.. code-block:: bash
+
+  source /opt/intel/compiler/VERSION/bin/compilervars.sh intel64 # if you use intel compilers
```
krenzland: This is specific to the lrz workstations

Documentation/compilation.rst
```diff
@@ -16,34 +16,63 @@ In order to run SeisSol, you need to first install:
...
+  touch seissol_env.sh
+
+Open the file with your favorite text editor (vi ./seissol_env.sh) and add following lines:
+
+.. code-block:: bash
 
   # User specific aliases and functions
-  export PATH=$HOME/bin:$PATH
-  export LD_LIBRARY_PATH=$HOME/lib:$HOME/lib64:$LD_LIBRARY_PATH
+  export S3_HOME=$PWD/s3_software
```
krenzland: I think that the suggestion of the folder structure ~/{bin,lib,include} is sane, as it can be reused for multiple coding projects.

Nevertheless, this could (should?) be made configurable.

S3 is hard to understand, SEISSOL_HOME would be better imo

Documentation/compilation.rst
```diff
@@ -16,34 +16,63 @@ In order to run SeisSol, you need to first install:
...
-  export C_INCLUDE_PATH=$HOME/include:$C_INCLUDE_PATH 
-  export CPLUS_INCLUDE_PATH=$HOME/include:$CPLUS_INCLUDE_PATH
+  export SCONS_LIB_DIR=$S3_HOME/lib64/scons-2.2.0/
+  export C_INCLUDE_PATH=$S3_HOME/include:$C_INCLUDE_PATH
+  export CPLUS_INCLUDE_PATH=$S3_HOME/include:$CPLUS_INCLUDE_PATH
+  export PKG_CONFIG_PATH=$S3_HOME/lib/pkgconfig
+
+  export FC=mpiifort  # or mpifort
+  export CXX=mpiicpc  # or mpicxx
+  export CC=mpiicc    # or mpicc
```
krenzland: Probably not a good idea to set this in the bashrc file as it overrides user defined settings
Could be source of strange errors, better to set in each call separately

Documentation/compilation.rst
```diff
@@ -16,34 +16,63 @@ In order to run SeisSol, you need to first install:
...
 ----------------
 
 .. code-block:: bash
 
   wget http://prdownloads.sourceforge.net/scons/scons-2.2.0.tar.gz
-  tar -xaf scons-2.2.0.tar.gz
+  tar -xvf ./scons-2.2.0.tar.gz
   cd scons-2.2.0
-  python setup.py install --prefix=$HOME
+  python setup.py install --prefix=$S3_HOME
```
krenzland: Nobody should install scons manually, but should use pip
See e.g.[Scons docu](https://scons.org/doc/production/HTML/scons-user.html#idm126)
(--user version strongly preferred!)

Documentation/compilation.rst
```diff
@@ -52,78 +81,88 @@ Installing HDF5
...
-  wget https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8/hdf5-1.8.11/src/hdf5-1.8.11.tar.bz2
-  tar -xaf hdf5-1.8.11.tar.bz2
-  cd hdf5-1.8.11
-  CC=mpiicc FC=mpiifort ./configure --enable-parallel --prefix=$HOME --with-zlib --disable-shared --enable-fortran 
-  #NOTE: This is with Intel MPI, if you have to use a different MPI, please adjust mpiicc and mpiifort accordingly to the scripts that call icc and ifort!
-  #MPICH note: CC=mpicc FC=mpif90 ./configure --enable-parallel --prefix=$HOME --with-zlib --disable-shared --enable-fortran
-  make
+  git clone https://github.com/mortenpi/hdf5.git
+  cd hdf5
+  git checkout 8a275ab7831002f3e  # take a stable version (optional)
```
krenzland: Why the change to github repository? As far as i know, the previously used archive was a stable release

Documentation/compilation.rst
```diff
@@ -52,78 +81,88 @@ Installing HDF5
...
-  tar -xaf netcdf-4.3.0.tar.gz
-  cd netcdf-4.3.0
-  CPPFLAGS=-I$HOME/include LDFLAGS=-L$HOME/lib CC=mpiicc ./configure --enable-shared=no --prefix=$HOME 
-  #MPICH note: CPPFLAGS=-I$HOME/include LDFLAGS=-L$HOME/lib CC=mpicc ./configure --enable-shared=no --prefix=$HOME 
-  #NOTE: Check for this line to make sure netCDF is build with parallel I/O: 
-  #"checking whether parallel I/O features are to be included... yes" This line comes at the very end (last 50 lines of configure run)!
-  make
+  tar -xaf ./netcdf-*.tar.gz
+  cd netcdf-4.4.1.1
+  CPPFLAGS="-I$S3_HOME/include -fPIC" ./configure --prefix=$S3_HOME --enable-shared=no
```
krenzland: why `-fPIC`?

Documentation/compilation.rst
```diff
@@ -52,78 +81,88 @@ Installing HDF5
...
 
-   git clone https://github.com/hfp/libxsmm
-   cd libxsmm
-   make generator
-   #add libxsmm/bin to path
+  git clone https://github.com/hfp/libxsmm
+  cd libxsmm
+  git checkout b6de187f832a723295a  # take a stable version (optional) 
+  make generator
+  cp ./bin/libxsmm_gemm_generator $S3_HOME/bin/
```
krenzland: Here, a symlink would be better (ln instead of cp)

Documentation/compilation.rst
```diff
@@ -52,78 +81,88 @@ Installing HDF5
...
 
-   tar -xvf metis-5.1.0.tar.gz
-   cd metis-5.1.0
-   #edit ./include/metis.h IDXTYPEWIDTH to be 64 (default is 32).
-   make config cc=mpiicc cxx=mpiicpc prefix=$HOME #(must confirm compiler!!!)
-   make install
-   cd ..
+  wget http://glaros.dtc.umn.edu/gkhome/fetch/sw/metis/metis-5.1.0.tar.gz
+  tar -xvf ./metis-5.1.0.tar.gz
+  cd metis-5.1.0
```
krenzland: why did you skip editing the metis.h file? Is is not necessary?

Documentation/compilation.rst
```diff
@@ -176,7 +209,13 @@ Get your executable with
 
 .. code-block:: bash
 
-   scons -j 32 buildVariablesFile=build/options/supermuc_mac_cluster.py
+   scons -j 4 buildVariablesFile=build/options/supermuc_mac_cluster.py
```
krenzland: Here (and in general): why -j4?
Maybe would be good idea to use placeholder variable for this

Documentation/compilation.rst
```diff
@@ -176,7 +209,13 @@ Get your executable with
...
+
+   # choose a correct executable file in case if you compile multiple
+   # versions of Seissol
+   ln -s $PWD/build/SeisSol_<version> $S3_HOME/bin/Seissol
```
krenzland: This can lead to confusion very quickly, probably better to use the information contained in the exec. name

Documentation/compilation.rst
ravil-mobile: Completely agree with you. It was setting from my laptop.  I'll fix the script and push updates

Documentation/compilation.rst
ravil-mobile: I checked installation of metis on several machines. You don't need to edit the header at all

krenzland: As far as I've understood this, it is not required for compilation but rather for very large graphs (where the 32bit length becomes restrictive).
Probably @uphoffc knows more about this.

uphoffc: Yes the 64 bits are required for large meshes. Might make sense to mark it as optional and to provide a patch in the documentation, such that manual editing is no longer required.
Regards,
Carsten

Thomas-Ulrich:
Is there anything to be done with this pull request?

uphoffc:
I think Ravil wanted to change some things according to the comments, which he has not done yet.

PR #67, opened 2019-05-16
Added PSpaMM installation to doc

uphoffc:


krenzland:
Is this actually used for something? Couldn't find anything in documentation or source code.

uphoffc:
Yes, https://github.com/SeisSol/SeisSol/blob/master/generated_code/generate.py#L94

krenzland:
Ah, okay, that's on me.
Still, I'm not sure if adding this as a dependency is the best way here - why not as a submodule?

The other dependencies of this kind make sense as these can be sometimes found on clusters (e.g. hdf5/metis).

uphoffc:
A submodule would also be possible, although I don't like the abuse of submodules as a package manager.

krenzland:
Neither do I.
I'll merge this; but I'm not happy with this.
W should try and find a better solution in the future.

uphoffc:
Yes, but we also should find a solution for a lot of other submodules, like glm, yaml-cpp, maybe ImpalaJIT

-> Package manager that works on hpc clusters

PR #69, opened 2019-05-28
Carsten/stats

uphoffc:
- Moved summary (median, min, max, mean, std) function to own class.
- Summary created for time spent in compute kernels.
- Removed load imbalance based on flops and introduced load imbalance based on time.
- Introduced standard error for linear regression analysis of compute kernels -> sufficient to calculate confidence intervals.

### DIFF
--- a/src/Monitoring/LoopStatistics.cpp
+++ b/src/Monitoring/LoopStatistics.cpp
@@ -44,10 +44,12 @@
 #include <netcdf_par.h>
 #endif
 
+#include "Numerical_aux/Statistics.h"
+
 #ifdef USE_MPI  
 void seissol::LoopStatistics::printSummary(MPI_Comm comm) {
   unsigned nRegions = m_times.size();
-  double* sums = new double[5*nRegions];
+  auto sums = std::vector<double>(5*nRegions);
   double totalTimePerRank = 0.0;
   for (unsigned region = 0; region < nRegions; ++region) {
     double x = 0.0, x2 = 0.0, xy = 0.0, y = 0.0;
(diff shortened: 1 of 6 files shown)


src/Monitoring/LoopStatistics.cpp
```diff
@@ -74,12 +77,45 @@ void seissol::LoopStatistics::printSummary(MPI_Comm comm) {
...
+
+  const auto loadImbalance = 1.0 - summary.mean / summary.max;
+  logInfo(rank) << "Load imbalance:" << 100.0 * loadImbalance << "%";
+
+  MPI_Allreduce(MPI_IN_PLACE, sums.data(), 5*nRegions, MPI_DOUBLE, MPI_SUM, comm);
+
+  auto regressionCoeffs = std::vector<double>(2*nRegions);
+  auto stderror = std::vector<double>(nRegions, 0.0);
+  for (unsigned region = 0; region < nRegions; ++region) {
+    double x  = sums[5*region + 0];
```
krenzland: all variables here should be consts

src/Monitoring/LoopStatistics.cpp
```diff
@@ -74,12 +77,45 @@ void seissol::LoopStatistics::printSummary(MPI_Comm comm) {
...
+      if (sample.numIters > 0) {
+        double error = sample.time - (constant + slope * sample.numIters);
+        stderror[region] += error * error;
+      }
+    }
+  }
 
   if (rank == 0) {
-    MPI_Reduce(MPI_IN_PLACE, sums, 5*nRegions, MPI_DOUBLE, MPI_SUM, 0, comm);
+    MPI_Reduce(MPI_IN_PLACE, stderror.data(), nRegions, MPI_DOUBLE, MPI_SUM, 0, comm);
```
krenzland: use stderror.size instead of nRegions

src/Monitoring/LoopStatistics.cpp
```diff
@@ -74,12 +77,45 @@ void seissol::LoopStatistics::printSummary(MPI_Comm comm) {
...
   } else {
-    MPI_Reduce(sums, 0L, 5*nRegions, MPI_DOUBLE, MPI_SUM, 0, comm);
+    MPI_Reduce(stderror.data(), 0L, nRegions, MPI_DOUBLE, MPI_SUM, 0, comm);
```
krenzland: use stderror.size instead of nRegions

src/Monitoring/LoopStatistics.cpp
```diff
@@ -88,28 +124,27 @@ void seissol::LoopStatistics::printSummary(MPI_Comm comm) {
...
       double y  = sums[5*region + 3];
       double N  = sums[5*region + 4];
 
-      double det = N*x2 - x*x;
-      double regressionCoeffs[2];
-      regressionCoeffs[0] = (x2*y - x*xy) / det;
-      regressionCoeffs[1] = (-x*y + N*xy) / det;
-    
+      double xm = x / N;
+      double xv = x2 - 2*x*xm + xm*xm;
```
krenzland: Is there a reason why you've expanded the binomial?
Makes it imo a bit hard for "pattern matching" this with the formula

uphoffc: Yes, because I may reuse the sums I already computed and sent around via MPI.

krenzland:
@uphoffc, can you please address my comments so that we can merge?

PR #70, opened 2019-06-11
Fix pinning for async i/o.

krenzland:
This fixes some pinning issues with the asyn IO.

Old behavior: I/O thread is pinned to some thread, depending on the implementation in the async library.

New behavior: I/O thread is pinned to threads that are not utilized by OpenMP and thus uses identical pinning as the communication thread.

Related changes in the ASYNC library are [here](https://github.com/TUM-I5/ASYNC/commit/5a55cdf94fad642232e5c48acd92beea8542c3ae)

### DIFF
--- a/src/Checkpoint/Manager.h
+++ b/src/Checkpoint/Manager.h
@@ -120,14 +120,12 @@ public:
 	void setUp()
 	{
 		setExecutor(m_executor);
-#if defined(_OPENMP) && defined(USE_MPI)
 		auto freeCpus = parallel::getFreeCPUsMask();
 		logInfo(seissol::MPI::mpi.rank()) << "Checkpoint thread affinity:" << parallel::maskToString(parallel::getFreeCPUsMask());
 		if (parallel::freeCPUsMaskEmpty(freeCpus)) {
 		  logError() << "There are no free CPUs left. Make sure to leave one for the I/O thread(s).";
 		}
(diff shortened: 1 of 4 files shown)


krenzland:
> Do you think it makes sense to introduce an intermediate class, as the setUp function is duplicated 3 times?
In principle yes but I'm not sure if the mental overhead is worth it.
If we further extend the setup, we should definitely do something about this, right now I prefer the small duplication.

Is it a good idea to error when there's no core left to pin to? There is probably no use case where you want to use async-i/o and pin openmp threads to all available threads, right?

uphoffc:
I think if someone is using asynchronuous I/O then he should be aware about pinning and multi-threading and stuff. If the code then fails, it's imho better to fail then to continue with weird pinning, and likely wasted CPU-hours.

However, the code would give the same error also for synchronuous I/O, where one might not leave a core free. Could you maybe add a "if necessary" query to async? Then the error could be thrown only if when an async thread is actually used.

uphoffc:
You could also check "CPU_COUNT(&set) == 0" (see freeCPUsMaskEmpty) inside setAffinityIfNecessary. The latter function could then throw an exception which may be caught by SeisSol and then logError is called.

krenzland:
Thanks, that's a good idea.
I've implemented the first approach; we are only interested in the affinity cpu-mask if we perform any pinning. Otherwise this might be confusing.

PR #72, opened 2019-06-18
Fix typo in metis comp.

krenzland:


Thomas-Ulrich:
well I guess it depends on the compiler name. 
e.g. using mpi.ibm, it is called mpicc.
So for me it is not a typo.

krenzland:
That's a good point, but shouldn't it be mpicc instead of mpiCC then?

Thomas-Ulrich:
mpicc for c files and mpiCC for cpp files?

krenzland:
Ah, you're right.
Apparently it's been a long time since I've compiled on phase 2 ;)

I think we should change this to use ${CC} ${CXX} environment variables in the future but I'll close this for now.

PR #78, opened 2019-07-04
Optimization of FL103

krenzland:
Two optimizations for FL103:

- Newton iteration now checks convergence for each spatial quad node separately.

- Number of time integration points can now be changed to a value independent of the polynomial order of the DG scheme.

### DIFF
--- a/src/Kernels/DynamicRupture.cpp
+++ b/src/Kernels/DynamicRupture.cpp
@@ -90,8 +90,8 @@ void seissol::kernels::DynamicRupture::setTimeStepWidth(double timestep)
     timeWeights[timeInterval] = subIntervalWidth;
   }*/
 #else
-  seissol::quadrature::GaussLegendre(timePoints, timeWeights, CONVERGENCE_ORDER);
-  for (unsigned point = 0; point < CONVERGENCE_ORDER; ++point) {
+  seissol::quadrature::GaussLegendre(timePoints, timeWeights, NUMBER_OF_TEMPORAL_INTEGRATION_POINTS);
+  for (unsigned point = 0; point < NUMBER_OF_TEMPORAL_INTEGRATION_POINTS; ++point) {
     timePoints[point] = 0.5 * (timestep * timePoints[point] + timestep);
     timeWeights[point] = 0.5 * timestep * timeWeights[point];
   }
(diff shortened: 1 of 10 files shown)


src/Kernels/DynamicRupture.cpp
```diff
@@ -158,6 +158,6 @@ void seissol::kernels::DynamicRupture::flopsGodunovState( DRFaceInformation cons
   o_nonZeroFlops += kernel::godunovState::nonZeroFlops(faceInfo.minusSide, faceInfo.faceRelation);
   o_hardwareFlops += kernel::godunovState::hardwareFlops(faceInfo.minusSide, faceInfo.faceRelation);
   
-  o_nonZeroFlops *= CONVERGENCE_ORDER;
-  o_hardwareFlops *= CONVERGENCE_ORDER;
+  o_nonZeroFlops *= NUMBER_OF_TEMPORAL_INTEGRATION_POINTS;
```
krenzland: @uphoffc Is this number of flops correct here?

Thomas-Ulrich:
Did you actually tested that it is faster (e.g. on tpv104)?
Because on the one hand, you are indeed running less iterations, but on the other hand you loose the vectorialization on this probably most expensive part of the function.

krenzland:
> Did you actually tested that it is faster (e.g. on tpv104)?
No, but we will do so before merging.

> Because on the one hand, you are indeed running less iterations, but on the other hand you loose the vectorialization on this probably most expensive part of the function.
I'm not sure whether the loop is actually vectorized in the current implementation, as the number of iterations isn't constant in this function, I think. We will check the performance and will then decide.

Thomas-Ulrich:
> Initial tests suggest that this version is slower. However, I cannot exclude any side effects.
> We should run a detailed benchmark before merging.

That s what I was expected. Problably there is no strong variability in the number of iteration of the loop that could justify breaking the vectorialization.

AliceGabriel:
In the original implementation the number of iterations was hard coded ( to *cough* 3 or 5)
- i think there is not much variation in practice 

How about looping through the fvw without ADER at all?

--
Dr. Alice-Agnes Gabriel
bit.ly/AAG8LMU

Thomas-Ulrich:
> How about looping through the fvw without ADER at all?

This is possible thanks to this commit (If I understood correctly)
@krenzland
Add preprocessor flag to control num. of time quad points for dr

(using 1 time quad points)

krenzland:
Thanks for the comments!
 I should have marked this as WIP; which is apparently a feature that Github offers ("draft pull request").

In hindsight I'm not surprised that the Newton method converges in a similar speed for all quad. points because we have quite a large resolution near the fault and thus similar values.

The plan for this PR is:
- Test changes again on a more reliable system. I don't trust the NG enough to make any decisions based on its performance.
- Merge the part that deals with the number of quad. points regardless of the other "optimization" as it offers a potential avenue for optimization and by default doesn't change the behavior of SeisSol.

PR #83, opened 2019-07-08
Fix compile flags for netcdf.

krenzland:
Add fPIC flag to netcdf, needed to compile ASAGI.


uphoffc:
Don't you need to add -fPIC to hdf5 and asagi, too, in order for this change to make sense?

krenzland:
True, forgot the flags for HDF5.

Not sure if the flag is needed for asagi. Doesn't matter anyway, I'll include this in a separate commit to the asagi repo.

PR #84, opened 2019-07-10
Remove lsw16

Thomas-Ulrich:
remove LSW2, and move LSW16 (optimized + forced time rupture) to FL=2.
Add instant healing feature (was in former LSW2).


AliceGabriel:
Approved from my side - but can the documentation and any possible examples
using FL=16 or FL=2 please be updated accordingly?

uphoffc:
This change removes code duplication, which is good.

However, 
- This change (in the current implementation) requires you to always specify forced_rupture_time, which might not be necessary for every scenario.
- It break backwards compatibility to old parameter files for no good reason.

Thomas-Ulrich:
You are right Carsten.
Therefore I ve rewritten the pull request:
I now keep the tag 16, but I combine 2 and 16 into a single routine.

PR #88, opened 2019-07-16
Include the documentation build into the travis pipeline

sebwolf-de:


.travis.yml
```diff
@@ -31,6 +33,7 @@ install:
  - cd libxsmm-master && make generator && sudo cp bin/libxsmm_gemm_generator /usr/bin && cd ..
 
 script:
+ - sphinx-build -n -b dummy Documentation Documentation/_build/html
```
krenzland: Is there a reason why you don't use the makefile in the Documentation folder?
(not sure how it's actually build by readthedocs)

uphoffc: Yes, "make" would be better.

PR #91, opened 2019-07-17
Tp

sebanger:
Thermal pressurization implementation


Thomas-Ulrich:
Do you have an example file to test it?

SConstruct
```diff
@@ -378,8 +378,10 @@ if env['scalasca'] == 'kernels_2.x':
 #
 
 # enforce restrictive C/C++-Code
-env.Append(CFLAGS   = ['-Wall', '-Werror', '-ansi'],
-           CXXFLAGS = ['-Wall', '-Werror', '-ansi'])
+#env.Append(CFLAGS   = ['-Wall', '-Werror', '-ansi'],
```
krenzland: Two issues with that:
-  Please don't commit commented out code. That's what Git is for.
- While in principle I agree that this should be changed this is probably something that should be discussed separately.

src/Numerical_aux/typesdef.f90
```diff
@@ -931,7 +931,22 @@ REAL                         , POINTER :: BF_GP_Hex(:,:)                   !<Val
...
-     REAL, POINTER                          :: Strength(:,:) => NULL()                   !< save strength since it is used for bimaterial
+     REAL, POINTER                          :: Strength(:,:) => NULL()          !< save strength since it is used for bimaterial
+     INTEGER                                :: thermalPress                     !< thermal pressurization switch
+     REAL                                   :: alpha_th                         !< thermal diffusion parameter for TP
+     REAL                                   :: alpha_hy                         !< hydraulic diffusion parameter for TP
+     REAL                                   :: rho_c                            !< heat capacity for TP
+     REAL                                   :: TP_lambda                        !< pore pressure increase per unit increase
+     INTEGER                                :: TP_nz                            !< number of grid points to solve advection for TP
+     REAL                                   :: TP_dlDwn                         !< grid space distance for TP_grid
```
Thomas-Ulrich: Why not using more explicit variable names?
DISC%DynRup%TP_hwid vs TP_grid_half_width?
DISC%DynRup%TP_dlDwn vs TP_grid_dz? or TP_dz? given the fact that there is nz?
DISC%DynRup%TP_nz vs TP_grid_nz?
DISC%DynRup%TP_Dwnmax vs DISC%DynRup%TP_max_wavenumber
update_RSF vs update_State_Variable_RSF

src/Reader/readpar.f90
```diff
@@ -1329,6 +1338,22 @@ SUBROUTINE readdr(IO, EQN, DISC, BND, IC)
...
+             DISC%DynRup%ThermalPress = thermalPress !switches TP on (1) or off(0)
+             IF (DISC%DynRup%ThermalPress.EQ.1) THEN !additional parameters
+                 logInfo0(*) 'Thermal pressurization assumed'
+                 !pyhsical
+                 DISC%DynRup%TP_hwid = TP_hwid
+                 DISC%DynRup%alpha_th = alpha_th
+                 DISC%DynRup%alpha_hy = alpha_hy
+                 DISC%DynRup%rho_c = rho_c
+                 DISC%DynRup%TP_lambda = TP_lambda
+                 EQN%Temp_0 = IniTemp
```
Thomas-Ulrich: Why not using a consistent name in the namelist and the structure?
Also don't you want to use easi for initialisation of these parameters?

src/Reader/readpar.f90
```diff
@@ -1329,6 +1338,22 @@ SUBROUTINE readdr(IO, EQN, DISC, BND, IC)
              DISC%DynRup%RS_iniSlipRate1 = RS_iniSlipRate1! V_ini1, initial sliding velocity
              DISC%DynRup%RS_iniSlipRate2 = RS_iniSlipRate2! V_ini2, initial sliding velocity
              DISC%DynRup%t_0      = t_0       ! forced rupture decay time
+             DISC%DynRup%ThermalPress = thermalPress !switches TP on (1) or off(0)
+             IF (DISC%DynRup%ThermalPress.EQ.1) THEN !additional parameters
+                 logInfo0(*) 'Thermal pressurization assumed'
+                 !pyhsical
```
Thomas-Ulrich: typo

src/Physics/Evaluate_friction_law.f90
```diff
@@ -1607,4 +1626,106 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
...
+    LOGICAL       :: has_converged                                            !check convergence
+    INTEGER       :: FL, nSRupdates, i, nBndGP
+    REAL          :: RS_sr0, RS_a(nBndGP)                                     !constants
+    REAL          :: SRtest(nBndGP), LocSR(nBndGP), LocSV(nBndGP)
+    REAL          :: n_stress(nBndGP), sh_stress(nBndGP), invZ
+    REAL          :: NR(nBndGP), dNR(nBndGP), tmp(nBndGP), tmp2(nBndGP), tmp3(nBndGP)
+    REAL          :: mu_f(nBndGP), dmu_f(nBndGP)                              !calculated here in routine
+    REAL          :: AlmostZero = 1D-45, aTolF = 1e-8
+    !-------------------------------------------------------------------------!
+    INTENT(IN)    :: FL, nSRupdates, LocSR, RS_sr0, LocSV, RS_a, n_stress,&
```
Thomas-Ulrich: FL is not needed

src/Physics/Evaluate_friction_law.f90
```diff
@@ -1607,4 +1626,106 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
...
+                     sh_stress, invZ
+    INTENT(OUT)   :: SRtest, NR, has_converged
```
Thomas-Ulrich: NR, has_converged not needed

src/Physics/Evaluate_friction_law.f90
```diff
@@ -1607,4 +1626,106 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
...
+    IMPLICIT NONE
+    !-------------------------------------------------------------------------!
+    ! Argument list declaration
+    TYPE(tInputOutput)       :: IO
+    INTEGER                  :: nBndGP
+    REAL                     :: RS_f0, RS_b, RS_a(nBndGP), RS_sr0, RS_fw, RS_srW(nBndGP), RS_sl0(nBndGP) !constant input parameters
+    REAL                     :: SV0(nBndGP), time_inc, SR_tmp(nBndGP), LocSR(nBndGP)                  !changing during iterations
+    REAL                     :: flv(nBndGP), fss(nBndGP), SVss(nBndGP), LocSV(nBndGP)                 !calculated in this routine
+    !-------------------------------------------------------------------------!
+    INTENT(IN)    :: IO, RS_f0, RS_b, RS_a, RS_sr0, RS_fw, RS_srW, &
```
Thomas-Ulrich: IO and LocSR not used

src/Physics/Evaluate_friction_law.f90
```diff
@@ -1539,17 +1544,31 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
...
-         flv = RS_f0 -(RS_b-RS_a)* LOG(tmp/RS_sr0)
-         fss = RS_fw + (flv - RS_fw)/(1.0D0+(tmp/RS_srW)**8)**(1.0D0/8.0D0)
-         SVss = RS_a * LOG(2.0D0*RS_sr0/tmp * ( EXP(fss/RS_a)-EXP(-fss/RS_a))/2.0D0)
-         LocSV=Svss*(1.0d0-EXP(-tmp*time_inc/RS_sl0))+EXP(-tmp*time_inc/RS_sl0)*SV0
-         !Mu from LocSR
-         tmp = 0.5D0*(LocSR)/RS_sr0 * EXP(LocSV/RS_a)
-         LocMu    = RS_a * LOG(tmp+SQRT(tmp**2+1.0D0))
-         ! update stress change
+         CALL update_RSF (IO, nBndGP, RS_f0, RS_b, RS_a, RS_sr0, RS_fw, RS_srW, RS_sl0, SV0, time_inc, SR_tmp, LocSR, LocSV)
+         S = LocMu*(P - P_f)
```
Thomas-Ulrich: S  is only used for TP. I would therefore move it inside the IF loop?

src/Physics/Evaluate_friction_law.f90
```diff
@@ -1607,4 +1626,106 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
...
+    ! steady-state state variable
+    ! For compiling reasons we write SINH(X)=(EXP(X)-EXP(-X))/2
+    SVss = RS_a * LOG(2.0D0*RS_sr0/SR_tmp * (EXP(fss/RS_a)-EXP(-fss/RS_a))/2.0D0)
+
+    ! exact integration of dSV/dt DGL, assuming constant V over integration step
+    LocSV = Svss*(1.0D0-EXP(-SR_tmp*time_inc/RS_sl0))+EXP(-SR_tmp*time_inc/RS_sl0)*SV0
+
+
+    IF (ANY(IsNaN(LocSV)) .EQV. .TRUE.) THEN
+       STOP
```
Thomas-Ulrich: add something like logError(*) 'NaN detected', time?

src/Reader/readpar.f90
uphoffc: easi is only well suited for parameters that (potentially) vary in space. Otherwise it's more overhead than useful.
What are these parameters and do they potentially vary?

src/Physics/Evaluate_friction_law.f90
```diff
@@ -57,8 +57,16 @@ MODULE Eval_friction_law_mod
   INTERFACE Eval_friction_law
      MODULE PROCEDURE Eval_friction_law
   END INTERFACE
+  !INTERFACE update_RSF
```
Thomas-Ulrich: remove commented out code

src/Numerical_aux/typesdef.f90
sebanger: the name conventions were introduced by Stephanie. I didn't replace it. 
but those ones are short and still explainable

uphoffc: I agree with Thomas. We are not in the 70s anymore where we need cryptic abbreviations to save a few bytes.

src/Physics/Evaluate_friction_law.f90
```diff
@@ -57,8 +57,16 @@ MODULE Eval_friction_law_mod
...
+     !MODULE PROCEDURE update_RSF
+  !END INTERFACE
+    !INTERFACE Newton_Raphson
+     !MODULE PROCEDURE Newton_Raphson
+  !END INTERFACE
   !---------------------------------------------------------------------------!
   PUBLIC  :: Eval_friction_law
+  PRIVATE :: update_RSF
+  PRIVATE :: Newton_Raphson
```
uphoffc: These names are not sensible:
- RSF is an unnecessary abbreviation and breaks with the rate_and_state naming scheme.
- Both names suggest to be more general then they actually are. In particular Newton_Raphson is a method that may be applied to many problems,

src/Physics/Evaluate_friction_law.f90
```diff
@@ -1417,7 +1428,7 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
     !absolute tolerance on the function to be optimzed
     ! This value is quite arbitrary (a bit bigger as the expected numerical error) and may not be the most adapted
     !aTolF = 5e-15
-    aTolF = 1e-8
+    !aTolF = 1e-8
```
uphoffc: Delete the comment if unnecessary.

src/Physics/ini_model_DR.f90
```diff
@@ -659,6 +688,53 @@ SUBROUTINE friction_LSW6(DISC,EQN,MESH,BND)
...
+  !-------------------------------------------------------------------------!
+  ! Local variable declaration
+  INTEGER                        :: j
+  REAL                           :: dlDwn, hwid, Dwnmax, Dwnmax_norm
+  REAL, PARAMETER                :: pi=3.141592653589793
+  !-------------------------------------------------------------------------!
+  INTENT(INOUT) :: DISC,EQN
+  !-------------------------------------------------------------------------!
 
+  !values currently from bicycle code -> how can we optimize that?
```
Thomas-Ulrich: what is a bicycle code?

src/Reader/readpar.f90
sebanger: At the beginning TP parameters should not vary (just to make life easier).
There will be an updated version when everything is working where some parameters can vary.

src/Physics/Evaluate_friction_law.f90
sebanger: this is already in the master branch with Nan detected. I haven't pulled it yet.

src/Reader/readpar.f90
sebanger: where?

src/Physics/ini_model_DR.f90
sebanger: I guess Stephanie referred to this as the spectral BIEM by Noda & Lapusta 2010

src/Physics/Evaluate_friction_law.f90
```diff
@@ -1539,17 +1544,31 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
...
+         !update LocMu for next strength determination, only needed for last update
+         ! X in Asinh(x) for mu calculation
+         tmp = 0.5D0/RS_sr0 * EXP(LocSV/RS_a)
+         tmp2 = LocSR*tmp
+         ! mu from LocSR
+         LocMu  = RS_a*LOG(tmp2+SQRT(tmp2**2+1.0D0))
+
+         ! update stress change
+         LocTracXY = -((EQN%InitialStressInFaultCS(:,4,iFace) + XYStressGP(:,iTimeGP))/ShTest)*LocMu*(P-P_f)
+         LocTracXZ = -((EQN%InitialStressInFaultCS(:,6,iFace) + XZStressGP(:,iTimeGP))/ShTest)*LocMu*(P-P_f)
```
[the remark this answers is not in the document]
sebanger: because P_f is getting updated in the ThermalPressure subroutine

src/Reader/readpar.f90
Thomas-Ulrich: pyhsical

src/Physics/ini_model_DR.f90
Thomas-Ulrich: ok I guess bi-cycle (two cycles?)

src/Numerical_aux/typesdef.f90
Thomas-Ulrich: Hi,
So after working on TP, I realized that the notations are the same as in the matlab script on the TP SCEC benchmark.  Therefore, it might make sense to keep them.

SConstruct
```diff
@@ -378,8 +378,10 @@ if env['scalasca'] == 'kernels_2.x':
...
+#           CXXFLAGS = ['-Wall', '-Werror', '-ansi'])
```
Thomas-Ulrich: please remove commented lines

src/Physics/Evaluate_friction_law.f90
Thomas-Ulrich: sorry but this is not resolved.
What about updateStateVariable instead of update_RSF
and IterativelyInvertSR instead of Newton_Raphson?

sebanger:
I made all changes. Please review again.

SConstruct
```diff
@@ -369,8 +369,8 @@ if env['scalasca'] == 'kernels_2.x':
 #
 
 # enforce restrictive C/C++-Code
-env.Append(CFLAGS   = ['-Wall', '-Werror', '-ansi'],
-           CXXFLAGS = ['-Wall', '-Werror', '-ansi'])
+env.Append(CFLAGS   = ['-Wall', '-ansi'],
```
Thomas-Ulrich: This has to be reverted

src/Physics/Evaluate_friction_law.f90
```diff
@@ -1455,4 +1469,105 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
...
+    !-------------------------------------------------------------------------!
+    ! Argument list declaration
+    LOGICAL       :: has_converged                                            !check convergence
+    INTEGER       :: nSRupdates, i, nBndGP
+    REAL          :: RS_sr0, RS_a(nBndGP)                                     !constants
+    REAL          :: SRtest(nBndGP), LocSR(nBndGP), LocSV(nBndGP)
+    REAL          :: n_stress(nBndGP), sh_stress(nBndGP), invZ
+    REAL          :: NR(nBndGP), dNR(nBndGP), tmp(nBndGP), tmp2(nBndGP), tmp3(nBndGP)
+    REAL          :: mu_f(nBndGP), dmu_f(nBndGP)                              !calculated here in routine
+    REAL          :: AlmostZero = 1D-25, aTolF = 5d-14 !1e-8
```
Thomas-Ulrich: Is there a reason for changing the values of AlmostZero and aTolF?

uphoffc:
> I made all changes. Please review again.

Closing conversations != making changes.

SConstruct
```diff
@@ -369,8 +369,8 @@ if env['scalasca'] == 'kernels_2.x':
...
+           CXXFLAGS = ['-Wall', '-ansi'])
```
uphoffc: This, too.

src/Reader/readpar.f90
```diff
@@ -729,7 +730,7 @@ subroutine readpar_faultElementwise(EQN,BND,IC,DISC,IO,CalledFromStructCode)
     DISC%DynRup%DynRup_out_elementwise%refinement = refinement                 ! read info of desired refinement level : default 0
 
     !Dynamic shear stress arrival output currently only for linear slip weakening friction laws
-    IF (OutputMask(11).EQ.1) THEN
+    IF (OutputMask(11).GE.1) THEN
```
Thomas-Ulrich: Why this change?

src/ResultWriter/ini_faultoutput.f90
```diff
@@ -104,6 +105,19 @@ SUBROUTINE fill_variables_list(DISC, VariableList)
...
+       IF (j.EQ.12.AND.DISC%DynRup%DynRup_out_atPickpoint%OutputMask(6).NE.1) CYCLE
+       IF (j.EQ.13.AND.DISC%DynRup%DynRup_out_atPickpoint%OutputMask(6).NE.1) CYCLE
+       IF (j.EQ.14.AND.DISC%DynRup%DynRup_out_atPickpoint%OutputMask(7).NE.1) CYCLE
+       IF (j.EQ.15.AND.DISC%DynRup%DynRup_out_atPickpoint%OutputMask(8).NE.1) CYCLE
+       IF (j.EQ.16.AND.DISC%DynRup%DynRup_out_atPickpoint%OutputMask(9).NE.1) CYCLE
+       IF (j.EQ.17.AND.DISC%DynRup%DynRup_out_atPickpoint%OutputMask(10).NE.1) CYCLE
+       IF (j.EQ.18.AND.DISC%DynRup%DynRup_out_atPickpoint%OutputMask(11).NE.1) CYCLE
+       IF (j.EQ.19.AND.DISC%DynRup%DynRup_out_atPickpoint%OutputMask(12).NE.1) CYCLE
+       IF (j.EQ.20.AND.DISC%DynRup%DynRup_out_atPickpoint%OutputMask(12).NE.1) CYCLE
+
```
uphoffc: Please rewrite this code in a more sensible way.

src/ResultWriter/faultoutput.f90
```diff
@@ -535,174 +541,179 @@ SUBROUTINE calc_FaultOutput( DynRup_output, DISC, EQN, MESH, MaterialVal, BND, t
...
+                xac = (/ xV(2)-xV(1), yV(2)-yV(1), zV(2)-zV(1) /)
+             CASE(2)
+                xab = (/ xV(2)-xV(1), yV(2)-yV(1), zV(2)-zV(1) /)
+                xac = (/ xV(4)-xV(1), yV(4)-yV(1), zV(4)-zV(1) /)
+             CASE(3)
+                xab = (/ xV(4)-xV(1), yV(4)-yV(1), zV(4)-zV(1) /)
+                xac = (/ xV(3)-xV(1), yV(3)-yV(1), zV(3)-zV(1) /)
+             CASE(4)
+                xab = (/ xV(3)-xV(2), yV(3)-yV(2), zV(3)-zV(2) /)
+                xac = (/ xV(4)-xV(2), yV(4)-yV(2), zV(4)-zV(2) /)
```
uphoffc: Where do all these changes come from? Please don't "just" change the formatting and indentation.

src/ResultWriter/faultoutput.f90
```diff
@@ -535,174 +541,179 @@ SUBROUTINE calc_FaultOutput( DynRup_output, DISC, EQN, MESH, MaterialVal, BND, t
...
+          IF (DynRup_output%OutputMask(9).EQ.1) THEN
+              OutVars = OutVars + 1
+              DynRup_output%OutVal(iOutPoints,1,OutVars) = DISC%DynRup%output_PeakSR(iBndGP,iFace) !OutVars = 16
+          ENDIF
+          IF (DynRup_output%OutputMask(10).EQ.1) THEN
+              OutVars = OutVars + 1
+              DynRup_output%OutVal(iOutPoints,1,OutVars) = DISC%DynRup%output_rupture_time(iBndGP,iFace) !OutVars = 17
+          ENDIF
+          IF (DynRup_output%OutputMask(11).EQ.1) THEN
+              OutVars = OutVars + 1
```
uphoffc: Where do all these changes come from? Please don't "just" change the formatting and indentation.

src/ResultWriter/faultoutput.f90
```diff
@@ -535,174 +541,179 @@ SUBROUTINE calc_FaultOutput( DynRup_output, DISC, EQN, MESH, MaterialVal, BND, t
...
-                  OutVars = OutVars + 1
-                  DynRup_output%OutVal(iOutPoints,1,OutVars)  = cos1 * DISC%DynRup%output_Slip1(iBndGP,iFace) - sin1* DISC%DynRup%output_Slip2(iBndGP,iFace)
-                  OutVars = OutVars + 1
-                  DynRup_output%OutVal(iOutPoints,1,OutVars)  = sin1 * DISC%DynRup%output_Slip1(iBndGP,iFace) + cos1 * DISC%DynRup%output_Slip2(iBndGP,iFace)
-              ENDIF
+            IF (compute_VR.EQ..TRUE.) THEN
+
+            ! The rupture velocity is based on the gradient of the rupture time
+            ! The gradient is not expected to be accurate for the side gauss points
+            !! We consider then the next internal gauss point if the closest gauss point is external
```
uphoffc: Where do all these changes come from? Please don't "just" change the formatting and indentation.

src/ResultWriter/faultoutput.f90
```diff
@@ -535,174 +541,179 @@ SUBROUTINE calc_FaultOutput( DynRup_output, DISC, EQN, MESH, MaterialVal, BND, t
...
-                  Vr=1d0/Slowness
-                ENDIF
-                DEALLOCATE(projected_RT)
-                iBndGP = DynRup_output%OutInt(iOutPoints,1)
-                ENDIF
-
-                OutVars = OutVars + 1
-                DynRup_output%OutVal(iOutPoints,1,OutVars)  = Vr
+            !calculation of the spatial derivatives of the rupture time
+            chi  = MESH%ELEM%BndGP_Tri(1,iBndGP)
```
uphoffc: Where do all these changes come from? Please don't "just" change the formatting and indentation.

src/Physics/Evaluate_friction_law.f90
sebanger: I used the same values as in the 2d code. I can redo this and look if there is a difference.

src/Reader/readpar.f90
sebanger: because it can be for 11 and 12 (TP)

Thomas-Ulrich:
I think to keep the value unchanged would be better.
For instance decreasing aTolF from 5d-14 to 1e-8 will lead to
additionnal Newton Iterations, which are probably not required, and
might affect the scaling of the code with RS.
Changing AlmostZero should not have large effects, but better keeping
the same value and keeping RS unchanged.

src/ResultWriter/ini_faultoutput.f90
sebanger: to be honest I do not know how to do that. What does sensible way mean?

Thomas-Ulrich:
1) This is not what you wrote.

2) This if is only for DS output, not related to TP.

src/ResultWriter/faultoutput.f90
sebanger: that means I should redo this?

src/ResultWriter/ini_faultoutput.f90
uphoffc: More sensible means:
- Less code repetition
- Not checking 20 if statements in each iteration

E.g.

maskCount = (/ 2, 3, 1, 2, 3, 2, 1, 1, 1, 1, 1, 2 /)

k = 1
do i=lbound(maskCount),ubound(maskCount)
  if (DISC%...%OutputMask(i) .EQ. 1)
    do j=0,maskCount(i)-1
       WRITE(VariableList,'(a,a3,a,a1)') TRIM(VariableList_temp),',"',TRIM(VName(k+j)),'"'
    end do
  end if
  k = k + maskCount(i)
end do

src/Numerical_aux/typesdef.f90
krenzland: Just because other people use bad naming doesn't imply that we should also use bad naming.
Call the variable TP_max_wavenumber and (maybe) add a comment that it corresponds to TP_Dwnmax.

src/Reader/readpar.f90
```diff
@@ -1032,6 +1042,23 @@ SUBROUTINE readdr(IO, EQN, DISC, BND, IC)
              DISC%DynRup%RS_iniSlipRate1 = RS_iniSlipRate1! V_ini1, initial sliding velocity
              DISC%DynRup%RS_iniSlipRate2 = RS_iniSlipRate2! V_ini2, initial sliding velocity
              DISC%DynRup%t_0      = t_0       ! forced rupture decay time
+             DISC%DynRup%ThermalPress = thermalPress !switches TP on (1) or off(0)
+             IF (DISC%DynRup%ThermalPress.EQ.1) THEN !additional parameters
+                 logInfo0(*) 'Thermal pressurization assumed'
+                 !physical
+                 DISC%DynRup%TP_grid_half_width = TP_grid_half_width
```
Thomas-Ulrich: what is the difference between TP_grid_half_width and TP_grid_space_distance?

Thomas-Ulrich:
@uphoffc or @krenzland : any idea why the build fails in Travis and not on the master branch?
'Vr' has also 2 characters in the master branch, right?

src/Reader/readpar.f90
sebanger: ok maybe a bit misleading. I can change TP_grid_half_width into TP_half_width_shear_zone

Thomas-Ulrich:
ok and then TP_grid_space_distance by TP_grid_half_width?

src/Reader/readpar.f90
sebanger: no TP_grid_space_distance was formely TP_dlDwn which means it is the logarithmic space distance of the grid in the z-direction.

sebanger: it has nothing to do with half_width and the space distance is logarithmic. Ok, I could do TP_grid_log_space_distance.

src/Physics/Evaluate_friction_law.f90
```diff
@@ -1249,13 +1254,11 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
 
     !TU 7.07.16: if the SR is too close to zero, we will have problems (NaN)
     !as a consequence, the SR is affected the AlmostZero value when too small
-    AlmostZero = 1d-45
+    AlmostZero = 1d-8
```
Thomas-Ulrich: this should remain 1d-45, right?

src/Physics/Evaluate_friction_law.f90
```diff
@@ -1457,4 +1471,106 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
...
+    !-------------------------------------------------------------------------!
+    ! Argument list declaration
+    LOGICAL       :: has_converged                                            !check convergence
+    INTEGER       :: nSRupdates, i, nBndGP
+    REAL          :: RS_sr0, RS_a(nBndGP)                                     !constants
+    REAL          :: SRtest(nBndGP), LocSR(nBndGP), LocSV(nBndGP)
+    REAL          :: n_stress(nBndGP), sh_stress(nBndGP), invZ
+    REAL          :: NR(nBndGP), dNR(nBndGP), tmp(nBndGP), tmp2(nBndGP), tmp3(nBndGP)
+    REAL          :: mu_f(nBndGP), dmu_f(nBndGP)                              !calculated here in routine
+    REAL          :: AlmostZero = 1D-8, aTolF = 5e-14
```
Thomas-Ulrich: same here?

src/Physics/Evaluate_friction_law.f90
sebanger: those are the values that worked for TP. Alice also doubted why we need such a low value.

src/ResultWriter/FaultWriterExecutor.cpp
```diff
@@ -95,5 +95,5 @@ void seissol::writer::FaultWriterExecutor::execInit(const async::ExecInfo &info,
 }
 
 char const * const seissol::writer::FaultWriterExecutor::LABELS[] = {
-	"SRs", "SRd", "T_s", "T_d", "P_n", "u_n", "Mud", "StV", "Ts0", "Td0", "Pn0", "Sls", "Sld", "Vr", "ASl","PSR", "RT", "DS"
+	"SRs", "SRd", "T_s", "T_d", "P_n", "u_n", "Mud", "StV", "Ts0", "Td0", "Pn0", "Sls", "Sld", "Vr", "ASl","PSR", "RT", "DS", "P_f", "Tmp"
```
Thomas-Ulrich: please update the documentation. 
https://github.com/SeisSol/SeisSol/blob/master/Documentation/fault-output.rst

src/Physics/Evaluate_friction_law.f90
uphoffc: The question is rather why one needs an arbitrary value here which is not constrained by anything....

src/ResultWriter/FaultWriterExecutor.cpp
sebanger: is a new pull request

Thomas-Ulrich: ok

src/Physics/Evaluate_friction_law.f90
Thomas-Ulrich: I think this does not make sense and will lead to a slower code, because if the converged value of SR is let say 10e-20, and we reinitialize every time SR at 1e-8, then we loop over the 60 iterations. That's might be why the TP code is about 30% slower (I just tested on something close to tpv103, order3).
I think aTolF should be one or 2 order of magnitude larger than AlmostZero to avoid that.

Thomas-Ulrich: What about AlmostZero = 1D-12, aTolF = 1e-8
or AlmostZero = 5D-18, aTolF = 5e-14?
Can you test that and see if it works?

src/Physics/Evaluate_friction_law.f90
Thomas-Ulrich: still holds. Please remove from inputs of the subroutine.

src/Physics/Evaluate_friction_law.f90
Thomas-Ulrich: idem

src/Physics/Evaluate_friction_law.f90
Thomas-Ulrich: idem

Thomas-Ulrich:
Hi,
I think the branch is ready for being merged. Any comments?
Thomas.

Thomas-Ulrich:
For some reason, SlipRateOutputType = 1 gives unexpected results, whereas SlipRateOutputType = 0 works fine on the TPV105_test benchmark.... To be further investigated.

AliceGabriel:
Thomas, its great to have found this, could you be more specific (maybe
after the holidays) - what exactly is unexpected ?

Alice

krenzland:
@Thomas-Ulrich, @uphoffc Should we merge this branch or should we wait until we fix the "unexpected results"?

Thomas-Ulrich:
Hi all,

I found and fixed the bug in the output when using `SlipRateOutputType = 1`.
The problem was the strength was not accounting for the change in pressure due to TP.
Please review the 5 lasts commits before we can merge to master.

Thomas.

Thomas-Ulrich:
The forced update only changed this commit:
https://github.com/SeisSol/SeisSol/pull/91/commits/b104af6cab1b35b69ba41c2f9be2e6e65c4e3827
by removing the if statement instead of commenting.

Issue #116, opened 2019-09-03
"Did not find dynamic rupture surfaces"


Thomas-Ulrich:
Taufiq and I got the error on setups that were previously working.
Therefore, it seems that this commit does not do his job properly:
https://github.com/SeisSol/SeisSol/commit/bc00758f0fa0bc034f6eae6a9c0a38705264d82a
Was it properly tested?

Thomas.

krenzland:
Please provide an example of setups that fail and the error message.

Otherwise we cannot reproduce this issue.

uphoffc:
You should not need another example. Just run a dynamic rupture test-case on many nodes until one node does not get dynamic rupture faces from the graph partitioner.

krenzland:
Carsten, the fact that it occurs on many nodes is an information that is not provided by the bug report...
Generally it's good practice to attach information about the environment and situations that are not working.

Thomas-Ulrich:
Sorry I was a bit lazy. But it seems the origin of the bug is known now ;-)

sebwolf-de:
This was tested, but only on small simulations. We thought this error message might be helpful. Now it seems that it breaks more things, than helps. I will remove it again.

sebwolf-de:
The originally intended functionality seems to be harder to implement, and we might not even need it.

sebwolf-de:
https://github.com/SeisSol/SeisSol/tree/dynrup-hotfix
Does your simulation run on this branch?

Thomas-Ulrich:
yes removing out the 2 lines fix the problem.

PR #117, opened 2019-09-03
Remove the error "Did not find dynamic rupture surfaces"

sebwolf-de:
as it causes too much trouble.

This has been tested for the TPV12 scenario.


sebwolf-de:
fix #116

PR #120, opened 2019-09-06
add strike convention for flat faults

Thomas-Ulrich:
defines a convention strike orientation for flat faults.
This may lead to SRd incoherently switching sign, but then at least the SR magnitude can be outputted.

### DIFF
--- a/src/Numerical_aux/create_fault_rotationmatrix.f90
+++ b/src/Numerical_aux/create_fault_rotationmatrix.f90
@@ -90,10 +90,14 @@ CONTAINS
     n = MESH%Fault%geoNormals(1:3,iFace)
     ! s = MESH%Fault%geoTangent1(1:3,iFace)
     ! t = MESH%Fault%geoTangent2(1:3,iFace)
-
     strike_vector(1) = n(2)/sqrt(n(1)**2+n(2)**2)
     strike_vector(2) = -n(1)/sqrt(n(1)**2+n(2)**2)
     strike_vector(3) = 0.0D0
+    
+    !convention for flat faults
+    if (norm2(strike_vector).eq.0) then
+       strike_vector(1)=1d0
+    endif
    
     dip_vector(1) = -strike_vector(2)*n(3)
     dip_vector(2) = strike_vector(1)*n(3)


src/Numerical_aux/create_fault_rotationmatrix.f90
```diff
@@ -90,10 +90,14 @@ SUBROUTINE create_fault_rotationmatrix(rotmat,iFace,EQN,MESH,iRotmat)
     n = MESH%Fault%geoNormals(1:3,iFace)
     ! s = MESH%Fault%geoTangent1(1:3,iFace)
     ! t = MESH%Fault%geoTangent2(1:3,iFace)
-
     strike_vector(1) = n(2)/sqrt(n(1)**2+n(2)**2)
     strike_vector(2) = -n(1)/sqrt(n(1)**2+n(2)**2)
     strike_vector(3) = 0.0D0
+    
+    !convention for flat faults
+    if (norm2(strike_vector).eq.0) then
```
krenzland: Does this actually work in practice? I'd expect something like `(norm2(strike_vector) < eps)`, where eps is some small number

Thomas-Ulrich: > Looks generally fine to me; not sure whether this convention makes sense - but this is probably something that you guys have to decide!

The problem is that the strike direction is not constrained for a strike fault.
So I guess the best would be to use the direction of a neighbor fault element where the strike is constrained to define the strike orientation for the flat fault element.
But this is seems complicated to implement.
So the best is to define such an arbitrary convention to have an output, and then only visualise the absolute SR or absolute slip if part of the fault is flat.

> Does this actually work in practice? I'd expect something like (norm2(strike_vector) < eps), where eps is some small number

I think 0 can be accurately represented by a float or a double so there is not need to do norm2<eps.

uphoffc:
I don't really like this solution, as I wonder how the user is supposed to know whether the convention applies or whether it does not. In particular, surprises may appear when the norm of the strike vector is not exactly zero. That is, the normal (1e-15, 0, 1) would give the strike vector (0, -1, 0) which is different from the strike vector (1,0,0) with normal (0,0,1).
Note that (1e-15, 0, 1) is normalized as (1e-15)^2 + 1^2 = 1 in double precision...

Looking at the neighbouring elements would be more sophisticated, but also not universally applicable. Just think of a paraboloid.

Maybe the definition just sucks in this case...

krenzland:
What's the status of this pull request? Is it okay to close it or do you need this for discussions?

Thomas-Ulrich:
ok let's close it for now

PR #124, opened 2019-09-17
Initial support for CMake builds

krenzland:
This pull request implements an initial support for building SeisSol with CMake instead of Scons.
While not everything works perfectly, compiling SeisSol with Intel Compiler and HDF5/NetCDF/Asagi/ParMETIS works locally and on NG.
Right now I do not want to offer support for CMake but this is probably going to replace Scons in the  future, for various reasons (including better IDE-support,  support for tools like clang-tidy out of the box, easier maintainability, quicker compilation, etc)
I want to merge this right now because I want to use this for daily development - this way I can see whether I find any issues for practical usage.
This guarantees that the end-users will get a **stable** build tool that was **extensively tested**.

I will later provide support for other compiler, other libraries, documentation, etc.
Right now, nothing is going to change for the users. Scons is still the **only** supported build tool.


uphoffc:
Please add a message in the CMakeLists.txt that CMake is experimental and not supported.

krenzland:
> Please add a message in the CMakeLists.txt that CMake is experimental and not supported.

Good idea, added this.

krenzland:
> Please clean up the memlayout.py and arch.py double occurances in site_scons and generated_code
Fixed. The dir generated_code is now a python package and exports arch. Not the best software design but good enough for this purpose.

uphoffc:
How do you distinguish whether the parallel or the serial version of netcdf and hdf5 is required?

krenzland:
I don't ;)
This wasn't a problem thus far.

Anyway, I'll fix this; see issue #125.

PR #138, opened 2019-10-09
Carsten/viscoelastic

uphoffc:
I repaired the matrix-based implementation of viscoelastic attenuation (for comparison with viscoelastic2).

The implementation itself is likely unnecessary, however it is a prototype for aderdg schemes requiring a source term. Most of the kernel code is shared with the elastic implementation.


Thomas-Ulrich:
Could you maybe add a line in:
https://seissol.readthedocs.io/en/latest/attenuation.html
describing such alternative implementation?
E.g. what is the difference (not in details), when to use, which one should be more robust, faster?

generated_code/aderdg.py
```diff
@@ -131,3 +151,57 @@ def addNeighbor(self, generator):
   @abstractmethod
   def addTime(self, generator):
     pass
+
+class ADERDGStandard(ADERDGBase):
```
krenzland: Name is confusing but I've got no better suggestion.
Maybe we should provide a comment for this class that explains its purpose (e.g. linear hyperbolic cons. law of form [...])

generated_code/aderdg.py
```diff
@@ -131,3 +151,57 @@ def addNeighbor(self, generator):
...
+  def sourceMatrix(self):
+    return None
```
krenzland: Is there a reason why this isn't in ADERDGBase?

generated_code/elastic.py
```diff
@@ -62,51 +57,5 @@ def numberOfQuantities(self):
   def numberOfExtendedQuantities(self):
     return self.numberOfQuantities()
```
krenzland: Why is this needed/why isn't this in a super class?

generated_code/viscoelastic.py
```diff
@@ -1,12 +1,12 @@
...
+#!/usr/bin/env python3
 ##
 # @file
 # This file is part of SeisSol.
 #
 # @author Carsten Uphoff (c.uphoff AT tum.de, http://www5.in.tum.de/wiki/index.php/Carsten_Uphoff,_M.Sc.)
 #
 # @section LICENSE
-# Copyright (c) 2015, SeisSol Group
+# Copyright (c) 2016-2018, SeisSol Group
```
krenzland: 2019 ;)

generated_code/viscoelastic.py
```diff
@@ -37,104 +37,75 @@
...
+      'star': ['star(0)', 'star(1)', 'star(2)'],
+    }
+    self.db.update( parseXMLMatrixFile('{}/matrices_viscoelastic.xml'.format(matricesDir), clones) )
+
+    star_spp = self.db.star[0].spp().as_ndarray()
+    star_rows, star_cols = star_spp.shape
+    aniso_cols = star_cols - self.numberOfElasticQuantities
+    star_spp_new = np.zeros((self.numberOfQuantities(), self.numberOfQuantities()), dtype=bool)
+    star_spp_new[0:star_rows,0:star_cols] = star_spp
+    for mech in range(1,numberOfMechanisms):
```
krenzland: Does this what you expect? range(1, 6) for example results in [1, 2, 3, 4, 5].

generated_code/viscoelastic.py
```diff
@@ -37,104 +37,75 @@
...
+      offset0 = self.numberOfElasticQuantities
+      offsetm = self.numberOfElasticQuantities + mech*aniso_cols
+      star_spp_new[0:star_rows,offsetm:offsetm+aniso_cols] = star_spp[0:star_rows,offset0:offset0+aniso_cols]
```
krenzland: Please add a comment here. I'm not sure why you're building the matrix in this way.

generated_code/viscoelastic.py
```diff
@@ -37,104 +37,75 @@
...
+    for dim in range(3):
+      self.db.star[dim] = Tensor(self.db.star[dim].name(), star_spp_new.shape, spp=star_spp_new)
+
+    source_spp = np.zeros((self.numberOfQuantities(), self.numberOfQuantities()), dtype=bool)
+    ET_spp = self.db['ET'].spp().as_ndarray()
+    for mech in range(numberOfMechanisms):
+      offset = self.numberOfElasticQuantities + mech*aniso_cols
+      r = slice(offset, offset+aniso_cols)
+      source_spp[r,0:aniso_cols] = ET_spp
+      source_spp[r,r] = np.identity(aniso_cols, dtype=bool)
```
krenzland: Also add a small comment here that explains the matrix structure.

src/Equations/elastic/Kernels/Local.cpp
```diff
@@ -52,6 +52,10 @@
 #include <stdint.h>
 #include <cstring>
 
+#include <Kernels/common.hpp>
+GENERATE_HAS_MEMBER(ET)
+GENERATE_HAS_MEMBER(sourceMatrix)
```
krenzland: This macro is also called in Time kernel - maybe we should do this in one place to avoid duplicate code generation (minor issue)

src/Equations/viscoelastic/Model/Setup.cpp
```diff
@@ -74,67 +75,91 @@ void getTransposedViscoelasticCoefficientMatrix( real            i_omega,
...
+void getTransposedSourceCoefficientTensor( seissol::model::Material const& material, T& sourceMatrix )
+{
+  sourceMatrix.setZero();
+
+  //       | E_1^T |
+  // E^T = |  ...  |
+  //       | E_L^T |
+  for (unsigned mech = 0; mech < NUMBER_OF_RELAXATION_MECHANISMS; ++mech) {
+    unsigned offset = 9 + mech * 6;
+    real const* theta = material.theta[mech];
```
krenzland: const real* is preferable. (I know your reasoning but imo "everyone does it" is a good reason in this case to prefer the other way)

src/Equations/viscoelastic/Model/Setup.cpp
```diff
@@ -159,58 +184,30 @@ void seissol::model::setMaterial( double* i_materialVal,
...
+  seissol::transformations::symmetricTensor2RotationMatrix(i_normal, i_tangent1, i_tangent2, o_T);
+  seissol::transformations::inverseSymmetricTensor2RotationMatrix(i_normal, i_tangent1, i_tangent2, o_Tinv);
   
-  seissol::transformations::tensor1RotationMatrix(i_normal, i_tangent1, i_tangent2, o_T.block<3,3>(6, 6));
-  seissol::transformations::inverseTensor1RotationMatrix(i_normal, i_tangent1, i_tangent2, o_Tinv.block<3,3>(6, 6));
+  seissol::transformations::tensor1RotationMatrix(i_normal, i_tangent1, i_tangent2, o_T, 6, 6);
+  seissol::transformations::inverseTensor1RotationMatrix(i_normal, i_tangent1, i_tangent2, o_Tinv, 6, 6);
   
   for (unsigned mech = 0; mech < NUMBER_OF_RELAXATION_MECHANISMS; ++mech) {
     unsigned origin = 9 + mech * 6;
```
krenzland: should be const

src/Model/common.hpp
```diff
@@ -149,12 +149,12 @@ void seissol::model::getTransposedElasticGodunovState( Material const&
...
-  for (unsigned i = 0; i < QgodLocal.shape(1); ++i) {
-    for (unsigned j = 0; j < QgodLocal.shape(0); ++j) {
+  for (unsigned i = 0; i < 9; ++i) {
+    for (unsigned j = 0; j < 9; ++j) {
       QgodLocal(i,j) = -QgodNeighbor(i,j);
     }
   }  
-  for (unsigned idx = 0; idx < QgodLocal.shape(0) && idx < QgodLocal.shape(1); ++idx) {
+  for (unsigned idx = 0; idx < 9; ++idx) {
     QgodLocal(idx,idx) += 1.0;
```
krenzland: Why is the size now hardcoded?

auto_tuning/proxy/SConstruct
```diff
@@ -168,7 +167,7 @@ env.Append(CXXFLAGS=['-std=c++11'])
 if env['compiler'] == 'gcc':
   env.Append(CXXFLAGS = ['-Werror=vla'])
 
-if env['equations'] in ['elastic', 'viscoelastic2']:
+if env['equations'] in ['elastic', 'viscoelastic', 'viscoelastic2']:
   env.Append(CPPDEFINES=['ENABLE_MATRIX_PREFETCH'])
```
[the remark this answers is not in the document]
uphoffc: Dunno, mostly historical reasons. Shall I delete it?

generated_code/aderdg.py
```diff
@@ -66,15 +69,17 @@ def __init__(self, order, multipleSimulations, matricesDir):
     self.Q = OptionalDimTensor('Q', 's', multipleSimulations, 0, qShape, alignStride=True)
     self.I = OptionalDimTensor('I', 's', multipleSimulations, 0, qShape, alignStride=True)
 
-    Ashape = (self.numberOfQuantities(), self.numberOfExtendedQuantities())
-    self.AplusT = Tensor('AplusT', Ashape)
-    self.AminusT = Tensor('AminusT', Ashape)
+    Apm_spp = self.flux_solver_spp()
```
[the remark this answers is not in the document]
uphoffc: Aplus or Aminus

generated_code/aderdg.py
uphoffc: Is there a reason why it should be in ADERDGBase?

generated_code/elastic.py
uphoffc: Yes we can move that to ADERDGStandard.

src/Model/common.hpp
uphoffc: Because shape is independent of the memory layout and is too large for viscoelastic.
I thought about it and there is no good reason to not hard-code the 9. (The 9 is already hard-coded due to QgodNeighbor(8,8).)

auto_tuning/proxy/SConstruct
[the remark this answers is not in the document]
krenzland: I guess - I've checked and it's used only a few times.

generated_code/aderdg.py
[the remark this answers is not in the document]
krenzland: Can we maybe call it A_spp? This is a bit clearer for me than Apm

generated_code/aderdg.py
krenzland: No, you're right.

src/Model/common.hpp
krenzland: Okay, fine.

generated_code/aderdg.py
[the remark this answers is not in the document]
uphoffc: No, that's different.

generated_code/aderdg.py
uphoffc: How about ADERDGLinear or LinearADERDG?

generated_code/viscoelastic.py
uphoffc: Yes, basically mech 0 is repeated several times.

generated_code/aderdg.py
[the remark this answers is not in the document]
krenzland: True. Still, I'd prefer a more descriptive name here.

generated_code/aderdg.py
krenzland: Sounds good!

generated_code/viscoelastic.py
uphoffc: Satisfied?

src/Equations/viscoelastic2/Physics/InitialField.cpp
```diff
@@ -28,11 +29,11 @@ seissol::physics::Planarwave::Planarwave(real phase)
...
   e_interoperability.fitAttenuation(rho, mu, lambda, Qp, Qs, material);
 
-  std::complex<real> planeWaveOperator[NUMBER_OF_QUANTITIES*NUMBER_OF_QUANTITIES];
+  std::complex<double> planeWaveOperator[NUMBER_OF_QUANTITIES*NUMBER_OF_QUANTITIES];
   model::getPlaneWaveOperator(material, m_kVec.data(), planeWaveOperator);
 
-  using Matrix = Eigen::Matrix<std::complex<real>, NUMBER_OF_QUANTITIES, NUMBER_OF_QUANTITIES, Eigen::ColMajor>;
-  using Vector = Eigen::Matrix<std::complex<real>, NUMBER_OF_QUANTITIES, 1, Eigen::ColMajor>;
+  using Matrix = Eigen::Matrix<std::complex<double>, NUMBER_OF_QUANTITIES, NUMBER_OF_QUANTITIES, Eigen::ColMajor>;
+  using Vector = Eigen::Matrix<std::complex<double>, NUMBER_OF_QUANTITIES, 1, Eigen::ColMajor>;
```
sebwolf-de: Typedef vector is locally defined but not used.

PR #146, opened 2019-10-15
Skip mini SeisSol when using one MPI rank.

krenzland:


src/Geometry/MeshReaderFBinding.cpp
```diff
@@ -291,16 +291,20 @@ void read_mesh_puml_c(const char* meshfile, const char* checkPointFile, bool has
...
+
+	  const auto summary = seissol::statistics::parallelSummary(tpwgt);
+	  logInfo(rank) << "Node weights: mean =" << summary.mean
+			<< " std =" << summary.std
+			<< " min =" << summary.min
+			<< " median =" << summary.median
+			<< " max =" << summary.max;
+	} else {
+	  logInfo(rank) << "Only running on one rank, using node weight of 1.";
+	}
```
uphoffc: Isn't this info just noise?

PR #148, opened 2019-10-21
Use OpenMP for AnalysisWriter.

krenzland:


krenzland:
Good point. But: the problem is that I also want to print out the position of the maximum error; this doesn't work with the reduction clause. I could theoretically do some parts with reduction and some without but I think doing everything in the same way is a bit more neat. It's not performance critical anyway. 

Also as far as I know, array reductions are only supported by GCC6+ (not sure)

src/ResultWriter/AnalysisWriter.cpp
```diff
@@ -62,14 +53,43 @@ void seissol::writer::AnalysisWriter::printAnalysis(double simulationTime) {
...
 
     // Note: We iterate over mesh cells by id to avoid
     // cells that are duplicates.
-    for (unsigned int meshId = 0; meshId < elements.size(); ++meshId) {
+    std::vector<std::array<double, 3>> quadraturePointsXyz(numQuadPoints);
+
+    alignas(ALIGNMENT) real numericalSolutionData[tensor::dofsQP::size()];
+    alignas(ALIGNMENT) real analyticalSolutionData[numQuadPoints*numberOfQuantities];
+#ifdef _OPENMP
+#pragma omp parallel for firstprivate(quadraturePointsXyz, numericalSolutionData, analyticalSolutionData)
```
uphoffc: Why firstprivate for numericalSolutionData and analyticalSolutionData?

krenzland: No reason. Fixed.

PR #157, opened 2019-11-04
Sebastian/improve-bc

sebwolf-de:
We found an easier way to implement the free surface boundary. The neighbor kernel can be omitted for the free surface to safe some (not a lot) computation time.
I tested the new implementation with the LOH3 benchmark and get similar results as with the old implementation.


src/Equations/elastic/Kernels/Neighbor.cpp
```diff
@@ -140,14 +140,7 @@ void seissol::kernels::Neighbor::computeNeighborsIntegral(  NeighborData&
...
         nfKrnl.execute(data.cellInformation.faceRelations[l_face][1], data.cellInformation.faceRelations[l_face][0], l_face);
-      } else { // fall back to local matrices in case of free surface boundary conditions
-        kernel::localFlux lfKrnl = m_lfKrnlPrototype;
-        lfKrnl.Q = data.dofs;
-        lfKrnl.I = i_timeIntegrated[l_face];
-        lfKrnl.AplusT = data.neighboringIntegration.nAmNm1[l_face];
-        lfKrnl._prefetch.I = faceNeighbors_prefetch[l_face];
-        lfKrnl.execute(l_face);
-      }
+      }
```
uphoffc: Please also remove m_lfKrnlPrototype in NeighborBase.h and Neighbor::setGlobalData.

src/Equations/viscoelastic2/Kernels/Neighbor.cpp
```diff
@@ -114,15 +114,7 @@ void seissol::kernels::Neighbor::computeNeighborsIntegral(  NeighborData&
...
-      } else { // fall back to local matrices in case of free surface boundary conditions
-        kernel::localFluxExt lfKrnl = m_lfKrnlPrototype;
-        lfKrnl.Qext = Qext;
-        lfKrnl.I = i_timeIntegrated[l_face];
-        lfKrnl.AplusT = data.neighboringIntegration.nAmNm1[l_face];
-        lfKrnl._prefetch.I = faceNeighbors_prefetch[l_face];
-        lfKrnl._prefetch.Q = faceNeighbors_prefetch[l_face];
-        lfKrnl.execute(l_face);
-      }
+      }
```
uphoffc: Please also remove m_lfKrnlPrototype in NeighborBase.h and Neighbor::setGlobalData.

src/Model/common.hpp
```diff
@@ -158,22 +158,26 @@ void seissol::model::getTransposedElasticGodunovState( Material const&
     QgodLocal(idx,idx) += 1.0;
   }
   
-  applyBoundaryConditionToElasticFluxSolver(faceType, QgodNeighbor);
+  applyBoundaryConditionToElasticFluxSolver(faceType, QgodLocal);
```
uphoffc: Should we set QGodNeighbor to zero or NaN? In the first case, the solver would still work if QGodNeighbor is used by accident, in the second case, it would immediately fail such that the cause can be (quickly?) found?

src/Model/common.hpp
```diff
@@ -158,22 +158,26 @@ void seissol::model::getTransposedElasticGodunovState( Material const&
...
-    // Gamma^T * Fneighbor
-    for (unsigned j = 0; j < QgodNeighbor.shape(1); ++j) {
-      for (unsigned i = 0; i < 9; ++i) {
-        QgodNeighbor(i,j) *= Gamma[i];
+    //set traction at interface to zero
+    int traction_indices[3] = {0, 3, 5};
+    int other_indices[6] =  {1, 2, 5, 6, 7, 8};
+    for (int i = 0; i < 9; i++) {
+      for (int j = 0; j < 3; j++) {
+        QgodLocal(i,traction_indices[j]) = 0;
```
uphoffc: Why do you scale the columns? I think you need to scale the rows.

src/Model/common.hpp
```diff
@@ -158,22 +158,26 @@ void seissol::model::getTransposedElasticGodunovState( Material const&
...
+      }
+      for (int j = 0; j < 6; j++) {
+        QgodLocal(i, other_indices[j]) = 2* QgodLocal(i, other_indices[j]);
```
uphoffc: Same here

src/Model/common.hpp
```diff
@@ -158,22 +158,26 @@ void seissol::model::getTransposedElasticGodunovState( Material const&
...
 {
   if (type == freeSurface) {
-    // Gamma is a diagonal matrix
-    real Gamma[] = { -1.0, 1.0, 1.0, -1.0, 1.0, -1.0, 1.0, 1.0, 1.0 };
-    // Gamma^T * Fneighbor
-    for (unsigned j = 0; j < QgodNeighbor.shape(1); ++j) {
-      for (unsigned i = 0; i < 9; ++i) {
-        QgodNeighbor(i,j) *= Gamma[i];
+    //set traction at interface to zero
+    int traction_indices[3] = {0, 3, 5};
```
uphoffc: I'd rather use std::array.

src/Model/common.hpp
```diff
@@ -158,22 +158,26 @@ void seissol::model::getTransposedElasticGodunovState( Material const&
...
+    int other_indices[6] =  {1, 2, 5, 6, 7, 8};
+    for (int i = 0; i < 9; i++) {
+      for (int j = 0; j < 3; j++) {
```
uphoffc: If you use std::array you may write for(auto& j : traction_indices)
Saves you hardcoding "3", which is a potential error when this code is modified.

src/Model/common.hpp
krenzland: I'd vote for signaling NaNs here. You can activate floating point exceptions and then SeisSol would crash immediately if it would be used by accident.

src/Model/common.hpp
krenzland: Also works for normal arrays, as std::begin and std::end is implemented for c-style arrays with known size.

src/Model/common.hpp
sebwolf-de: First I was also scaling the rows - without success. QgodLocal is actually the transpose of R*chi*R^-1.

src/Model/common.hpp
```diff
@@ -57,10 +63,12 @@ namespace seissol {
                                            enum ::faceType                      faceType,
                                            Tloc&                                QgodLocal,
                                            Tneigh&                              QgodNeighbor );
-
+    
     template<typename T>
-    void applyBoundaryConditionToElasticFluxSolver( enum ::faceType type,
-                                                    T&              QgodNeighbor );
+    void getTransposedBoundaryGodunovState( T& QgodLocal,
```
uphoffc: Wouldn't it make sense if free surface appears in the function name?

src/Model/common.hpp
```diff
@@ -114,65 +122,121 @@ void seissol::model::getTransposedElasticCoefficientMatrix( seissol::model::Elas
...
-  for (unsigned idx = 0; idx < 9; ++idx) {
-    QgodLocal(idx,idx) += 1.0;
   }
-  
-  applyBoundaryConditionToElasticFluxSolver(faceType, QgodNeighbor);
+
+  QgodLocal.setZero();
+  std::array<std::array<int, 2>, 3> traction_indices = {{{0,0}, {1,3}, {2,5}}};
+  std::array<std::array<int, 2>, 3> velocity_indices = {{{0,6}, {1,7}, {2,8}}};
+  //Eigen does not have fancy slicing options :(
```
uphoffc: ?? https://eigen.tuxfamily.org/dox-devel/group__TutorialSlicingIndexing.html

uphoffc: See "Array of indices"

src/Model/common.hpp
```diff
@@ -114,65 +122,121 @@ void seissol::model::getTransposedElasticCoefficientMatrix( seissol::model::Elas
...
+  R(7,7) = -sqrt(neighbor.mu/neighbor.rho);
+  R(8,2) = sqrt(local.mu/local.rho);
+  R(8,6) = -sqrt(neighbor.mu/neighbor.rho);
+
+  if(faceType == freeSurface) {
+    getTransposedBoundaryGodunovState(QgodLocal, QgodNeighbor, R);
+
+  } else {
+    Matrix99 R_inv = Matrix99::Zero();
+    R_inv(0,0) = 1/(local.lambda + 2*local.mu) - (neighbor.lambda + 2*neighbor.mu)*sqrt((local.lambda + 2*local.mu)/local.rho)/((local.lambda + 2*local.mu)*(local.lambda + 2*local.mu)*((neighbor.lambda + 2*neighbor.mu)*sqrt((local.lambda + 2*local.mu)/local.rho)/(local.lambda + 2*local.mu) + sqrt((neighbor.lambda + 2*neighbor.mu)/neighbor.rho)));
```
uphoffc: Is this generated code? I'm just wondering whether it would be more readable if one codes Rinv by inverting the three independent 2x2 matrices.

src/Model/common.hpp
```diff
@@ -114,65 +122,121 @@ void seissol::model::getTransposedElasticCoefficientMatrix( seissol::model::Elas
...
+
+    Matrix99 chi = Matrix99::Zero();
+    chi(0,0) = 1.0;
+    chi(1,1) = 1.0;
+    chi(2,2) = 1.0;
+
+    auto godunov = ((R*chi)*R_inv).eval();
+
+    // QgodLocal = I - QgodNeighbor
+    for (unsigned i = 0; i < QgodLocal.shape(1); ++i) {
```
uphoffc: Please write 9 or godunov.cols() or something like that for indices i and j. The code is going to fail whenever Qgodlocals shape is larger than 9 (e.g. for viscoelastic).

sebwolf-de:
I've rewritten the Godunov State and Free Surface BC to make it easier to read and better extensible to new models like anisotropy and poroelasticity (yet to come)

src/tests/Model/GodunovState.t.h
```diff
@@ -0,0 +1,193 @@
...
+ **/
+#include <cxxtest/TestSuite.h>
+#include <cassert>
+#include <cmath>
+
+#include <Model/common.hpp>
+#include <Model/Setup.h>
+#include <Model/common_datastructures.hpp>
+
+#if defined(DOUBLE_PRECISION)
```
uphoffc: You could also use std::numeric_limits<real>::epsilon() instead of these defines

src/Model/common.hpp
sebwolf-de: Thanks for pointing it out. Apparently the git mirror provides version 3.3.90 although the eigen webpage states, that 3.3.7 is the latest release.

src/Model/common.hpp
```diff
@@ -40,12 +40,18 @@
 #ifndef MODEL_COMMON_HPP_
 #define MODEL_COMMON_HPP_
 
+#include <Eigen/Eigen>
 #include <cmath>
 #include <Initializer/typedefs.hpp>
 #include <generated_code/init.h>
+#include <Geometry/MeshDefinition.h>
+#include <Numerical_aux/Transformation.h>
+#include <iostream>
```
krenzland: unused include?

src/Model/common.hpp
```diff
@@ -114,65 +122,114 @@ void seissol::model::getTransposedElasticCoefficientMatrix( seissol::model::Elas
...
+    R_inv(1,7) = inv_01(local.mu, neighbor.mu, sqrt(local.mu/local.rho), sqrt(neighbor.mu/neighbor.rho));
+    R_inv(2,5) = inv_00(local.mu, neighbor.mu, sqrt(local.mu/local.rho), sqrt(neighbor.mu/neighbor.rho));
+    R_inv(2,8) = inv_01(local.mu, neighbor.mu, sqrt(local.mu/local.rho), sqrt(neighbor.mu/neighbor.rho));
+
+    Matrix99 chi = Matrix99::Zero();
+    chi(0,0) = 1.0;
+    chi(1,1) = 1.0;
+    chi(2,2) = 1.0;
+
+    auto godunov = ((R*chi)*R_inv).eval();
```
krenzland: should be const auto?

src/tests/Model/GodunovState.t.h
```diff
@@ -0,0 +1,193 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ **/
```
krenzland: Why did you copy this useless header? Authors can be found in git, license is not needed per file

src/tests/Model/GodunovState.t.h
```diff
@@ -0,0 +1,193 @@
...
+            test_relative(QgodNeighbor(i,j), solution[j][i]);
+        }
+      }
+    }
+
+    void test_nan(init::QgodLocal::view::type QgodNeighbor) {
+      for (int i = 0; i < 9; i++) {
+        for (int j = 0; j < 9; j++) {
+          //CXXTEST TS_ASSERT_IS_NAN is broken
+          assert(std::isnan(QgodNeighbor(i,j)));
```
krenzland: Replace assert with cxxassert, otherwise this would crash the test when it fails

src/tests/Model/SConscript
```diff
@@ -0,0 +1,45 @@
...
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+
```
krenzland: The header is ~10 times longer than the source code...

src/Model/common.hpp
uphoffc: auto const ;-P

PR #158, opened 2019-11-05
Update of cmake

ravil-mobile:
Processing the users' input, users' input error checking and compiler flag settings are moved to separate files. This is supposed to provide some consistency between Seissol and Seissol-proxy cmake files


CMakeLists.txt
```diff
@@ -78,11 +58,11 @@ add_custom_command(
      "--equations" ${EQUATIONS}
      "--matricesDir" ${CMAKE_CURRENT_SOURCE_DIR}/generated_code/matrices
      "--outputDir" ${CMAKE_CURRENT_BINARY_DIR}/src/generated_code
-     "--arch" ${ARCH}
+     "--arch" ${ARCH_STRING}
      "--order" ${ORDER}
      "--numberOfMechanisms" ${NUMBER_OF_MECHANISMS}
-     "--memLayout" "auto"
-     "--multipleSimulations" ${NUMBER_OF_MULTIPLE_SIMULATIONS}
+     "--memLayout" ${MEMEORY_LAYOUT}
```
krenzland: MEMORY_LAYOUT is misspelled here

CMakeLists.txt
```diff
@@ -1,35 +1,25 @@
...
+#                     PLASTICITY, NUMBER_OF_FUSED_SIMULATIONS,
+#                     MEMEORY_LAYOUT, COMMTHREAD, LOG_LEVEL,
+#                     ACCELERATOR_TYPE
+#
+#       derived input: REAL_SIZE_IN_BYTES, ALIGNMENT, ARCH_STRING, NUMBER_OF_QUANTITIES
+#
+include(cmake/process_users_input.cmake)
+set(HARDWARE_SPECIFIC_DEFINITIONS "ALIGNMENT=${ALIGNMENT}"
+                                  "REAL_SIZE=${REAL_SIZE_IN_BYTES}"
+                                  "ALIGNED_REAL_SIZE=${REAL_SIZE_IN_BYTES}")
```
krenzland: Maybe rename this to HARDWARE_DEFINITIONS as this is shorter and just as clear?

cmake/cpu_arch_flags.cmake
```diff
@@ -0,0 +1,62 @@
...
+
+        if (compiler STREQUAL "Intel")
+            set(CPU_ARCH_SPECIFIC_FLAGS "-xMIC-AVX512" "-fma" PARENT_SCOPE)
+        elseif(compiler STREQUAL "GNU")
+            set(CPU_ARCH_SPECIFIC_FLAGS "-march=skylake-avx512" PARENT_SCOPE)
+        endif()
+    
+    endif()
+
+endfunction()
```
krenzland: Can you please remove some of the white space here?
I'm not that happy with all these if statements - it's a bit hard to read. Especially, when we might add more compilers at some point.
Maybe this could be rewritten as a generator expression. Not sure if that's cleaner though...
Another option could be switching the if statements, i.e. first check compilers and then check the architecture?

krenzland: Also maybe we can rename the function to get_arch_flags?

cmake/process_users_input.cmake
```diff
@@ -0,0 +1,143 @@
...
+
+
+set(EQUATIONS "elastic" CACHE STRING "Equation set used")
+set(EQUATIONS_OPTIONS elastic viscoelastic viscoelastic2)
+set_property(CACHE EQUATIONS PROPERTY STRINGS ${EQUATIONS_OPTIONS})
+
+
+set(ARCH "hsw" CACHE STRING "Type of the target architecture")
+set(ARCH_OPTIONS noarch wsm snb hsw knc knl skx)
+set(ARCH_ALIGNMENT   16  16  32  32  64  64  64)  # size of a vector registers in bytes for a given architecture
```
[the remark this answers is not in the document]
ravil-mobile: I guess before we used a lookup table where names were corresponded with alignment (in bytes). I think it is better to keep this info as close as possible. It will be easy for an extension.  For example, as soon as you decide to add a new cpu architecture you know that you have to manually add its alignment and you will be able to see where you have to add it (just next line)

Documentation/compilation.rst
```diff
@@ -87,9 +87,25 @@ Installing Libxsmm
...
+ 
+   git clone https://github.com/uphoffc/ImpalaJIT.git
+   cd ImpalaJIT
+   mkdir build
+   cd build 
+   cmake .. -DCMAKE_INSTALL_PREFIX=$HOME
+   make
+   make install
+
+
```
krenzland: We should move the cmake config to a separate file. Could you please start this (with a huge warning DO NOT USE)? I would then write the rest of the documentation :)

cmake/FindMKL.cmake
```diff
@@ -0,0 +1,122 @@
...
+#
+################################################################################
+# NOTE: this file was originally developer by:
+#
+# \file      cmake/FindMKL.cmake
+# \author    J. Bakosi
+# \copyright 2012-2015, Jozsef Bakosi, 2016, Los Alamos National Security, LLC.
+# \brief     Find the Math Kernel Library from Intel
+# \date      Thu 26 Jan 2017 02:05:50 PM MST
+#
```
krenzland: Can we include this? What's the license of the original file?

ravil-mobile: The author said that it was BSD-3 license

PR #163, opened 2019-11-17
Create corner_frequency

NicoSchlw:


uphoffc:
Shouldn't this file have the ending .py?

krenzland:
> Shouldn't this file have the ending .py?

And/or also a shebang like `#!/usr/bin/env python3`

NicoSchlw:
I'm sorry, I hope it is right now

PR #164, opened 2019-11-18
fix vectorization in RSF103

daisy20170101:
I changed the vectorization for RSF variables into a loop over nBndGP, so that RSF variables become scalar. The changes are only in src/Physics/Evaluate_friction_law.f90

I have a test scenario on NG: /hppfs/work/pr63qo/di52lak2/Mexico/Launch_SeisSol/Test
In this test, vectorization RSF fails with NaN but scalar RSF works 

Hope this could help with your bugs @sebanger


src/Physics/Evaluate_friction_law.f90
```diff
@@ -1300,31 +1302,33 @@ SUBROUTINE rate_and_state_nuc103(TractionGP_XY,TractionGP_XZ,        & ! OUT: tr
          !                                      mu_ss = mu_w + [mu_lv - mu_w] / [ 1 + (V/Vw)^8 ] ^ (1/8) ]
          !                                      mu_lv = mu_0 - (b-a) ln (V/V0)
          !
-         LocP   = NorStressGP(:,iTimeGP)
+         LocP   = 0.8*NorStressGP(iBndGP,iTimeGP)
```
krenzland: Where does this factor come from?

daisy20170101: The title may be misleading. 
It is just for Sebastian and our group to test whether it fix his bugs. 
It doesn't need to be merged. 

The factor can be moved.

PR #165, opened 2019-11-18
update openquake installation

daisy20170101:
update installation instruction


uphoffc:
How does this fit with the other parts of the README? I think the gmpe-smtk version does not use OpenQuake? (git checkout 4f008173e89f6e4ba4450fb43e95ffdf51a7c2ba^)

daisy20170101:
It is required as I run it, e.g.

sm_database.py:from openquake.hazardlib.gsim.base import (SitesContext, DistancesContext,
sm_database.py:from openquake.hazardlib.site import Site, SiteCollection
sm_database.py:from openquake.hazardlib.geo.point import Point
sm_database.py:from openquake.hazardlib.scalerel import PeerMSR

Thomas-Ulrich:
Hi, 
Carsten is right, the installation process is currently such as you don't need to install openquake, but you only rely on some routines of gmpe-smtk, which did not depend no openquake until recently.
Therefore, an outdated version of gmpe-smtk is picked using ```git checkout 4f008173e89f6e4ba4450fb43e95ffdf51a7c2ba^```

daisy20170101:
OK, I see. Then should I close it?

Thomas-Ulrich:
I think you can either write an alternative installation procedure which relies on the installation of openquake (which might be better, it is not very good to rely on the deprecated gmpe-smtk), or you can close the issue.
In any case, I would not remove the current installation procedure, as it might be difficult to install openquake on HPC servers, and the current installation procedure does not require to install anything.
I would write that as an alternative.

daisy20170101:
I didn't change the current installation, except for adding the additional section. 
I could remove it to the very end as an alternative option then.

Thomas-Ulrich:
Hi,
I ve updated the README. please check it before we merge it.
Thomas.

PR #169, opened 2019-11-25
Implement elastic-acoustic coupling

krenzland:
This pull request includes:

- Support for acoustic materials (mu=0)
- Support for coupling between elastic and acoustic materials with an exact Riemann solver (modified implementation of Carsten's hack)
- Three new boundary condition types: a free surface with gravitation for the acoustic equation, an analytical boundary condition (initial condition/analytical solution is imposed at boundary) and a user-configurable boundary condition (with easi)
- Two test cases for elastic-acoustic coupling (a Scholte wave and a version of Snell's law), using the analytical boundary condition type
- A test case for the free-surface with gravity boundary condition and the user-defined bc, due to Lauren.
- Some small changes (improvements?) to some related parts of the code

The user-configurable boundary condition is currently not documented as I'm currently evaluating, whether it actually makes sense to use it. It allows users to define all boundary conditions that use a state at a fictitious ghost cell that can be expressed as a linear affine mapping of the values of the interior.
For example, to impose a free-slip boundary condition with normal velocity = 1, a user defines:

```yaml
!ConstantMap
    map:
      map_u_u: -1.0
      const_u: 1
```
The user then adds the line
```fortran
&Equations
BoundaryFileName = 'boundary.yaml'
```
to the parameter file.

There are still some TODOs that are related to performance optimizations - I'll perform those later. The important thing is that for now all new scenarios (seem to) work.
I am going to push the Maple notebooks related to the new matrices and some meshes and settings for the new scenarios in separate commits/pull request to the relevant repos soon.

The performance of the kernels shouldn't be slower with my updates. To check this, I've ran the proxy on NG for elastic order 6 with 100k elements and got ~1750 HW flops, which aligns with Carsten's results.


src/Equations/elastic/Kernels/Local.cpp
```diff
@@ -70,17 +72,25 @@ void seissol::kernels::Local::setGlobalData(GlobalData const* global) {
   m_volumeKernelPrototype.kDivM = global->stiffnessMatrices;
   m_localFluxKernelPrototype.rDivM = global->changeOfBasisMatrices;
   m_localFluxKernelPrototype.fMrT = global->localChangeOfBasisMatricesTransposed;
+
+  m_nodalLfKrnlPrototype.rDivMMultV2nTo2m = global->rDivMMultV2nTo2m;
+
+  m_projectKrnlPrototype.V3mTo2nFace = global->V3mTo2nFace;
+  m_projectRotatedKrnlPrototype.V3mTo2nFace = global->V3mTo2nFace;
+
+  dirichletBoundary = seissol::kernels::DirichletBoundary();
```
uphoffc: -> Constructor?

src/Equations/elastic/Kernels/Time.cpp
```diff
@@ -146,24 +152,53 @@ void seissol::kernels::Time::computeAder( double                      i_timeStep
...
   for (unsigned der = 1; der < CONVERGENCE_ORDER; ++der) {
     krnl.execute(der);
 
     // update scalar for this derivative
     intKrnl.power *= i_timeStepWidth / real(der+1);    
     intKrnl.execute(der);
   }
+
+
+#if NUMBER_OF_RELAXATION_MECHANISMS == 0
```
uphoffc: Does it crash or not compile for viscoelastic? If not, why the ifdef?

src/Geometry/PUMLReader.cpp
```diff
@@ -363,7 +363,7 @@ void seissol::PUMLReader::getMesh(const PUML::TETPUML &puml)
 
 			int bcCurrentFace = (boundaryCond[i] >> (j*8)) & 0xFF;
 			int faultTag = bcCurrentFace;
-			if (bcCurrentFace > 6) {
+			if (bcCurrentFace > 8) {
```
uphoffc: This change might break some existing setups. You need to inform the guys and the girls.

uphoffc: Maybe we should make this change more drastic, i.e. bcCurrentFace > 32, 64 or 128. Then there is some space for future bcs.

src/Initializer/time_stepping/LtsLayout.cpp
```diff
@@ -59,34 +59,32 @@ seissol::initializers::time_stepping::LtsLayout::LtsLayout():
...
-  std::vector<Element> const& elements = i_mesh.getElements();
-  for( unsigned int l_cell = 0; l_cell < elements.size(); l_cell++ ) {
-    m_cells.push_back( elements[l_cell] );
-  }
-  std::vector<Fault> const& fault = i_mesh.getFault();
-  for( unsigned int i = 0; i < fault.size(); ++i ) {
-    m_fault.push_back( fault[i] );
-  }
+  m_cells = i_mesh.getElements();
+  m_fault = i_mesh.getFault();
```
uphoffc: Are you sure that copying is unnecessary? (I don't recall why it was in there in the first place...)

src/Initializer/time_stepping/LtsWeights.cpp
```diff
@@ -118,7 +118,7 @@ int seissol::initializers::time_stepping::LtsWeights::getBoundaryCondition( int
                                                                             unsigned cell,
                                                                             unsigned face ) {
   int bcCurrentFace = ((boundaryCond[cell] >> (face*8)) & 0xFF);
-  if (bcCurrentFace > 6) {
+  if (bcCurrentFace > 7) {
```
uphoffc: I think you had 8 there in PUMLReader?? Maybe one should put this logic in a separate function, maybe as part of FaceType

generated_code/NodalBoundaryConditions.py
```diff
@@ -0,0 +1,115 @@
...
+    # Dirichlet boundary, as ghost cell dofs are already rotated
+    identity_rotation = np.double(aderdg.transformation_spp())
+    identity_rotation[0:9, 0:9] = np.eye(9)
+    identity_rotation = Tensor('identityT',
+                               aderdg.transformation_spp().shape,
+                               identity_rotation,
+                               )
+    include_tensors.add(identity_rotation)
+
+    rDivM_mult_V2nTo2m = tensor_collection_from_constant_expression(
```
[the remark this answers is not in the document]
krenzland: rDivM_mult_V2nTo2m => project2nFaceTo3m

src/Model/common.hpp
```diff
@@ -119,115 +130,131 @@ void seissol::model::getTransposedElasticCoefficientMatrix( seissol::model::Elas
...
+
+  } else {
+    R(0,0) = local.lambda;
+    R(1,0) = local.lambda;
+    R(2,0) = local.lambda;
+    R(6,0) = std::sqrt((local.lambda + 2 * local.mu) / local.rho);
+
+    R(3,1) = 1.0;
+    R(5,2) = 1.0;
+  }
```
sebwolf-de: Isnt this if statement just duplicated code? In the end, if the material is not elastic, you would just add an 2*epsilon.

src/Model/common.cpp
```diff
@@ -38,3 +38,9 @@
  **/
  
 #include <Model/common.hpp>
+#include <cmath>
+#include <iostream>
+
+bool seissol::model::testIfElastic(real mu) {
```
sebwolf-de: I'd prefer isAcoustic, as there are a several non-acoustic materials (visco, poro, acoustic)

src/Initializer/time_stepping/LtsLayout.cpp
krenzland: Dunno. It works without copying - I remember that the mesh reader was freed quite early but currently it's freed only at the end of the simulation. Not sure what's the best way here...

PR #171, opened 2019-11-29
Add TP parameters to documentation

sebanger:


sebwolf-de:
Thanks for the update.
I have some more general concerns about the documentary. In the code snippet, there is the OutputMask array, in the text we talk about the iOutputMask. Additionally there is a malformatted U_n. 
Maybe you could fix these on the way ;-)

sebwolf-de:
For the namelists "Pickpoint" and "Elementwise" the parameter is called OutputMask.
For the namelist "Output" the parameter is called iOutputMask. 
So for this context we just need to talk about OutputMask.

sebanger:
> 
> 
> For the namelists "Pickpoint" and "Elementwise" the parameter is called OutputMask.
> For the namelist "Output" the parameter is called iOutputMask.
> So for this context we just need to talk about OutputMask.

should fix this now

sebwolf-de:
Anyway, is it a good idea to merge this now, when the TP branch has not been merged.

sebanger:
The TP branch should be merged by end of next week. So the documentation can be already merged

PR #174, opened 2019-12-19
Remove incorrect assertion that breaks DEBUG mode on elastic.

krenzland:
(See title)


uphoffc:
Should be something like assert( ((uintptr_t)o_timeDerivatives)      % ALIGNMENT == 0 || o_timeDerivatives == NULL );

PR #176, opened 2020-01-09
Sr imposed on dr boundary

Thomas-Ulrich:
This new friction law allows imposing Slip (via SR) on the boundary condition.
The most direct use is (a very expensive way) for computing the final traction distribution from a slip distribution (potential application for the Norcia earthquake).

### DIFF
--- a/Documentation/dynamic-rupture.rst
+++ b/Documentation/dynamic-rupture.rst
@@ -154,7 +154,7 @@ Linear-Slip Weakening Friction
 Rate-and-State Friction
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-Slip-rate imposed on a DR Boundary Condition
+Slip-rate imposed on a DR boundary condition
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 This friction law allows imposing slip-rate on a dynamic rupture boundary (potentially any kinematic models, but the current implementation is limited, see below).
 The FL id for this friction law is 33.
--- a/Documentation/dynamic-rupture.rst
+++ b/Documentation/dynamic-rupture.rst
@@ -158,15 +158,15 @@ Slip-rate imposed on a DR boundary condition
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 This friction law allows imposing slip-rate on a dynamic rupture boundary (potentially any kinematic models, but the current implementation is limited, see below).
 The FL id for this friction law is 33.
-The potential advantages of this approch, compared to a multi point-sources representation are:
-
-- offset discontinuity at the free-surface accurately accounted for
-- wavefield (including final stress or displacement) not impacted by the discretization in the near-field.
+The advantage of this approach compared to a multi point-sources representation is that the fault slip is not condensed to points. 
+Therefore the discontinuity of the displacement across the fault can be accurately accounted for, and more generally the wavefield is accurate in the near-field.
 
 The current implementation allows imposing a slip distribution on the DR Boundary using the same arbitrary smooth-step SR function everywhere on the fault.
 The slip distribution is imposed simultaneously everywhere on the fault, smoothly over a time ``t_0``, where ``t_0`` is a parameter of the ``DynamicRupture`` namelist.
 This is an expensive way of getting the final stress distribution from a given slip distribution.
-The slip distribution is defined using easi by the ``strike_slip`` and ``dip_slip`` variables. Below is an example of input file for defining the slip distribution:
+The slip distribution is defined using easi by the ``strike_slip`` and ``dip_slip`` variables. 
+Warning: the direction of positive ``strike_slip`` and ``dip_slip`` is based on the convention of Seissol (e.g. positive strike_slip for right-lateral faulting). 
+Below is an example of input file for defining the slip distribution:
 
 .. code-block:: YAML
 
@@ -186,11 +186,4 @@ The slip distribution is defined using easi by the ``strike_slip`` and ``dip_sli
           map:
             strike_slip: 0.05
             dip_slip: 0.05
-    #T_n,T_s,T_d are given to avoid crash when SeisSol try defining 'faultParameterizedByTraction' 
-    [T_n,T_s,T_d]: !ConstantMap
-      map:
-        T_n: 0.0
-        T_s: 0.0
-        T_d: 0.0
-
 
--- a/Documentation/dynamic-rupture.rst
+++ b/Documentation/dynamic-rupture.rst
@@ -153,3 +153,44 @@ Linear-Slip Weakening Friction
 
 Rate-and-State Friction
 ^^^^^^^^^^^^^^^^^^^^^^^
+
+Slip-rate imposed on a DR Boundary Condition
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+This friction law allows imposing slip-rate on a dynamic rupture boundary (potentially any kinematic models, but the current implementation is limited, see below).
+The FL id for this friction law is 33.
+The potential advantages of this approch, compared to a multi point-sources representation are:
+
+- offset discontinuity at the free-surface accurately accounted for
+- wavefield (including final stress or displacement) not impacted by the discretization in the near-field.
+
+The current implementation allows imposing a slip distribution on the DR Boundary using the same arbitrary smooth-step SR function everywhere on the fault.
+The slip distribution is imposed simultaneously everywhere on the fault, smoothly over a time ``t_0``, where ``t_0`` is a parameter of the ``DynamicRupture`` namelist.
+This is an expensive way of getting the final stress distribution from a given slip distribution.
+The slip distribution is defined using easi by the ``strike_slip`` and ``dip_slip`` variables. Below is an example of input file for defining the slip distribution:
+
+.. code-block:: YAML
+
+    !Switch
+    [strike_slip, dip_slip]: !Any
+      components:
+       - !AxisAlignedCuboidalDomainFilter
+          limits:
+            x: [-1000, 0]
+            y: [-1e10, 1e10]
+            z: [-5000, -3000]
+          components: !ConstantMap
+            map:
+              strike_slip:   0.01
+              dip_slip: 0.1
+       - !ConstantMap
+          map:
+            strike_slip: 0.05
+            dip_slip: 0.05
+    #T_n,T_s,T_d are given to avoid crash when SeisSol try defining 'faultParameterizedByTraction' 
+    [T_n,T_s,T_d]: !ConstantMap
+      map:
+        T_n: 0.0
+        T_s: 0.0
+        T_d: 0.0
+
+
--- a/src/Physics/Evaluate_friction_law.f90
+++ b/src/Physics/Evaluate_friction_law.f90
@@ -64,6 +64,7 @@ MODULE Eval_friction_law_mod
   PRIVATE :: no_fault
   PRIVATE :: Linear_slip_weakening_bimaterial
   PRIVATE :: Linear_slip_weakening_TPV1617
+  PRIVATE :: ImposedSlipRateOnDRBoundary
   PRIVATE :: rate_and_state
   PRIVATE :: rate_and_state_vw
   PRIVATE :: rate_and_state_nuc101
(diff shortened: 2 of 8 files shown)


Documentation/dynamic-rupture.rst
```diff
@@ -153,3 +153,44 @@ Linear-Slip Weakening Friction
...
+            z: [-5000, -3000]
+          components: !ConstantMap
+            map:
+              strike_slip:   0.01
+              dip_slip: 0.1
+       - !ConstantMap
+          map:
+            strike_slip: 0.05
+            dip_slip: 0.05
+    #T_n,T_s,T_d are given to avoid crash when SeisSol try defining 'faultParameterizedByTraction'
```
uphoffc: I don't like a mandatory work-around for something that can be easily fixed. You just should not call faultParameterizedByTraction because it is unnecessary, right?

src/Physics/Evaluate_friction_law.f90
```diff
@@ -135,6 +136,15 @@ SUBROUTINE Eval_friction_law(    TractionGP_XY,TractionGP_XZ,        & ! OUT: up
                                 resampleMatrix,                            &
                                 DISC,EQN,MESH,MPI,IO)                          
                                 
+        CASE(33)
```
uphoffc: Wasn't 33 used for tpv33 beforehand?

src/Physics/Evaluate_friction_law.f90
```diff
@@ -526,6 +536,109 @@ SUBROUTINE Linear_slip_weakening_TPV1617(TractionGP_XY,TractionGP_XZ,       & !
...
+         Gnucprev = EXP((prevtime-Tnuc)**2/(prevtime*(prevtime-2.0D0*Tnuc)))
+         Gnuc=EXP((tn-Tnuc)**2/(tn*(tn-2.0D0*Tnuc))) - Gnucprev
+         Gnuc=Gnuc/time_inc
+      ELSE
+         Gnuc=0.0D0
+      ENDIF
+
+      !EQN%NucleationStressInFaultCS (4 and 6) contains the slip in FaultCS
+      LocTracXY(:)  = XYStressGP(:,iTimeGP) + eta * EQN%NucleationStressInFaultCS(:,4,iFace)*Gnuc
+      LocTracXZ(:) =  XZStressGP(:,iTimeGP) + eta * EQN%NucleationStressInFaultCS(:,6,iFace)*Gnuc
```
uphoffc: Should be "-" because otherwise you change the plus-side and minus-side convention.

src/Physics/ini_model_DR.f90
```diff
@@ -297,6 +311,13 @@ SUBROUTINE DR_basic_ini(DISC,EQN,MESH,BND)                 ! global variables
       call rotateStressToFaultCS(EQN,MESH,DISC%Galerkin%nBndGP,nuc_xx,nuc_yy,nuc_zz,nuc_xy,nuc_yz,nuc_xz,EQN%NucleationStressInFaultCS,faultParameterizedByTraction)
       deallocate(nuc_xx,nuc_yy,nuc_zz,nuc_xy,nuc_yz,nuc_xz)
     end if
+
+    if (EQN%FL == 33) then !ImposedSlipRateOnDRBoundary
+        allocate(EQN%NucleationStressInFaultCS(DISC%Galerkin%nBndGP,6,MESH%Fault%nSide))
+        call rotateStressToFaultCS(EQN,MESH,DISC%Galerkin%nBndGP,nuc_xx,nuc_yy,nuc_zz,nuc_xy,nuc_yz,nuc_xz,EQN%NucleationStressInFaultCS, .TRUE.)
```
uphoffc: Are you sure that this is correct? Vectors are not rotated in the same way as tensors!

src/ResultWriter/faultoutput.f90
```diff
@@ -388,6 +389,21 @@ SUBROUTINE calc_FaultOutput( DynRup_output, DISC, EQN, MESH, MaterialVal, BND, t
           ELSE
              P_f = 0.0
           ENDIF
+
+          if (EQN%FL.eq.33) then 
+             !case of ImposedSlipRateOnDRBoundary 'friction law': we add the additional stress to the fault output
+             !to show the imposed SR
+             Tnuc = DISC%DynRup%t_0
+             eta = (w_speed(2)*rho*w_speed_neig(2)*rho_neig) / (w_speed(2)*rho + w_speed_neig(2)*rho_neig)
+             IF (time.LE.Tnuc) THEN
```
uphoffc: One should create a subroutine for Gnuc if used at several places.

src/ResultWriter/faultoutput.f90
```diff
@@ -388,6 +389,21 @@ SUBROUTINE calc_FaultOutput( DynRup_output, DISC, EQN, MESH, MaterialVal, BND, t
...
+                Gnuc = EXP((time-Tnuc)**2/(time*(time-2.0D0*Tnuc)))
+             else
+                Gnuc=1d0
+             endif
+             S_XY = S_XY + eta * EQN%NucleationStressInFaultCS(iBndGP,4,iFace)*Gnuc
+             S_XZ = S_XZ + eta * EQN%NucleationStressInFaultCS(iBndGP,6,iFace)*Gnuc
```
uphoffc: "-"

src/ResultWriter/faultoutput.f90
```diff
@@ -510,6 +526,17 @@ SUBROUTINE calc_FaultOutput( DynRup_output, DISC, EQN, MESH, MaterialVal, BND, t
...
 
+          if (EQN%FL.eq.33) then 
+             !case of ImposedSlipRateOnDRBoundary 'friction law': we plot the Stress from Godunov state, because we want to see the traction change from the imposed slip distribution
+             TracMat(4)=LocMat(4)
+             TracMat(6)=LocMat(6)
+             !if DISC%DynRup%SlipRateOutputType .eq. 1 switch back to 0 
+             if (DISC%DynRup%SlipRateOutputType .eq. 1) then
+                LocSRs = dot_product(SideVal2(8) * NormalVect_s + SideVal2(9) * NormalVect_t, strike_vector) - dot_product(SideVal(8) * NormalVect_s + SideVal(9) * NormalVect_t, strike_vector)
+                LocSRd = dot_product(SideVal2(8) * NormalVect_s + SideVal2(9) * NormalVect_t, dip_vector   ) - dot_product(SideVal(8) * NormalVect_s + SideVal(9) * NormalVect_t, dip_vector   )
+             endif
```
uphoffc: Please don't duplicate this code.

Documentation/dynamic-rupture.rst
```diff
@@ -153,3 +153,44 @@ Linear-Slip Weakening Friction
...
 Rate-and-State Friction
 ^^^^^^^^^^^^^^^^^^^^^^^
+
+Slip-rate imposed on a DR boundary condition
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+This friction law allows imposing slip-rate on a dynamic rupture boundary (potentially any kinematic models, but the current implementation is limited, see below).
+The FL id for this friction law is 33.
+The potential advantages of this approch, compared to a multi point-sources representation are:
+
+- offset discontinuity at the free-surface accurately accounted for
```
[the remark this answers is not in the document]
Thomas-Ulrich: I mean if there is surface rupture, you can model accurately the discontinuity of the displacement across the fault, which is not the case with points sources.

Documentation/dynamic-rupture.rst
```diff
@@ -153,3 +153,44 @@ Linear-Slip Weakening Friction
...
+- wavefield (including final stress or displacement) not impacted by the discretization in the near-field.
```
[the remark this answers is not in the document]
Thomas-Ulrich: yes

Documentation/dynamic-rupture.rst
```diff
@@ -153,3 +153,44 @@ Linear-Slip Weakening Friction
...
+
+The current implementation allows imposing a slip distribution on the DR Boundary using the same arbitrary smooth-step SR function everywhere on the fault.
+The slip distribution is imposed simultaneously everywhere on the fault, smoothly over a time ``t_0``, where ``t_0`` is a parameter of the ``DynamicRupture`` namelist.
+This is an expensive way of getting the final stress distribution from a given slip distribution.
```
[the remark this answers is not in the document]
Thomas-Ulrich: Currently yes. I agree that it is not very smart to use seissol for a static calculation.
But it seems that the tools for doing that, that our collaborators at INGV and Umich tried rely on simplifying assumptions (e.g. homogeneous, no topography, if boundary element methods) and it seems that the resulting tractions on the fault are not usable because too noisy. Therefore it is useful to have something expensive that is working, and it might be an advantage that it relies on the exact same mesh.

Documentation/dynamic-rupture.rst
Thomas-Ulrich: Yes but I don't like adding if conditions and making the code more complex for a feature which will be only marginally used ;-)

src/Physics/Evaluate_friction_law.f90
Thomas-Ulrich: no

src/Physics/ini_model_DR.f90
Thomas-Ulrich: ha yes indeed, wrong.

Thomas-Ulrich: Ha wait, no, it is correct, because faultParameterizedbyTraction is set True, and therefore it uses create_fault_rotation_matrix function, which rotates vectors and not a tensor.

uphoffc: No that is not true. create_fault_rotation_matrix also returns a transformation matrix for a tensor and not for a vector. Moreover, Stress is still multiplied with iT.

uphoffc: What the function rotateStressToFaultCS does if faultParameterizedbyTraction=True is the following: The input stress tensor is given in a coordinate system with directions normal, strike, dip. The input stress tensor is then transformed to the physical space (x,y,z) and afterwards transformed to a fault-aligned space (with the same normal but different tangents!).

Thomas-Ulrich: ok, you are right. I will fix it.

src/ResultWriter/faultoutput.f90
Thomas-Ulrich: where would you put such a function within the code? 
a new file in ``SeisSol/src/Numerical_aux``? 
Apart from that, I think I fixed everything. Thx for the useful comments!

uphoffc: You should put it where you would look for it first :) 
I'd put it in the Physics folder.

Thomas-Ulrich: Hi, 
Let me know how you like the function name I ve given. 
I will then also change the RS friction nucleation with the new function.

uphoffc: Yes looks fine.

Thomas-Ulrich:
I think I 've done all change requested.
This friction law may not be very useful in its current state (yet we still have a few applications already) but can be seen as a demonstrator/a base for future developments, depending of our needs, which should be straightforward.

I would, therefore, suggest merging it to master.

src/Physics/ini_model_DR.f90
```diff
@@ -363,6 +381,50 @@ SUBROUTINE rotateStressToFaultCS(EQN,MESH,nBndGP,s_xx,s_yy,s_zz,s_xy,s_yz,s_xz,s
...
+    do i = 1, MESH%Fault%nSide
+          NormalVect_n = MESH%Fault%geoNormals(1:3,i)
+          NormalVect_s = MESH%Fault%geoTangent1(1:3,i)
+          NormalVect_t = MESH%Fault%geoTangent2(1:3,i)
+ 
+          strike_vector(1) = NormalVect_n(2)/sqrt(NormalVect_n(1)**2+NormalVect_n(2)**2)
+          strike_vector(2) = -NormalVect_n(1)/sqrt(NormalVect_n(1)**2+NormalVect_n(2)**2)
+          strike_vector(3) = 0.0D0
+          dip_vector = NormalVect_n .x. strike_vector
+          dip_vector = dip_vector / sqrt(dip_vector(1)**2+dip_vector(2)**2+dip_vector(3)**2)
```
uphoffc: This is a duplicate from create_fault_rotationmatrix. I'd introduce another subroutine in create_fault_rotationmatrix.f90 which returns these three vectors, which is called from here and from create_fault_rotationmatrix

Thomas-Ulrich: done

PR #180, opened 2020-01-14
Thomas/docu up

Thomas-Ulrich:


Thomas-Ulrich:
but how to fix/rerun travis to merge?

krenzland:
The issue is that we try to download netcdf but the file does not exist. I'm going to fix this soon

PR #184, opened 2020-01-23
update viewreceiver qt5 py3

Thomas-Ulrich:
I (manually) applied 2to3 and changes some lines to make the code run with pyqt5.
I did not test all the features, but it seems to work fine.


sebwolf-de:
Thanks for the update, especially with the upcoming end of life of Python2.

Issue #140, opened 2019-10-10
SeisSol_debug_generatedKernels_sskx_hybrid_none_9_3 crashes


Thomas-Ulrich:
Hi,

When trying to debug something, I ve realized that SeisSol_debug_generatedKernels_sskx_hybrid_none_9_3 (no plasticity) will always (tested on 2 different setups) crash after the LTS initialization:

```
Thu Oct 10 10:29:19, Info:  Deriving clusters ids for min. time step width / multiRate: 0.000119396 / 2 
Thu Oct 10 10:29:20, Info:  Number of elements in time clusters: 
Thu Oct 10 10:29:20, Info:  0: 2587 
Thu Oct 10 10:29:20, Info:  1: 212960 
Thu Oct 10 10:29:20, Info:  2: 670034 
Thu Oct 10 10:29:20, Info:  3: 223240 
Thu Oct 10 10:29:20, Info:  4: 100687 
Thu Oct 10 10:29:20, Info:  5: 33874 
Thu Oct 10 10:29:20, Info:  6: 42686 
Thu Oct 10 10:29:20, Info:  7: 290906 
Thu Oct 10 10:29:20, Info:  8: 89033 
Thu Oct 10 10:29:20, Info:  maximum theoretical speedup (compared to GTS): 7.9566 per cell LTS, 5.28691 with the used clustering. 
Thu Oct 10 10:29:20, Info:  Number of elements in dynamic rupture time clusters: 
Thu Oct 10 10:29:20, Info:  0 (dr): 600 
Thu Oct 10 10:29:20, Info:  1 (dr): 76902 
Thu Oct 10 10:29:20, Info:  2 (dr): 126385 
Thu Oct 10 10:29:20, Info:  3 (dr): 0 
Thu Oct 10 10:29:20, Info:  4 (dr): 0 
Thu Oct 10 10:29:20, Info:  5 (dr): 0 
Thu Oct 10 10:29:20, Info:  6 (dr): 0 
Thu Oct 10 10:29:20, Info:  7 (dr): 0 
Thu Oct 10 10:29:20, Info:  8 (dr): 0 
SeisSol_debug_generatedKernels_sskx_hybrid_none_9_3: src/Initializer/tree/Layer.hpp:109: void *seissol::initializers::Layer::bucket(const seissol::initializers::Bucket &): Assertion `m_buckets != __null && m_buckets[handle.index] != __null' failed.
```

For those on pr45fi, here is an example:
[...]

PR #193, opened 2020-02-12
Remove check for invertability

sebwolf-de:


uphoffc:
??

uphoffc:
Is the issue fixed or what??

Thomas-Ulrich:
I think there is an underlying issue.
See my edit on the penultimate comment of this issue:
https://github.com/SeisSol/SeisSol/issues/140

sebwolf-de:
We compared the old implementation e599e8b with the one implemented here. We used all materials given by Thomas' Korea setup and calculated the Godunov States. The solutions with the inverse compared to the old one has an error of 10^-{16}.

uphoffc:
Ah ok. Is now an SVD or something like that used to invert the matrix?

sebwolf-de:
[matrix.tar.gz](https://github.com/SeisSol/SeisSol/files/4191378/matrix.tar.gz)

sebwolf-de:
We leave it to the Eigen internals to decide what they use. Talk to you after lunch ;-)

Thomas-Ulrich:
Hi,
I just run Kadek's setup and it does not crash and produces outputs that looks ok.

krenzland:
> Ah ok. Is now an SVD or something like that used to invert the matrix?

It's LU according to the documentation

PR #195, opened 2020-02-12
Fix single precision wavefield

ravil-mobile:
Updated reference to xdmfwriter which allows to store data in either single or double precision. However, geometry (vertices, etc.) is going to get stored in double precision. This is a temporary solution and going to be reverted back as soon as SeisSol gets rid off the fortran code.

Additionally, Doxyfile was move to a separate folder in order to force developers to keep working on and improving the source code documentation


krenzland:
Thanks!
I've quickly tried this, and the wave field writer is indeed working for single precision. Unfortunately, the free surface writer is now broken for single precision (which works on master)

ravil-mobile:
It is a pity. Let me have a look at the free surface writer then. Also, the reference to seissol-master branch of xdmfwriter is going to be removed and updated to reference to the top of dev branch

krenzland:
I think the reason is that the free surface writer always writes out doubles. So we could (and should!) probably change this to use the same precision as for the other writers.
Not sure what the fault output writes - I suspect this one would be more annoying to fix, as it still uses some fortran somewhere.

Also, why the dev branch and not master?

PR #196, opened 2020-02-17
Update supermuc.rst

daisy20170101:
update compilation options on NG due to changes in HDF5, Metis, etc

### DIFF
--- a/Documentation/supermuc.rst
+++ b/Documentation/supermuc.rst
@@ -71,16 +71,17 @@ Supermuc-NG
 
   ##### module load for SeisSol
   module load scons gcc cmake/3.6 python/3.6_intel slurm_setup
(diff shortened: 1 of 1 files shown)


Documentation/supermuc.rst
```diff
@@ -156,7 +157,8 @@ set compiler options:
   metis='yes'
   netcdfDir=os.environ['NETCDF_BASE']
   hdf5Dir=os.environ['HDF5_BASE']
-
+  metisDir=os.environ['PARMETIS_BASE']
```
Thomas-Ulrich: is this used by any script?

daisy20170101: There seems an update in module HDF6 and Metis on NG and some names are changed.

It seems that metisDir is needed in the compilation.

PR #198, opened 2020-02-19
add transient nucleation stress to fault output

Thomas-Ulrich:
in case of RS friction law

### DIFF
--- a/src/ResultWriter/faultoutput.f90
+++ b/src/ResultWriter/faultoutput.f90
@@ -556,13 +556,7 @@ CONTAINS
               IF (EQN%FL.EQ.103) THEN
                  NucleationStressXYZ = MATMUL(T(1:6,1:6), EQN%NucleationStressInFaultCS(iBndGP,:,iFace))
                  Tnuc = DISC%DynRup%t_0
-                 IF (time.LE.0) THEN
-                    Gnuc = 0d0
-                 ELSEIF (time.LT.Tnuc) THEN
-                    Gnuc = exp((time-Tnuc)**2/(time*(time-2*Tnuc)))
-                 ELSE
-                    Gnuc = 1d0
-                 ENDIF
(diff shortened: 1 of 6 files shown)


Thomas-Ulrich:
I've added a second feature to the pull request.
Now we can use s_ij in conjunction with Tnuc_i or 
T_i in conjunction with nuc_ij.
This may simplify some of our setups.

Thomas-Ulrich:
Any review on this?

src/ResultWriter/faultoutput.f90
```diff
@@ -550,12 +551,29 @@ SUBROUTINE calc_FaultOutput( DynRup_output, DISC, EQN, MESH, MaterialVal, BND, t
...
+              IF (EQN%FL.EQ.103) THEN
+                 NucleationStressXYZ = MATMUL(T(1:6,1:6), EQN%NucleationStressInFaultCS(iBndGP,:,iFace))
+                 Tnuc = DISC%DynRup%t_0
+                 IF (time.LE.0) THEN
+                    Gnuc = 0d0
+                 ELSEIF (time.LT.Tnuc) THEN
+                    Gnuc = exp((time-Tnuc)**2/(time*(time-2*Tnuc)))
+                 ELSE
+                    Gnuc = 1d0
+                 ENDIF
```
uphoffc: Why don't you use the Calc_Smoothstep function here?

Thomas-Ulrich: because it is called Calc_SmoothStepIncrement, that is it compute the increment of nucleation stress and not the nucleation stress.

uphoffc: https://github.com/SeisSol/SeisSol/blob/master/src/Physics/NucleationFunctions.f90#L86

Thomas-Ulrich: ups

PR #199, opened 2020-02-21
Unify Initialfield + use material from easi

sebwolf-de:
1) Initial Field for elastic was a mess and not comprehensible.
2) I moved all the initial fields to one common place. The formulation of a  planar wave is the same for all linear PDEs anyways.
3) Do not hard-code the material parameters, but use the ones from the first cell (assuming homogeneous material) -> Convergence tests can be done with the material described by the easi files.


src/Physics/InitialField.cpp
```diff
@@ -1,148 +1,70 @@
...
+  Eigen::ComplexEigenSolver<Matrix> ces;
+  ces.compute(op);
+  
+  auto eigenvalues = ces.eigenvalues();
+  std::vector<size_t> sortedIndices(NUMBER_OF_QUANTITIES);
+  std::iota(sortedIndices.begin(), sortedIndices.end(), 0);
+  std::sort(sortedIndices.begin(), sortedIndices.end(), [&eigenvalues](size_t a, size_t b) {
+    return eigenvalues[a].real() < eigenvalues[b].real();
+  });
+
```
krenzland: Could you maybe add a comment that explains that sortedIndices contains a sorted list of eigenvalues-indices. Maybe also why you're sorting them?

Thomas-Ulrich:
Hi, 
maybe you could add a documentation page about how to set up an Initialfield in seissol and what are the limitations of the current implementation?
Thomas.

AliceGabriel:
Agreed - nice changes but needs documenting and best 1-2 examples.. Thanks!

src/Physics/InitialField.cpp
uphoffc: There is even a comment in the code: sort eigenvalues so that we know which eigenvalue corresponds to which mode
:-D

sebwolf-de: I've added that after Lukas's comment ;-)

Documentation/parameters.par
```diff
@@ -43,7 +52,7 @@ SlipRateOutputType=0        ! 0: (smoother) slip rate output evaluated from the
 &Elementwise
 printIntervalCriterion = 2       ! 1=iteration, 2=time
 printtimeinterval_sec = 0.2      ! Time interval at which output will be written
-OutputMask = 1 1 1 1 1 1 1 1 1 1 1   ! turn on and off fault outputs
+OutputMask = 1 1 1 L1 1 1 1 1 1 1 1   ! turn on and off fault outputs
```
uphoffc: L?

krenzland:
Sounds good to me. I agree with @AliceGabriel and @Thomas-Ulrich that we should add some documentation to this. But considering that we didn't have documentation before this pull-request I don't see any reason why we shouldn't merge this bug fix now.
Nevertheless, it would be great if you could open an issue regarding the documentation of initial conditions so that we track the need for documentation.

PR #201, opened 2020-03-17
fix compiling rconv

Thomas-Ulrich:


Thomas-Ulrich:
I've added to the pull request an unrelated parameter check for anelasticity.

PR #204, opened 2020-03-27
Ravil/spack

ravil-mobile:
Automatic installation of SeisSol dependencies with Spack


Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,391 @@
+Installation with Spack
+=======================
+
+Installation of any HPC application can be tricky. SeisSol is not an exception. 
+To considerably alleviate the installation process, we provide few scripts which 
+relies on `Spack <https://github.com/spack/spack/wiki>`_, which is a new HPC
```
sebwolf-de: rely

ravil-mobile: thanks!

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,391 @@
...
+Examples
+~~~~~~~~
+
+.. code-block:: bash
+
+  # 1. with intel compiler suite
+  spack install seissol-env +mpi +asagi ^intel-mpi@2018.2.199 %intel@17.0.2
+
+  # 2. with gcc compiler suite
+  spack install seissol-env +mpi +asagi ^openmpi@3.1.5 %gcc@8.3.0
```
sebwolf-de: When I execute this line (tried on LRZ workstation and personal notebook) I get the following error:
seissol-env requires cmake version 3.12.0:3.16.2, but spec asked for 3.16.5
I have installed cmake 3.15.2 and 3.16.4 respectively.

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,391 @@
...
+.. code-block:: bash
+
+  module load seissol-env-develop-<compiler>-<hash>
+
+  # if you compile seissol-env with a compiler installed with Spack
+  # you may be required to load that compiler as well
+  module load <compiler>
+
+
+After that, you can compiler SiesSol using either CMake or
```
sebwolf-de: typo

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,391 @@
...
+
+
+After that, you can compiler SiesSol using either CMake or 
+:ref:`Scons <compiling-seissol>`.
+
+
+SeisSol-Utils
+-------------
+
+By default, the scrip installs:
```
sebwolf-de: typo

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,391 @@
...
+
+- pumgen (without a Simmetrix support)
+- gmsh (without a GPU support)
+- gmsh2gambit
+- cube_c
+- rconv
+- SiesSol Cookbook, which contains some examples to run
```
sebwolf-de: typo

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,391 @@
...
+  cd $BENCHMAKRS
+
+
+
+Tips and Tricks
+---------------
+
+.. _spack_known_issues:
+
+1. Spack builds the entire dependency graph before compaling and installing.
```
sebwolf-de: typo

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,391 @@
...
+1. Spack builds the entire dependency graph before compaling and installing. 
+The graph includes all libs and packages which are necessary to build your 
+application, including packages like: *tar, gzip, zlib,  autoconf, 
+cmake, automake, pkgconf, m4, ncurses, etc*. Packages like these do not 
+affect performance of your application but help Spack to install it. 
+Therefore, it is not necessary to install them again and again. You can 
+install such  packages only once and mark them as Default 
+`(External) <https://spack-tutorial.readthedocs.io/en/latest/tutorial_configuration.html#external-packages>`_.
+and Non-Buildable. It can speed-up installation of SeisSol-Env and SeisSol-utils 
+considerably. You will need to modify and edit **~/.spack/packages.yaml** file.
```
sebwolf-de: Wouldnt it be a good idea to reverse the process here? Make it the default to not build all these basic packages and tell people to also build tar, gzip, ... when they encounter problems.

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,391 @@
...
+
+
+Known Issues
+------------
+
+1. Spack is a really live project with dozens of commits per day. It is 
+diffucult for us to keep the same pace with Spack. A new version of Spack
```
sebwolf-de: typo

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,391 @@
...
+may not work because of new added features what we may not be aware of. 
+Therefore, it may be necessary to use an older version of Spack. You
+can simply do it by moving the HEAD of your locally installed Spack
+repository to an old commit:
+
+.. code-block:: bash
+
+  cd $SPACK_ROOT
+  git checkout <a previous SPACK commit>
```
sebwolf-de: Maybe provide todays spack commit hash?

spack_support/packages/seissol-env/package.py
```diff
@@ -0,0 +1,117 @@
+# Copyright 2013-2020 Lawrence Livermore National Security, LLC and other
```
sebwolf-de: I guess this file is written by you, Ravil, based on some sample taken from Lawrence Livermore? So maybe alter the LICENCE header accordingly. But I am no expert on legal stuff.

spack_support/packages/seissol-utils/package.py
```diff
@@ -0,0 +1,108 @@
+# Copyright 2013-2020 Lawrence Livermore National Security, LLC and other
```
sebwolf-de: Same as for seissol-env

spack_support/packages/seissol-env/package.py
ravil-mobile: It is the standard license agreement enforced by spack. Basically, this line is automatically generated by spack when one creates an installation script. I checked other scripts from hdf5 and openmpi. It is the same there

Documentation/installation-with-spack.rst
ravil-mobile: As far as I know there is only one way how to force spack to not build some packages: to add them to your ~/.spack/packages.yaml and mark them as Non-Buildable

ravil-mobile: I would like to do the same as you suggested but I couldn't find a solution to enforce spack to prevent building these packages inside of an installation script

Documentation/installation-with-spack.rst
ravil-mobile: Honestly, I am not sure about it. I cannot check all possible combinations with respect to all possible installation options. Therefore, I am not sure that today's snippet of Spack is a perfect reference

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,392 @@
...
+*NOTE*: mpi is a virtual package, a user must specify a concrete implementation
+of the standard
+
+Examples
+~~~~~~~~
+
+.. code-block:: bash
+
+  # 1. with intel compiler suite
+  spack install seissol-env +mpi +asagi %intel@17.0.2 ^intel-mpi@2018.2.199
```
sebwolf-de: Shouldn't we include the ^cmake@3.16.0 part after our discussion on slack?

Documentation/installation-with-spack.rst
sebwolf-de: Okay, so let's leave it like this.

Documentation/installation-with-spack.rst
sebwolf-de: Okay, I see there are just to many options to try all of them. But maybe we can add something like: On SuperMUC commit xxxxxx works fine for the combinations described above.

ravil-mobile: It is a good idea. Additionally, we can provide package.yaml file with not-buildable packages (an use it as an exaple) and a specify a particular commit for Spack. Let me see whether I can do it for SuperMUC because the admins did whatever they could to prevent compiling on production nodes but maybe i will manage to find a way around

ravil-mobile:
Hi all,

One of the last issues before the merge was a request to test the installation of NG. Finally I've got access to NG. I talked to one of the admins about how to use Spack there from the user's perspective. The short answer was to wait and play with Spack on the Linux cluster.

The NG is going to be on maintenance and it will support Spack@15 in the future (but nobody knows when). The LRZ people use LinuxCluster as a playground for Spack. In other words, they test how users interact with Spack, collect issues and make notes about what should be done on the NG in order to fully support Spack for users like us. They told me that everything would work more or less the same on NG as it is on the Linux Cluster right now. 

So, I will try to install everything on the Linux Cluster and document. The best strategy with respect to NG is just to wait.

One question that I wanted to ask is: maybe we can move Spack support to a separate repository? What do you think?

Thomas-Ulrich:
Can you just install and use your own spack on supermuc? E.g. this is 
what I used for installing python vtk on NG:git clone 
git@github.com:spack/spack.git
cd spack/bin
#load all the required compilers, e.g. module load gcc
./spack compilers
#Install spack on the computer with internet access, then download the 
required packages using:
./spack mirror create -d spack-mirror-2019-10-17/ -D 
vtk+python^hdf5+hl+mpi^python@3.6.0 py-numpy^python@3.6.0 
py-lxml^python@3.6.0 py-h5py^python@3.6.0
#copy spack-mirror-2019-10-17/ to supermuc and then run:
./spack mirror add mvtk spack-mirror-2019-10-17/
./spack install py-numpy^python@3.6.0 py-lxml^python@3.6.0 
py-h5py^python@3.6.0 vtk+python^hdf5+hl+mpi^python@3.6.0
./spack activate py-numpy^python@3.6.0 py-lxml^python@3.6.0 
py-h5py^python@3.6.0
#Here are other python modules that might be interesting to install: 
py-matplotlib^python@3.6.0 py-scipy^python@3.6.0

ravil-mobile:
Hi @Thomas-Ulrich,

Yes, you can.  I guess you ended up with Spack installing lots of low-level stuff (even m4, tar, zlib). 

However, it should not be like this in a cluster. Admins of a cluster "mark" all these low-level packages as well as compilers and different mpi installations only once "in a system level" which becomes a default setup for all users. Additionally, the admins setup default paths to some caches: for builds and sources, as well as a tmp folder for Spack.

For example, one can load Spack on the linux cluster as following
```
module load user_spack
```
which will set up SPACK_ROOT to /lrz/sys/spack/user/release/20.1/spack
there is a folder called config with (as the name suggest) all Spack configurations for the linux cluster.
The most interesting is *upstreams.yaml* and a filed inside called *install_tree*

Basically, it say where spack should look for a package first. If it can find a binary there it will just copy it to my local spack folder. Otherwise, spack will install the package from sources. It will look at user's defined mirrors (basically what you proposed) and try to find sources. If Spack cannot find source files there. It will try to fetch it from the Internet.

ravil-mobile:
Your solution is fine because there is no official Spack support on NG. I can try to hack the Spack that they use but I don't think that is a good idea. Also I can try to setup Spack for a project but this solution is not portable between projects. Maybe it is better to wait till they officially provide Spack support. What do you think?

brietzke:
optimal would be to bring seissol/package.py to upstream spack

ravil-mobile:
Hi, thanks for joining our conversation. Yes, you right. But I don't know whether spack reviewers will agree upon this style: https://github.com/SeisSol/SeisSol/blob/e4602bede0aab8509f59b019db94d8889152b267/spack_support/packages/seissol-env/package.py#L68-L105

Basically Im looping through all dependencies and add them to the environment variables. It is convenient for me as an user because I eventually need to load only one module. However, I guess it is not what spack developers want a script to perform. But on another hand, it is a `BundlePackage`.

Before up-streaming seissol-env and seissol-utils I need to upstream other packages (you can have a look here: https://github.com/SeisSol/SeisSol/tree/ravil/spack/spack_support/packages). Some of them are not ours. Do you have any idea what I should do in this case? For example, what should I do with `cxxtest` and `impalajit`?

Thanks in advance

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,392 @@
...
+Additionally, a user can customize each individual dependency using 
+Spack 
+`recursive syntax <https://spack.readthedocs.io/en/latest/basic_usage.html#specs-dependencies>`_. 
+
+
+Options
+~~~~~~~
+
+- *asagi* [**default=on**, off] - installs asagi 
+- *building_tools* [**default=on**, off] - installs scons and cmake
```
sebwolf-de, line 218: Here you say, default=on, above you reason that scons and cmake are optional.

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,392 @@
...
+
+  module load seissol-env-develop-<compiler>-<hash>
+
+  # if you compile seissol-env with a compiler installed with Spack
+  # you may need to load that compiler as well
+  module load <compiler>
+
+
+After that, you can compile SeisSol using either CMake or 
+:ref:`Scons <compiling-seissol>`.
```
sebwolf-de, line 260: compiling-seissol has now cmake as the default.

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,392 @@
...
+
+
+SeisSol-Utils
+-------------
+
+By default, the script installs:
+
+- pumgen (without a Simmetrix support)
+- gmsh (without a GPU support)
+- gmsh2gambit
```
sebwolf-de, line 270: With the new gmsh funcionality of pumgen, gmsh2gambit might be obsolete.

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,392 @@
...
+- cube_c
+- rconv
+- SeisSol Cookbook, which contains some examples to run
```
sebwolf-de, line 273: SeisSol Cookbook is now SeisSol/Examples, I guess.

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,392 @@
...
+
+As in case of *seissol-env*, you need **scons** and, therefore, **python3** for 
+compiling. However, installation of these packages is optional to save 
+the set-up time.
+
+
+Options
+~~~~~~~
+
+- *benchmarks* [on, **default=off**] - installs SeisSol benchmarks. Make sure that you have access to the SeisSol LRZ-gitlab account.
```
sebwolf-de, line 283: Also here: Everything is merged in SeisSol/Examples.

Documentation/installation-with-spack.rst
```diff
@@ -0,0 +1,392 @@
...
+
+.. code-block:: bash
+
+  module load seissol-utils-develop-<compiler>-<hash>
+
+  # to access the Cookbook
+  cd $COOKBOOK
+
+  # to access the Benchmakrs
+  cd $BENCHMAKRS
```
sebwolf-de, line 316: Also this is just SeisSol/Examples now

spack_support/packages/seissol-utils/package.py
```diff
@@ -0,0 +1,154 @@
...
+
+    variant('cookbook', 
+            default=False, 
+            description="fetches cookbook. Be sure that you "
+                        "have access to the SeisSol LRZ-gitlab repo.")
+
+    variant('benchmarks', 
+            default=False, 
+            description="fetches benchmarks. Be sure that you "
+                        "have access to the SeisSol LRZ-gitlab repo.")
```
sebwolf-de, line 34: Also here: Move to SeisSol/Examples.

spack_support/packages/seissol-utils/package.py
```diff
@@ -0,0 +1,154 @@
...
+
+    resource(name='benchmarks', 
+             git='https://gitlab.lrz.de/seissol.org/benchmarks',
+             when='+benchmarks', 
+             placement='benchmarks')
+
+    depends_on("hdf5@1.8.21 +fortran +shared +mpi")
+    depends_on('netcdf-c@4.4.0 +shared +mpi')
+    depends_on('pumgen')
+    depends_on('glm@0.9.7.1')
```
sebwolf-de, line 52: We removed glm recently.

spack_support/packages/seissol-utils/package.py
```diff
@@ -0,0 +1,154 @@
...
+                cmake(".")
+                make()
+
+    def install(self, spec, prefix):
+
+        if "+cookbook" in spec:
+            install_tree("cookbook", prefix.cookbook)
+
+        if "+benchmarks" in spec:
+            install_tree("benchmarks", prefix.benchmarks)
```
sebwolf-de, line 114: You already know, what I mean here ;-)

sebwolf-de (commented):
Hi, good steps forward. I guess a lot of people already use spack-support successfully. So IMO we can move this to another repository as Ravil supposed and think about upstreaming this there.

PR #208, opened 2020-04-07
Carsten/multiple materials

uphoffc:
Hi,

this is not really a pull request (did not test too much) but rather I'd like to start a discussion.

An open question how to store multiple material types within the LTS tree. My idea would be that one may specify multiple types for a field of the tree. The time cluster then stores an int which indicates which type is actually stored in the time cluster. This is essentially the same method that std::variant uses for a type-safe union. Therefore, the interface to the tree could be std::variant (or similar). That is, when I request a material from the time cluster, then I get returned a std::variant<ElasticMaterial*, AnisotropicMaterial*, ...>.

The implications on CellLocalMatrices of this approach are shown in this pull request. Basically, one can use std::visit then to implement multiple dispatch, e.g. to choose the correct Riemann solver.

What do you think?

Regards,
Carsten

PS: Requires c++17 btw.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 3.13)
 
 if(POLICY CMP0075)
   cmake_policy(SET CMP0075 NEW)
@@ -29,7 +29,7 @@ set(HARDWARE_DEFINITIONS "ALIGNMENT=${ALIGNMENT}"
 # TODO: Move one dir up.
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
@@ -260,7 +260,7 @@ if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
   # using Intel C++
   # todo remove std?, is ffreestanding needed?
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -qopenmp")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -qopenmp")
 
   # only if openmp
   set(CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} -qopenmp")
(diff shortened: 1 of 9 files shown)


krenzland:
Just some thoughts:

Looks like a good idea in principle. It probably will get a bit hairy when including this in the LTS tree. std::variant has the size of the largest member, so we have the options: further level of indirection (std::variant of pointers to materials), wasting some space (std::variant of materials) or rolling our own std::variant with variable space (really, really ugly).
Do we actually want to store multiple materials in one lts tree in the future? Once proper coupling is implemented, we might want to have one material per LTS tree, solving this problem.
Still need to design an interface to access the material

C++17 is fine, Actor model will require this as well.

P.S: You can open WIP pull requests, those cannot be merged.

uphoffc:
Yes, but you still need the multiple dispatch mechanisms even if you have separate trees. (Or you template everything... which sucks.)

So I think the basic idea of std::variant is the following:
1) Generate a type id for every table
2) Create a software vtable
3) Call right function according to vtable

So in pseudo-code this is roughly equivalent to
struct variant {
  union {
    Type1 type1;
    Type2 type2;
    Type3 type3;
  }
  int index;
};

void visit(visitor, variant) {
  void (Visitor::*vtable)[]{
    &visitor.operator()(Type1&);
    &visitor.operator()(Type2&);
    &visitor.operator()(Type3&);
  };
 
  visitor.(*vtable[variant.index])();
}

I likely got the syntax wrong... 

In our case we can assume that every time cluster stores the same type (i.e. one elastic cluster, one poroelastic cluster). So it would make sense to dispatch only once and run over the whole cluster. I.e.

auto visit(visitor, cluster) {
  void (Visitor::*vtable)[]{
    &visitor.operator()(Type1&);
    &visitor.operator()(Type2&);
    &visitor.operator()(Type3&);
  };
 
  return *vtable[cluster.index];
}

and then
auto visitFun = visit(visitor, cluster);
for (auto& element : cluster) {
  visitor.visitFun(element);
}

So one needs to figure out how to actually implement something like this. But I guess its possible as it is possible for std::variant.

krenzland:
I think of std::variant as a union + enum tag + a lot of magic.
But yes, we still need to dispatch - but we can neglect the storage part!
We can dispatch once per cluster but also need to handle the interfaces between clusters.

Not sure what the best way of implementing is - we probably have to template the lts trees somehow and then do our own tagged union interface. I don't really like that idea but it should work.

uphoffc:
I found a way to support multiple dispatch which is as nice as std::visit but without std::variant:

https://godbolt.org/z/nqRY6M

The basic idea is to store the type of an object in its vtable (therefore only classes with vtable are supported but that is the case for Material at least). In principle one could also store the type in an integer (without vtable). That's not as nice but would be a trivial extension.

In essence, the approach is not much different from specifying enum class MaterialType and adding a getMaterialType() to every Material type (as in https://github.com/SeisSol/SeisSol/blob/master/src/Model/common_datastructures.hpp#L50).
However, instead of an enum we use a type-list as the root class, i.e.

```

krenzland:
This sounds quite good in principle. I haven't had the time to play around with this in much detail.
But overall I like to move typeids to the type system. They belong there.
It still looks like quite a bit of magic but that doesn't have to be a negative aspect ;)

uphoffc:
"But overall I like to move typeids to the type system."
Which type system?

PR #213, opened 2020-05-13
Unit test for NRF Reader

sebwolf-de:
Hi, I've written a unit test for the NRF Reader to test my remove-glm branch.
I think it is a good idea to have smaller - and faster to review pull requests. So I'd like to have it reviewed and merged now and not together with the remove-glm stuff later.


Thomas-Ulrich:
If the aim of these unit tests is also to test that the geometric transformations that give the tangent and normal vectors are correct, then it might be a good idea to test a more complicated fault orientation.

sebwolf-de:
I just want to check that the contents of the nrf file are correctly read into SeisSol. In my opinion the test for tangents and normals should be in rconv.

Thomas-Ulrich:
ah yes that's right.

PR #221, opened 2020-06-08
Some small fixes

sebwolf-de:
This PR fixes some small issues, which we encountered recently:
* Remove glm include dir from CMake
* Fix Free Surface Godunov state
* CPP/Fortran interoperability array ordering

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -227,7 +227,6 @@ target_include_directories(SeisSol-lib PUBLIC
 
 target_include_directories(SeisSol-lib SYSTEM PUBLIC
         ${CMAKE_CURRENT_SOURCE_DIR}/submodules/eigen3
-        ${CMAKE_CURRENT_SOURCE_DIR}/submodules/glm
 )
 
 target_compile_definitions(SeisSol-lib PUBLIC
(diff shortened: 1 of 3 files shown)


Thomas-Ulrich:
Did these issues were affecting fully elastic setups? When were they introduced?

sebwolf-de:
The one with the free surface is from November, but this was a minor problem. 
The one with the wavespeeds was introduced in April.
Cmake does not affect anything.

krenzland:
Regarding the free surface:
It seems to be only a small error. This is not great but if the Godunov state is slightly wrong the solution shouldn't be affected too much!

krenzland:
Okay, I ran a quick convergence test with Snell's law for elastic-acoustic interfaces and got the following results:

> Without fix
log(0.584158/0.00798384)/log(2)) = 6.193132015607311
WIth fix:
log(0.585889/0.00798296)/log(2) = 6.197559771311222

This scenario uses two materials but no free-surface conditions.

sebwolf-de:
Regarding Materials and Free Surface:
I compared LOH1 receiver-data between this PR and commit 19037b (before we started breaking things). They are identical up to an relative error of 1e-14.
Screenshot from 2020-06-16 09-28-35

PR #226, opened 2020-07-17
Add normalize button to Viewrec

sebwolf-de:
Hi, I have to compare traces of SeisSol with the ones from collaborators who use other codes. Sometimes it is unnecessarily complicated to think about comparisons of magnitudes and we are just interested in the wave form. That's why I added a normalize button to viewrec. Maybe that's useful for the one or other...


krenzland:
Is normalization by L_inf norm the usual practice?

sebwolf-de:
At least thats what we did for our code comparison

krenzland:
Okay. We can come back to this later if we'd need other norm options.

sebwolf-de:
Yes, you could switch to a Dropdown to distinguish between different norms...

PR #227, opened 2020-07-18
Ravil/clustering info

ravil-mobile:
Add clustering info into the output. Updated reference to xdmfwriter submodule

### DIFF
--- a/src/ResultWriter/WaveFieldWriter.h
+++ b/src/ResultWriter/WaveFieldWriter.h
@@ -202,7 +202,7 @@ public:
 	 * @param timeTolerance The tolerance in the time for ignoring duplicate time steps
 	 */
 	void init(unsigned int numVars, int order, int numAlignedDOF,
-			const MeshReader &meshReader,
+			const MeshReader &meshReader,  const unsigned *ClusteringFieldMap,
 			const double* dofs,  const double* pstrain, const double* integrals,
 			unsigned int* map,
 			int refinement, int* outputMask, double* outputRegionBounds,
--- a/src/ResultWriter/WaveFieldWriter.h
+++ b/src/ResultWriter/WaveFieldWriter.h
@@ -202,7 +202,7 @@ public:
 	 * @param timeTolerance The tolerance in the time for ignoring duplicate time steps
 	 */
 	void init(unsigned int numVars, int order, int numAlignedDOF,
-			const MeshReader &meshReader,  const unsigned *ClusteringFieldMap,
+			const MeshReader &meshReader,  const std::vector<unsigned> &LtsClusteringData,
 			const double* dofs,  const double* pstrain, const double* integrals,
 			unsigned int* map,
 			int refinement, int* outputMask, double* outputRegionBounds,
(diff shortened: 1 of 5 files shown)


ravil-mobile:
Hi Thomas,

1) Yes, I can change naming if it requires.
2) I also noticed some alignment issues, however, it seems to me that it appears only on github. I haven't seen any problems with it in both Clion and Vim. 
3) Yes, it is getting written only once before a simulation starts. It is difficult to introduce any kind of switch (or at least it will require some rework of the source code and our submodules). I don't know why but somebody hard-coded the initialization of results writers in the original source code in a way: 
```
// WaveFieldWriterExecutor.h:189
m_waveFieldWriter->init(variables, std::vector<const char*>());

// XdmfWriter.h:154
void init(const std::vector<const char*> &cellVariableNames, const std::vector<const char*> &vertexVariableNames, bool useVertexFilter = true, bool writePartitionInfo = true);
```

For example, *writePartitionInfo* is optional for XdmfWriter but SeisSol always set to *true*. There is no way to switch it on or off. Probably, it is a headache to propagate these options to the SeisSol parameter text file since it has to go through Fortran.

src/Solver/Interoperability.cpp
```diff
@@ -802,11 +802,18 @@ void seissol::Interoperability::initializeIO(
 
   constexpr auto numberOfQuantities = tensor::Q::Shape[ sizeof(tensor::Q::Shape) / sizeof(tensor::Q::Shape[0]) - 1];
 
+  // record the clustering info i.e., distribution of elements within an LTS tree
+  const std::vector<Element>& MeshElements = seissol::SeisSol::main.meshReader().getElements();
+  std::vector<unsigned> ClusteringFieldMap(MeshElements.size());
+  for (const auto& Element: MeshElements) {
+    ClusteringFieldMap[Element.localId] = m_ltsLut.cluster(Element.localId);
```
krenzland, line 809: Are you sure that this tracks the global cluster id and not the local one?  (They can differ!)
It probably tracks the global one but I'm not sure.

src/ResultWriter/WaveFieldWriter.h
```diff
@@ -202,7 +202,7 @@ class WaveFieldWriter : private async::Module<WaveFieldWriterExecutor, WaveField
 	 * @param timeTolerance The tolerance in the time for ignoring duplicate time steps
 	 */
 	void init(unsigned int numVars, int order, int numAlignedDOF,
-			const MeshReader &meshReader,
+			const MeshReader &meshReader,  const unsigned *ClusteringFieldMap,
```
krenzland, line 205: It's probably more idiomatic to pass a const ref to the std::vector instead.

src/Solver/Interoperability.cpp
ravil-mobile: I've fixed it. 

Regarding tabs and white spaces. I can see that it is getting messy on github. I set my IDE to use spaces instead of tabs. Which setting do you use?

src/ResultWriter/WaveFieldWriter.h
ravil-mobile: Done

src/Solver/Interoperability.cpp
uphoffc: Think much of the C++ code uses two spaces. But in total it's a mess.

PR #230, opened 2020-08-20
Added fty to SeisSol as a submodule

ravil-mobile:
* reads parameters.par from disk and stores the info
  in a hash table i.e., YAML::Node

* FTY throws an exception if a NAMELIST block is not closed. Example: 
```
...
Thu Aug 20 18:37:02, Info:  Using OMP with #threads/rank: 6 
Thu Aug 20 18:37:02, Info:  The stack size ulimit is unlimited. 
Thu Aug 20 18:37:02, Error: ParsingError: cannot find the end of block: &Debugging 
Backtrace:
./SeisSol_Release_dhsw_elastic_6() [0x424538]
./SeisSol_Release_dhsw_elastic_6() [0x40f117]
...
```

* In case of success, SeisSol prints data retrieved from parameters.par stored in a YAML::Node:
```
...
Thu Aug 20 18:37:34, Info:  The stack size ulimit is unlimited. 
Thu Aug 20 18:37:34, Info:  Input Parameters:
 equations:
  materialfilename: material.yaml
boundaries:
  bc_fs: 1
  bc_dr: 1
  bc_of: 1
dynamicrupture:
  fl: 2
  modelfilename: fault.yaml
  gpwise: 1
  xref: 0.0
  yref: -1.0e5
  zref: 0
  rf_output_on: 1
  outputpointtype: 5
elementwise:
  printintervalcriterion: 2
  printtimeinterval_sec: 1.
  outputmask: 1 1 1 1 1 1 1 1 1  1
  refinement_strategy: 2
  refinement: 1
pickpoint:
  printtimeinterval: 1
  outputmask: 1 1 1 1
  noutpoints: 9
  ppfilename: FaultReceivers5.dat
meshnml:
  meshgenerator: PUML
  meshfile: mesh/tpv5.h5
discretization:
  order: 4
  material: 1
  cfl: 0.5
  fixtimestep: 5
  dgmethod: 1
  clusteredlts: 2
output:
  outputfile: output/tpv5
  ioutputmask: 1 1 1 1 1 1 1 1 1
  ioutputmaskmaterial: 1 1 1
  format: 10
  timeinterval: 0.25
  printintervalcriterion: 2
  pickdt: 0.005
  pickdttype: 1
  faultoutputflag: 1
  nrecordpoints: 6
  rfilename: Receivers5.dat
abortcriteria:
  endtime: 5.0 
...
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,3 +34,6 @@
 [submodule "submodules/eigen3"]
 	path = submodules/eigen3
 	url = ../../eigenteam/eigen-git-mirror
+[submodule "submodules/fty"]
+	path = submodules/fty
+	url = https://github.com/ravil-mobile/fty.git
--- a/src/Initializer/MemoryManager.h
+++ b/src/Initializer/MemoryManager.h
@@ -85,6 +85,7 @@
 #include <Initializer/DynamicRupture.h>
 #include <Initializer/Boundary.h>
 #include <Initializer/ParameterDB.h>
+#include <yaml-cpp/yaml.h>
 
 namespace seissol {
   namespace initializers {
@@ -161,6 +162,8 @@ class seissol::initializers::MemoryManager {
     Boundary m_boundary;
 
     EasiBoundary m_easiBoundary;
+
+    YAML::Node m_inputParams;
 
     /**
      * Corrects the LTS Setups (buffer or derivatives, never both) in the ghost region
@@ -290,6 +293,10 @@ class seissol::initializers::MemoryManager {
 
     inline EasiBoundary* getEasiBoundaryReader() {
       return &m_easiBoundary;
+    }
+
+    inline void setInputParams(const YAML::Node& Params) {
+      m_inputParams = Params;
     }
 };
 
--- a/src/SeisSol.cpp
+++ b/src/SeisSol.cpp
@@ -167,4 +167,16 @@ void seissol::SeisSol::finalize()
 	logInfo(rank) << "SeisSol done. Goodbye.";
 }
 
+
+void seissol::SeisSol::setInputParams(const YAML::Node& Params) {
+  m_inputParams = Params;
+  const int rank = MPI::mpi.rank();
+  if (rank == 0) {
+    logInfo() << "Input Parameters:\n"
+              << m_inputParams;
+  }
+
+  m_memoryManager.setInputParams(m_inputParams);
+}
+
 seissol::SeisSol seissol::SeisSol::main;
--- a/src/SeisSol.h
+++ b/src/SeisSol.h
@@ -60,6 +60,8 @@
 
 #include "ResultWriter/AnalysisWriter.h"
 
+#include <yaml-cpp/yaml.h>
+
 class MeshReader;
 
 namespace seissol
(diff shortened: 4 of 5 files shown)


uphoffc:
Why would you store the parsed namelist file in YAML::node if the only thing you need is

`std::unordered_map<std::string, std::unordered_map<std::string, std::string>>`

Moreover, if you think of XML, using YAML::Node is equivalent employing a DOM parser. However, I'd rather use SAX-style parsing and directly translate the key-value pairs to internal data structures. E.g. (that's not the way to do it, just to get the idea):

```c++
void parseBlock() {
  ...
  auto [key, value] = parseLine();
  callBlockTranslator(key, value);
  ...
}

struct PickpointBlockTranslator {
  blockUserCallback(std::string_view key, std::string_view value) {
    if (key == "printtimeinterval") {
       if (value.to_lower() == "yes" || value.to_lower() == "true") {
          block.printtimeinterval = true;
       } else {
          block.printtimeinterval = false;
       } else if (key == "outputmask") {
           ....
       }
       ....
    }
  }
  class PickPointBlock {
    bool printtimeinterval;
    std::array<double, 4> outputmask;
    int noutpoints;
    std::string ppfilename;
  };
  PickPointBlock block;
};
```

Above can be generalized easily, e.g. using pointers-to-members: https://blog.molecular-matters.com/2014/01/23/translating-a-human-readable-json-like-data-format-into-c-structs/

I implemented a data translator recently for TOML files (for a DOM parser, but that doesn't change much). Besides directly loading the data directly into internal structures, you can also easily add configuration file validation with such an approach, see e.g. https://github.com/TEAR-ERC/tandem/blob/master/app/poisson.cpp#L63

ravil-mobile:
Hi, Thanks for your comments

> Why would you store the parsed namelist file in YAML::node if the only thing you need is

I think we will use yaml text format for input data once fortran will be deprecated. So, the easiest way to read yaml fill be by using yaml-cpp. **_FTY just mimics_** that (as if) we are loading from a yaml file. In fact, FTY lib is intended to get deprecated in the future i.e., it is just a temporary solution. 

> std::unordered_map<std::string, std::unordered_map<std::string, std::string>>

It depends. Maybe somebody will come to a better (more logical) structure of the input file which may result in additional nesting of attributes

krenzland:
I think that Ravil's general approach is good for our purposes for the following reasons:

- This is only supposed to be a temporary solution. While I do agree that your TOML parser is a very good implementation it's overkill to do something like this here. The time would be better spent trying to entangle the Fortran mess and to move more and more of the configuration to C++.
- Using YAML Nodes as a data structure does seem weird but it has the advantages, that it allows us to:
   - Add more hierarchical information to our configuration (i.e. not a flat list of key,values). This allows us to reorganize the configuration a bit.
   - We can add an additional, (at first) optional parameters.yaml that can be easily merged with the current configuration. This is something that I am interested in. For example, collaboration partners have indicated that they want to be able to configure the initial conditions/boundary conditions for performing parameter studies. This approach would make this trivial.

But I do agree with your concerns. Temporary solutions often become permanent and we should strife to avoid this here. But in general, it's quite hard to write code that's worse than our current solution.

uphoffc:
Ok, but what kind of implementation do you want to have eventually? If you want to use it I can also make the Schema classes a standalone library and add callbacks for a SAX parser?

sebwolf-de:
I guess we can close this PR. Since #625, we already have fty as a submodule.

PR #234, opened 2020-09-14
Ravil/gpu merge

ravil-mobile:
This request is the first stage of merging the GPU branch
* Updated reference to yateto
* Extended cmake input parameters
* Merged with Lukas's cmake branch
* Fixed Travis CI issues with cmake


.travis.yml
```diff
@@ -15,54 +15,57 @@ before_install:
...
+    fi
+        dirname="build_${equation}_${SEISSOL_PRECISION}_${SEISSOL_BUILD_TYPE}"
+        echo "mkdir ${dirname}"
+        mkdir -p $dirname && cd $dirname
+        CMAKE_PREFIX_PATH=~ PKG_CONFIG_PATH=~/lib/pkgconfig/: cmake -DNETCDF=OFF -DMETIS=ON -DCOMMTHREAD=ON -DASAGI=OFF -DHDF5=ON -DCMAKE_BUILD_TYPE=$SEISSOL_BUILD_TYPE -DTESTING=ON -DLOG_LEVEL=warning -DLOG_LEVEL_MASTER=info -DHOST_ARCH=hsw -DPRECISION=$SEISSOL_PRECISION -DEQUATIONS=$equation -DNUMBER_OF_MECHANISMS=$mechanisms -DGEMM_TOOLS_LIST=LIBXSMM ..
+        make -j $(nproc)
+        make test
+        cd ..
+    done
+    set +u
```
krenzland, line 71: Is there a reason why you have to set +u at the end? I'd assume that it's not doing anything here?

Documentation/compilation.rst
```diff
@@ -5,15 +5,14 @@ In order to run SeisSol, you need to first install:
 
 -  Python (>= 3.5)
 -  Numpy (>= 1.12.0)
--  SCons (>= 3.0, for instructions see below)
 -  hdf5 (>= 1.8, for instructions see below)
 -  netcdf (C-Release) (>= 4.4, for instructions see below)
--  Intel compiler (>= 17.0, icc, icpc, ifort) or GCC (>= 5.0, gcc, g++, gfortran)
+-  Intel compiler (>= 18.0, icc, icpc, ifort) or GCC (>= 7.0, gcc, g++, gfortran)
```
krenzland, line 10: Should we maybe require gcc >= 8 here? As this one is used by travis.

Documentation/compilation.rst
```diff
@@ -5,15 +5,14 @@ In order to run SeisSol, you need to first install:
...
 -  Some MPI implementation (e.g. OpenMPI)
 -  ParMETIS for partitioning
 -  libxsmm (libxsmm\_gemm\_generator) for small matrix multiplications
 -  PSpaMM (pspamm.py) for small sparse matrix multiplications (required only on Knights Landing or Skylake)
--  CMake (for compiling submodules ImpalaJIT and yaml-cpp)
+-  CMake (>3.15), for compiling submodules ImpalaJIT and yaml-cpp, and for SeisSol itself
```
krenzland, line 15: If cmake >3.15 is needed, this should be stated in the root level cmakefile -> otherwise it will fail with weird errors

Documentation/compilation.rst
```diff
@@ -29,23 +28,23 @@ Add the following lines to your .bashrc (vi ~/.bashrc).
...
 ----------------
 
 .. code-block:: bash
 
-  wget http://prdownloads.sourceforge.net/scons/scons-3.0.5.tar.gz
-  tar -xaf scons-3.0.5.tar.gz
-  cd scons-3.0.5
-  python setup.py install --prefix=$HOME
-  cd ..
+  wget -qO- https://github.com/Kitware/CMake/releases/download/v3.16.4/cmake-3.16.4-Linux-x86_64.tar.gz | tar -xvz -C "/" && mv "/cmake-3.16.4-Linux-x86_64" "${HOME}/bin/cmake"
```
krenzland, line 42: This uses cmake 3.16 - but documentation says 3.15 is needed. Maybe we should decide on one of both for docs/travis

cmake/FindGemmTools.cmake
```diff
@@ -49,6 +49,12 @@ foreach(component ${_GEMM_TOOLS_LIST})
     elseif ("${component}" STREQUAL "Eigen")
         # already included by default!
 
+    elseif ("${component}" STREQUAL "GemmForge")
+        execute_process(COMMAND python -c "import gemmforge; gemmforge.print_cmake_path()"
```
krenzland, line 53: It would be better style to use https://cmake.org/cmake/help/v3.16/module/FindPython.html for this and specify that we want to have python3. Your version would use python2 on most systems. (I assume gemmforge is python3, as python2 is basically dead)

cmake/cpu_arch_flags.cmake
```diff
@@ -47,6 +47,11 @@ function(get_arch_flags architecture compiler)
             set(CPU_ARCH_FLAGS "-mcpu=thunderx2t99" PARENT_SCOPE)
         endif()
 
+    # IBM power 9
+    elseif ("${HOST_ARCH}" STREQUAL "power9")
+        set(HAS_REDZONE OFF PARENT_SCOPE)
+        set(CPU_ARCH_FLAGS "-mtune=power9" PARENT_SCOPE)
```
krenzland, line 53: Should be -march=power9 (as on all archs except on arm!). Or is this similar to arm? In this case -mcpu should be used...

cmake/process_users_input.cmake
```diff
@@ -21,20 +21,31 @@ set_property(CACHE ORDER PROPERTY STRINGS ${ORDER_OPTIONS})
...
 
 
-set(ARCH "hsw" CACHE STRING "Type of the target architecture")
-set(ARCH_OPTIONS noarch wsm snb hsw knc knl skx thunderx2t99)
-set(ARCH_ALIGNMENT   16  16  32  32  64  64  64 16)  # size of a vector registers in bytes for a given architecture
-set_property(CACHE ARCH PROPERTY STRINGS ${ARCH_OPTIONS})
+set(HOST_ARCH "hsw" CACHE STRING "Type of the target host architecture")
+set(HOST_ARCH_OPTIONS noarch wsm snb hsw knc knl skx thunderx2t99 power9)
+# size of a vector registers in bytes for a given architecture
+set(HOST_ARCH_ALIGNMENT   16  16  32  32  64  64  64           16     16)
```
krenzland, line 32: Spacing is really strange here

cmake/process_users_input.cmake
```diff
@@ -21,20 +21,31 @@ set_property(CACHE ORDER PROPERTY STRINGS ${ORDER_OPTIONS})
...
+set_property(CACHE HOST_ARCH PROPERTY STRINGS ${HOST_ARCH_OPTIONS})
+
+
+set(DEVICE_ARCH "none" CACHE STRING "Type of the target compute architecture")
+set(DEVICE_ARCH_OPTIONS    none nvidia amd_gpu)
+set(DEVICE_ARCH_ALIGNMENT  none     64     128)
```
krenzland, line 38: Please avoid introducing new aligned lines, as changing them destroys the git diffs

cmake/process_users_input.cmake
```diff
@@ -67,13 +78,8 @@ set(LOG_LEVEL_MASTER_OPTIONS "debug" "info" "warning" "error")
 set_property(CACHE LOG_LEVEL_MASTER PROPERTY STRINGS ${LOG_LEVEL_MASTER_OPTIONS})
 
 
-set(ACCELERATOR_TYPE "NONE" CACHE STRING "type of accelerator")
-set(ACCELERATOR_TYPE_OPTIONS NONE GPU)
-set_property(CACHE ACCELERATOR_TYPE PROPERTY STRINGS ${ACCELERATOR_TYPE_OPTIONS})
-
-
 set(GEMM_TOOLS_LIST "LIBXSMM,PSpaMM" CACHE STRING "choose a gemm tool(s) for the code generator")
```
krenzland, line 81: Maybe eigen should be the default here as it works on all platforms?

generated_code/aderdg.py
```diff
@@ -210,46 +211,73 @@ def addInit(self, generator):
...
-    qShape = (self.numberOf3DBasisFunctions(), self.numberOfQuantities())
-    dQ0 = OptionalDimTensor('dQ(0)', self.Q.optName(), self.Q.optSize(), self.Q.optPos(), qShape, alignStride=True)
-    power = Scalar('power')
-    derivatives = [dQ0]
-    generator.add('derivativeTaylorExpansion(0)', self.I['kp'] <= power * dQ0['kp'])
-    for i in range(1,self.order):
-      derivativeSum = Add()
+  def addLocal(self, generator, platforms):
+    for platform in platforms:
+      name_prefix = generate_kernename_prefix(platform)
```
krenzland, line 216: Typo in method name

generated_code/aderdg.py
```diff
@@ -40,6 +40,7 @@
 
 import numpy as np
 from abc import ABC, abstractmethod
+from common import *
```
krenzland, line 43: We should try and avoid from common import *, especially if we are only importing one function. Better to stay explicit when importing something into the global space.

krenzland: Same of course for other instances of this line.

generated_code/aderdg.py
```diff
@@ -174,15 +175,15 @@ def addInit(self, generator):
...
 
   @abstractmethod
-  def addNeighbor(self, generator):
+  def addNeighbor(self, generator, platforms):
     pass
 
   @abstractmethod
-  def addTime(self, generator):
+  def addTime(self, generator, platforms):
     pass
```
krenzland, line 187: Shouldn't platforms be a class member? In this case we do not need to pass it around every time. Mabye also true for the generator...

generated_code/aderdg.py
krenzland: Not sure if I like that pattern though -> it leads to a lot of duplicate code. Don't have a better idea right now...

generated_code/generate.py
```diff
@@ -73,22 +75,38 @@
...
-arch = useArchitectureIdentifiedBy(cmdLineArgs.arch)
+
+if cmdLineArgs.device_arch == 'none':
+  compute_arch = cmdLineArgs.host_arch
+  compute_sub_arch = None
+  host_arch = cmdLineArgs.host_arch
+else:
+  compute_arch = cmdLineArgs.device_arch
+  compute_sub_arch = cmdLineArgs.device_sub_arch
+  host_arch = cmdLineArgs.host_arch
```
krenzland, line 94: host_arch should be set outside of the if clause

src/Initializer/GlobalData.cpp
```diff
@@ -47,6 +47,14 @@
 
 namespace init = seissol::init;
 
+class CustomCopyPolicy {
+public:
+  float_t* copy(float_t const* first, float_t const* last, float_t*& mem) {
```
krenzland, line 52: Where does float_t here actually come from?

krenzland: And: What's the difference to real that we already have?

krenzland (changes requested):
Looks good mostly! Note: I did not comment on my own cmake changes, someone else should probably do that

.travis.yml
ravil-mobile: It is actually needed. *u* in the following line _set -euo pipefail_ forces bash to treat unset variables and parameters other than the special parameters ‘@’ or ‘*’ as an error when performing parameter expansion.There are some unused parameters in bash functions used by Travis. Therefore, I had to restore the settings back. This Travis bug has been reported and mentioned in some threads dedicated to Travis CI. Currently, it is the solution.

cmake/process_users_input.cmake
ravil-mobile: But it is easy to understand which alignment belongs to which architecture. It is done on purpose

cmake/process_users_input.cmake
ravil-mobile: I didn't get you

.travis.yml
krenzland: I see. I knew what +euo pipefail does - I never write a bash script without, as not setting it can hide errors. Didn't know that that breaks travis. Thanks for figuring this out!

Thomas-Ulrich:
Hi, Just trying to build this branch on Shaheen and getting:

```
CMake Error at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
   Could NOT find ParMETIS (missing: PARMETIS_TEST_RUNS) (found version "4.0")
 Call Stack (most recent call first):
   /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
   cmake/FindParMETIS.cmake:120 (find_package_handle_standard_args)
   CMakeLists.txt:218 (find_package)
```
I run the following command:
```
ccmake .. -DCMAKE_PREFIX_PATH=/sw/xc40cle7/parmetis/4.0.3/cle7_gnu8.3.0;/sw/xc40cle7/metis/5.1.0/cle7_gnu8.3.0/
```
With the following module loaded:
```
x_ulricht@cdl2:/project/k1343/ulrich/SeisSol/build_cmake> module display metis
-------------------------------------------------------------------
/sw/xc40cle7/modulefiles/metis/5.1.0:

module-whatis    METIS : Serial Graph Partitioning and Fill-reducing Matrix Ordering 
module           swap PrgEnv-gnu PrgEnv-gnu 
setenv           METIS_DIR /sw/xc40cle7/metis/5.1.0/cle7_gnu8.3.0 
prepend-path     PATH /sw/xc40cle7/metis/5.1.0/cle7_gnu8.3.0/bin 
prepend-path     MANPATH /sw/xc40cle7/metis/5.1.0/cle7_gnu8.3.0/share/man 
prepend-path     LD_LIBRARY_PATH /sw/xc40cle7/metis/5.1.0/cle7_gnu8.3.0/lib 
append-path      INCLUDE /sw/xc40cle7/metis/5.1.0/cle7_gnu8.3.0/include 
-------------------------------------------------------------------
x_ulricht@cdl2:/project/k1343/ulrich/SeisSol/build_cmake> module display parmetis/4.0.3 
-------------------------------------------------------------------
/sw/xc40cle7/modulefiles/parmetis/4.0.3:

module-whatis    ParMETIS : Parallel Graph Partitioning and Fill-reducing Matrix Ordering 
setenv           PARMETIS_DIR /sw/xc40cle7/parmetis/4.0.3/cle7_gnu8.3.0 
prepend-path     PATH /sw/xc40cle7/parmetis/4.0.3/cle7_gnu8.3.0/bin 
prepend-path     MANPATH /sw/xc40cle7/parmetis/4.0.3/cle7_gnu8.3.0/share/man 
prepend-path     LD_LIBRARY_PATH /sw/xc40cle7/parmetis/4.0.3/cle7_gnu8.3.0/lib 
append-path      INCLUDE /sw/xc40cle7/parmetis/4.0.3/cle7_gnu8.3.0/include 
-------------------------------------------------------------------
````
I wonder if I m doing something wrong or not?

File: generated_code/common.py [hunk 15]
```diff
@@ -0,0 +1,2 @@
+def generate_kernename_prefix(target):
+  return f'{target}_' if target == 'gpu' else ''
```
krenzland, line 2: typo in name -> l in kernel is missing

krenzland (changes requested):
Would be great if you actually look at the proposed changes... most were from my last review.

krenzland:
> Hi, Just trying to build this branch on Shaheen and getting:
> PS: The problem was that libmetis was not detected. I then needed to manually set it to: /sw/xc40cle7/metis/5.1.0/cle7_gnu8.3.0/lib/libmetis.so

Did this work on master? I don't see anything here that actually changes how metis is handled. I don't have access on Shaheen and can't reproduce.

Thomas-Ulrich:
> > Hi, Just trying to build this branch on Shaheen and getting:
> > PS: The problem was that libmetis was not detected. I then needed to manually set it to: /sw/xc40cle7/metis/5.1.0/cle7_gnu8.3.0/lib/libmetis.so
> 
> Did this work on master? I don't see anything here that actually changes how metis is handled. I don't have access on Shaheen and can't reproduce.

Hi, I used the branch because yaml-cpp and ImpalaJIT were not built, and I ve seen that this branch compiles also these libs.
But no, this does not work either on master.

generated_code/aderdg.py
```diff
@@ -210,46 +211,73 @@ def addInit(self, generator):
...
-    derivatives = [dQ0]
-    generator.add('derivativeTaylorExpansion(0)', self.I['kp'] <= power * dQ0['kp'])
-    for i in range(1,self.order):
-      derivativeSum = Add()
+  def addLocal(self, generator, targets):
+    for target in targets:
+      name_prefix = generate_kernename_prefix(target)
+      volumeSum = self.Q['kp']
+      for i in range(3):
+        volumeSum += self.db.kDivM[i][self.t('kl')] * self.I['lq'] * self.starMatrix(i)['qp']
```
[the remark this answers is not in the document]
uphoffc: This is an abomination. We already have a concept to distinguish architectures which is the Arch class -- you just need to call generate twice giving a CPU Arch object in one call and a GPU Arch object in another call.

This also relates to the comment in the yateto PR: Why don't you make "target" part of Arch?

cmake/FindGemmTools.cmake
ravil-mobile: Ideally, we need to check a presence of python3 at top of the main CMake file.

generated_code/aderdg.py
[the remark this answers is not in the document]
ravil-mobile: Not all tensor expression must be generated for GPU inside a kernel. Consider Dynamic rupture: only 1 out of 3 kernels can be generated for GPU.

krenzland:
>  But no, this does not work either on master.

Okay, then I'd propose we postpone this issue. I'll add it to my todo-list and come back to you once this PR is merged/NG is up again.

Documentation/compilation.rst
```diff
@@ -29,23 +28,23 @@ Add the following lines to your .bashrc (vi ~/.bashrc).
...
 ----------------
 
 .. code-block:: bash
 
-  wget http://prdownloads.sourceforge.net/scons/scons-3.0.5.tar.gz
-  tar -xaf scons-3.0.5.tar.gz
-  cd scons-3.0.5
-  python setup.py install --prefix=$HOME
-  cd ..
+  cd $(mktemp -d) && wget -qO- https://github.com/Kitware/CMake/releases/download/v3.10.2/cmake-3.10.2-Linux-x86_64.tar.gz | tar -xvz -C "." && mv "./cmake-3.10.2-Linux-x86_64" "${HOME}/bin/cmake"
```
krenzland, line 42: The whole command should be set in () brackets. This opens a subshell - the user remains in the same directory.

krenzland (changes requested):
Looks mostly good.
I still really dislike the loop over targets inside of each generator. Imo it should be outside and each generator should be called with one target.

krenzland (approved):

PR #241, opened 2020-10-05
update rconv installation and documentation

Thomas-Ulrich:


uphoffc:
Doesn't build if proj4 was not found even though there used to be a "noproj" option in the scons script. I would also delete the scons script if there is a cmakelists

Thomas-Ulrich:
1) Ok, but this is not related with the changes in my pull request ;-)
Concerning the `noproj` option, I just checked and it seems using :
```
-m `+proj=lonlat +datum=WGS84 +units=m` works.
```
So maybe we could drop this option?
2) then I also delete the SConscript?

uphoffc:
Just add the following to CMakeLists:

if(NOT PROJ4_FOUND)
    target_compile_definitions(SeisSol-rconv PRIVATE -Dnoproj)
    message(WARNING "Compiling without proj.4 support.")
endif()

Thomas-Ulrich:
ps: I really like your warning `prepare for trouble`.

uphoffc:
:-D

If you want to drop the "noproj" option then you should add REQUIRED to find_package(PROJ4...).
For simple cases it might be sufficient to use rconv without proj... if you know what you're doing...

Thomas-Ulrich:
This goes at the very end of the CMakeLists? Or it does not matter
```
if(NOT PROJ4_FOUND)
    target_compile_definitions(SeisSol-rconv PRIVATE -Dnoproj)
    message(WARNING "Compiling without proj.4 support.")
endif()
```

krenzland:
Hm, is this line in the documentation

>  `export LD_LIBRARY_PATH=path-to-proj.4/build/lib:$LD_LIBRARY_PATH` 

still useful? Iirc cmake should set the RPATH of the executable (i.e. where it looks for dyn. libraries) automatically. Not sure if this works though. Also should not work when installing.

(Not relevant to this PR, just wondering)

Thomas-Ulrich:
@krenzland: you are right, I didn't do that and it worked.

Thomas-Ulrich:
Ok I pushed all the required changes.

uphoffc (approved):

PR #243, opened 2020-10-08
Thomas/sim modeler cad

Thomas-Ulrich:
add SimModeler workflow to documentation.
(to have an up to date user documentation for the training workshop).


Documentation/cad-models.rst
```diff
@@ -4,10 +4,20 @@ CAD models
...
 Gocad.
 
-A simple tutorial
------------------
 
-See :doc:`generating-a-cad-model-using-gocad-basic-tutorial`.
+SimModeler CAD workflow
+-----------------------
+
+Since September 2019, SimModeler features powerfull tools for processing discrete data (geometry in the form of meshes), which allow building structural models without the need to rely on additionnal CAD software.
```
sebwolf-de, line 11: powerful~~l~~ addition~~n~~al

Documentation/cad-models.rst
```diff
@@ -4,10 +4,20 @@ CAD models
...
+We illustrate the SimModeler CAD workflow by building the structural model of the Palu earthquake dynamic rupture scenario (Ulrich et al., 2019).
+See :doc:`simmodelerCAD-workflow`.
+
+GOCAD CAD workflow
+------------------
+
+GOCAD is the tool we historically used to to process complex geophysical data into structural models.
+Since then, SimModeler developped tools for processing discrete data, in particular a discrete surface intersection algorithm, which is much faster and more reliable than the one from GoCAD.
```
sebwolf-de, line 19: develop~~p~~ed

Documentation/cad-models.rst
```diff
@@ -4,10 +4,20 @@ CAD models
...
+We therefore recommand the use of the SimModeler workflow. Because GoCAD may still be useful for fine processing of surface data, we detail the full GoCAD workflow at: :doc:`generating-a-cad-model-using-gocad-basic-tutorial`.
```
sebwolf-de, line 20: recomm_e_nd

Documentation/cad-models.rst
```diff
@@ -4,10 +4,20 @@ CAD models
 The following help pages describe how to build a structural model with
```
sebwolf-de, line 4: Maybe add some remark about when to use Simmodeler and when to use Gocad here.

Documentation/simmodelerCAD-workflow.rst
```diff
@@ -0,0 +1,141 @@
...
+The scripts first resample the 2D fault trace and then can smooth them.
+Finally, the smoothed and resampled traces are swept towards negative and positive (if the topography has positive elevation) z.
+
+createFaultFromCurve.py takes 3 main arguments: ``filename``, ``dipType`` and ``dipDesc``.  
+
+- ``filename`` is the name of the ASCII file describing the trace.  
+- ``dipType`` allow switching between a constant (0), an along-depth dependant (1) or an along-strike dependent (2).
+- ``dipDesc`` gives either the dip angle value (dipType=0) or the 1D variation of the dip angle (dipType=1 or 2).
+
+.. figure:: https://www.geophysik.uni-muenchen.de/~ulrich/dipDescription.png
```
Thomas-Ulrich, line 74: should we move all image of the doc to imgur?

sebwolf-de: Why should we move it there?

Thomas-Ulrich: because we don't want large files on a git repository?
And because if someday I leave LMU and my personal account gets closed, the images won't be available anymore?

Documentation/cad-models.rst
Thomas-Ulrich: ok you are right, I checked also this file with Grammarly now. Thx.

Documentation/cad-models.rst
Thomas-Ulrich: I think gocad can still be useful for some complex surface processing, such as surface smoothing with constrains. But this may be too technical, so I leave that unchanged.

Thomas-Ulrich: Ok I added (e.g. surface smoothing with constraints) as an example below.

Documentation/simmodelerCAD-workflow.rst
sebwolf-de: You are right, but we have already some pictures in the repo, see Documentation/LatexFigures 
I guess https://git-lfs.github.com/ might be interesting.

Thomas-Ulrich: I know for Documentation/LatexFigures, this wouldn't have append if we would have merged the cookbook through pull request (anyway, the files are not extremely large).
https://git-lfs.github.com/ looks like a good option.

sebwolf-de: So I suppose you check in your pictures to the git and then we do a git lfs migrate afterwards.

Thomas-Ulrich:
Done

sebwolf-de (approved):

Thomas-Ulrich:
Hi,
By mistake (I wonder how), I included 2 files to the pull request, and happily, they broke the build so I could notice them.
I then reverted them to origin/master in a new commit and pushed them to the branch.

krenzland:
Maybe we should perform a "squash" merge in this case (to keep things a bit cleaner).
 @sebwolf-de does this still look like the version that you've reviewed?

sebwolf-de (approved):

PR #244, opened 2020-10-09
Adding scripts to view fault outputs

Bo-Li-LMU:
Here I uploaded the scripts to view the fault outputs of SeisSol DR models. 
Python script: to plot the output for one simulation. Need to add the folder named "submodules", which includes the "pythonXdmfReader", to your working directory
Matlab script: to plot a list of models with same structure.


krenzland:
Without actually reviewing the PR, I can see the following issues:
- You've commited many pyc files
- You actually didn't add new submodules, but rather the entire content of the submodules (if I see it correctly)
- There should be no need to add submodule dirs to the path, you can change the python path in the script.

Bo-Li-LMU:
Removed the submodules folder and updated with the link to the pythonXdmfReader that Thomas created before

git-taufiq:
I got this error when trying this code to plot Palu training output. Could you check the data shape for En and Mo_rate?

`start plotting the moment rate & calculating the magnitude
Traceback (most recent call last):
  File "Faultoutputview.py", line 34, in <module>
    Mo_rate[:,0]=Mo_rate[:,0]+En[:,1]
ValueError: operands could not be broadcast together with shapes (86560,) (43360,)`

Bo-Li-LMU:
Ah, you know yesterday we discussed this. The previous script is to deal with the -Enf* output with same time step. I already updated the code to account for different time steps, you can use the new python one

git-taufiq:
I think you have not push it yet, it is still the previous version in your repo.

krenzland:
As there hasn't been an update in over 2 years, I'll close this PR for now. Feel free to reopen if this script is actively being developed.

PR #247, opened 2020-10-12
refactor script

Thomas-Ulrich:
Sorry for the previous pull request (was not clean).

This is a script for generating a vtk coastline file.
I just did some refactoring to be able to apply a translation on the coordinates.
This does not need a thorough review IMO.

### DIFF
--- a/postprocessing/visualization/tools/CreateVtkCoastLineFromGmt.py
+++ b/postprocessing/visualization/tools/CreateVtkCoastLineFromGmt.py
@@ -1,92 +1,97 @@
 ###
-#Author: Thomas Ulrich, LMU, 24.09.2015
-#Read a coastline file from GMT
-#Create a vtk file from that input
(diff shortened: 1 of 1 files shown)


sebwolf-de:
I guess np.savetxt does the same as your for loops...

sebwolf-de (approved):

Thomas-Ulrich:
> I guess np.savetxt does the same as your for loops...

The updated code uses np.savetxt.
(actually I needed to add the translation vector to the geometry array, so I used numpy instead of lists and then I could use savetxt),

krenzland:
Your code would be clearer if you'd use the following pattern:
```python
with open('pathToFile.ext', 'w') as f:
    f.write(...)
```

This automatically closes the file handle for you - which avoids errors. Probably no need to change here.

Thomas-Ulrich:
thx I will try to remember that.

PR #248, opened 2020-10-12
Added a search for proj4 via pkg-config for rconv cmake

ravil-mobile:
Rationale: some versions of *proj4* don't generate cmake-find files. Instead, proj4 generates proj.pc file which follows pkg-config format.  To cope with it and support older versions of proj4 (<4.9.3), it is necessary to import pkg-config module inside of CMakeLists.txt and try to find the package using it.


sebwolf-de (approved):
lgtm

ravil-mobile:
Shall we merge?

uphoffc:
Why don't you use INCLUDE_DIRS from pkg-config? Then you don't need to duplicate target_...

https://cmake.org/cmake/help/v3.0/module/FindPkgConfig.html

PR #254, opened 2020-10-14
update compilation and supermuc doc

Thomas-Ulrich:


Documentation/supermuc.rst
```diff
@@ -116,9 +112,9 @@ set compiler options:
 
 ::
 
-  $ export FC=mpif90
-  $ export CXX=mpiCC
-  $ export CC=mpicc
+  $ export FC=mpiifort
+  $ export CXX=mpiicpc
+  $ export CC=mpiicc
```
krenzland, line 117: Just a note: With cmake it is no longer required to use the mpi wrapper compilers, e.g. using gcc directly also works.

krenzland (approved):

PR #257, opened 2020-10-16
Update documentation

git-taufiq:
update documentation, adding tpv34 and palu example


Thomas-Ulrich:
Thx for this pull request. 
I see many things that could be integrated into the documentation.
But now the problem is the overlap with 
https://seissol.readthedocs.io/en/latest/simmodelerCAD-workflow.html
I, therefore, suggest you try merging with the existing page before we review the pull request.
You could also open a new pull request with everything except Palu.

git-taufiq:
Thanks! yes sure, the Palu page mainly about meshing and they are already covered in  <https://seissol.readthedocs.io/en/latest/simmodelerCAD-workflow.html> that pulled about the same time. So I will make a new pull request without Palu page. I will close this pull.

PR #258, opened 2020-10-16
add script to refine temporally and spatially a kinematic model

Thomas-Ulrich:
This is the script I designed to refine temporally and spatially the kinematic model of Northridge.
Basically create more source points by interpolation, and smooth also the SR.
Needed for the training.

Thomas.

### DIFF
--- a/preprocessing/science/refine_srf.py
+++ b/preprocessing/science/refine_srf.py
@@ -1,0 +1,171 @@
+import numpy as np
+import pyproj
+import matplotlib.pyplot as plt
+import scipy.ndimage
+from scipy import interpolate
+from math import floor
+import os
+
+
+class FaultPlane:
+    def __init__(self, nx, ny):
+        self.nx = nx
+        self.ny = ny
+        self.ndt = 0
+        self.lon = np.zeros((ny, nx))
+        self.lat = np.zeros((ny, nx))
+        self.x = np.zeros((ny, nx))
+        self.y = np.zeros((ny, nx))
+        self.depth = np.zeros((ny, nx))
+        self.t0 = np.zeros((ny, nx))
+        self.slip1 = np.zeros((ny, nx))
+        self.strike = np.zeros((ny, nx))
+        self.dip = np.zeros((ny, nx))
+        self.rake = np.zeros((ny, nx))
+        self.aSR = 0
+        self.PSarea = 0
+        self.dt = 0
+        self.myt = 0
+
+    def init_aSR(self):
+        self.aSR = np.zeros((self.ny, self.nx, self.ndt))
+
+    def compute_xy_from_latlon(self, proj_string):
+        if args.proj != None:
+            lla = pyproj.Proj(proj="latlong", ellps="WGS84", datum="WGS84")
+            myproj = pyproj.Proj(proj_string)
+            self.x, self.y = pyproj.transform(lla, myproj, self.lon, self.lat)
+        else:
+            print("no proj string specified!")
+            self.x, self.y = self.lon, self.lat
+
+    def compute_latlon_from_xy(self, proj_string):
+        if args.proj != None:
+            lla = pyproj.Proj(proj="latlong", ellps="WGS84", datum="WGS84")
+            myproj = pyproj.Proj(proj_string)
+            self.lon, self.lat = pyproj.transform(myproj, lla, self.x, self.y)
+        else:
+            self.lon, self.lat = self.x, self.y
+
+    def compute_time(self):
+        self.myt = np.linspace(0, (self.ndt - 1) * self.dt, self.ndt)
+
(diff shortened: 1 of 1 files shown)


preprocessing/science/refine_srf.py
```diff
@@ -0,0 +1,174 @@
...
+                if ndt1 == 0:
+                    continue
+                if ndt1 > p1.ndt:
+                    print("this script assumes that ndt is not larger than the first ndt in the file", ndt1, p1.ndt)
+                    raise
+                while True:
+                    line = fid.readline()
+                    lSTF.extend(line.split())
+                    if len(lSTF) == ndt1:
+                        p1.aSR[j, i, 0:ndt1] = np.array([float(v) for v in lSTF])
```
sebwolf-de, line 110: Here we just read the slip rate in u_1 direction, right? What about the slip rate in u_2 and u_3 direction?

preprocessing/science/refine_srf.py
```diff
@@ -0,0 +1,174 @@
...
+        self.strike = np.zeros((ny, nx))
+        self.dip = np.zeros((ny, nx))
+        self.rake = np.zeros((ny, nx))
+        self.aSR = 0
+        self.PSarea = 0
+        self.dt = 0
+        self.myt = 0
+
+    def init_aSR(self):
+        self.aSR = np.zeros((self.ny, self.nx, self.ndt))
```
sebwolf-de, line 31: Only one sliprate, but according to srf documentation it is possible, that we have also three slip rates per subfault.

preprocessing/science/refine_srf.py
```diff
@@ -0,0 +1,174 @@
...
+        aSRa[:, :, k] = scipy.ndimage.zoom(p.aSR[:, :, k], spatial_zoom, order=spatial_order)
+
+    # interpolate temporally the AST
+    for j in range(p2.ny):
+        for i in range(p2.nx):
+            f = interpolate.interp1d(p.myt, aSRa[j, i, :], kind="quadratic")
+            p2.aSR[j, i, :] = f(p2.myt)
+            if p2.slip1[j, i] < 0:
+                p2.aSR[j, i, :] = 0
+                continue
```
sebwolf-de, line 149: Why do we do this? Is not completly clear to me. Maybe add a comment.

preprocessing/science/refine_srf.py
```diff
@@ -0,0 +1,174 @@
...
+                while True:
+                    line = fid.readline()
+                    lSTF.extend(line.split())
+                    if len(lSTF) == ndt1:
+                        p1.aSR[j, i, 0:ndt1] = np.array([float(v) for v in lSTF])
+                        break
+    return p1
+
+
+def upsampleFault(p, spatial_order, spatial_zoom, temporal_zoom):
```
sebwolf-de, line 115: Could be a member of the FaultPlane.

sebwolf-de (commented):
Good morning, looks good overall!
You are not consistent in the use of camelCase and snake_case regarding the naming of methods.

preprocessing/science/refine_srf.py
Thomas-Ulrich: After the cubic interpolation, slip1 may get negative which does not make sense.

preprocessing/science/refine_srf.py
Thomas-Ulrich: `Although	the	general	format	given	here	allows	for	slip	in	all	three	directions	(u1, u2, u3),	
most	cases	will	have	only	one	(u1)	or	two	(u1,	u2)	non-zero	slip	components.`
I will add a warning.

Thomas-Ulrich:
Thx Sebastian for the nice review! I think I answered it now.

preprocessing/science/refine_srf.py
```diff
@@ -0,0 +1,210 @@
...
+                        print("this script assumes that dt is the same for all sources", dt, self.dt)
+                        raise
+                    while True:
+                        line = fid.readline()
+                        lSTF.extend(line.split())
+                        if len(lSTF) == ndt1:
+                            self.aSR[j, i, 0:ndt1] = np.array([float(v) for v in lSTF])
+                            break
+
+    def upsample_fault(self, p, spatial_order, spatial_zoom, temporal_zoom):
```
sebwolf-de, line 149: I'd reverser the order: p2 = p.upsample_fault(...)

sebwolf-de (commented):
Good changes: Nice that also v2.0 is now also supported!

preprocessing/science/refine_srf.py
Thomas-Ulrich: adressed by last commits?

sebwolf-de (approved):

PR #261, opened 2020-10-19
Also compute relative error

sebwolf-de:
Hi, I'd like to also compute relative errors, when doing a convergence analysis.


src/ResultWriter/AnalysisWriter.cpp
```diff
@@ -200,11 +225,11 @@ void seissol::writer::AnalysisWriter::printAnalysis(double simulationTime) {
         } else {
           MPI_Recv(centerRecv, 3, MPI_DOUBLE,  errLInfRecv[i].rank, i, comm, MPI_STATUS_IGNORE);
         }
-        logInfo(mpi.rank()) << "L1  , var[" << i << "] =\t" << errL1MPI[i];
-        logInfo(mpi.rank()) << "L2  , var[" << i << "] =\t" << std::sqrt(errL2MPI[i]);
-        logInfo(mpi.rank()) << "LInf, var[" << i << "] =\t" << errLInfRecv[i].val
+        logInfo(mpi.rank()) << "L1  , var[" << i << "] =\t" << errL1MPI[i] << "	" << errL1MPI[i] / analyticalL1MPI[i];
```
krenzland, line 228: Why are you using this kind of spacing? Why no tabs? Would be also good use the same spacing everywhere?

Thomas-Ulrich:
Could you add in the documentation how to run such convergence analysis?
Is it related to this cube setup with periodic boundaries?

src/ResultWriter/AnalysisWriter.cpp
```diff
@@ -215,10 +240,10 @@ void seissol::writer::AnalysisWriter::printAnalysis(double simulationTime) {
...
             center);
 
-      logInfo() << "L1, var[" << i << "] =\t" << errL1Local[i];
-      logInfo() << "L2, var[" << i << "] =\t" << std::sqrt(errL2Local[i]);
-      logInfo() << "LInf, var[" << i << "] =\t" << errLInfLocal[i]
-          << "\tat [" << center[0] << ",\t" << center[1] << ",\t" << center[2] << "\t]";
+      logInfo() << "L1, var[" << i << "] =\t" << errL1Local[i] << "\t" << analyticalL1Local[i];
+      logInfo() << "L2, var[" << i << "] =\t" << std::sqrt(errL2Local[i]) << "\t" << std::sqrt(analyticalL2Local[i]);
+      logInfo() << "LInf, var[" << i << "] =\t" << errLInfLocal[i << "\t" << analyticalLInfLocal[i]]
+          << "\tat [" << center[0] << "	" << center[1] << "	" << center[2] << "\t]"
```
Thomas-Ulrich, line 246: not fixed tab issue here

krenzland:
> Could you add in the documentation how to run such convergence analysis?
> Is it related to this cube setup with periodic boundaries?

This is done automatically when running a simulation with certain initial conditions.
I don't think that it is documented currently but it's not in scope of this PR.
So I'd make this a separate issue - normal users won't be interested in convergence analysis anyway.

Thomas-Ulrich:
> > Could you add in the documentation how to run such convergence analysis?
> > Is it related to this cube setup with periodic boundaries?
> 
> This is done automatically when running a simulation with certain initial conditions.
> I don't think that it is documented currently but it's not in scope of this PR.
> So I'd make this a separate issue - normal users won't be interested in convergence analysis anyway.

Everything you put in the documentation is something you do not need to give an answer to later one, so the effort is not necessarily unworthy. But in the end, it is up to you.

krenzland:
No, I agree that we should add this. I just want to avoid the situation where a lot of extra work is attached to every PR. We want to merge stuff as fast as possible.

When you modify something that is documented, I agree that we should modify the documentation in the same PR.

src/ResultWriter/AnalysisWriter.cpp
sebwolf-de: Fixed, thanks for pointing this out!

sebwolf-de:
There is a PR in the pipeline connected to https://github.com/SeisSol/Examples/issues/10. Once the licence debate is over, I plan to merge a script (originally written by Carsten) to compute convergence rates and so forth. This will include docu.

Thomas-Ulrich:
> No, I agree that we should add this. I just want to avoid the situation where a lot of extra work is attached to every PR. We want to merge stuff as fast as possible.
> 
> When you modify something that is documented, I agree that we should modify the documentation in the same PR.

Ok I understand, but 
Is it a lot of extra work? Probably this won't exceed 10 lines of doc.
It is clear that opening an issue has little chance to lead to a documentation.

krenzland:
> Is it a lot of extra work? Probably this won't exceed 10 lines of doc.
> It is clear that opening an issue has little chance to lead to a documentation.

No, not really. (Well, it's still ~10 minutes to write and actually test the documentation).
 But it's something that I noticed for a few past PRs that a lot of additional changes get tacked on top of them. PRs should be (imo) as small as possible to ensure that we can merge it.

But I generally agree. Maybe the best idea is to write documentation when you need to explain something. I've explained the convergence procedure at least once and haven't written the documentation...

PR #272, opened 2020-11-17
Qt4 -> Qt5 fix for viewrec

krenzland:


uphoffc (approved):
Wow that was a lot of effort :D

PR #273, opened 2020-11-19
Ravil/gpu merge

ravil-mobile:
Merged gpu_integration branch with the current master.


sebwolf-de (commented):
Hi, 
thanks for this PR towards GPU integration into SeisSol. I've marked some minor remarks in the code. 
I see some code duplication , eg `Seissol::time_stepping::TimeCluster::computeNeighboringIntegration` which has its own implemenation for CPU and GPU execution now. The same is in the kernels with e.g. `computeIntegrals` and `computeBatchedIntegrals`. If I have a look at the local kernel, then all of the 'fancy' BCs don't seem to work and would have to be reimplemented for the batched version. Can we unify this, so that a change in one of the kernels, does not have to be done twice for GPU and CPU version?

sebwolf-de (commented):
Thanks for the changes, still I found some minor remarks.
BTW: how did you test your changes?

sebwolf-de (approved):

PR #276, opened 2020-12-01
Thomas/variouschanges

Thomas-Ulrich:
- remove pythonXdmfReader (replaced by pip module https://pypi.org/project/seissolxdmf/).
- update the 3 scripts which used it.
- precise off-fault receivers (no OutputMask)
- fix typo in compilation.rst
- refactor asagi.rst and add example.
- document memory requirements.


Documentation/asagi.rst
```diff
@@ -81,42 +100,32 @@ followed by running ``fix_submodules`` to clone and set up the submodules.
...
+A typical example which generates a 2D ASAGI file can be found 
+`here <https://github.com/SeisSol/SeisSol/tree/master/preprocessing/science/generating_ASAGI_file.py>`__.
 
 using asagiconv
 ~~~~~~~~~~~~~~~
 
 Asagiconv (Located
 `here <https://github.com/SeisSol/SeisSol/tree/master/preprocessing/science/asagiconv>`__)
-allow querying data, visualising and exporting to netcdf data from the
+allow querying data, vizualising and exporting to NetCDF data from the
```
sebwolf-de, line 120: typo

sebwolf-de (commented):
LGTM, but should we add a comment somewhere in the docu that you need to install the xdmf reader via pip?

Documentation/asagi.rst
Thomas-Ulrich: > LGTM, but should we add a comment somewhere in the docu that you need to install the xdmf reader via pip?

I've added a comment in the fault output section. Ideally, there would be a more general output page, where I would have added that. But I don't want to do this now. Anyway, the idea is that the user can now search for `seissolxdmf` in the readthedocs and find the link.

sebwolf-de (approved):

PR #279, opened 2020-12-04
Improved cmake error message for metis

ravil-mobile:
Added an additional FindMETIS.cmake file which checks only METIS. It uses hints of both `metis` and `parmetis` env. variables to find metis.h and libmetis.(a/so). Additionally, it checks whether METIS was compiled with IDXTYPEWIDTH = 64. Fix #279


sebwolf-de (approved):
LGTM

PR #280, opened 2020-12-10
Ravil/gpu plasticity

ravil-mobile:
Added GPU plasticity kernels to SeisSol.

* Compared numerical values with CPU version using a tpv12 setup


src/Initializer/LTS.h
```diff
@@ -138,8 +138,8 @@ struct seissol::initializers::LTS {
     tree.addBucket(displacementsBuffer,                         PAGESIZE_HEAP,      MEMKIND_TIMEDOFS );
 
 #ifdef ACL_DEVICE
-    tree.addVar(   localIntegrationOnDevice,   LayerMask(Ghost),  1,      seissol::memory::DeviceGlobalMemory );
-    tree.addVar(   neighIntegrationOnDevice,   LayerMask(Ghost),  1,      seissol::memory::DeviceGlobalMemory );
+    tree.addVar(   localIntegrationOnDevice,           LayerMask(Ghost),  1,      seissol::memory::DeviceGlobalMemory );
```
sebwolf-de, line 141: Can you remove these changes? They clutter the git history and they are no functional changes.

src/Kernels/DeviceAux/cuda/PlasticityAux.cu
```diff
@@ -0,0 +1,235 @@
+#include <Kernels/DeviceAux/PlasticityAux.h>
+#include <init.h>
+#include <cmath>
+
+#if REAL_SIZE == 8
```
sebwolf-de, line 5: Is this the way to do it? Disclaimer: I'm no expert on Cuda.

sebwolf-de (commented):
Hi,
thanks for the changes, no big problems, just some small questions.
How did you validate your results?
Also the GPU stuff is not documented anywhere. Can you write a section like https://seissol.readthedocs.io/en/latest/supermuc.html just for a GPU cluster. This could also be a space where our users can be informed which parts of the code are already ported and which parts are not yet ported.

src/Kernels/DeviceAux/cuda/PlasticityAux.cu
ravil-mobile: Hi, Could you tell me which part of the code you refer to?

Do you mean this one?
https://github.com/SeisSol/SeisSol/blob/27140584ba70cf88c46f26f348414117a82938b8/src/Kernels/DeviceAux/cuda/PlasticityAux.cu#L5-L11

ravil-mobile:
>How did you validate your results?

I usually run a CPU version of  the code, record the output and do the same for a GPU version. Then I visually compare results using `paraview` and the use `h5diff`

src/Initializer/LTS.h
ravil-mobile: yes, sure

src/Kernels/DeviceAux/cuda/PlasticityAux.cu
sebwolf-de: Yes these defines. Is this a standard way of writing stuff in cuda?

ravil-mobile: So, I agree that it is ugly but it is necessary in order to support both double and single precision.
Here is a definition of `max` function in cuda
`__device__ ​ double fmax ( double , double ) `  [reference double precision](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__DOUBLE.html#group__CUDA__MATH__DOUBLE)
`__device__ ​ float fmaxf ( float  x, float  y )`  [reference single precision](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html#group__CUDA__MATH__SINGLE)

I don't know why CUDA devs didn't use templates in this case. Officially, CUDA supports templates but maybe math functions are implemented as intrinsics

ravil-mobile:
> Also the GPU stuff is not documented anywhere. Can you write a section like https://seissol.readthedocs.io/en/latest/supermuc.html just.

Done

Documentation/gpus.rst
```diff
@@ -0,0 +1,89 @@
...
+
+.. code-block:: bash
+
+   git clone --recurse-submodules https://github.com/SeisSol/SeisSol.git seissol-gpu
+
+Compile SeisSol with (e.g.)
+
+.. code-block:: bash
+
+    mkdir -p seissol-gpu/buid && cd seissol-gpu/buid
```
sebwolf-de, line 64: typo, should be `build`, right?

sebwolf-de (changes requested):
Hi, I just found one small typo, except for that we can merge :D

Documentation/gpus.rst
ravil-mobile: Thanks. It is fixed now.

sebwolf-de (approved):

PR #285, opened 2021-02-01
Add support for AMD rome

krenzland:
(Squash merge please)

Add support for AMD Rome architecture.
No support for Intel compiler (it works theoretically but why would you?)


sebwolf-de (approved):
LGTM

PR #287, opened 2021-02-04
Sebastian/error codes

### LINKED ISSUE #286: Feature request: Non-zero exit codes
SeisSol exits with exit-code 0 from Fortran, which means everything was done correctly, even though an error occurred. Try for example running SeisSol without specifying a parameter file.
I suggest we change this to make the error behaviour more verbose. Especially, we will need this for the MUQ adapter, where SeisSol will be called as a dependency.

In Fortran parts the error handling is done by 
```
logError(*) "error message"
stop
```
If we replace `stop` by `call exit(1)` we generate a proper exit code.

In the C++ parts we call the error Logger, which then calls either `abort()` or `MPI_Abort()`.
`MPI_Abort` exits with exit code `-1`, but `std::abort` does exit with `SIGABRT (134)`.

TL;DR
Errors in Fortran parts lead to exit code 0.
Errors in CPP parts lead to exit code -1 (if compiled with MPI) or 134 (if not compiled with MPI).
To Decide:
- [ ] Think about which exit codes we want to throw. I suggest 134 everywhere.

sebwolf-de:
Fix #286

### DIFF
--- a/src/Geometry/MeshReaderCBinding.f90
+++ b/src/Geometry/MeshReaderCBinding.f90
@@ -140,7 +140,7 @@ contains
                                     disc%galerkin%clusteredLts                  )
         else
             logError(*) 'Unknown mesh reader'
-            stop
+            call exit(134)
         endif
 
         ! Set additional SeisSol variables
--- a/src/Geometry/allocate_mesh.f90
+++ b/src/Geometry/allocate_mesh.f90
@@ -106,7 +106,7 @@ CONTAINS
     
     IF (allocStat .NE. 0) THEN
         logError(*) 'allocate_mesh_level0_1: could not allocate the whole mesh!'
-        STOP
+        call exit(134)
     END IF
     !
   END SUBROUTINE allocate_mesh_level0_1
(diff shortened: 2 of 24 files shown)


krenzland:
Is there a reason why you didn't include the return code directly in the Fortran function logError?

sebwolf-de:
Yes, it's a macro and I didn't want to put more effort than search replace into it :D

krenzland:
`#define logError(format)   if (LOGLEVEL .ge. 0) write(FORTRAN_STDERR, '(A, I8, A)', advance='no') 'Rank: ', myrank, ' | Error   |'; if (LOGLEVEL .ge. 0) write(FORTRAN_STDERR, format);  call exit(134)
`
Should work, right?
Here: https://github.com/SeisSol/SeisSol/blob/master/src/Initializer/preProcessorMacros.fpp#L72

sebwolf-de:
Nope, doesn't work. The preprocessor replaces `logError(*) 'asdf'` with `if (LOGLEVEL) ...; call exit(134) 'asdf'` which is a syntax error.

krenzland (approved):
I'm still not sure whether 134 is the correct exit code (or whether it actually matters) but that's no reason not to merge.

PR #289, opened 2021-02-10
Remove old scons stuff

krenzland:
Also solves the issue that you find all file twice in IDEs - once in src and once in autotuning/proxy/src/seissol_src.


sebwolf-de:
If you are interested in removing scons completly from the repo: https://github.com/SeisSol/SeisSol/tree/sebastian/cmake_cube_c 
Although I like the idea of the PR, I would not merge directly to give people, who might still use scons, time to complain. Say until Friday afternoon, maybe?

krenzland:
I'd be surprised if scons would still work to be honest. If people complain, they can still use an old version - we're not going to support two build systems.

I'm generally interested in purging scons completely, but this PR only deals with the main seissol src. I don't want to deal with these other things currently.

krenzland:
Oh, and please open a pr for the cmake_cube_c! That sounds useful.

sebwolf-de (approved):

PR #291, opened 2021-02-11
Sebastian/cmake cube c

sebwolf-de:
This adds the possibilty to build the cube generator with cmake.


krenzland:
Should we then directly delete the corresponding scons build files?

preprocessing/meshing/cube_c/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+cmake_minimum_required(VERSION 3.10)
+
+# set the project name
+project(cube_c)
+
+# specify the C++ standard
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
+
+cmake_policy(SET CMP0074 NEW)
```
krenzland, line 10: This line does not make sense here. The policy is only supported for Cmake version > 3.12
See: https://cmake.org/cmake/help/latest/policy/CMP0074.html
I also think that we should provide a quick comment in these cases. I know that we do not do this currently but it would make things a bit clearer. What do you think?

preprocessing/meshing/cube_c/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+cmake_minimum_required(VERSION 3.10)
+
+# set the project name
+project(cube_c)
+
+# specify the C++ standard
+set(CMAKE_CXX_STANDARD 11)
```
krenzland, line 7: Can we switch to 17 here? SeisSol main requires it anyway.

preprocessing/meshing/cube_c/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+target_link_libraries(cubeGenerator PUBLIC OpenMP::OpenMP_CXX)
+
+#add some compiler specific flags
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  # using GCC
+  target_compile_options(cubeGenerator PUBLIC -fopenmp -pedantic $<$<OR:$<COMPILE_LANGUAGE:CXX>,$<COMPILE_LANGUAGE:C>>:-Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas>)
+  target_link_libraries(cubeGenerator PUBLIC "-fopenmp")
+elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
+  # using Intel C++
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -qopenmp")
```
krenzland, line 53: Is there a reason why we do not enable warnings for intel compiler?

krenzland (changes requested):
Thank you! This is a very helpful change. Can you please look at my comments and also remove the old scons files?

preprocessing/meshing/cube_c/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+elseif("${LOG_LEVEL}" STREQUAL "info")
+  target_compile_definitions(cubeGenerator PUBLIC LOG_LEVEL=2)
+elseif("${LOG_LEVEL}" STREQUAL "warning")
+  target_compile_definitions(cubeGenerator PUBLIC LOG_LEVEL=1)
+elseif("${LOG_LEVEL}" STREQUAL "error")
+  target_compile_definitions(cubeGenerator PUBLIC LOG_LEVEL=0)
+endif()
+
+#build and link libraries and executable
+set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
```
krenzland, line 30: Imo it's better to set this to `set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/../../../cmake" ${CMAKE_MODULE_PATH})` and delete the symlink.

preprocessing/meshing/cube_c/CMakeLists.txt
sebwolf-de: Warnings for intel compiler are a bit complicated: https://stackoverflow.com/questions/27343360/how-to-turn-on-icc-icpc-warnings <- I've used what they recommend.

krenzland (approved):

PR #292, opened 2021-02-12
Nothing to be proud of

Thomas-Ulrich:
Commit about the dark side of SeisSol.
The hard-limit on the clustering with attenuation (update of the docu).
The moment rate function files :-)
(I just added a small feature to decrease the size of the concatenated moment rate file, as sampled every 1e-5s over 900s it can be 100mb big).


sebwolf-de (approved):
LGTM

sebwolf-de:
There seems to be a problem with libxsmm in the pipeline. @krenzland any ideas?

krenzland:
Not really, no. I'll try to re-trigger it automatically. If that doesn't work, we can pin libxsmm to a working version.

krenzland:
Worked.

PR #297, opened 2021-02-17
minor improvement to extractDataFromUnstructuredOutput.py

Thomas-Ulrich:
Some minor improvement to this script for extracting time steps:
- default to hdf5 and float as it drastically reduces file size.
- write in the current folder and not in the folder of the input file.
- add `last` option to write the last time step without knowing the index.
-fix for `-` sign in prefix

### DIFF
--- a/postprocessing/visualization/tools/extractDataFromUnstructuredOutput.py
+++ b/postprocessing/visualization/tools/extractDataFromUnstructuredOutput.py
@@ -12,7 +12,7 @@ parser.add_argument("--add2prefix", help="string to append to prefix for new fil
 parser.add_argument("--Data", nargs="+", metavar=("variable"), default=(""), help="Data to resample (example SRs)")
 parser.add_argument("--downsample", help="write one out of n output", type=int)
 parser.add_argument("--precision", type=str, choices=["float", "double"], default="float", help="precision of output file")
-parser.add_argument("--backend", type=str, choices=["hdf5", "posix"], default="hdf5", help="backend used: posix (.bin file), hdf5 (.h5)")
+parser.add_argument("--backend", type=str, choices=["hdf5", "raw"], default="hdf5", help="backend used: raw (.bin file), hdf5 (.h5)")
 parser.add_argument("--last", dest="last", default=False, action="store_true", help="output last time step")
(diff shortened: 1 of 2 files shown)


postprocessing/visualization/tools/extractDataFromUnstructuredOutput.py
```diff
@@ -11,8 +11,9 @@
 parser.add_argument("--add2prefix", help="string to append to prefix for new file", type=str, default="_resampled")
 parser.add_argument("--Data", nargs="+", metavar=("variable"), default=(""), help="Data to resample (example SRs)")
 parser.add_argument("--downsample", help="write one out of n output", type=int)
-parser.add_argument("--float", dest="float", default=False, action="store_true", help="convert to float")
-parser.add_argument("--hdf5", dest="hdf5", default=False, action="store_true", help="output as hdf5")
+parser.add_argument("--double", dest="double", default=False, action="store_true", help="do not convert to float")
+parser.add_argument("--posix", dest="posix", default=False, action="store_true", help="output as posix")
```
sebwolf-de, line 15: Maybe make these two arguments a string: like ```"--precision", type=str, choices=["float", "double"],  default="float"```

sebwolf-de (commented):
Looks good to me, just one minor remark. But this is just for IMHO better usability.

postprocessing/visualization/tools/extractDataFromUnstructuredOutput.py
krenzland: Also the name posix is slightly confusing (has nothing to do with the actual file format). Not sure what (if anything) would be better. Maybe 'raw'?

Thomas-Ulrich: I used posix as it is the name used in xdmfWriterBackend. Do you think posix makes sense in xdmfWriterBackend?
I can use raw (at the end, probably this option will never be used anyway :-p (well I used it as some point when samoa was only reading bin files)).

Thomas-Ulrich: anyway, I've adressed all comments, thx for the review!

sebwolf-de (approved):

Issue #288, opened 2021-02-09
Code of Conduct


krenzland:
Hi all,

in our meeting today we came to the conclusion that it would be nice to have a code of conduct (CoC).
@AliceGabriel suggested something similar to the PETSc CoC which is based on the [contributor covenant.](https://www.contributor-covenant.org/version/2/0/code_of_conduct/).
Of course, we are open for other suggestions!

Personally, I like the suggested one, as it's basically "be nice to each other" and does not include anything which could lead to controversies. I especially like that it separates behavior in private and when representing this project. I also like that conflicts should be handled anonymous.

I suggest that we add this (or a similar CoC). I also suggest that we nominate ~2 people from LMU/TUM each to represent both communities as contact persons.This CoC would then apply to all repositories hosted by this Github organization.

What do you think? If no one complains in about a week or so, I'll open a pull request to this repo.

Cheers,
Lukas

Thomas-Ulrich:
Honestly, I doubt it would be of any help but I m not against trying.

krenzland:
I also don't see it as a necessity but I do not see any downsides. If we have one, we signal that we care about inclusion/openness to all collaborators, etc.

ravil-mobile:
hmmm, it is a pity. I collected some information about Code of Conduct, Contribution, etc from other projects like llvm, singularity, spack, docker, etc but I forgot to press om `comment`. Ok, I will repeat the process again

krenzland:
@ravil-mobile I would like to create a PR for a CoC soon. Should we wait for your contribution or can we go ahead with the contributor covenant?

For all others, please state possible objections in this issue! This should be a community effort and not something that is decided by a small minority. Thanks!

ravil-mobile:
Sorry for delay. I will do it today

ravil-mobile:
### Code of conduct
[llvm](https://llvm.org/docs/CodeOfConduct.html) - it is very nice. I like this style

[singularity](https://github.com/sylabs/singularity-cri/blob/master/CODE_OF_CONDUCT.md) - a template 

[spack](https://github.com/spack/spack/blob/develop/.github/CODE_OF_CONDUCT.md) - a template 

[docker](https://github.com/docker/code-of-conduct/blob/master/code-of-conduct-EN.md) - I guess it is a good example

[rocm](https://github.com/RadeonOpenCompute/rocm-spack/blob/develop/.github/CODE_OF_CONDUCT.md) - a template 

[...]

PR #300, opened 2021-02-25
Add code of conduct

krenzland:
As outlined in issue #288, we want to have a code of conduct. I propose that we use (maybe a modified version) of the one contained in this repository. It's the current version of the contributor covenant, one of the most widely adopted code of conducts.

Should we extend this by stating clearly who the community leaders are? I've created a mailing list seissol-maintainers@lists.lrz.de which we could use as a contact address. Otherwise it might be a bit unclear who is actually enforcing the code of conduct.

What do you think?

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,129 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the
+  overall community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or
+  advances of any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or email
+  address, without their explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Enforcement Responsibilities
+
+Community leaders are responsible for clarifying and enforcing our standards of
+acceptable behavior and will take appropriate and fair corrective action in
+response to any behavior that they deem inappropriate, threatening, offensive,
+or harmful.
+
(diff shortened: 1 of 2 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,130 @@
+i# Contributor Covenant Code of Conduct
```
sebwolf-de, line 1: Typo ;-)

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,130 @@
...
+not aligned to this Code of Conduct, and will communicate reasons for moderation
+decisions when appropriate.
+
+## Scope
+
+This Code of Conduct applies within all community spaces, and also applies when
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
```
sebwolf-de, line 57: Should we add, that this code of conduct holds for all repositories within the SeisSol project?

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at coraline c/o 
+idolhands dot com.
```
sebwolf-de, line 63: Fix Email adress here.

sebwolf-de (commented):
Hi Lukas, thanks for pushing this forward. Some small comments on the text of the code of conduct, but no major objections from my side.
About the email list: Who are the recipients? I think it is reasonable to keep the list of responsibles as small as possible i.e. not all SeisSol developers.

CODE_OF_CONDUCT.md
krenzland: Ah, there it is. I was actually unable to find the email address in the template...

CODE_OF_CONDUCT.md
krenzland: No, I'd just push it to other repositories later - otherwise it becomes very confusing. Maybe add it to the website as well (at some point).

krenzland: At least if we interpret community as the seissol community - which is fair here, I think, because this is obviously the main repository. Or what do you think?

sebwolf-de: So, let's have it at the main repo + website and link to it from all the other repos?

krenzland: Yup, that's a good idea. Not sure whether we need to link it from the other repos. I guess, most people come to the main repo first.

krenzland:
>  About the email list: Who are the recipients? I think it is reasonable to keep the list of responsibles as small as possible i.e. not all SeisSol developers.

Right now it's only me. I agree, we should not include all developers. Not sure, who is a good choice. Probably should also include @baderml and @AliceGabriel.

CODE_OF_CONDUCT.md
ravil-mobile: Hi. @krenzland thanks for your work. Maybe we can do something what [spack](https://github.com/spack/spack#code-of-conduct) did. I mean we can make a link to code of conduct in our main README.md. Would it be better if the file was placed in `<root>/.github` directory?

krenzland: Yes, it's a good idea to add this to the README.md.

About the .github directory: Github detects the file also in this way and it is a bit more visible when it is in the root.

sebwolf-de (approved):

Issue #295, opened 2021-02-15
cmake SeisSol compilation - which SuperMucNG modules required (potentially wrong documentation)


AliceGabriel:
**Describe the bug**
I am following along a new Seissol on SuperMucNG user. SeisSol's documented description on using cmake to compile fails, due to 'malfunctioning' mpiicc.

The documentation is here https://seissol.readthedocs.io/en/latest/supermuc.html?highlight=supermuc#supermuc-ng.
I tried two things: loading the recommended modules (leading to error 1 below) and also hand-loading the modules I think are required as follows:
di34pac4@login01:~/SeisSol/build-release> module list
Currently Loaded Modulefiles:
 1) admin/1.0            6) mkl/2019_gcc        11) inspector_xe/2019      16) parmetis/4.0.3-intel-impi-i64-r64          
 2) tempdir/1.0          7) mpi.intel/2019_gcc  12) devEnv/GCC/7-IntelMPI  17) metis/5.1.0-intel-i64-r64                  
 3) spack/release/19.2   8) itac/2019           13) cmake/3.14.4           18) numactl/2.0.11                             
 4) lrz/1.0              9) advisor_xe/2019     14) python/3.6_intel       19) hdf5/1.8.21-impi-cxx-frt-threadsafe        
 5) gcc/7               10) amplifier_xe/2019   15) libszip/2.1.1          20) netcdf/4.6.1-intel-impi-hdf5v1.8-parallel  

Trying to run a simple test
di34pac4@login01:~/SeisSol/build-release> mpiicc a.c 

<img width="339" alt="Screenshot 2021-02-15 at 22 59 32" src="https://user-images.githubusercontent.com/9365956/107996770-78cda580-6fe1-11eb-91fa-4ed33d500f13.png">

leads to
/lrz/sys/intel/impi2019u6/impi/2019.6.154/intel64/bin/mpiicc: line 545: icc: command not found

**Expected behavior**
Compiling.

**To Reproduce**
checkout current master and follow the documentation.

**Screenshots/Console output**

[...]

PR #304, opened 2021-03-04
update supermuc installation docu

Thomas-Ulrich:
adresses #295?


krenzland:
Did you test this today? Just asking because they changed some modules around :(

Thomas-Ulrich:
today, yes, but maybe they have made some updates since then. I will test again...

Thomas-Ulrich:
I double-checked, the procedure still works.

krenzland:
Why is this PR updating easi? That includes the checks for NaNs, right=

Thomas-Ulrich:
> Why is this PR updating easi? That includes the checks for NaNs, right?

Hi, because it was never merged (sorry for occulting this change).

PR #308, opened 2021-03-04
Make README shields clickable

krenzland:


ravil-mobile (approved):
Hi @krenzland. I like this change. Let's merge it!

PR #309, opened 2021-03-10
exit to MPIABORT

Thomas-Ulrich:
I changed the `call exit(134)` to `call MPI_ABORT(MPI%commWorld, 134)`.
Same files as Sebastian previous PR.
Then I added the `USE mpi_f08` at the beginning of each file modified.
But, then I removed it in files where the compilation was complaining (because MPI_COMM_WORLD was already included in another module?).

(in the process of fixing the compilation, I changed some of the MPI_COMM_WORLD to MPI%commWorld, which should be equivalent, right?).

### DIFF
--- a/azure-pipelines-seissol-template.yml
+++ b/azure-pipelines-seissol-template.yml
@@ -18,7 +18,6 @@ jobs:
                         sudo apt-get install gcc-8 g++-8 gfortran-8
                         sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 100
                         sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-8 100
-                        sudo update-alternatives --install /usr/bin/gfortran gfortran /usr/bin/gfortran-8 100
                         gcc --version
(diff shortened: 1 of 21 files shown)


krenzland:
In my opinion, we should delete the support for compiling without MPI anyway. Someone is probably going to disagree with that though.

krenzland:
I've had a quick look why the pipeline doesn't compile. It looks like the standard package from ubuntu compiles MPI with an older gfortran. We explictely use a newer gfortran version. So there seems to be a mismatch. Not sure what the solution is. We can try to see whether we can use an older version of gfortran (i.e. the default one that ubuntu ships!). Should work in theory.

@Thomas-Ulrich  Can I commit to your branch? Maybe I can fix it.

Thomas-Ulrich:
Thank you! yes, feel free to commit to the branch!

krenzland:
It's fixed!

Thomas-Ulrich:
Great then it is ready for review too!

sebwolf-de:
>  I changed some of the MPI_COMM_WORLD to MPI%commWorld
I think we should make this consistent.

Thomas-Ulrich:
Hi, 
I've updated most of the MPI_COMM_WORLD to MPI%commWorld, except in 2 files:
src/Numerical_aux/dgbasis.f90
src/Numerical_aux/operators.f90
Because adding the MPI to the function interface would require changing a lot of functions calling these basic functions.

sebwolf-de (approved):

PR #311, opened 2021-03-11
imporved ctest output in case of a failure

ravil-mobile:
* CTEST_OUTPUT_ON_FAILURE=1
* removed unnecessary env. vars from pipeline scripts

### DIFF
--- a/test_singularity.sh
+++ b/test_singularity.sh
@@ -2,12 +2,6 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


ravil-mobile:
Hi @VMPW. Could you merge this changes if you are agree? I think CTEST_OUTPUT_ON_FAILURE=1 will make it clear why pipeline gets broken

PR #306, opened 2021-03-04
slip and aging laws

Thomas-Ulrich:
This is work in collaboration with Kadek.
This PR merges the slip and aging laws with the already existing strong velocity weakening rate and state.
It also introduces a fix to prevent negative state variables (also slip in LSW), that could arise due to the resampling matrix approach. Without the fix, NaN arises after some time in slip and aging laws.
So far, I've quickly tested that it runs and gives reasonable levels of slip.
A proper benchmarking is still pending, but I think the code can already be reviewed.

PR #314, opened 2021-03-17
Fix eta (plasticity)

uphoffc:
Mostly added comments, but fix one actual problem:

The computation

```
dudt_pstrain[q] =
            mufactor * (prev_degreesOfFreedom[q] - degreesOfFreedom[q * NUMBER_OF_ALIGNED_BASIS_FUNCTIONS]);
```

can be interpreted as finite differenceing of (A.1)
```
dsigma_{ij}/dt = -2mu dstrain_{ij}/dt
```
**But** the factor 1/timeStepWidth is missing. That is actually fine for
```
pstrain[q] += dudt_pstrain[q];
```
because timeStepWidth * dudt_pstrain[q] / timeStepWidth = dudt_pstrain[q].

However, for the computation of eta a factor of sqrt(1/timeStepWidth^2) = 1 / timeStepWidth is missing, therefore the multiplication with timeStepWidth is **wrong**.

I derived the new implementation of dstrain_{ij}/dt from Equation (10), Wollherr et al, see comments. [Shouldn't make much difference for very small time-steps for the strain](https://www.wolframalpha.com/input/?i=plot+x%2F%280.1*%281-exp%28-x%2F0.1%29%29%29+from+0+to+1e-2) from 0 to 1e-2), but Eta should now be scaled correct.

@ravil-mobile Where do I need to change the GPU code?
@Thomas-Ulrich Could you test this for your scenario?


ravil-mobile:
Hi,

Here is the kernel which updates [Pstrains](https://github.com/SeisSol/SeisSol/blob/b014d1180d53a5231e92bb5062a45711fea9aade/src/Kernels/DeviceAux/cuda/PlasticityAux.cu#L171-L209) and this is the corresponding [launcher](https://github.com/SeisSol/SeisSol/blob/b014d1180d53a5231e92bb5062a45711fea9aade/src/Kernels/DeviceAux/cuda/PlasticityAux.cu#L212-L229) which is called from [here](https://github.com/SeisSol/SeisSol/blob/b014d1180d53a5231e92bb5062a45711fea9aade/src/Kernels/Plasticity.cpp#L230-L236).

From my side, I can clone this branch, adjust the GPU kernel, make PR to this branch and assign you as a reviewer. What do you think?

uphoffc:
You can also just push to this branch. Thanks!

Thomas-Ulrich:
Hi, I used the changes you send me yesterday, which should be comparable to this PR, and the final eta and eps^p are consistent.

AliceGabriel:
Hi,
Thomas, Casper and I discussed this yesterday and Thomas validated the fix
in eta calculation by comparing mu*J2(eps^p)*dV)  with sum(mu*eta*dV)
 which are two ways to calculate the equivalent plastic moment.
--> Thomas could you compare your current patch with this PR ?
--> Could you add the Mp output based on eta next to the Mw output please?

Specifically to the LMU team: Very importantly, this means previous mapping
of the plastic strain tensor into eta are *wrong*  and have to be repeated.

Alice

AliceGabriel:
Hi
Taufiq tested with our Ridgecrest setup but the test fails with NaN.
@Taufiqurrahman
Taufiqurrahman ***@***.***> please share your
setup and problem here.
Alice

Thomas-Ulrich:
Hi,
I wonder if the problem of Taufiq could be also this new change added by PR #306
https://github.com/SeisSol/SeisSol/blob/master/src/Physics/Evaluate_friction_law.f90#L1075
Could be worth running once more without the max (if you are using RS, I don't remember).
(yet, I think this change makes sense).

src/Kernels/Plasticity.cpp
```diff
@@ -93,35 +97,44 @@ namespace seissol::kernels {
...
+    // Compute tau_c for every node
     for (unsigned ip = 0; ip < tensor::meanStress::size(); ++ip) {
       taulim[ip] = std::max((real) 0.0, plasticityData->cohesionTimesCosAngularFriction -
                                         meanStress[ip] * plasticityData->sinAngularFriction);
     }
 
     bool adjust = false;
     for (unsigned ip = 0; ip < tensor::yieldFactor::size(); ++ip) {
+      // Compute yield := (t_c / tau - 1) r for every node,
+      // where r = 1 - exp(-timeStepWidth / T_v)
```
Thomas-Ulrich, line 137: what you call r is relaxTime at L140?
but if r = 1 - exp(-timeStepWidth / T_v), it is not a time, right?

Thomas-Ulrich: Yes I think it not a time, this variable should be named differently. e.g. one_minus_exp_dt_Tv
But I now understand that what you send me is equivalent to the changes in the PR if dt/Tv
because 1-exp(-dt/Tv)~ dt/Tv

Thomas-Ulrich:
After discussing with Taufiq, it turns out that he did apply the patch on a deprecated SeisSol.
Taufiq, can you use the latest master?
(if you have setup specific changes on your branch, it might be a good habit to rebase your branch on master from time to time (or maybe simpler merge master to your branch).

git-taufiq:
Thanks, Thomas. Previously, I quickly tried the patch on not the recent version. It works now without NaN error.  https://github.com/SeisSol/SeisSol/pull/314/commits/26654d3e5d4b1ff6ffd73cf3d3cf142e5385ed05
I will update you with the output soon.

Thomas-Ulrich:
So I quickly compared the maximum eta value (on 5s of the Aqaba scenario, sampling rate 2.5s for the wavefield), with the patch you first send me:
0 0.0 0.0
1 0.0 0.0015739776606070739
2 0.0 0.002788263926546942
And with this PR:
0 0.0 0.0
1 0.0 0.001583104531013265
2 0.0 0.0027961887334781734
Results are coherent, small differences are expected because 1-exp(-dt/Tv)!= dt/Tv for the largest dt.

uphoffc:
Shall we merge then?

Thomas-Ulrich:
can you address my comment with relaxTime?

uphoffc:
> can you address my comment with relaxTime?

Changed it to one_minus_integrating_factor

Thomas-Ulrich (approved):

Thomas-Ulrich:
> > can you address my comment with relaxTime?
> 
> Changed it to one_minus_integrating_factor
Thank you!

src/Solver/time_stepping/TimeCluster.h
```diff
@@ -198,7 +198,7 @@ class seissol::time_stepping::TimeCluster
     double m_tv;
     
     //! Relax time for plasticity
-    double m_relaxTime;
+    double m_one_minus_integrating_factor;
```
Thomas-Ulrich, line 201: Seems like this line needs to be changed?
https://github.com/SeisSol/SeisSol/blob/faffbb2281dcd2fce3f7261dc903fe566f1dc4ec/src/Solver/time_stepping/TimeCluster.cpp#L640
How could it pass the compilations test?

uphoffc: Could pass if no plasticity version is compiled (USE_PLASTICITY)

git-taufiq: Got the same error related to m_relaxTime, this could be related to this commit changing relaxTime to one_minus_integrating_factor:   https://github.com/SeisSol/SeisSol/commit/25dedbdef13079485f3b7bec37a58e0d144a28a2
testing the commit just before renaming works fine:
https://github.com/SeisSol/SeisSol/commit/693e2d327901acd5169cb691fc0e74f37d694ea8

Issue #80, opened 2019-07-05
WaveFieldWriter works only in double precision


uphoffc:
https://github.com/SeisSol/SeisSol/blob/master/src/ResultWriter/WaveFieldWriter.cpp#L113
https://github.com/SeisSol/SeisSol/blob/master/src/Solver/Interoperability.cpp#L686

Thomas-Ulrich:
can't we substitute double by real in these 2 routines, try to compile SeisSol and change the declaration of all routines that complains about this change?

uphoffc:
Yes, almost. E.g. this line is not going to complain, so one needs to be a bit more careful:
https://github.com/SeisSol/SeisSol/blob/master/src/ResultWriter/WaveFieldWriter.cpp#L400

But in principle it's not that hard, but has to be done and tested.

ravil-mobile:
Could you, please, describe the problem according to the form that we established for issues?

Thomas-Ulrich:
fixed by #315

PR #315, opened 2021-03-18
Ravil/writers

ravil-mobile:
These changes are intended to fix result writers while computing with single precision. An important point that geometry is still handled with double precision but data can be written to hdf5 files with single precision fp. format. It led to some changes in XdmfWriter sub-module (changes can be seen in this [PR](https://github.com/TUM-I5/XdmfWriter/pull/1)). 

DR-single-output
WF-single-output

I've tested only TPV5 benchmark so far. However, a more detailed evaluation is required.

**Important:** this PR is not going to fix *check-pointing*. It will be done in a different PR. This PR is only focused on result writers


sebwolf-de:
Hi, LGTM, but should we wait for the "more detailed evaluation" before merging?

ravil-mobile:
Thanks @sebwolf-de. I am going to fix tests and test some other scenarios. I think It would be nice if @Thomas-Ulrich can check this out.

uphoffc:
Don't think we need to wait on more detailed testing. While I'm not sure whether double -> real was substituted everywhere, but this PR doesn't break things for double precision

Thomas-Ulrich:
Thx for this PR! 
I will test the patch in the next days.

ravil-mobile:
Hi all, 

I fixed a unit test and everything is working now. I will focus on suggestions of @sebwolf-de and @uphoffc about XdmfWriter.

> Don't think we need to wait on more detailed testing. While I'm not sure whether double -> real was substituted everywhere, but this PR doesn't break things for double precision

agree

Thomas-Ulrich:
Hi, the plasticity (eta, eps_xx) outputs are still not working in single precision.
Fault, surface, and wavefield output looks fine

ravil-mobile:
Hi all, 

Here is just some info. I rebased this branch with the current master to keep the project history clean and (forced) pushed it here. So, all my changes are at the top of the thread pinning stuff.

Thomas-Ulrich:
ok now it is working fine!

ravil-mobile:
Thanks @Thomas-Ulrich!

Can somebody approve these changes?

src/ResultWriter/FaultWriterC.cpp
```diff
@@ -43,15 +43,26 @@
 extern "C"
 {
 
-void fault_hdf_init(const int* cells, const double* vertices,
-		int nCells, int nVertices,
-		int* outputMask, const double** dataBuffer, const char* outputPrefix,
-		double interval, char const* xdmfWriterBackend)
+void fault_hdf_init(const int* cells,
```
Thomas-Ulrich, line 46: is it only formatting changes?
Better uncommit that?

src/ResultWriter/FaultWriterC.cpp
```diff
@@ -43,15 +43,26 @@
...
+                    const real** dataBuffer,
+                    const char* outputPrefix,
+                    double interval,
+                    char const* xdmfWriterBackend)
 {
   auto type = seissol::writer::backendType(xdmfWriterBackend);
 	seissol::SeisSol::main.faultWriter().init(reinterpret_cast<const unsigned int*>(cells),
-		vertices, nCells, nVertices,
-		outputMask, dataBuffer, outputPrefix, interval, type);
+                                            vertices,
```
Thomas-Ulrich, line 58: same

src/ResultWriter/FaultWriterExecutor.h
```diff
@@ -45,10 +45,9 @@
 #endif // USE_MPI
 
 #include "xdmfwriter/XdmfWriter.h"
-
```
Thomas-Ulrich, line 48: same

src/ResultWriter/FaultWriterExecutor.h
```diff
@@ -45,10 +45,9 @@
...
 #include "async/ExecInfo.h"
-
```
Thomas-Ulrich, line 50: same

src/ResultWriter/FreeSurfaceWriterExecutor.cpp
```diff
@@ -78,14 +78,20 @@ void seissol::writer::FreeSurfaceWriterExecutor::execInit(const async::ExecInfo
...
+		m_xdmfWriter = new xdmfwriter::XdmfWriter<xdmfwriter::TRIANGLE, double, real>(param.backend,
+		                                                                              outputName.c_str(),
+		                                                                              param.timestep);
 
 #ifdef USE_MPI
 		m_xdmfWriter->setComm(m_comm);
 #endif // USE_MPI
 
 		m_xdmfWriter->init(variables, std::vector<const char*>());
-		m_xdmfWriter->setMesh(nCells, static_cast<const unsigned int*>(info.buffer(CELLS)), nVertices, static_cast<const double*>(info.buffer(VERTICES)), param.timestep != 0);
```
Thomas-Ulrich, line 88: same

src/ResultWriter/FaultWriterC.cpp
Thomas-Ulrich: (sorry I know it is annoying)

ravil-mobile: why? The old formatting was difficult to read.

Thomas-Ulrich: because it is then difficult to track changes?
We could automate the formatting at some point as we did for PUMGEN?
I had a similar review in another PR.

Thomas-Ulrich: anyway, that was just my 2 cents, you got my approval if you don't want to change it.

Thomas-Ulrich (approved):

src/ResultWriter/FaultWriterC.cpp
ravil-mobile: >because it is then difficult to track changes?

Really?

>  We could automate the formatting at some point as we did for PUMGEN?

I also proposed clang-format for SeisSol a long ago but it was rejected (or it is better to say postponed). But it is inevitable and at tome point it is going to be introduced. I am personally for introducing clang-format to SeisSol as soon as possible. I know that it will introduce lot's of changes in one commit and lot's people will complain whose branches are a bit behind of the master. But on the other hand, there will never be perfect time to seamlessly introduce clang-format to SeisSol.

Thomas-Ulrich:
fixes #80

PR #316, opened 2021-03-19
Fix pinning for GCC

krenzland:
Our current pinning method does not work for the GCC compiler.
Or rather, with its standard OpenMP runtime. It sets the affinity mask for the entire process from the OpenMP pinning mask at the beginning of the process. For intel/clang compiler, this only happens after the first OpenMP call.

This PR implements a different strategy:

- Find affinity mask for all OpenMP threads on every process
- Find affinity mask for all OpenMP threads on every *node*
- Pin communication thread to all cores that are not used by OpenMP.

Furthermore, if the option NUMA_AWARE_PINNING is used, we only pin the communication thread to the cores that are not used by OpenMP and that are in on of the numa domains used by OpenMP.
This option needs libnuma which is available basically everywhere (module numactl most often).

Note that the pinning should still work without NUMA_AWARE_PINNING but might be not optimal, depending on how the MPI runtime pins the threads.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,6 +199,14 @@ if (COMMTHREAD)
   target_compile_definitions(SeisSol-lib PUBLIC USE_COMM_THREAD)
 endif()
 
+if (NUMA_AWARE_PINNING)
+  target_compile_definitions(SeisSol-lib PUBLIC USE_NUMA_AWARE_PINNING)
+  find_package(NUMA REQUIRED)
+
+  target_include_directories(SeisSol-lib SYSTEM PUBLIC ${NUMA_INCLUDE_DIR})
+  target_link_libraries(SeisSol-lib PUBLIC ${NUMA_LIBRARY})
+endif()
+
 #set(HDF5_PREFER_PARALLEL True)
 if (NETCDF)
   find_package(NetCDF REQUIRED)
--- a/cmake/process_users_input.cmake
+++ b/cmake/process_users_input.cmake
@@ -67,7 +67,7 @@ set(MEMORY_LAYOUT "auto" CACHE FILEPATH "A file with a specific memory layout or
 
 option(COMMTHREAD "Use a communication thread for MPI+MP." OFF)
 
-option(NUMA_AWARE_PINNING "Use libnuma to pin threads to correct NUMA nodes" OFF)
+option(NUMA_AWARE_PINNING "Use libnuma to pin threads to correct NUMA nodes" ON)
 
 
 set(LOG_LEVEL "warning" CACHE STRING "Log level for the code")
--- a/cmake/process_users_input.cmake
+++ b/cmake/process_users_input.cmake
@@ -65,8 +65,9 @@ set(NUMBER_OF_FUSED_SIMULATIONS 1 CACHE STRING "A number of fused simulations")
 
 set(MEMORY_LAYOUT "auto" CACHE FILEPATH "A file with a specific memory layout or auto")
 
-
 option(COMMTHREAD "Use a communication thread for MPI+MP." OFF)
+
+option(NUMA_AWARE_PINNING "Use libnuma to pin threads to correct NUMA nodes" OFF)
 
 
 set(LOG_LEVEL "warning" CACHE STRING "Log level for the code")
--- a/src/Parallel/Pin.cpp
+++ b/src/Parallel/Pin.cpp
@@ -43,6 +43,12 @@
 #include <sys/sysinfo.h>
 #include <sched.h>
 #include <sstream>
+#include <set>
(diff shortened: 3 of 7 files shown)


CMakeLists.txt
```diff
@@ -199,6 +199,14 @@ if (COMMTHREAD)
...
 endif()
 
+if (NUMA_AWARE_PINNING)
+  target_compile_definitions(SeisSol-lib PUBLIC USE_NUMA_AWARE_PINNING)
+  find_package(NUMA REQUIRED)
+
+  target_include_directories(SeisSol-lib SYSTEM PUBLIC ${NUMA_INCLUDE_DIR})
+  target_link_libraries(SeisSol-lib PUBLIC ${NUMA_LIBRARY})
+endif()
+
```
uphoffc, line 209: I would invert that NO_NUMA_AWARE_PINNING

-> should be the default but people can disable it if they're lazy

cmake/process_users_input.cmake
```diff
@@ -65,9 +65,10 @@ set(NUMBER_OF_FUSED_SIMULATIONS 1 CACHE STRING "A number of fused simulations")
 
 set(MEMORY_LAYOUT "auto" CACHE FILEPATH "A file with a specific memory layout or auto")
 
-
 option(COMMTHREAD "Use a communication thread for MPI+MP." OFF)
 
+option(NUMA_AWARE_PINNING "Use libnuma to pin threads to correct NUMA nodes" OFF)
```
uphoffc, line 70: -> ON

src/Parallel/Pin.cpp
```diff
@@ -43,6 +43,12 @@
...
 #include <sched.h>
 #include <sstream>
+#include <set>
+#include "Parallel/MPI.h"
+
+#ifdef USE_NUMA_AWARE_PINNING
+#include "numa.h"
+#endif
 
 seissol::parallel::Pinning::Pinning() {
```
uphoffc, line 53: Process mask could be deleted now?

src/Parallel/Pin.cpp
```diff
@@ -74,9 +80,42 @@ cpu_set_t seissol::parallel::Pinning::getWorkerUnionMask() const {
...
+  for (int cpu = 0; cpu < get_nprocs(); ++cpu) {
+    const bool isFree = !CPU_ISSET(cpu, &nodeOpenMpMask);
+    if (isFree) {
+      const int numaNode = numa_node_of_cpu(cpu);
+      const bool isValidNumaNode = numaDomainsOfThisProcess.count(numaNode) != 0;
+      if (isValidNumaNode) {
+        CPU_SET(cpu, &freeMask);
+      }
+    }
+  }
```
uphoffc, line 107: Does SeisSol crash if there is no free cpu? If yes, then it's fine.

(Could be a pathological situation where there is no free core associated with the numa domain. But that's a sign that the affinity settings are messed up.)

src/Parallel/Pin.cpp
```diff
@@ -103,3 +142,28 @@ std::string seissol::parallel::Pinning::maskToString(cpu_set_t const& set) {
...
 }
+
+cpu_set_t seissol::parallel::Pinning::getNodeMask() const {
+  const auto workerMask = getWorkerUnionMask();
+
+  // We have to use this due to the insanity of std::vector<bool>
+  auto workerMaskArray = std::vector<char>( get_nprocs(), 0);
+  for (int cpu = 0; cpu < get_nprocs(); ++cpu) {
+    workerMaskArray[cpu] = CPU_ISSET(cpu, &workerMask);
+  }
```
uphoffc, line 153: My solution is much nicer :D But doesn't matter here.

src/SeisSol.cpp
```diff
@@ -108,10 +108,12 @@ bool seissol::SeisSol::init(int argc, char* argv[])
 
 #ifdef _OPENMP
   logInfo(rank) << "Using OMP with #threads/rank:" << omp_get_max_threads();
-  logInfo(rank) << "Overall affinity            :" << parallel::Pinning::maskToString(
+  logInfo(rank) << "Overall affinity (this process)      :" << parallel::Pinning::maskToString(
       pinning.getProcessMask());
```
uphoffc, line 112: Necessary?

CMakeLists.txt
krenzland: I've kept that as NUMA_AWARE_PINNING because otherwise this would be the only option that is negated. I've made it the default though.

src/SeisSol.cpp
krenzland: No, is actually really confusing when using gcc..

krenzland:
Thanks, Carsten!

uphoffc (approved):

PR #317, opened 2021-03-30
Removed some warnings (gcc, gfortran)

ravil-mobile:
* updated reference to utils and xdmfwriter submodules

The majority of warnings are coming from the following 3 sources:
- Variable<real[kernels::size<tensor::Qane>()]> dofsAne;
- double omega[NUMBER_OF_RELAXATION_MECHANISMS];
- double theta[NUMBER_OF_RELAXATION_MECHANISMS][3];
It is going to be addressed by a dedicated PR


sebwolf-de (approved):
Great

PR #319, opened 2021-04-07
Compute L2 errors in viewrec

sebwolf-de:
I added a feature to view L2 Errors with viewrec, when comparing traces.

### DIFF
--- a/postprocessing/visualization/receiver/src/View.py
+++ b/postprocessing/visualization/receiver/src/View.py
@@ -204,15 +204,15 @@ class View(QWidget):
             p.plot(wf.time, waveform, label=str(nWf))
             p.set_xlabel('t (s)')
           p.set_ylabel(name)
-          #print L2 error
+          #print L2 difference
           if nWf > 0 and not self.diff.isChecked():
             time_union = numpy.union1d(wf.time, wf_ref.time)
             wf_interp = scipy.interpolate.interp1d(wf.time, waveform)
             ref_interp = scipy.interpolate.interp1d(wf_ref.time, wf_ref.waveforms[name])
             wf_union = wf_interp(time_union)
             ref_union = ref_interp(time_union)
-            error = numpy.sqrt(scipy.integrate.trapz((wf_union - ref_union)**2))
-            p.text(0.05, 1-0.1*nWf, "L2 error={:.2e}".format(error), transform = p.transAxes)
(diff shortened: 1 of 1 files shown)


uphoffc (commented):
I wonder whether one should name it "L2 error", because the first one might not necessarily be the ground truth. Maybe "Difference in L2 norm"?

postprocessing/visualization/receiver/src/View.py
```diff
@@ -201,6 +204,15 @@ def plot(self):
...
           p.set_ylabel(name)
+          #print L2 error
+          if nWf > 0 and not self.diff.isChecked():
+            time_union = numpy.union1d(wf.time, wf_ref.time)
+            wf_interp = scipy.interpolate.interp1d(wf.time, waveform)
+            ref_interp = scipy.interpolate.interp1d(wf_ref.time, wf_ref.waveforms[name])
+            wf_union = wf_interp(time_union)
+            ref_union = ref_interp(time_union)
+            error = numpy.sqrt(scipy.integrate.trapz((wf_union - ref_union)**2))
+            p.text(0.05, 1-0.1*nWf, "L2 error={:.2e}".format(error), transform = p.transAxes)
```
Thomas-Ulrich, line 215: f strings are now recommanded over {}.format

uphoffc (approved):

postprocessing/visualization/receiver/src/View.py
```diff
@@ -201,6 +204,15 @@ def plot(self):
...
             p.set_xlabel('t (s)')
           p.set_ylabel(name)
+          #print L2 error
+          if nWf > 0 and not self.diff.isChecked():
+            time_union = numpy.union1d(wf.time, wf_ref.time)
+            wf_interp = scipy.interpolate.interp1d(wf.time, waveform)
+            ref_interp = scipy.interpolate.interp1d(wf_ref.time, wf_ref.waveforms[name])
+            wf_union = wf_interp(time_union)
+            ref_union = ref_interp(time_union)
+            error = numpy.sqrt(scipy.integrate.trapz((wf_union - ref_union)**2))
```
Thomas-Ulrich: Sorry for this late comment, but actually, you could have specified the x parameter of scipy.integrate.trapz.
Then you can compare the diff across several couples of simulations.
(else your diff is strongly dependent on the time samples in each simulation).

PR #322, opened 2021-04-16
Added a correct processes pinning for GPU computing to docu

ravil-mobile:
A correct process pinning for GPU computing has been added to docu.


krenzland (approved):
LGTM

PR #325, opened 2021-04-21
Added pybind11 to seissol-proxy

ravil-mobile:
pybind11 is used to provide Python interface to proxy (but not only!). Different py-runners can be added, for example, for performance monitoring.

Unfortunately, everything needs to be compiled as position independent: with -fPIC flag. I don't know whether it is going to affect performance (I didn't noticed). However, the feature is optional.

Added two extra variables in cmake: EXTRA_FLAGS and ENABLE_PIC_COMPILATION.
* EXTRA_FLAGS is equal to `-fPIC` if PROXY_PYBINDING=ON
* ENABLE_PIC_COMPILATION can be used to signal our sub-modules to generate PIC code.


auto_tuning/proxy/src/poxy_main.cpp
```diff
@@ -0,0 +1,40 @@
+#include <utils/args.h>
+#include "proxy_common.hpp"
+#include<iostream>
```
sebwolf-de, line 3: Missing blank

auto_tuning/proxy/src/proxy-runners/requirements.txt
```diff
@@ -0,0 +1,2 @@
+matplotlib
```
sebwolf-de, line 1: Why do we need matplotlib?

auto_tuning/proxy/src/proxy-runners/run-all.py
```diff
@@ -0,0 +1,57 @@
...
+    output['NZ GFlops'].append(result.non_zero_gflops)
+    output['kernel_str'].append(pb.Aux.kernel_to_string(kernel))
+    bar.next()
+
+if args.output_type == 'j':
+  json_output = json.dumps(output)
+  print(json_output)
+else:
+  for key, value in output.items():
+    print(f'{key}:\t', value)
```
sebwolf-de, line 57: You could use pandas, it can output json and pretty print tables.

auto_tuning/proxy/src/proxy_bindings.cpp
```diff
@@ -0,0 +1,49 @@
+#include "proxy_common.hpp"
+#include <pybind11/pybind11.h>
+namespace py = pybind11;
+
+ProxyOutput runProxy(ProxyConfig config);
```
sebwolf-de, line 5: Why do you not add it to `proxy_common.hpp` or another header?

auto_tuning/proxy/src/poxy_main.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+ProxyOutput runProxy(ProxyConfig config);
```
sebwolf-de, line 5: Why do you not add it to proxy_common.hpp or another header?

sebwolf-de (commented):
Hi, in general I very much like this PR, but I couldn't compile it with intel compilers.

auto_tuning/proxy/src/proxy-runners/run-all.py
ravil-mobile: I like your idea!

auto_tuning/proxy/src/proxy_bindings.cpp
ravil-mobile: done

auto_tuning/proxy/src/poxy_main.cpp
ravil-mobile: done

auto_tuning/proxy/src/poxy_main.cpp
ravil-mobile: done

auto_tuning/proxy/src/poxy_main.cpp
```diff
@@ -0,0 +1,40 @@
+#include <utils/args.h>
```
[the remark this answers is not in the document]
ravil-mobile: done

auto_tuning/proxy/src/proxy-runners/requirements.txt
ravil-mobile: I added an optional output from matplotlib

cmake/process_users_input.cmake
```diff
@@ -218,3 +219,11 @@ endfunction()
 
 cast_log_level_to_int(LOG_LEVEL LOG_LEVEL)
 cast_log_level_to_int(LOG_LEVEL_MASTER LOG_LEVEL_MASTER)
+
+if (PROXY_PYBINDING)
+    set(EXTRA_FLAGS -fPIC)
```
krenzland, line 224: Can you rename EXTRA_FLAGS to something like EXTRA_CXX_FLAGS or so? Was pretty confused by the name.

CMakeLists.txt
```diff
@@ -9,15 +9,15 @@ project(SeisSol LANGUAGES C CXX Fortran)
...
+#       switches: HDF5, NETCDF, METIS, MPI, OPENMP, ASAGI, SIONLIB, MEMKIND, PROXY_PYBINDING, ENABLE_PIC_COMPILATION
 #
 #       user's input: HOST_ARCH, DEVICE_ARCH, DEVICE_SUB_ARCH,
 #                     ORDER, NUMBER_OF_MECHANISMS, EQUATIONS,
 #                     PRECISION, DYNAMIC_RUPTURE_METHOD,
 #                     PLASTICITY, NUMBER_OF_FUSED_SIMULATIONS,
 #                     MEMORY_LAYOUT, COMMTHREAD,
 #                     LOG_LEVEL, LOG_LEVEL_MASTER,
-#                     GEMM_TOOLS_LIST
+#                     GEMM_TOOLS_LIST, EXTRA_FLAGS
```
krenzland, line 20: It's a bit confusing what EXTRA_FLAGS mean (not defined anywhere). I also think that a name similar to EXTRA_CXX_FLAGS would be clearer.

CMakeLists.txt
```diff
@@ -332,12 +333,22 @@ add_executable(SeisSol-bin src/main.cpp)
...
 set_target_properties(SeisSol-bin PROPERTIES OUTPUT_NAME "SeisSol_${EXE_NAME_PREFIX}${PLASTICITY_NAME_SUFFIX}")
 
-add_executable(SeisSol-proxy
-        auto_tuning/proxy/src/proxy_seissol.cpp
-        auto_tuning/proxy/src/flop_counter.cpp
-)
-target_link_libraries(SeisSol-proxy PUBLIC SeisSol-lib)
-set_target_properties(SeisSol-proxy PROPERTIES OUTPUT_NAME "SeisSol_proxy_${EXE_NAME_PREFIX}")
+# SeisSol proxy-core
+add_library(proxy-core auto_tuning/proxy/src/proxy_seissol.cpp
```
krenzland, line 338: Could we please keep using the prefix SeisSol- for all targets? Makes it a bit more organized.

auto_tuning/proxy/src/proxy-runners/run-all.py
```diff
@@ -0,0 +1,78 @@
+import os, sys, inspect
+currentdir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
```
krenzland, line 2: Just as a hint: Instead of inspect you can also use `__file__`. Inspect is fine as well, of course.

auto_tuning/proxy/src/proxy-runners/run-all.py
```diff
@@ -0,0 +1,78 @@
...
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-c', '--cells', default=50000, type=int, help="num cells in a time cluster")
+parser.add_argument('-t', '--timesteps', default=30, type=int, help="num time steps/repeats")
+parser.add_argument('--output_type', default='c', type=str, help="console(p), json (j), matplotlib (m)")
+parser.add_argument('-o', '--output_dir', default='.', type=str, help="relative output directory")
+args = parser.parse_args()
+
+if not args.output_type in ['j', 'c', 'p']:
+    print(f'notice: output only to terminal')
```
krenzland, line 24: You can make this a bit more clearer by using the argument `choices` from `parser.add_argument()`
https://stackoverflow.com/a/46385352

auto_tuning/proxy/src/proxy-runners/simple-runner.py
```diff
@@ -0,0 +1,21 @@
...
+sys.path.insert(0, parentdir)
+
+import argparse
+import proxy_bindings as pb
+
+parser = argparse.ArgumentParser()
+parser.add_argument('-c', '--cells', default=100000, type=int, help="num cells in a time cluster")
+parser.add_argument('-t', '--timesteps', default=20, type=int, help="num time steps/repeats")
+parser.add_argument('-k', '--kernel', default='all', type=str, help="kernel types")
+args = parser.parse_args()
```
krenzland, line 13: Here you can also use `choices` to validate that the user is using the correct kernels.

auto_tuning/proxy/src/proxy_common.hpp
```diff
@@ -0,0 +1,117 @@
...
+  local,
+  neigh,
+  ader,
+  localwoader,
+  neigh_dr,
+  godunov_dr
+};
+
+struct ProxyConfig {
+  unsigned cells{static_cast<unsigned>(1e5)};
```
krenzland, line 20: You can also use `1e5U instead of the static cast`

auto_tuning/proxy/src/proxy_common.hpp
```diff
@@ -0,0 +1,117 @@
...
+  double hardwareFlopPerCycle{};
+  double bytesPerCycle{};
+  double nonZeroGFlops{};
+  double hardwareGFlops{};
+  double gibPerSecond{};
+};
+
+ProxyOutput runProxy(ProxyConfig config);
+
+struct Aux {
```
krenzland, line 45: Why is that a struct? It only has static members. Should be a namespace imo. Or is it needed for pybind?

auto_tuning/proxy/src/proxy-runners/run-all.py
```diff
@@ -0,0 +1,78 @@
...
+
+kernels = [pb.Kernel.all,
+           pb.Kernel.local,
+           pb.Kernel.neigh,
+           pb.Kernel.ader,
+           pb.Kernel.localwoader,
+           pb.Kernel.neigh_dr,
+           pb.Kernel.godunov_dr]
```
krenzland, line 32: We could also use the helper function from Aux that you export. This way, the code would be more resilient.

auto_tuning/proxy/src/proxy-runners/run-all.py
```diff
@@ -0,0 +1,78 @@
...
+        bar.next()
+
+# prepare a unique file suffix
+time_obj = time.localtime()
+now = time.strftime("%B-%H-%M-%S_%Y", time_obj)
+
+# prepare a directory where to write data
+output_dir = os.path.join(os.getcwd(), args.output_dir)
+if not os.path.exists(output_dir):
+    os.makedirs(output_dir)
```
krenzland, line 57: `os.makedirs(output_dir, exist_ok=True)`

auto_tuning/proxy/src/proxy-runners/run-all.py
```diff
@@ -6,19 +6,22 @@
...
+parser.add_argument('--output_type', default='c', type=str, help="console(p), json (j), matplotlib (m)")
+parser.add_argument('-o', '--output_dir', default='.', type=str, help="relative output directory")
 args = parser.parse_args()
 
-if not args.output_type in ['j', 'c']:
-  print(f'error: expected `j` of `c` as output type, given `{args.output_type}`')
-  parser.print_help()
-  sys.exit(-1)
+if not args.output_type in ['j', 'c', 'p']:
+    print(f'notice: output only to terminal')
```
sebwolf-de: You, can use the `choices` option: https://docs.python.org/3/library/argparse.html#choices to make the help redundant.
Also I don't completly understand your logic. You give options `p, j, m` in the help, but then ask for `j, c, p` later.

krenzland (changes requested):
Very nice! I've added some comments but the overall design is great.
Maybe we should document this at some point? (Separate PR is fine, of course, as the proxy is currently not documented)

auto_tuning/proxy/src/proxy-runners/run-all.py
sebwolf-de: Is there a lock for reviews?

krenzland: Apparently not :)

CMakeLists.txt
ravil-mobile: sure. Btw, is it necessary to have `SeisSol` as a prefix? I mean why can we just use prefix `seissol` (it is just easy to type)

auto_tuning/proxy/src/proxy-runners/run-all.py
ravil-mobile: Can you elaborate it? I didn't get it

auto_tuning/proxy/src/proxy_common.hpp
ravil-mobile: Yes, I can do it. I used `struct` as kinda namespace.

> Or is it needed for pybind?

No, you can bind methods as well

auto_tuning/proxy/src/proxy-runners/run-all.py
krenzland: You can export this one function here:
https://github.com/SeisSol/SeisSol/blob/2d9a950251d5ea74303a9613ddc887dacbb0816e/auto_tuning/proxy/src/proxy_common.hpp#L62
And then use this for a list of valid kernels?

auto_tuning/proxy/src/proxy-runners/run-all.py
ravil-mobile: Completely forgot about `choices`. Thanks for reminding!

> You give options p, j, m in the help, but then ask for j, c, p later

sorry, a bunch of typos

auto_tuning/proxy/src/proxy-runners/run-all.py
ravil-mobile: Agree

auto_tuning/proxy/src/proxy_common.hpp
ravil-mobile: It doesn't work for me. 1e5 if floating point value. Prefix `U` is meant to be for integer types. Could you share a code snippet if it works for you with gcc@10?

krenzland: Ah, right. My mistake. I'd use `100'000U;`, but your option is also good then.

auto_tuning/proxy/src/proxy_common.hpp
ravil-mobile: I remembered why I did in this way. So, if it a struct it is going to be a used-derived type with a bunch of inline static methods and I can safely include it in other source files avoiding multiple definitions. If it is a bunch of functions then I need to create another translation unit and add it to cmake. In general, I would leave it as it is.

In terms of namespaces, I agree but the entire proxy was written without namespaces. I guess it is better to introduce proxy namespace in a separate PR.

auto_tuning/proxy/src/proxy-runners/run-all.py
ravil-mobile: removed it

auto_tuning/proxy/src/proxy-runners/simple-runner.py
ravil-mobile: done

auto_tuning/proxy/src/proxy-runners/run-all.py
ravil-mobile: done

auto_tuning/proxy/src/proxy-runners/run-all.py
ravil-mobile: replaced with 

```
plotting_options = ['csv, json', 'matplotlib']
...
parser.add_argument('--output_type', choices=plotting_options, default='csv', help='format to save')
```
done

CMakeLists.txt
ravil-mobile: done

CMakeLists.txt
ravil-mobile: done

cmake/process_users_input.cmake
ravil-mobile: done

auto_tuning/proxy/src/proxy-runners/run-all.py
ravil-mobile: done

CMakeLists.txt
krenzland: In principle it doesnt matter but wouldn't it be annoying to change it now? Imo it's fine as long as it is consistent

sebwolf-de (approved):

auto_tuning/proxy/src/proxy_common.hpp
krenzland: Okay, sure. Doesn't matter. There shouldn't be any problem when defining inline functions imultiple times. That's exactly what inline should allow.

krenzland (approved):

PR #329, opened 2021-04-23
Ci merge

VMPW:
Added more submodule set-url statements (worked with less on branch) but failed in master, so now all submodules have their path set explicitly now.


ravil-mobile:
Hi, thanks for fixing it. Maybe you can squash everything in one commit and force push?

krenzland (approved):

krenzland:
You can also squash directly when merging.

Issue #330, opened 2021-04-23
Docker SeisSol fails with HDF5 / MPI error message when output destination directory doesn't exist


jbiemiller:
**Describe the bug**
When first using the Docker version of SeisSol to run training examples, the Northridge kinematic rupture example failed with an HDF5 / MPI error (see below for error text).

The issue was that I hadn't created the 'output' subdirectory where output files were being directed. Fixed by simply creating the 'output' directory and re-running, but it would be helpful if the error message indicated the missing directory issue.

**Expected behavior**
Kinematic rupture model of Northridge

**To Reproduce**
Run Northridge example from Training materials with Docker SeisSol implementation, but don't create the 'output' directory before running.

**Screenshots/Console output**

Rank:        0 | Info    | <--------------------------------------------------------->
Rank:        0 | Info    | <     Start inioutput_SeisSol ...                         >
Rank:        0 | Info    | <--------------------------------------------------------->
Fri Apr 23 17:00:12, Info:  Initializing free surface integrator. 
Fri Apr 23 17:00:12, Info:  Initializing free surface integrator. Done. 
Fri Apr 23 17:00:12, Info:  Initializing free surface output. 
Fri Apr 23 17:00:12, Error: An HDF5 error occurred in the XDMF writer (/tmp/SeisSol/submodules/xdmfwriter/backends/HDF5.h: 174)
Backtrace:
SeisSol_Release_dhsw_4_elastic(+0x674ca) [0x560f667404ca]
SeisSol_Release_dhsw_4_elastic(+0xdc560) [0x560f667b5560]
SeisSol_Release_dhsw_4_elastic(+0xdef91) [0x560f667b7f91]
SeisSol_Release_dhsw_4_elastic(+0xfd296) [0x560f667d6296]
SeisSol_Release_dhsw_4_elastic(+0x113f0c) [0x560f667ecf0c]
SeisSol_Release_dhsw_4_elastic(+0x14d57b) [0x560f6682657b]
SeisSol_Release_dhsw_4_elastic(+0x14d908) [0x560f66826908]
SeisSol_Release_dhsw_4_elastic(+0x118b83) [0x560f667f1b83]
SeisSol_Release_dhsw_4_elastic(+0x116aa5) [0x560f667efaa5]
SeisSol_Release_dhsw_4_elastic(+0x61241) [0x560f6673a241]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xeb) [0x14e35dea509b]
[...]

PR #331, opened 2021-04-26
Update xdmf writer.

krenzland:
As mentioned in #330, currently SeisSol fails with a very strange error, when an output file is not found. This PR should address that issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -298,7 +298,7 @@ endif()
 # C++ compiler settings
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
   target_compile_options(SeisSol-lib PUBLIC -fopenmp -pedantic $<$<OR:$<COMPILE_LANGUAGE:CXX>,$<COMPILE_LANGUAGE:C>>:-Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas>)
-  target_link_libraries(SeisSol-lib PUBLIC "-fopenmp")
+  target_link_libraries(SeisSol-lib PUBLIC "-fopenmp" "-lstdc++fs")
 
   # using GCC
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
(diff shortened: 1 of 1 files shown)


sebwolf-de (approved):
LGTM, tested with HDF5 and Posix backend.

Thomas-Ulrich:
Cool, thx for fixing that!
Why don't you use in the test the 'parentPath` variable that you just created ;-)

ravil-mobile:
Wait. XdmfWriter CI pipeline got broken: `https://travis-ci.org/github/TUM-I5/XdmfWriter/builds/768393262`

Let's fix it first

krenzland:
Are you going to or should I?
Wasn't aware that we have a pipeline for this :)

(Fix is trivial: Need to use c++17)

ravil-mobile:
I'll fix it.

krenzland:
Thanks, Ravil. I've just noticed that we'll run into problems:

> Using this library may require additional compiler/linker options. GNU implementation prior to 9.1 requires linking with -lstdc++fs and LLVM implementation prior to LLVM 9.0 requires linking with -lc++fs. 

Apparently we need to add additional linker flags to use the fs library. I've used gcc10 locally and thus did not need it.
Not sure whether we need any flags with the Intel compiler. @sebwolf-de did you use intel compiler?

sebwolf-de:
I've used gcc 9.3

sebwolf-de:
It does not compile with Intel 19 though

krenzland:
> It does not compile with Intel 19 though

If you already have the setup, what exactly is the error?

sebwolf-de:
```
/work_fast/ga35kum/SeisSol/master/submodules/xdmfwriter/backends/Base.h(47): catastrophic error: cannot open source file "filesystem"
  #include <filesystem>
                       ^
```

krenzland:
Which gcc modules are loaded?

krenzland:
I've added a commit that adds the necessary flags.

sebwolf-de (approved):

ravil-mobile:
It is cool that it is working. I knew about `stdc++fs` because I used it for another project. Actually I was surprised at the beginning that you managed to compile seissol without it. But then I assumed that you probably used federa32 with gcc10 and std::filesystem probably became a build-in feature.

krenzland:
Yes, that's exactly it. I'm too used to modern compilers now :)

PR #332, opened 2021-04-26
Remove compile option for plasticity

krenzland:
Note: This PR is very much still in progress.

This PR removes the compile time option for plasticity and replaces it with a runtime option.
By using templates + constexpr if, we can create two optimized paths for the neighbor integration kernel: One with plasticity, one without.
In this way, this change comes without much (or any) performance cost.

This is very poorly tested.

@ravil-mobile Can you please have a look at the GPU version? I've just deleted the USE_PLASTICITY macros in your kernels/init. I think it should be fine to just do an i`f(usePlasticity)` in some places but I wasn't really sure. Also, I have currently no GPU, so I can't test this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,11 +108,6 @@ add_custom_command(
        COMMENT "Codegen for tensor stuff."
        )
 
-if (PLASTICITY)
-  target_compile_definitions(SeisSol-lib PUBLIC USE_PLASTICITY)
-  set(PLASTICITY_NAME_SUFFIX "_plasticity")
-endif()
-
 if (PLASTICITY_METHOD STREQUAL "ip")
   target_compile_definitions(SeisSol-lib PUBLIC USE_PLASTICITY_IP)
 elseif (PLASTICITY_METHOD STREQUAL "nb")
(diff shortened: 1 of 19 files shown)


ravil-mobile:
Hi, I looked through. As for me everything is fine. I will cone and try to compile with CUDA

ravil-mobile:
Hi, everything compiles for GPUs. I compared outputs of tpv12-13 obtained with CPUs and GPUs and they are ok.

krenzland:
Thanks, Ravil! Did you try also without plasticity? I'm not sure if that case is handled correctly, currently.

ravil-mobile:
I tried tpv5 as well. It work as well. Output also looks correct.

This
https://github.com/SeisSol/SeisSol/blob/71e39d89a257af0d5e8fe8e2f20334afaeba2147/src/Solver/time_stepping/TimeCluster.cpp#L720-L731

and this
https://github.com/SeisSol/SeisSol/blob/71e39d89a257af0d5e8fe8e2f20334afaeba2147/src/Initializer/MemoryManager.cpp#L760-L767

should be enough. I would suggest to add an `if-statement` in the last code snippet for plasticity. It is not strictly necessary but I guess it will be easy to understand the code. Otherwise, one can ask why do you perform recording and later do not use these data

krenzland:
Okay, will do that. Thanks!

krenzland:
@ravil-mobile I've added your suggested modification. Could you quickly check if it compiles on GPUs? Thanks!

This PR is now ready for review. Please try this with your favorite scenario. It should work.

ravil-mobile:
Hi @krenzland. Yes, sure. I will give it a try.

ravil-mobile:
Everything looks good.

ravil-mobile (approved):

PR #334, opened 2021-05-04
Add fileystem detection

uphoffc:
Filesystem flags didn't work on my system


krenzland (commented):
Looks mostly good, I'm just thinking whether it might be a bit more elegant to do this using the find package mechanism? Code would be identical but the error message would indicate that a library wasn't found.
Shouldn't require much more code.

krenzland:
Oh, and the clang flags were wrong. I took them from here: https://en.cppreference.com/w/cpp/filesystem but haven't tested it. I have tested it afterwards and noticed that it does not work on my system but I'm using a more recent version of clang. Not sure what flags are needed for older clang versions, but we shouldn't bother imo.

uphoffc:
> Looks mostly good, I'm just thinking whether it might be a bit more elegant to do this using the find package mechanism?

How exactly?

krenzland:
Basically using find_package(StdFilesystem REQUIRED) and to create a FindStdFilesystem. (Name is arbitrary and pretty terrible imo!)
Then we can just link this like all other packages.

uphoffc:
It's a library.

krenzland (approved):

PR #335, opened 2021-05-04
Fixes #333

### LINKED ISSUE #333: Inconsistent number of elements in the surface output
**Describe the bug**
I have performed a series of simulations with various point source locations under the same mesh. However, the total number of elements in each surface output slightly differ with each.

**Expected behavior**
I expect the total number of elements in the output remains the same when not altering the mesh file or any other related parameters.

uphoffc:


src/Solver/FreeSurfaceIntegrator.cpp
```diff
@@ -233,10 +233,15 @@ void seissol::solver::FreeSurfaceIntegrator::initializeSurfaceLTSTree(  seissol:
 {
   seissol::initializers::LayerMask ghostMask(Ghost);
 
+  auto const is_duplicate = [&ghostMask, ltsLut](unsigned ltsId) {
+    return ltsId != ltsLut->ltsId(ghostMask, ltsLut->meshId(ghostMask, ltsId));
+  };
+
```
krenzland, line 239: 1) Name should be formatted like isDuplicate
2) Probably should be called isUnique, as it is only used in the negated form

krenzland (changes requested):
Looks good! One minor comment. Haven't tested it but I don't think it breaks anything.

krenzland (approved):

PR #340, opened 2021-05-21
fix: split MPI communicator to exclude non DR nodes from the reduce

Thomas-Ulrich:
sorry about that

### DIFF
--- a/src/ResultWriter/magnitude_output.f90
+++ b/src/ResultWriter/magnitude_output.f90
@@ -66,7 +66,7 @@ MODULE magnitude_output_mod
 
 CONTAINS
 
-  SUBROUTINE magnitude_output(MaterialVal,DISC,MESH,MPI,IO)
+  SUBROUTINE magnitude_output(MaterialVal,DISC,MESH,MPI,IO,DR_comm)
     !< routine outputs the magnitude for each MPI domain that contains a subfault
     !-------------------------------------------------------------------------!
     IMPLICIT NONE
@@ -79,16 +79,15 @@ CONTAINS
     !-------------------------------------------------------------------------!
     ! Local variable declaration                                              !
     INTEGER                         :: iElem,iSide,nSide,iFace
-    INTEGER                         :: stat, UNIT_MAG, iErr
+    INTEGER                         :: stat, UNIT_MAG, iErr, DR_comm
     REAL                            :: magnitude, magnitude0
     REAL                            :: MaterialVal(:,:)
     LOGICAL                         :: exist
     CHARACTER (LEN=5)               :: cmyrank
     CHARACTER (len=200)             :: MAG_FILE
     !-------------------------------------------------------------------------!
-    INTENT(IN)    :: DISC, MESH, MPI, IO
+    INTENT(IN)    :: DISC, MESH, MPI, IO, DR_comm
     !-------------------------------------------------------------------------!
-
     !
     ! Compute output
     magnitude = 0.0D0
@@ -105,7 +104,7 @@ CONTAINS
        ENDIF
     ENDDO
 #ifdef PARALLEL
-    CALL MPI_REDUCE(magnitude,magnitude0,1,MPI%MPI_AUTO_REAL,MPI_SUM,0, MPI%commWorld,iErr)
+    CALL MPI_REDUCE(magnitude,magnitude0,1,MPI%MPI_AUTO_REAL,MPI_SUM,0, DR_comm,iErr)
 #else
     magnitude0 = magnitude
 #endif
--- a/src/ResultWriter/magnitude_output.f90
+++ b/src/ResultWriter/magnitude_output.f90
@@ -76,10 +76,11 @@ CONTAINS
     TYPE(tUnstructMesh)             :: MESH
     TYPE(tMPI)                      :: MPI
     TYPE(tInputOutput)              :: IO
+    integer                         :: DR_comm !< dynamic rupture communicator
     !-------------------------------------------------------------------------!
     ! Local variable declaration                                              !
     INTEGER                         :: iElem,iSide,nSide,iFace
-    INTEGER                         :: stat, UNIT_MAG, iErr, DR_comm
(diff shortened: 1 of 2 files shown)


src/ResultWriter/magnitude_output.f90
```diff
@@ -79,16 +79,15 @@ SUBROUTINE magnitude_output(MaterialVal,DISC,MESH,MPI,IO)
     !-------------------------------------------------------------------------!
     ! Local variable declaration                                              !
     INTEGER                         :: iElem,iSide,nSide,iFace
-    INTEGER                         :: stat, UNIT_MAG, iErr
+    INTEGER                         :: stat, UNIT_MAG, iErr, DR_comm
```
uphoffc, line 82: Should be TYPE(MPI_Comm) :: DR_comm

src/Solver/calc_seissol.f90
```diff
@@ -104,6 +104,7 @@ SUBROUTINE calc_SeisSol(time,timestep,pvar,cvar,EQN,MESH,DISC,SOURCE,BND,IC, &
     INTEGER                       :: plotNr, minl(1), iter
     INTEGER                       :: iElem, iSide, i, iObject, MPIIndex
     INTEGER(KIND=4)               :: now(3)
+    integer                       :: DR_comm                        !< dynamic rupture communicator
```
uphoffc, line 107: Should be TYPE(MPI_Comm) :: DR_comm

src/Solver/calc_seissol.f90
```diff
@@ -215,8 +216,12 @@ SUBROUTINE calc_SeisSol(time,timestep,pvar,cvar,EQN,MESH,DISC,SOURCE,BND,IC, &
 #endif
     ENDIF
 
+#ifdef USE_MPI
+    CALL MPI_Comm_split(MPI%commWorld, EQN%DR, 1, DR_comm, iErr)
```
uphoffc, line 220: This will only work if rank 0 has dynamic rupture faces!

uphoffc (changes requested):
This will only work if EQN%DR = 1 for rank 0. Otherwise, rank 0 ends up in the other group and the magnitude will be zero.

You might just remove EQN%DR.EQ.1 in Line 224 instead of splitting the communicator. Then magnitude_output is called by all processes.

Thomas-Ulrich:
Thanks for the review!
Could it work like that?
```
diff --git a/src/ResultWriter/magnitude_output.f90 b/src/ResultWriter/magnitude_output.f90
index bfc096b8..4c528004 100644
--- a/src/ResultWriter/magnitude_output.f90
+++ b/src/ResultWriter/magnitude_output.f90
@@ -76,10 +76,11 @@ CONTAINS
     TYPE(tUnstructMesh)             :: MESH
     TYPE(tMPI)                      :: MPI
     TYPE(tInputOutput)              :: IO
+    TYPE(MPI_Comm)                  :: DR_comm
     !-------------------------------------------------------------------------!
     ! Local variable declaration                                              !
     INTEGER                         :: iElem,iSide,nSide,iFace
-    INTEGER                         :: stat, UNIT_MAG, iErr, DR_comm
+    INTEGER                         :: stat, UNIT_MAG, iErr, rankDR
     REAL                            :: magnitude, magnitude0
     REAL                            :: MaterialVal(:,:)
     LOGICAL                         :: exist
@@ -105,10 +106,12 @@ CONTAINS
     ENDDO
 #ifdef PARALLEL
     CALL MPI_REDUCE(magnitude,magnitude0,1,MPI%MPI_AUTO_REAL,MPI_SUM,0, DR_comm,iErr)
+    CALL MPI_Comm_rank(DR_comm, rankDR, iErr)
 #else
     magnitude0 = magnitude
+    rankDR = 0
 #endif
-    IF (MPI%myrank.EQ.0) THEN
+    IF (rankDR.EQ.0) THEN

         WRITE(MAG_FILE, '(a,a4,a4)') TRIM(IO%OutputFile),'-MAG','.dat'
         UNIT_MAG = 299875
diff --git a/src/Solver/calc_seissol.f90 b/src/Solver/calc_seissol.f90
index 08ceabfd..2f7f0435 100644
--- a/src/Solver/calc_seissol.f90
+++ b/src/Solver/calc_seissol.f90
@@ -94,6 +94,7 @@ CONTAINS
     TYPE (tUnstructOptionalFields):: OptionalFields
     TYPE (tInputOutput)           :: IO
     TYPE (tMPI)                   :: MPI
+    TYPE(MPI_Comm)                :: DR_comm !< dynamic rupture communicator

     REAL                          :: time                 ! current time
     INTEGER                       :: timestep             ! index of time step
@@ -104,7 +105,6 @@ CONTAINS
     INTEGER                       :: plotNr, minl(1), iter
     INTEGER                       :: iElem, iSide, i, iObject, MPIIndex
     INTEGER(KIND=4)               :: now(3)
-    integer                       :: DR_comm                        !< dynamic rupture communicator
     REAL                          :: tol, InitBegin, InitEnd, LoopBegin, LoopEnd
     REAL                          :: MPITime, MPIdt, dt
     REAL                          :: CPUTime

```
The TYPE(MPI_Comm) does not work for me:
```
/hppfs/work/pr45fi/di73yeq4/SeisSol/src/ResultWriter/magnitude_output.f90(79): error #6457: This derived type name has not been declared.   [MPI_COMM]
    TYPE(MPI_Comm)                  :: DR_comm
```

uphoffc:
Yes that should work.

Maybe TYPE(MPI_COMM_WORLD) ?

Thomas-Ulrich:
> Yes that should work.
> 
> Maybe TYPE(MPI_COMM_WORLD) ?

does not work.
But here it is declared as integer:
https://github.com/SeisSol/SeisSol/blob/master/src/Numerical_aux/typesdef.f90#L739

uphoffc:
ok whatever, then just use integer

uphoffc (approved):

Thomas-Ulrich:
I can't merge.

PR #341, opened 2021-05-25
fix fault strength when tension with Rate and State frictions

Thomas-Ulrich:
Hi, following a question of Kadek about positive Pn0 in SeisSol, I've been looking a the friction laws and came to the conclusion that Rate and State friction laws currently wrongly handle tension, contrary to linear slip weakening.
Let me know what you think of the changes I propose to fix that.
We also need to check the effect of these changes on both TP and tpv104 benchmarks.

Thomas.

### DIFF
--- a/src/Physics/Evaluate_friction_law.f90
+++ b/src/Physics/Evaluate_friction_law.f90
@@ -935,7 +935,7 @@ MODULE Eval_friction_law_mod
          IF (DISC%DynRup%ThermalPress.EQ.1) THEN
              P_f = DISC%DynRup%TP(:,iFace,2)
          ELSE
-              P_f = 0.0
+              P_f = ZERO
          ENDIF
 
          DO j=1,nSVupdates   !This loop corrects SV values
@@ -944,7 +944,7 @@ MODULE Eval_friction_law_mod
              !1.update SV using Vold from the previous time step
              CALL updateStateVariable (EQN, DISC, iFace, nBndGP, SV0, time_inc, SR_tmp, LocSV)
              IF (DISC%DynRup%ThermalPress.EQ.1) THEN
-                 S = -LocMu*min(0, P - P_f)
+                 S = -LocMu*min(ZERO, P - P_f)
                  DO iBndGP = 1, nBndGP
                          !recover original values as it gets overwritten in the ThermalPressure routine
                          Theta_tmp = DISC%DynRup%TP_Theta(iBndGP, iFace,:)
@@ -957,7 +957,7 @@ MODULE Eval_friction_law_mod
              ENDIF
              !2. solve for Vnew , applying the Newton-Raphson algorithm
              !effective normal stress including initial stresses and pore fluid pressure
-             n_stress = min(0, P - P_f)
+             n_stress = min(ZERO, P - P_f)
              CALL IterativelyInvertSR (EQN, DISC, iFace, nBndGP, nSRupdates, LocSR, LocSV, &
                              n_stress, Shtest, invZ, SRtest, has_converged)
 
@@ -985,7 +985,7 @@ MODULE Eval_friction_law_mod
          CALL updateStateVariable (EQN, DISC, iFace, nBndGP, SV0, time_inc, SR_tmp, LocSV)
 
          IF (DISC%DynRup%ThermalPress.EQ.1) THEN
-             S = -LocMu*(P - P_f)
+             S = -LocMu*min(ZERO, P - P_f)
              DO iBndGP = 1, nBndGP
                           Theta_tmp = DISC%DynRup%TP_Theta(iBndGP, iFace,:)
                           Sigma_tmp = DISC%DynRup%TP_sigma(iBndGP, iFace,:)
@@ -1012,8 +1012,9 @@ MODULE Eval_friction_law_mod
          LocMu  = RS_a*LOG(tmp2+SQRT(tmp2**2+1.0D0))
 
          ! update stress change
-         LocTracXY = -((EQN%InitialStressInFaultCS(:,4,iFace) + XYStressGP(:,iTimeGP))/ShTest)*LocMu*(P-P_f)
-         LocTracXZ = -((EQN%InitialStressInFaultCS(:,6,iFace) + XZStressGP(:,iTimeGP))/ShTest)*LocMu*(P-P_f)
+         n_stress = min(ZERO, P - P_f)
+         LocTracXY = -((EQN%InitialStressInFaultCS(:,4,iFace) + XYStressGP(:,iTimeGP))/ShTest)*LocMu*n_stress
+         LocTracXZ = -((EQN%InitialStressInFaultCS(:,6,iFace) + XZStressGP(:,iTimeGP))/ShTest)*LocMu*n_stress
          LocTracXY = LocTracXY - EQN%InitialStressInFaultCS(:,4,iFace)
          LocTracXZ = LocTracXZ - EQN%InitialStressInFaultCS(:,6,iFace)
          !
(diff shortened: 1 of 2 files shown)


palgunadi1993:
Hi Thomas,
Thank you for opening the PR. I tested your changes on TPV104 and it still holds.

Rupture time contour:
Screen Shot 2021-05-25 at 16 21 28

Horizontal shear stress:
Screen Shot 2021-05-25 at 16 36 39

State variable:
Screen Shot 2021-05-25 at 16 37 15

Cheers,
Kadek

AliceGabriel:
Hi,
I am not sure about this change - is tensional normal stress something you
observed in any rsf friction law simulation?
I cc Capser here as our local RSF expert.
Kadek, it would be useful to compare SeisSol's old implementation with
SeisSol patched and it would be useful plotting the normal stress.

uphoffc:
If you don't have tension then the PR doesn't change anything, so there's no harm done. But if you have tension then your coefficient of friction would be negative :-/

palgunadi1993:
Hi,

Positive stress drop most of the time occurs at the intersection of two or more intersected fault/fracture planes when there is "unclamping" dynamic stress transfer. The problem of the positive normal stress is unrealistic high slip at the intersection. 

I have tried Thomas' implementation but with additional changes attached here. 
[normalStressChanges.log](https://github.com/SeisSol/SeisSol/files/6553374/normalStressChanges.log)

The old implementation seems to have still similar normal stress:
[Normal stress for old implementation of RSF]
Pn0oldImplentation

[Normal stress for new implementation of RSF]
Pn0newImplentation

The corresponding absolute slip:
The new implementation seems to depress the unrealistically high slip at the intersection.
[Absolute slip for old implementation of RSF]
ASloldImplentation

[Absolute slip for new implementation of RSF]
ASlnewImplentation

Please let me know what you think.

Best regards,
Kadek

Thomas-Ulrich:
Hi,
Just to clarify, the changes I propose are not meant to suppress positive normal stress, but only to prevent that positive normal stress is associated with fault strength.
In fact, if `n_stress = P - P_f>0` then fault strength in the IterativelyInvertSR routine (`ABS(n_stress)*mu_f`) is positive
https://github.com/SeisSol/SeisSol/blob/master/src/Physics/Evaluate_friction_law.f90#L1198
I think fault strength should be zero then, as it is implemented in the linear slip weakening law.

Also in the TP routine, if `P-Pf>0` then fault strength
`S = -LocMu*(P - P_f) <0` and therefore
`tauV = Sh*SR<0` !fault strenght*slip rate
https://github.com/SeisSol/SeisSol/blob/master/src/Physics/thermalpressure.f90#L97
This would decrease the fault temperature and therefore the pressure.

Thomas-Ulrich:
But, yes the changes that you propose make a lot of sense! yes, I did not see that P-Pf have to be changed also at the end of the routine. Thank you Kakek, great!

Thomas-Ulrich:
There was another one in faultoutput routine.

AliceGabriel:
If you want to test this two changes at once please also provide tpv105 (with thermal pressurisation)

palgunadi1993:
Hi,

I implemented the changes to TPV104 and TPV105, as noted by @uphoffc in these TPVs there is no tension so the changes are not harming the result. To see the result of the TPV's, you can check through scec-benchmark website (Kadek Palgunadi - SeisSol tension).
As Alice suggested, I use my simulation as an example where the scenario would have the wanted effect. 
In my simulation, tension occurs most of the time at the intersection between the main fault and fractures, especially when the Shmax orientation is parallel to the fracture's strike. Dynamic stress transfer from the main fault decreases the normal stress of the fracture at the intersection. Large dynamic stress transfer, on some occasions, is larger than the initial normal stress loading. Because the current implementation of tension is wrongly handled, it produces unrealistic high slip at the intersection. The wanted effect is to suppress the unrealistic high slip at the intersection.
ASloldImplentation120
Figure 1. Unrealistic slip at the intersection. The colorbar is saturated at 2m (the maximum value is up to 240m)

The proposed changes seem to suppress the unrealistic high slip at the intersection by regularizing the positive normal stress to zero (currently avoiding Mode I fracture). 
ASlnewImplentation120
Figure 2. The wanted result. The unrealistic slip at the intersection has vanished. The colorbar is saturated at 2m (the maximum value is up to 2.4m)

Please let me know what you think.

Best regards,
Kadek

Thomas-Ulrich:
Hi, so I think it is ready to be merged, but there are small problems that need to be resolved in the SCEC results uploaded by  @palgunadi1993 .
(e.g. the cplot does not match for x positive (maybe the TP setup that you use is not correct?), and the dip-slip components are inverted so we cannot really compare).

palgunadi1993:
Hi Thomas, Thanks for pointing that out, it was the problem with the script that I use to extract files from SeisSol output for TPV105 (for cplot). Now, I double-check and the is no problem anymore. 

Cplot:
cplot

Dip slip:
dipslip


But overall, as @Thomas-Ulrich  said I think it's ready to be merged.

Thank you Thomas!

palgunadi1993:
Hi Thomas,
I think I don't have rights to approve this. But it works fine as expected for the tension.

AliceGabriel:
Hi
I am not sure if my previous message was lost but in order to be able to approve  merge please
„  compare SeisSol's old implementation with SeisSol patched and it would be useful plotting the normal stress.“

--
Dr. Alice-Agnes Gabriel
bit.ly/AAG8LMU

On 6. Jun 2021, at 22:24, Kadek Hendrawan Palgunadi ***@***.***> wrote:

﻿
Hi Thomas,
I think I don't have rights to approve this. But it works fine as expected for the tension.

Thomas-Ulrich:
Thanks again Kadek for running tpv105 with and without the patch.
Here is the comparison on tpv105, at the receiver closest to the free surface:

faultst000dp030 (strike 0.0 km, dip 3.0 km)
Field: n-stress (Normal stress (MPa))
image

Some slight differences in the first 5s:
Field: h-slip-rate (Horizontal slip rate (m/s))
image
Field: v-shear-stress (Vertical shear stress (MPa))
image

Shall we merge?

palgunadi1993:
Hi,

Additional comparison for the expected result from my fracture network simulation. 
The shown figures are the final simulation time (t=9s) of the difference between SeisSol with and without patch (e.g. with - (subtract) without tension regularization). Unrealistic high slip occurs mostly when the strike of fractures parallel to the Shmax orientation. The figure is zoomed to the fracture that intersects with the main fault and the fracture's strike is parallel to the maximum horizontal stress loading.

Absolute slip difference (ASl)
ASlDiff

Normal stress difference (Pn0)
Pn0Diff

Traction in strike direction (Ts0)
Ts0Diff

Traction in dip direction (Td0)
Td0Diff

If you notice there is ripple-like artifact (e.g. Pn0, Ts0, and Td0) that corresponds to the SeisSol without the patch.
Ts0 without patch:
Ts0WithoutPatch

Ts0 with patch:
Ts0WithPatch

Thank you!

Best regards,
Kadek

AliceGabriel:
Hi, thank you for this thorough analysis! Yes let's merge !! 

But before closing this issue I would like @NicoSchlw to post his recent Ridgecrest model with a fault zone which seems to be affected, Nico should rerun his case with the updated SeisSol.

NicoSchlw:
Hi,

I get unreasonable high slip values at this complex fault intersection of the Ridgecrest model. The slip rate distribution also looks very weird.

Ridgecrest1
Ridgecrest2

Thomas-Ulrich:
HI, you mean, with the master branch implementation? If yes, can you run with the new implementation?

NicoSchlw:
Yes, I will run it again tomorrow

NicoSchlw:
Hi, I ran my model again with the new branch and it looks fine now. But I also increased the resolution.

Thomas-Ulrich:
Can somebody validate the PR?

palgunadi1993:
Hi Thomas,

I don't have the right to do that :)

AliceGabriel (approved):
all good from my side

PR #343, opened 2021-06-02
add filters to extractDataFromUnstructuredOutput.py

Thomas-Ulrich:
I just added some arguments to the script to extract a subset of the paraview output given coordinate ranges.
(was useful for processing the Palu surface output, might be useful for future setups).
I also changed a few things to fix errors identified by flake8.
Example of use:
```
python extractDataFromUnstructuredOutput.py /hppfs/work/pr45fi/ga24dib3/shared/2021-04-29_17-00_0/run_0/output/Sulawesi_WL_Batv1_CFL04b-surface.xdmf --Data U V W --do 100 --precision double --add resampled100s --xfilter -32000.0 24000.0 --yfilter -46000.0 58000.0 --no_ocean
```

Thomas.

### DIFF
--- a/postprocessing/visualization/tools/extractDataFromUnstructuredOutput.py
+++ b/postprocessing/visualization/tools/extractDataFromUnstructuredOutput.py
@@ -36,8 +36,6 @@ parser.add_argument(
     help="output only cells with z center coordinate in range zmin zmax",
     type=float,
 )
-parser.add_argument("--no_ocean", dest="no_ocean", default=False, action="store_true", help="remove sea-surface cells")
-
 args = parser.parse_args()
 
 sx = seissolxdmf.seissolxdmf(args.xdmfFilename)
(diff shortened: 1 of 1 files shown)


postprocessing/visualization/tools/extractDataFromUnstructuredOutput.py
```diff
@@ -15,23 +15,78 @@
...
+    ids = np.intersect1d(ids, id0) if len(ids) else id0
+if args.yfilter:
+    id0 = filter_cells(xyzc[:, 1], args.yfilter)
+    ids = np.intersect1d(ids, id0) if len(ids) else id0
+if args.zfilter:
+    id0 = filter_cells(xyzc[:, 2], args.zfilter)
+    ids = np.intersect1d(ids, id0) if len(ids) else id0
+if args.no_ocean:
+    id0 = np.where(np.abs(xyzc[:, 2]) > 1e-3)[0]
+    ids = np.intersect1d(ids, id0) if len(ids) else id0
```
sebwolf-de, line 68: This filter removes all cells with the z component of their respective center >0, right? Is this really the intended use here?

sebwolf-de: Yes, I also fear it is very specific. What if someone prefers enu  instead of ned orientation? Also you would remove parts, where the topography is above sea level. Since you can achieve this behavior with the z-filter anyways, I think we could remove the ocean filter.

Thomas-Ulrich: I've removed the option.

sebwolf-de (approved):

PR #344, opened 2021-06-07
Discussion: mneme (LTS tree replacement) in proxy

krenzland:
This is a quick and dirty rewrite of our proxy using the mneme library, which is a typesafe AoS/SoA container library.

Please have a look. I'd like to integrate this into SeisSol itself.

@ravil-mobile What do we need to do to ensure that it's going to work with the GPU version? Right now we're using std::arrays for some variables, this is likely going to cause problems. Are you aware of a workaround?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -34,3 +34,6 @@
 [submodule "submodules/Device"]
 	path = submodules/Device
 	url = ../../SeisSol/Device.git
+[submodule "submodules/mneme"]
+	path = submodules/mneme
+	url = ../../TEAR-ERC/mneme.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,6 +272,7 @@ target_include_directories(SeisSol-lib PUBLIC
 				   # todo: use reexport from easi?
 				   ${CMAKE_CURRENT_SOURCE_DIR}/submodules/easi/include
 				   ${CMAKE_CURRENT_SOURCE_DIR}/submodules/yateto/include
+                   ${CMAKE_CURRENT_SOURCE_DIR}/submodules/mneme/include
 				   ${CMAKE_CURRENT_BINARY_DIR}/src/
 )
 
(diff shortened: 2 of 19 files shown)


ravil-mobile:
Hi,

> Right now we're using std::arrays for some variables

Yes. It will. It is not a problem only in case of CUDA because we are adding support for AMD HIP and SYCL. Even if one decides to implement something similar to `std::array` one needs to make sure that GPU headers do not leak to SeisSol code. Otherwise, we need to make changes in CMake and use `nvcc`, `dcp++`, or `hipcc` for building the entire project. The best would be to dump everything to SoA because of coalesced memory access with is more efficient on GPUs. 

Here is another thing. One needs to provide another implementation for `Recorders.h`. I can contribute because in any way a proper abstraction should be made. Currently these recorders support only LTS tree for elastic WP and it should be extended to `viscoelastic2` at some point.

krenzland:
Ravil, just to clarify:
Is it not a problem for CUDA but is it a problem for CUDA&HIP&SYCL? You probably meant the latter, right? What exactly is the problem with std::array and CUDA?
Maybe the easiest way would be to write a custom object that works like std::array but does work on GPUs. We do not need much from std::array, basically a struct wrapping an array together with (optional) iterators is enough. We cannot use cstyle-arrays because then our template fails. Not sure why, probably some allocator problem.

Everything here is SoA. You can switch to AoS by literally changing one template - that's the beauty of this library.

ravil-mobile:
Hi, the main problem with `std::array` on GPUs is that there is no device function `data()` therefore I cannot get a pointer to the actual data. GPU API can copy any object bit-wise to GPU memory. But a compiler must generate both CPU and GPU methods to access fields if they are private.

If you use CUDA/HIP then a method declaration will look like:

```
__host__ __device__ T* data();
```
Only nvcc/hipcc compilers can recognize `__host__` and `__device__` keywords and generate two methods i.e. for CPU and GPU

 But if `__host__` and `__device__` keywords are in a header file then they will propagate to the main source code eventually. GCC and ICC are not aware about these grammar constructs and spit out a syntactic error. Therefore, we will have to compile the entire SeisSol code with nvcc/hipcc. Or, alternatively we have to add some `#ifdef` macros, etc. 

> Maybe the easiest way would be to write a custom object that works like std::array but does work on GPUs.

Good idea. I wanted to propose this one. I guess if one implements `std::array` as a structure and exposes a pointer to actual data then it will work.

ravil-mobile:
Good idea. I will test it whether it works. In theory `x[0]` is done with `T& operator[](size_t i)` which is also a method which should be implemented for both CPU and GPU. But I will check it to be sure.

vikaskurapati:
@davschneller do you know if this PR is still relevant or if we want to continue working on this?

Issue #256, opened 2020-10-15
GCC + rate and state friction very slow


uphoffc:
**Describe the bug**
DR is really slow with gcc 9.2.0

**Expected behavior**
Should as fast as with icc 19.1.0

**To Reproduce**
1. Which version do you use? master
2. Which build settings do you use? gcc 9.2.0 
3. On which machine does your problem occur? Sandy bridge (16 core) and Xeon gold (40c) machine
4. Provide parameter/material files. No.

**Screenshots/Console output**
gcc:
Thu Oct 15 15:46:17, Info:  "computeLocalIntegration" ( per element ): 3.39284e-07 (sample size: 437 , standard error: 8.62748e-10 )
Thu Oct 15 15:46:17, Info:  "computeNeighboringIntegration" ( per element ): 2.87025e-07 (sample size: 437 , standard error: 1.33964e-09 )
Thu Oct 15 15:46:17, Info:  "computeDynamicRupture" ( per element ): **0.000136992** (sample size: 409 , standard error: 4.52834e-07 )

icc:
Thu Oct 15 16:01:38, Info:  "computeLocalIntegration" ( per element ): 3.3736e-07 (sample size: 437 , standard error: 5.50422e-10 )
Thu Oct 15 16:01:38, Info:  "computeNeighboringIntegration" ( per element ): 2.90504e-07 (sample size: 437 , standard error: 1.22483e-09 )
Thu Oct 15 16:01:38, Info:  "computeDynamicRupture" ( per element ): **3.32489e-06** (sample size: 409 , standard error: 1.49403e-08 )

krenzland:
I have compared gcc vs Intel on Mahti as well. They use AMD Rome base CPUs. I've ran our Palu scenario which uses a fast-velocity-weakening friction law.
Performance with gcc was roughly 300GFLOP/s per node, with intel roughly 2k GFLOP/s per node.
Running the same setup without DR (Dr faces retagged as interior cells) I also reached 2k GFLOP/s per node with gcc.

Interestingly, the load balance was at roughly 80% for the gcc. I assume some of the performance loss comes also comes from the load imbalance resulting from the slower FL computation. So it's probably unfair to say that gfortran is 10 times slower. But it is definitely **significantly slower**!

AliceGabriel:
Just a quick note that this effects our training example of Palu, which will be next used Friday in a week and then again in the ChEESE training workshop (if we use the Docker image)....

krenzland:
[...]

PR #345, opened 2021-06-07
Fix flags for AMD Rome when using Intel compiler.

krenzland:
These are the flags that we used for our SC paper submission. They are tested.

We need to use Intel compiler currently because of #256.


sebwolf-de (approved):
LGTM

PR #346, opened 2021-06-08
Thomas/output plasticity

Thomas-Ulrich:
Hi, 
the first commit (working) allows selecting the plastic variable to be outputted (using a new iPlasticityMask variable).
The second commit (not fully working) allow outputting the plastic strain with subcell resolutions.
(at this point, the refinement level is the one from the wavefield output, but we could imaging a refinement_plasticity parameter in the parameter file).
The code almost works, that is it does not crash, but the refined data are nan.
Maybe someone familiar with the wavefield output writer could have a look at my code and fix the bug :-)


src/Numerical_aux/typesdef.f90
```diff
@@ -1084,6 +1084,9 @@ MODULE TypesDef
      LOGICAL                      ,POINTER  :: OutputMask(:)                    !< Mask for variable output
                                                                                 !< .TRUE.  = do output for this variable
                                                                                 !< .FALSE. = do no output for this variable
+     LOGICAL                      ,POINTER  :: PlasticityMask(:)                !< Mask for variable output
```
uphoffc, line 1087: Why is that a pointer when the size is known to be 7?

src/ResultWriter/WaveFieldWriter.cpp
```diff
@@ -280,9 +280,11 @@ void seissol::writer::WaveFieldWriter::init(unsigned int numVars,
 	// Set up for low order output flags
 	m_lowOutputFlags = new bool[WaveFieldWriterExecutor::NUM_LOWVARIABLES];
 	m_numIntegratedVariables = seissol::SeisSol::main.postProcessor().getNumberOfVariables();
+
 	for (size_t i = 0; i < WaveFieldWriterExecutor::NUM_PLASTICITY_VARIABLES; i++) {
-		m_lowOutputFlags[i] = (pstrain != 0L);
+		m_lowOutputFlags[i] = (pstrain != 0L) & (outputMask[numVars + i] != 0);
```
uphoffc, line 285: Should be &&
& = bitwise and
&& = logical and

src/ResultWriter/inioutput_seissol.f90
```diff
@@ -153,6 +153,16 @@ SUBROUTINE inioutput_SeisSol(time,timestep,pvar,cvar,EQN,IC,MESH,MPI,      &
...
     end do
+
+    !plastic strain
+    do i = 1, 7
+        if (io%PlasticityMask(i)) then
+            outputMaskInt(EQN%nVarTotal + i) = 1
+        else
+            outputMaskInt(EQN%nVarTotal + i) = 0
+        end if
+    end do
```
uphoffc, line 164: Why don't you pass on plasticityMask to c_interoperability_initializeIO? Would be much nicer

src/ResultWriter/WaveFieldWriter.cpp
```diff
@@ -280,9 +280,11 @@ void seissol::writer::WaveFieldWriter::init(unsigned int numVars,
 	// Set up for low order output flags
 	m_lowOutputFlags = new bool[WaveFieldWriterExecutor::NUM_LOWVARIABLES];
 	m_numIntegratedVariables = seissol::SeisSol::main.postProcessor().getNumberOfVariables();
+
 	for (size_t i = 0; i < WaveFieldWriterExecutor::NUM_PLASTICITY_VARIABLES; i++) {
```
uphoffc, line 284: NUM_LOWVARIABLES > NUM_PLASTICITY_VARIABLES, so the other entries are just left unitialised?

src/ResultWriter/WaveFieldWriter.cpp
```diff
@@ -280,9 +280,11 @@ void seissol::writer::WaveFieldWriter::init(unsigned int numVars,
...
-		m_lowOutputFlags[i] = (pstrain != 0L);
+		m_lowOutputFlags[i] = (pstrain != 0L) & (outputMask[numVars + i] != 0);
 	}
+
 	seissol::SeisSol::main.postProcessor().getIntegrationMask(&m_lowOutputFlags[WaveFieldWriterExecutor::NUM_PLASTICITY_VARIABLES]);
```
uphoffc, line 288: Doesn't this just overwrite your mask?

src/ResultWriter/WaveFieldWriter.cpp
```diff
@@ -313,16 +315,16 @@ void seissol::writer::WaveFieldWriter::init(unsigned int numVars,
...
-		if (pstrain) numLowVars += WaveFieldWriterExecutor::NUM_PLASTICITY_VARIABLES;
-
-		if (integrals) numLowVars += m_numIntegratedVariables;
-
-		for (int i = 1; i < numLowVars; i++)
-			addBuffer(0L, pLowMeshRefiner->getNumCells() * sizeof(real));
-
+		bool first = false;
+		for (unsigned int i = 0; i < WaveFieldWriterExecutor::NUM_LOWVARIABLES; i++) {
+			if (m_lowOutputFlags[i]) {
```
uphoffc, line 324: Will behave randomly when some lowOutputFlags are never set, see other comment.

uphoffc (changes requested):

uphoffc (commented):
Does this commit make any sense? Why do you want subsampling for an output that was explicitly created to bypass subsampling?

Thomas-Ulrich:
Hi, 
Just to give some context, I would the plastic strain output looks underresolved, while it is actually probably smooth. See e.g. this figure, showing the accumulated plastic strain in the latest Sumatra study: 
image
So yes, the file is named 'low', but we can change this name to plasticity if that's the problem ;-)

src/ResultWriter/WaveFieldWriter.cpp
Thomas-Ulrich: right

src/ResultWriter/inioutput_seissol.f90
Thomas-Ulrich: yes you are right

src/ResultWriter/WaveFieldWriter.cpp
Thomas-Ulrich: no, because it points the WaveFieldWriterExecutor::NUM_PLASTICITY_VARIABLES's element of the array, so will update only the `integral outputs`

src/ResultWriter/WaveFieldWriter.cpp
Thomas-Ulrich: I think all lowOutputFlags are set correctly.

src/ResultWriter/WaveFieldWriter.cpp
Thomas-Ulrich: I think they are initialized when calling seissol::SeisSol::main.postProcessor().getIntegrationMask

src/Numerical_aux/typesdef.f90
Thomas-Ulrich: good point.

src/ResultWriter/WaveFieldWriter.cpp
uphoffc: ok

src/ResultWriter/WaveFieldWriter.cpp
uphoffc: ok

src/ResultWriter/WaveFieldWriter.cpp
uphoffc: ok

src/ResultWriter/WaveFieldWriter.cpp
```diff
@@ -295,18 +295,22 @@ void seissol::writer::WaveFieldWriter::init(unsigned int numVars,
 	if (pstrain || integrals) {
 		logInfo(rank) << "Initialize low order output";
 
-		// Refinement strategy (no refinement)
-		refinement::IdentityRefiner<double> lowTetRefiner;
+		// Setup the tetrahedron refinement strategy
+		refinement::TetrahedronRefiner<double>* lowTetRefiner = createRefiner(refinement);
```
uphoffc, line 299: This refinement needs to be different from the regular wave field output. Otherwise the split in "low output" and "regular output" is pointless.

src/ResultWriter/WaveFieldWriter.h
```diff
@@ -91,6 +91,9 @@ class WaveFieldWriter : private async::Module<WaveFieldWriterExecutor, WaveField
 	/** The variable subsampler for the refined mesh */
 	refinement::VariableSubsampler<double>* m_variableSubsampler;
 
+	/** The variable subsampler for the plasticity output */
+	refinement::VariableSubsampler<double>* m_lowVariableSubsampler;
```
uphoffc, line 95: please use a std::unique_ptr here to ensure that the memory is freed. (Currently its never deleted.)

src/ResultWriter/WaveFieldWriter.cpp
```diff
@@ -427,13 +431,7 @@ void seissol::writer::WaveFieldWriter::write(double time)
...
 			real* managedBuffer = async::Module<WaveFieldWriterExecutor,
 					WaveFieldInitParam, WaveFieldParam>::managedBuffer<real*>(m_variableBufferIds[1] + nextId);
-
-#ifdef _OPENMP
-			#pragma omp parallel for schedule(static)
-#endif // _OPENMP
-			for (unsigned int j = 0; j < m_numLowCells; j++)
-				managedBuffer[j] = m_pstrain[m_map[j]
-						* WaveFieldWriterExecutor::NUM_PLASTICITY_VARIABLES + i];
+			m_lowVariableSubsampler->get(m_pstrain, m_map, i, managedBuffer);
```
uphoffc, line 434: That doesn't work because pstrain is a low order field and no high order information is available.

uphoffc (changes requested):
The problem is that the information you want to output is not computed in the first place. See
https://github.com/SeisSol/SeisSol/blob/0d2825fa3955cf1ffad0df55a65d2d868ed18776/src/Kernels/Plasticity.cpp#L171

src/ResultWriter/inioutput_seissol.f90
Thomas-Ulrich: I tried that, but now get:
```
/hppfs/work/pr45fi/di73yeq4/SeisSol/src/ResultWriter/inioutput_seissol.f90(171): error #6633: The type of the actual argument differs from the type of the dummy argument.   [PLASTICITYMASK]
        i_plasticityMask=io%PlasticityMask,     &
----------------------------^
```
(see last commit)

src/ResultWriter/WaveFieldWriter.cpp
Thomas-Ulrich: yes I know, see first post of this issue:
>(at this point, the refinement level is the one from the wavefield output, but we could imaging a refinement_plasticity parameter in the parameter file).

With this draft, I just wanted to get some help making it work, and then I can refine the code with a further refinement parameter.

src/ResultWriter/WaveFieldWriter.cpp
Thomas-Ulrich: ha ok so that my problem, thx!

Thomas-Ulrich: would it be difficult to get a high-order field?

uphoffc: Not for the stress tensor but for eta you need
eta += timeStepWidth * sqrt(0.5 dstrain_{ij}/dt dstrain_{ij}/dt)
which you'll need to do differently for high order

uphoffc: With quadrature, e.g.

src/ResultWriter/WaveFieldWriter.h
Thomas-Ulrich: I don't know how to do that.

uphoffc: You have to #include <memory>, then you replace
```c++
refinement::VariableSubsampler<double>* m_lowVariableSubsampler;
```
with
```c++
std::unique_ptr<refinement::VariableSubsampler<double>> m_lowVariableSubsampler;
```

and you replace
```c++
m_variableSubsamplerPStrain = new refinement::VariableSubsampler<double>(
			numElems, *tetRefiner, order, WaveFieldWriterExecutor::NUM_PLASTICITY_VARIABLES , numAlignedDOF);
```
with
```c++
m_variableSubsamplerPStrain = std::make_unique<refinement::VariableSubsampler<double>>(
			numElems, *tetRefiner, order, WaveFieldWriterExecutor::NUM_PLASTICITY_VARIABLES , numAlignedDOF);
```

Thomas-Ulrich: thx

Thomas-Ulrich:
ok, it is not working yet, and I m getting short of ideas why...
```
Thu Jul 01 13:47:20, Info:  Waiting for last wave field.
Thu Jul 01 13:47:20, Info:  Writing wave field at time 0.
SeisSol_Debug_sskx_4_elastic: /hppfs/work/pr45fi/di73yeq4/SeisSol/submodules/async/async/as/Thread.h:167: void async::as::Thread<Executor, InitParameter, Parameter>::sendBuffer(unsigned int, unsigned long) [with Executor = seissol::writer::FreeSurfaceWriterExecutor, InitParameter = seissol::writer::FreeSurfaceInitParam, Parameter = seissol::writer::FreeSurfaceParam]: Assertion `(Base<Executor, InitParameter, Parameter>::_buffer(id))' failed.
```
 if anybody wants to give some feedback, that would be great!

Thomas-Ulrich:
So It seems to be working as expected on tp12_13:
new (refinement=3, 32 times more cells):
image
old:
image

range of the data:
`minmax_SeisSol_data.py` computes  
`print(idt, np.amin(myData[:]), np.percentile(myData[:], 95), np.amax(myData[:]))`

new:
```
di73yeq4@login02:/hppfs/work/pr45fi/di73yeq4/tpv12_13> python ../TuSeisSolScripts/onHdf5/minmax_SeisSol_data.py output/tpv13e.xdmf  --Data eta
0 0.0 0.0 0.0
1 -0.018855873 0.0001909208098368249 0.126005
2 -0.029730557 0.0002469609316904098 0.18287298
di73yeq4@login02:/hppfs/work/pr45fi/di73yeq4/tpv12_13> python ../TuSeisSolScripts/onHdf5/minmax_SeisSol_data.py output/tpv13e.xdmf  --Data ep_yz
0 0.0 0.0 0.0
1 -0.023647618 3.239040393054716e-06 0.048905037
2 -0.034393962 1.72803063833271e-05 0.07112089
```
old:
```
di73yeq4@login02:/hppfs/work/pr45fi/di73yeq4/tpv12_13> python ../TuSeisSolScripts/onHdf5/minmax_SeisSol_data.py output/tpv13master-low.xdmf  --Data eta
0 0.0 0.0 0.0
1 0.0 0.0002432470500934869 0.018974442
2 0.0 0.00031244225101545453 0.023759468
di73yeq4@login02:/hppfs/work/pr45fi/di73yeq4/tpv12_13> python ../TuSeisSolScripts/onHdf5/minmax_SeisSol_data.py output/tpv13master-low.xdmf  --Data ep_yz
0 0.0 0.0 0.0
1 -0.0044504143 4.631079718819819e-06 0.010111902
2 -0.008216644 2.0416773622855544e-05 0.015443558
```

Peak values are much larger, as we do not plot cell average but now plot subcell data.
Eta can have negative values, because of the nodal to modal conversion.
(any idea to prevent that, in addition to what I've implemented?)

Note also that I've added the INTEGRATE_QUANTITY to cmake and documented this feature, and its broken state.
-> it's not compiling (the code is outdated).

Thomas-Ulrich:
Note that it seems `make clean` needs to be run before compiling. Else it compiles but crashes during initialization.

git-taufiq:
> Note that it seems `make clean` needs to be run before compiling. Else it compiles but crashes during initialization.

Hi Thomas, it seems I encounter this problem even though not always, ie during 'test' job it did not crash but unfortunately during 'general' job it crashed after long queue. It stopped at initializing refinement/class/subsampler: 

```
Rank:        0 | Info    | <--------------------------------------------------------->
Rank:        0 | Info    | <     Start inioutput_SeisSol ...                         >
Rank:        0 | Info    | <--------------------------------------------------------->
Tue Jul 27 08:03:35, Info:  Initializing free surface integrator.
Tue Jul 27 08:03:35, Info:  Initializing free surface integrator. Done.
Tue Jul 27 08:03:35, Info:  Wave field writer thread affinity: "----------|----------|----------|----------|-------7--|----------|----------|----------|----------|-----5"
Tue Jul 27 08:03:35, Info:  Initializing XDMF wave field output.
Tue Jul 27 08:03:35, Info:  Refinement Strategy is "Divide by 8"
Tue Jul 27 08:03:35, Info:  Refinement class initialized
Tue Jul 27 08:03:35, Info:  VariableSubsampler initialized
```
and error log only shows segmentation fault errors messages.
I did use `make clean` and generate a new build folder before compilation. Maybe I miss something @Thomas-Ulrich ?

Thomas-Ulrich:
Hi,

Sorry about that, I also got the error at some point, and I thought it is related with remains from a previous compilation and could be fixed with 'make clean', but maybe there is a latent issue, which won't be easy to debug....

Thomas.

Thomas-Ulrich:
Hi, 
I've been testing with the sumatra setup (small mesh, 3mio, o4 double), and found a configuration that allow reproducing the issue.
16 nodes, 4 nodes, 1 nodes ( --ntasks-per-node=2 for each). and refinement=1. Works fine.
2 nodes, 1 nodes ( --ntasks-per-node=2 for each). and refinement=2. Works fine.
4, 2 nodes (test), and refinement=3. Works fine
but crashes with 1 node  (both test and fat queue): 
test: crashes after `VariableSubsampler initialized`
fat: crashes after `Initializing XDMF wave field output. Done.`
So it might not be memory related?
But fat nodes have 740Gb of RAM vs test only 90Gb, so maybe not?
setup is here: /hppfs/work/pr45fi/di73yeq4/testSuma

git-taufiq:
I tested 16 nodes and --ntasks-per-node=1. and vary refinement 2 and 3. Run the same test setup can give me both crash and work. I would like to see setup /hppfs/work/pr45fi/di73yeq4/testSuma but do not have access to that project. Could you copy it to pr63qo?

Thomas-Ulrich:
A way to prevent the crashing of the high order output is to deactivate the writing of the info about the Lts clustering, with this change. I don't know why my changes interfere with this output, but it seems it is the reason for the crashes. 

```
diff --git a/src/ResultWriter/WaveFieldWriterExecutor.h b/src/ResultWriter/WaveFieldWriterExecutor.h
index 37d7ad81..4277c979 100644
--- a/src/ResultWriter/WaveFieldWriterExecutor.h
+++ b/src/ResultWriter/WaveFieldWriterExecutor.h
@@ -198,7 +198,7 @@ public:
                m_waveFieldWriter->setComm(m_comm);
 #endif // USE_MPI
 
-               m_waveFieldWriter->init(variables, std::vector<const char*>(), true, true, true);
+               m_waveFieldWriter->init(variables, std::vector<const char*>(), true, true, false);
                m_waveFieldWriter->setMesh(
                        info.bufferSize(param.bufferIds[CELLS]) / (4*sizeof(unsigned int)),
                        static_cast<const unsigned int*>(info.buffer(param.bufferIds[CELLS])),
                   
```

git-taufiq:
Unfortunately, both my tests with 16 nodes and 32 nodes crashed after 'VariableSubsampler initialized'. They worked well if only run with deactivating the 'OutputRegionBounds' feature in the parameter file.

PR #347, opened 2021-06-09
Added installation of SeisSol GPU on M100 to docu

ravil-mobile:


sebwolf-de:
Thanks for the documentation, unfortunately I can't test anything, since I do not have access to Marconi :D
Can we split the SuperMUC-NG and Marconi documentation into two pages?

Documentation/supermuc.rst
```diff
@@ -177,3 +181,267 @@ set compiler options, run cmake, and compile with:
...
+
+  $ make -j 
+  $ make install
+  $ cd ..
+
+8. Installing ParMetis:
+
+:: 
+
+  $ wget http://glaros.dtc.umn.edu/gkhome/fetch/sw/parmetis/parmetis-4.0.3.tar.gz
```
Thomas-Ulrich, line 347: it seems this block is not reckognized as a code block.

Documentation/supermuc.rst
```diff
@@ -177,3 +181,267 @@ set compiler options, run cmake, and compile with:
...
+3. Create any directory where you are going to configure and build libraries and packages. For example, 
+
+::
+
+  mkdir -p $HOME/Download
+  cd $HOME/Download
+
+4. Install UCX:
+
+::
```
Thomas-Ulrich, line 239: it seems this block is not recognized as a code block. (but maybe its a problem with GitHub code highlighting)

Documentation/supermuc.rst
ravil-mobile: Fixed. It was caused by trailing white spaces

Documentation/supermuc.rst
ravil-mobile: Fixed.

ravil-mobile:
Thanks @Thomas-Ulrich  and @sebwolf-de for checking this PR.

> Can we split the SuperMUC-NG and Marconi documentation into two pages?

I guess you would like to have 2 separate text files but I didn't figure out how to it. I tried `.. include::` directive but it didn't help.
Nevertheless, It appears that SuperMUC-NG and Marconi are split.  

image

sebwolf-de:
Sorry, for being unclear. I meant two distinct entries in the index. May I push to your branch?

ravil-mobile:
yes, sure

sebwolf-de:
What do you think?

uphoffc:
Hey, I wouldn't advise to put git links there.
E.g. instead of
```
$ git clone https://github.com/openucx/ucx.git
$ cd ucx
$ git checkout v1.10.x
```
do
```
wget https://github.com/openucx/ucx/archive/refs/tags/v1.10.1.tar.gz
tar -xvf v1.10.1.tar.gz
cd ucx-1.10.1
```
v1.10.x is a development version whereas v1.10.1 is a released version which makes the compilation reproducible

ravil-mobile:
Thanks @uphoffc, I'll change links from github to tar archives. Did you manage to compile UCX `v1.10.1` or it is also does not work? If not, which compiler and which OS did you use?

uphoffc:
Yes I could compile 1.10.1 on longhorn using gcc/7

Documentation/marconi.rst
```diff
@@ -0,0 +1,276 @@
...
+
+::
+
+  $ wget https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.10/hdf5-1.10.5/src/hdf5-1.10.5.tar.gz
+  $ tar -xvf ./hdf5-1.10.5.tar.gz
+  $ cd hdf5-1.10.5
+  $ ./autogen.sh
+  $ mkdir build && cd build
+
+  $ CPPFLAGS="-fPIC ${CPPFLAGS}" CC=mpicc CXX=mpicxx FC=mpif90 \
```
uphoffc, line 127: Shouldn't that be CFLAGS instead of CPPFLAGS?

ravil-mobile:
@sebwolf-de 

It looks a bit strange right now. 

image

I have another idea. We can put all `Compiling and Running SeisSol on Supercomputers` to separate toctree. I know that there will be issues with other supercomputers which will be worth documenting. For example, Cray systems i.e. one needs to remember to set `export CRAY_LINK_TYPE=dynamic` before compiling SeisSol GPU but only in case if login nodes do not have a GPU driver installed on them (I know it sounds strange but it is the current Piz Diant setup)

Documentation/marconi.rst
ravil-mobile: It seems to me that I copied and pasted it from `Installing HDF5`. I guess yes. In my case it didn't break my installation on M100.

sebwolf-de:
Yes, good idea

Documentation/marconi.rst
```diff
@@ -0,0 +1,277 @@
...
+  #!/bin/bash
+  #SBATCH --account=<you account>
+  #SBATCH --partition=m100_usr_prod
+  #SBATCH --qos=m100_qos_dbg
+  #SBATCH --time <time>
+  #SBATCH --nodes <number of nodes>
+  #SBATCH --ntasks-per-node=4
+  #SBATCH --cpus-per-task=32
+  #SBATCH --gres=gpu:4
+  #STABCH --gpu-bind=closest
```
uphoffc, line 231: typo STABCH

Documentation/marconi.rst
```diff
@@ -0,0 +1,277 @@
...
+  NUM_COMPUTE_CORES=$(expr $NUM_CORES - 1)
+
+  export OMP_NUM_THREADS=$NUM_COMPUTE_CORES
+  export OMP_PLACES="cores($NUM_COMPUTE_CORES)"
+  export PROC_BIND=spread
+
+  export DEVICE_STACK_MEM_SIZE=1.5
+  export UCX_MEMTYPE_CACHE=n
+  
+  mpirun --report-bindings --map-by ppr:$SLURM_NTASKS_PER_NODE:node:pe=$NUM_CORES \
```
uphoffc, line 250: isn't -np missing?

Documentation/marconi.rst
ravil-mobile: thanks. fixed

ravil-mobile:
I think it is better to `squash and merge` when this PR is ready. 

@sebwolf-de, I added a dedicate toctree for supercomputers. 

image

I guess it looks better right now

ravil-mobile:
Hi all, Can we merge this PR?

Documentation/compilation.rst
```diff
@@ -108,9 +108,6 @@ Installing GemmForge (for GPU)
 
 .. code-block:: bash
 
-   pip install gemmforge
```
Thomas-Ulrich, line 111: not blue ^^. Note that there is also a comment by Carsten that you did not adress.

ravil-mobile: Thanks, I'll fix this. 

> Note that there is also a comment by Carsten that you did not address

I've just realized that I wrote a reply but I forgot to press the button, and it turned out that I close my browser. My bad

Documentation/marconi.rst
ravil-mobile: Hi, 

Thanks for noticing. No it is not missing __in this particular case__ because OpenMPI on M100 was compiled with PMIX3.

```
--with-pmix=/opt/pmix/3.1.5
```

 As far as I understand PMIx is an API (or a protocol) which sits between a resource manager (e.g., slurm, torque) and an MPI implementation. In short, a resource manager sends all necessary info to PMIx and then it is delivered to MPI. This information can include hostnames of each involved compute node, maybe a mapping between PIDs and ranks, etc. 

I know that it is super specific and thus there is a line, right above the batch file, saying

```
In this particular case it is not necessary to provide a number of processes after **mpirun** because OpenMPI was compiled with PMIX (see step 5).
```

Documentation/compilation.rst
ravil-mobile: Hi, @Thomas-Ulrich. 

```
pip install gemmforge
```

The line is removed from the file.

Documentation/marconi.rst
```diff
@@ -0,0 +1,277 @@
...
+  export CMAKE_PREFIX_PATH=$SEISSOL_INSTALL
+  export CPATH=$SEISSOL_INSTALL/include:$CPATH
+  export CPLUS_INCLUDE_PATH=$SEISSOL_INSTALL/include:$CPLUS_INCLUDE_PATH
+  export C_INCLUDE_PATH=$SEISSOL_INSTALL/include:$C_INCLUDE_PATH
+
+2. Load basic modules, source launch file and create an install directory:
+
+::
+
+  $ source ./launch.sh
```
Thomas-Ulrich, line 40: Don't you think it would be better to remove the '$' signs?
When you follow such an installation procedure, you just want to copy the code block, not remove every '$', right?

ravil-mobile: Hi, actually  '$' signs help. Firstly, it helps when you want to show a content of a file using `cat` command. It makes clear what is a command and what is its output. For example:

```
$ cat $HOME/launch.sh  
#!/bin/bash  

VERSION="openmpi4-1-1_ucx1-10-1"  
export SEISSOL_INSTALL=$HOME/usr/local/${VERSION}
...
```

Secondly, it can prevent one to blindly copy a command and execute it. You still have a couple of seconds to have a look at and figure out whether you need to adjust it for your environment.

Thirdly, other project adopted this style e.g., LLVM, Docker, Spack, etc.

Thomas-Ulrich (approved):

PR #348, opened 2021-06-14
Added ampere arch + impalajit as an external lib. (optionally)

ravil-mobile:
Added extra switches to CMake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,11 @@ else()
   target_include_directories(SeisSol-lib PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/yaml-cpp-install/include/)
 endif()
 
-if (IMPALAJIT_FROM_SUBMODULES)
+find_package(ImpalaJIT 1.0 QUIET)
+if (ImpalaJIT_FOUND)
+  message(STATUS "Found ImpalaJIT: ${ImpalaJIT_DIR} (found version ${ImpalaJIT_VERSION})")
+  target_link_libraries(SeisSol-lib PRIVATE impalajit)
+else()
   ExternalProject_Add(ImpalaJIT-build
     SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/submodules/ImpalaJIT/
     INSTALL_DIR ${CMAKE_CURRENT_BINARY_DIR}/ImpalaJIT-install
@@ -171,10 +175,6 @@ if (IMPALAJIT_FROM_SUBMODULES)
   add_dependencies(SeisSol-lib ImpalaJIT-build)
   target_link_libraries(SeisSol-lib PUBLIC "${CMAKE_CURRENT_BINARY_DIR}/ImpalaJIT-install/lib/libimpalajit.a")
   target_include_directories(SeisSol-lib PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/ImpalaJIT-install/include/)
-else()
-  find_package(ImpalaJIT 1.0 REQUIRED)
-  message(STATUS "Found ImpalaJIT: ${ImpalaJIT_DIR} (found version ${ImpalaJIT_VERSION})")
-  target_link_libraries(SeisSol-lib PRIVATE impalajit)
 endif()
 
 # todo: if parallelization == omp/hybrid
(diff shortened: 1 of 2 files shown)


sebwolf-de:
Hi, I don't really understand the logic behind `IMPALAJIT_FROM_SUBMODULES`. 
I'd come up with the following logic (same as for yaml-cpp): 
* Try to find impalaJIT throught `find_package`
*  If not available (or wrong version) build from submodules.

krenzland:
Yeah. I'd argue for the same. It would be easier to find a correct version if impalajit would support version numbers. Then you make sure that the version installed on the system is not outdated.
Burn considering that no one is adding features to impalajit anymore...

ravil-mobile:
Hi, 

Thanks @sebwolf-de. I don't know why it didn't get into my mind and a better solution was a couple lines above. 

> if impalajit would support version numbers

It turns out that the original ImpalaJIT has a version. If you look at any of your SeisSol build directory you can find the following

```
$ cat ImpalaJIT-install/lib/pkgconfig/impalajit.pc
Name: ImpalaJIT
Description: JIT Compilation to realize flexible data access
URL: https://github.com/Manuel1605/ImpalaJIT
Version: 0.9

libdir=${prefix}/lib
includedir=${prefix}/include

Libs: -L${libdir} -limpalajit
Cflags: -I${includedir}
```

Version 1.0 corresponds to my fork of ImpalaJIT which I need to run DR on M100 and ARM+Nvidia nodes

krenzland:
Could you maybe add a prefix/suffix to your version? E.g. llvm-1.0?
Otherwise this is going to make updating impalajit impossible

ravil-mobile:
I can add cmake namespace or a component. Let me try. Btw, do we need to add it to docu or it is not necessary?

ravil-mobile:
Now, it looks like the following

```
target_link_libraries(SeisSol-lib PRIVATE LLVM::impalajit)
```

Cmake macro `write_basic_package_version_file` requires only digits for a version. Here is an example, taken from a generated file:

```
et(PACKAGE_VERSION "1.0.0")

if(PACKAGE_VERSION VERSION_LESS PACKAGE_FIND_VERSION)
  set(PACKAGE_VERSION_COMPATIBLE FALSE)
else()

  if("1.0.0" MATCHES "^([0-9]+)\\.")
    set(CVF_VERSION_MAJOR "${CMAKE_MATCH_1}")
  else()
    set(CVF_VERSION_MAJOR "1.0.0")
  endif()

...
```

krenzland:
I don't think that your approach solves the problem. It tries to find ANY ImpalaJIT with version 1.0 but then only links to LLVM::ImpalaJIT. If someone would publish ImpalaJIT with version 1.0 that does not use LLVM it wouldn't work.
Maybe you can rename your fork to ImpalaJIT-LLVM or similar? Then you can search for it.

ravil-mobile:
Ok. I will rename the fork. Actually I only need to rename `ImpalaJITConfig.cmake` to `ImpalaJIT-LLVMConfig.cmake` (and corresponding files like version, etc). I wanted to do it at the beginning but I didn't know whether all of you would be ok with it

krenzland:
Probably the best way  (in addition to renaming it to ImpalaJIT-LLVM or so) would be to add another compile time option to cmake, e.g. USE_IMPALA_JIT_LLVM or so. This would make it explicit and could avoid a lot of confusion. What do you think?

CMakeLists.txt
```diff
@@ -157,20 +157,24 @@ else()
...
-)
-add_dependencies(SeisSol-lib ImpalaJIT-build)
-target_link_libraries(SeisSol-lib PUBLIC "${CMAKE_CURRENT_BINARY_DIR}/ImpalaJIT-install/lib/libimpalajit.a")
-target_include_directories(SeisSol-lib PUBLIC ${CMAKE_CURRENT_BINARY_DIR}/ImpalaJIT-install/include/)
+if (USE_IMPALA_JIT_LLVM)
+  find_package(ImpalaJIT-LLVM 1.0 REQUIRED)
+  target_link_libraries(SeisSol-lib PRIVATE llvm::impalajit)
+else()
+  ExternalProject_Add(ImpalaJIT-build
+    SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/submodules/ImpalaJIT/
```
ravil-mobile, line 165: Is it OK now?

krenzland (approved):

PR #351, opened 2021-06-18
8d0 -> 8

uphoffc:
I've looked a bit in issue #256

The issue seems to be the heavy use of math functions which are apparently better vectorized with the Intel compiler. In fact, the Intel compiler has special vectorized math routines in the library SVML that is shipped with MKL I believe.

So if SVML is installed then adding the following arguments to gfortran could help
-mveclibabi=svml -ffast-math -funsafe-math-optimizations

Just adding
-ffast-math -funsafe-math-optimizations
also helps a little bit, and the change in this PR also helps a little bit.

### DIFF
--- a/src/Physics/Evaluate_friction_law.f90
+++ b/src/Physics/Evaluate_friction_law.f90
@@ -1120,7 +1120,7 @@ MODULE Eval_friction_law_mod
         ! low-velocity steady state friction coefficient
         flv = RS_f0 - (RS_b-RS_a)* LOG(SR_tmp/RS_sr0)
         ! steady state friction coefficient
-        fss = RS_fw + (flv - RS_fw)/(1.0D0+(SR_tmp/RS_srW)**8d0)**(1.0D0/8.0D0)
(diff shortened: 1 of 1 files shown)


ravil-mobile:
Hi, 

I wanted to suggest something for #256 today but I was not sure. #256 was related to a specific version of GCC compiler. I thought that maybe O2 or O1 optimization is better for the fortran part of SeisSol. Maybe O3 is too aggressive. What do you think?

krenzland:
Impressive! Do we actually want to to have -fast-math and -unsave-math-optimizations for the entire code, or would it be smarter to restrict it to this one function? To be fair, intel compiler has this as default, so it wouldn't make much sense to err on the side of caution here...

uphoffc:
I don't think that O1 or O2 will improve things, but please feel free to test. If fast-math is the default for icc then I don't see a reason to not enable it for gcc.

Anyhow, can we merge this PR?

krenzland (approved):

PR #353, opened 2021-06-20
Create rise_time_from_xdmf.py

NicoSchlw:
script to determine rise time from seissol fault.xdmf output

### DIFF
--- a/postprocessing/science/rise_time_from_xdmf.py
+++ b/postprocessing/science/rise_time_from_xdmf.py
@@ -1,0 +1,69 @@
+# Script to determine the rise time (time the absolute slip rate is above a certain threshold) from seissol fault.xdmf output
+# Output of this script is a binary file. If events = 1, the binary file can be exchanged with the rupture time fault output
+# and opened in paraview without creating a new xdmf file.
+
+import numpy as np
+import seissolxdmf
+import argparse
+import sys
+
+parser = argparse.ArgumentParser(description='calculate rise times from fault xdmf file')
+parser.add_argument('filename', help='path+filename.xdmf')
+parser.add_argument('--threshold', nargs=1, metavar=('threshold'), default=([0.01]), help='specify slip rate threshold' ,type=float)
+parser.add_argument('--dt', nargs=1, metavar=('dt'), default=([0.1]), help='insert time step of xdmf file' ,type=float)
+parser.add_argument('--events', nargs=1, metavar=('events'), default=([1]), help='number of events [1,2] in xdmf file' ,type=int)
+args = parser.parse_args()
+
+if args.events[0] != 1 and args.events[0] != 2:
+    sys.exit("Script only compatible with events = [1,2]")
+
+
+def get_rise_time(ASR_array, threshold=0.01, dt=0.1, events=1):
+    
+    RT = np.zeros_like(ASR_array)
+    
+    if events == 1:
+        for i in range(ASR_array.shape[1]):
+            count=0
+            for j in range(ASR_array.shape[0]):
+                if ASR_array[j,i] < threshold and count > 0: break
+                if ASR_array[j,i] <  threshold and count == 0: continue
+                if ASR_array[j,i] >= threshold: count += 1        
+            RT[:,i] = (count * dt)
+            
+    if events == 2:
+        mid = int(ASR_array.shape[0]/2)
+        for i in range(ASR_array.shape[1]):
+            count = 0
+            for j in range(0, mid):
+                if ASR_array[j,i] < threshold and count > 0: break
+                if ASR_array[j,i] < threshold and count == 0: continue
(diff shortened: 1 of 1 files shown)


postprocessing/science/rise_time_from_xdmf.py
```diff
@@ -0,0 +1,69 @@
...
+import sys
+
+parser = argparse.ArgumentParser(description='calculate rise times from fault xdmf file')
+parser.add_argument('filename', help='path+filename.xdmf')
+parser.add_argument('--threshold', nargs=1, metavar=('threshold'), default=([0.01]), help='specify slip rate threshold' ,type=float)
+parser.add_argument('--dt', nargs=1, metavar=('dt'), default=([0.1]), help='insert time step of xdmf file' ,type=float)
+parser.add_argument('--events', nargs=1, metavar=('events'), default=([1]), help='number of events [1,2] in xdmf file' ,type=int)
+args = parser.parse_args()
+
+if args.events[0] != 1 and args.events[0] != 2:
```
Thomas-Ulrich, line 17: better use `choices` in the argparse?
see e.g. https://github.com/SeisSol/SeisSol/blob/master/postprocessing/visualization/tools/extractDataFromUnstructuredOutput.py#L15

postprocessing/science/rise_time_from_xdmf.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def get_rise_time(ASR_array, threshold=0.01, dt=0.1, events=1):
+    
+    RT = np.zeros_like(ASR_array)
+    
+    if events == 1:
+        for i in range(ASR_array.shape[1]):
+            count=0
+            for j in range(ASR_array.shape[0]):
```
Thomas-Ulrich, line 28: I have the impression that this code might be quite slow because j is the first dimension in the array, so the ASR_array[:,i] is not contiguous. Because of that, I'm transposing after reading the array in this code:
https://github.com/SeisSol/SeisSol/blob/master/postprocessing/science/GroundMotionParametersMaps/ComputeGroundMotionParametersFromSurfaceOutput_Hybrid.py#L185

postprocessing/science/rise_time_from_xdmf.py
```diff
@@ -0,0 +1,69 @@
...
+if args.events[0] != 1 and args.events[0] != 2:
+    sys.exit("Script only compatible with events = [1,2]")
+
+
+def get_rise_time(ASR_array, threshold=0.01, dt=0.1, events=1):
+    
+    RT = np.zeros_like(ASR_array)
+    
+    if events == 1:
+        for i in range(ASR_array.shape[1]):
```
Thomas-Ulrich, line 26: I would move this double loop in a function (repeats 3 times in the function)

postprocessing/science/rise_time_from_xdmf.py
```diff
@@ -0,0 +1,69 @@
...
+                if ASR_array[j,i] >= threshold: count += 1       
+            RT[mid:,i] = (count * dt)
+            
+    return RT
+
+print("Loading file...")          
+sx = seissolxdmf.seissolxdmf(args.filename)
+
+print("Calculating absolute slip rate...")          
+SRs = sx.ReadData('SRs')
```
Thomas-Ulrich, line 57: consider creating ASR in a function so that the SRs and SRd object could be deleted when out of scope (they may occupy a lot of RAM).

postprocessing/science/rise_time_from_xdmf.py
```diff
@@ -0,0 +1,69 @@
...
+SRd = sx.ReadData('SRd')
+ASR = np.sqrt(SRs**2 + SRd**2)
+          
+print("Determining rise time...")
+RT = get_rise_time(ASR, threshold=args.threshold[0], dt=args.dt[0], events=args.events[0])
+
+print("Writing output file...")
```
Thomas-Ulrich, line 64: alternatively, you could write a new hdf5/xdmf file with just one or two time stamps using this code as a model:
https://github.com/SeisSol/SeisSol/blob/master/postprocessing/visualization/tools/extractDataFromUnstructuredOutput.py#L109

Thomas-Ulrich (changes requested):

Thomas-Ulrich:
Thx for this PR. 
Overall, I think it makes sense to process this output in a postprocessing script (rather than in the Fortran code) because the output may be strongly dependent on the threshold.

NicoSchlw:
Thanks for the feedback, I will rework the script soon!

postprocessing/science/rise_time_from_xdmf.py
Thomas-Ulrich: actually, this simpler code might give a simpler illustration: https://github.com/SeisSol/Meshing/blob/master/vizualizeBoundaryConditions/vizualizeBoundaryConditions.py

PR #356, opened 2021-06-23
Small correction in the documentation

hugosanrocks:
Small correction in the documentation regarding the orientation of the normal vector to the fault given in the example of reference. Some lines were added and the orientation of the normal vector was corrected.

### DIFF
--- a/Documentation/left-lateral-right-lateral-normal-reverse.rst
+++ b/Documentation/left-lateral-right-lateral-normal-reverse.rst
@@ -73,11 +73,19 @@ or logically
 Example
 ~~~~~~~
 
-We have a 60° dipping normal fault with 90° strike and zero rake with
-*enu* convention. (That is, the strike is east.) The normal of the fault
+We have a 60° dipping normal fault with 90° strike (West-East direction, pointing
+towards the East with *enu* convention) and 0° rake. The normal of the fault
 plane, which points from the foot wall to the hanging wall, is given by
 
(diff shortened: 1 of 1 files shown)


Documentation/left-lateral-right-lateral-normal-reverse.rst
```diff
@@ -73,11 +73,19 @@ or logically
...
-:math:`N:=\frac{1}{2}\begin{pmatrix}1 & 0 & \sqrt{3}\end{pmatrix}`
+:math:`n:=\frac{1}{2}\begin{pmatrix}0 & -\sqrt{3} & 1\end{pmatrix}`
+
+Such normal vector is obtained as the result of the cross product between the dip vector
+
+:math:`d:=\frac{1}{2}\begin{pmatrix}0 & -1 & -\sqrt{3}\end{pmatrix}`
+
+and the respective strike vector
+
+:math:`s:=\frac{1}{2}\begin{pmatrix}1 & 0 & 0\end{pmatrix}`
```
uphoffc, line 88: Without \frac{1}{2} here, right?

uphoffc (changes requested):

uphoffc (approved):

Documentation/left-lateral-right-lateral-normal-reverse.rst
hugosanrocks: Yes, totally right!

PR #357, opened 2021-06-28
New Module names for SuperMUC-NG

sebwolf-de:
After the SuperMUC-NG maintenance, some modules have new names. This PR reflects these changes.

### DIFF
--- a/Documentation/supermuc.rst
+++ b/Documentation/supermuc.rst
@@ -71,11 +71,10 @@ Building SeisSol
 ::
 
   ##### module load for SeisSol
-  module load scons gcc/9 cmake/3.14.4 python/3.6_intel
+  module load gcc/9 cmake python/3.6_intel
   module load libszip/2.1.1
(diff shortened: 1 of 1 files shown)


uphoffc:
You don't need to load the gcc module anymore as the Intel module will set C_INCLUDE_PATH and CPLUS_INCLUDE_PATH automatically. In fact, if you load the gcc and the intel module **simultaneously** the latter behaviour breaks the gcc compiler...

sebwolf-de:
If `gcc/9` is not loaded, we can't use `filesystem`:
```
catastrophic error: cannot open source file "filesystem"
  #include <filesystem>
```

Thomas-Ulrich:
seems like ```#SBATCH --workdir=<work directory>```
has to be changed to  ```#SBATCH --chdir=<work directory>```

Thomas-Ulrich (approved):

PR #359, opened 2021-07-07
Add option to enable address sanitzer in debug mode

krenzland:
Can use option -DADDRESS_SANITIZER_DEBUG to link with address sanitzer when in debug mode.
https://en.wikipedia.org/wiki/AddressSanitizer can help a lot with finding memory leaks and related issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,6 +263,12 @@ if (WITH_GPU)
   target_compile_definitions(SeisSol-lib PUBLIC ACL_DEVICE)
 endif()
 
+if (ADDRESS_SANITIZER_DEBUG)
+  target_link_libraries(SeisSol-lib PUBLIC debug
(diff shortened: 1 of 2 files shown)


ravil-mobile:
Do we have lot's memory leaks (Im just asking)? Btw, what is the difference between this tool and valgrind?

sebwolf-de:
Valgrind is an external tool, whereas ASan is directly compiled into your binary.
I've never tried valgrind, since it always took ages to initialize. ASan is much faster to use in that sense.
Also I don't know about valgrind in parallel environments...

sebwolf-de (approved):
LGTM

ravil-mobile:
@sebwolf-de, thanks for explaining.

> Also I don't know about valgrind in parallel environments...

krenzland:
Address sanitizer can also be a bit annoying with MPI.

PR #360, opened 2021-07-12
add script to compare seissol output (including reindexing)

Thomas-Ulrich:
This script allows computing the diff f2-f1 or the relative ratio (f2-f1)/f1 of 2 seissol output (volume or surface).
The steps to get to the results are
-remove duplicated nodes using pymesh function.
-sort the geometry array in the same way in both files (using function fuzzysort copied form stack-overflow, as with argsort, floating-point issues lead to inconsistent sort).
-reindex the connect arrays based on the new geometry array.
-intersect the connect arrays and reindex to have the same faces at the same index.

Tested on both surface and volume output. Example of use:
```
base) ulrich@ulrich-ThinkPad-T490s:~/trash/Nepal$ python ~/TuSeisSolScripts/onHdf5/compute_diff_seissol_data.py 4eWei_longerSRT-GME.xdmf 4ve2Weitest-GME.xdmf --Data PGV PGA --idt 0
removed 130 duplicates as out 388620
removed 130 duplicates as out 388620
776636 faces in common, n faces connect 1:776636, 2:776732 (diff: 0, 96)
PGV
0 -0.052160539042424015 4.28496811455999
PGA
0 -0.34455823068800795 32.62138288869466
done writing diff_4eWei_longerSRT-GME.xdmf
```


krenzland:
Just a small idea (will do proper review soon!): Might it be a good idea to append Diff to the output variable? If the diff variables have the same name as the original ones, paraview is going to use the same colorscheme for both.

postprocessing/science/compute_diff_seissol_data.py
```diff
@@ -0,0 +1,201 @@
...
+    required=True,
+    help="list of time step to differenciate (1st = 0); -1 = all",
+    type=int,
+)
+parser.add_argument(
+    "--Data",
+    nargs="+",
+    required=True,
+    metavar=("variable"),
+    help="Data to differenciate (example SRs)",
```
sebwolf-de, line 137: Can you also add an "all" option here?

sebwolf-de (approved):
Hi, thanks for this PR and sorry for not checking this for so long.
Definitely a useful tool, and I'll use this for debugging.
Still for #414 we want the (relative) L2 norm. I mean if you could introduce this feature here, it would be great, but I understand that this goes beyond. I would approve your PR nonetheless, as we do not want to put a lot of additional work on each PR.

postprocessing/science/compute_diff_seissol_data.py
Thomas-Ulrich: I added your commit from the DR branch, and improved it a little bit.

sebwolf-de (approved):

Issue #342, opened 2021-05-31
Strategies for setting initial conditions to avoid NaNs in rate-state friction with strong velocity-weakening (FL 103) models (modified TPV104)


jbiemiller:
**Describe the bug**
This is not really a bug, but rather a question about initializing models without any unphysical initial conditions that lead to NaNs. When using the rate-state friction with strong velocity-weakening law (FL 103) I’ve found that NaNs are detected for different sets of parameters and initial conditions. Using TPV104 (https://github.com/SeisSol/Examples/tree/master/tpv104) as an example, if the the rate state parameter a increases slightly in fault.yaml, from

rs_a:   return 0.01 + 0.01 * (1.0 - box_x*box_z);

to

rs_a:   return 0.011 + 0.01 * (1.0 - box_x*box_z);

then a NaN is detected on the first solve and Seissol quits. Does anyone have any advice on how to appropriately initialize rate-state friction models so that no NaNs occur? Are there any guidelines for choosing initial RSF parameters (and other parameters, initial conditions, prestresses, etc) to avoid NaNs? 

Thanks for any ideas or suggestions!

**Expected behavior**
Dynamic rupture model with similar behavior to TPV104 but slightly more strongly velocity-weakening material in the VW portion of the fault.

**To Reproduce**
Run TPV104 with slightly higher rate-state a value (increase a by 0.001).

I have been testing these models with the docker training version of seissol (https://github.com/SeisSol/Training) on Mac OS 10.14.6.

**Screenshots/Console output**

Rank:        0 | Info    | <--------------------------------------------------------->
Rank:        0 | Info    | <     Start calc_SeisSol ...                              >
Rank:        0 | Info    | <--------------------------------------------------------->
Mon May 31 10:12:17, Info:  Writing faultoutput at time 0.
Mon May 31 10:12:17, Info:  Writing faultoutput at time 0. Done.
Mon May 31 10:12:17, Info:  #max-updates since sync:  0  @  0 
Rank:        0 | Error   | NaN detected

**Additional context**

Thomas-Ulrich:
Hi,
I don't think it is a problem of 'unphysical initial conditions', but rather a problem of improper implementation.
Running SeisSol in double precision seems to fix many of the problems of nan in rate and state at t=0s, suggesting it is a cancelation issue. Here is what Duo came up with for using RS friction at low effective normal stress:

```
     flv = RS_f0 - (RS_b-RS_a)* LOG(SR_tmp/RS_sr0)
[...]

PR #361, opened 2021-07-12
to avoid zero in LocSV

daisy20170101:
As Alice asked, we should discuss how to address the NaN issue in FL103. Is adding a waterlevel for state variable (LocSV) a proper solution to this? This is an emergent issue for James but could happen to everyone.

### DIFF
--- a/src/Physics/Evaluate_friction_law.f90
+++ b/src/Physics/Evaluate_friction_law.f90
@@ -1124,7 +1124,7 @@ MODULE Eval_friction_law_mod
         ! steady-state state variable
         ! For compiling reasons we write SINH(X)=(EXP(X)-EXP(-X))/2
         SVss = RS_a * LOG(2.0D0*RS_sr0/SR_tmp * (EXP(fss/RS_a)-EXP(-fss/RS_a))/2.0D0)
-
+        SVss = max(1e-8,SVss)
         ! exact integration of dSV/dt DGL, assuming constant V over integration step
         LocSV = Svss*(1.0D0-EXP(-SR_tmp*time_inc/RS_sl0))+EXP(-SR_tmp*time_inc/RS_sl0)*SV0
     END SELECT  


jbiemiller:
I tried a similar fix to this as detailed by Thomas in https://github.com/SeisSol/SeisSol/issues/342#issuecomment-855731427. It helped allow for a-b < -0.006 in a TPV104-based setup (FL 103) without generating NaNs.

sebwolf-de:
Hi, this PR is a draft, is this still relevant?

Thomas-Ulrich:
I think it is still relevant, but the solution in #342 may fix more issues of nan with RS friction.
Maybe we should merge/discuss these changes.

davschneller:
Hi, again the question (sorry for "digging up" this issue): does it still have relevance? And what exactly was the issue here? That `SVss` may be NaN (https://github.com/SeisSol/SeisSol/blob/980699631d54ff8ba5924495188b7b18ec51a1cd/src/DynamicRupture/FrictionLaws/FastVelocityWeakeningLaw.h#L65-L70 ), or only `LocSV`?

If it's due to `SVss` with a positive (!) input failing (due to the `exp(x) - exp(-x)` in `log(exp(x) - exp(-x))` becoming negative), there may be a solution; well, actually there may be two of them:

* After the port, we can now also go back to using `sinh` directly (at least numpy seems to use a rather refined implementation for x\approx 0 there). Thus `log(2*sinh(x))`
* Alternatively/additionally, we can also re-formulate `exp(x) - exp(-x)` into `exp(-x) * (1 - exp(2x))` and use the `expm1` function which implements `expm1(y) = 1 - exp(y)`. Giving us `log(exp(-x) * expm1(2x))`.
(note: the latter method _may_ potentially be re-formulated into just `log(expm1(2x)) - x`. But I haven't tested that too thoroghly)

In both cases, we may now compute *`log(exp(x) - exp(-x))` (after reformulating it) for x = 10e-31, in single precision.

Needless to say, that's not accounting for any cancellation, and thus negative input values, happening earlier already—then this fix attempt won't help.

davschneller:
Closing here; cf. #1504 (and some other PRs in the meantime) which hopefully make RSF+FVW a bit more stable.

PR #364, opened 2021-07-13
fix GME script (cumtrapz)

Thomas-Ulrich:
in #350 I removed the 'initial=0.' option of cumtrapz, as I thought it is unnecessary.
But it turns out that the array returned is then of size n-1 and not n, which breaks the script.
This PR fixes the issue.


sebwolf-de (approved):
LGTM

PR #372, opened 2021-08-04
fix Mw not printed/ print Mw by default

Thomas-Ulrich:
- fix Mw not printed on log when rankDR.NE.0 
- change default  to magnitude_output_on=1


sebwolf-de (approved):
Seems good to me :)

uphoffc:
What happens if DR is off or IO%OutputFile is not set?

Thomas-Ulrich:
> What happens if DR is off or IO%OutputFile is not set?
mw is not printed.
https://github.com/SeisSol/SeisSol/blob/master/src/Solver/calc_seissol.f90#L224
the file is named `data-MAG.dat`
https://github.com/SeisSol/SeisSol/blob/master/src/Reader/readpar.f90#L2586

uphoffc:
ok

PR #373, opened 2021-08-05
Travelling wave

sebwolf-de:
This PR adds a new Initial Condition, a travelling wave:
You can specify origin, direction vector and the modes you wish to use in the parameters file.


src/Physics/InitialField.cpp
```diff
@@ -104,6 +108,48 @@ void seissol::physics::SuperimposedPlanarwave::evaluate( double time,
...
 }
 
+seissol::physics::TravellingWave::TravellingWave(const CellMaterialData& materialData, const TravellingWaveParameters& travellingWaveParameters)
+  //Set phase to 0.5*M_PI, so we have a zero at the origin
+  //The wave travels in direction of kVec
+  //2*pi / magnitude(kVec) is the wave length of the wave
+  : Planarwave(materialData, 0.5*M_PI, travellingWaveParameters.kVec, travellingWaveParameters.varField, travellingWaveParameters.ampField),
+  //origin is a point on the wavefront at time zero
+    m_origin(travellingWaveParameters.origin)
+{
```
krenzland, line 118: I don't like this part. Hell is going to break loose if the material is not homog. Maybe add a warning to evaluate()
It probably doesn't make sense to implement this for non-homog. materials.

Documentation/parameters.par
```diff
@@ -15,10 +15,16 @@ cICType = 'Zero'                  ! Initial Condition
...
 !Planarwave  A planar wave for convergence tests, needs periodic boundary conditions
+!Travelling  A planar wave travelling wave (one period of a plane wave), needs more parameters to be specified
 !Scholte     A Scholte wave to test elastic-acoustic coupling
 !Snell       Snells law to test elastic-acoustic coupling
 !Ocean       An uncoupled ocean test case for acoustic equations
 
+!The following parameters are only needed for the travelling wave IC:
+origin = 0 0 0.5             ! Origin of the wave 
+kVec = 6.283 0 0             ! Gives direction of wave propagation, the wavelength of the travelling wave is 2*pi / norm(kVec)
+ampField = 2 0 0 0 0 0 0 1 0 ! Amplification of the different wave mod
```
krenzland, line 26: I don't understand this. What does the amp. field mean exactly? It probably states which variables you want to init, right?

krenzland (changes requested):
Looks overall good. (Except for the .par nonsense, but I don't believe I can blame you for that.)

src/Physics/InitialField.cpp
sebwolf-de: I think "hell is breaking loose" is a bit overstated here, see attached video. There I've set a higher velocity in the middle and a lower velocity at the upper and lower ends. As you see, the initial condition is as it is supposed to be. Also the wave propagation makes sense to me. Only the error computations at the end do not make sense.
output

Documentation/parameters.par
sebwolf-de: Compare to gamma in Eq. 35 in https://library.seg.org/doi/10.1190/1.2965027, but I'll explain better ;-)

Documentation/initial-condition.rst
```diff
@@ -21,12 +21,36 @@ After the simulation is finished the errors between the analytic solution and th
...
+------------------------
 
 Superimposed three planar waves travelling in different directions. 
 This is especially interesting in the case of directional dependent properties such as for anisotropic materials.
 
+Travelling wave
+---------------
+
+Impose one period of a sinusoidal planar wave as initial condition, see for example the video below.
+There, we impose a P wave travelling to the left. The wave speed at the top and bottom is :math:`2 m/s` and :math:`2.83 m/s` in the middle.
```
Thomas-Ulrich, line 34: I would add between the 2 sentences. 
>Note that because of the heterogeneities in the elastic properties, the initially planar wave deviates from planarity.

src/Initializer/typedefs.hpp
```diff
@@ -457,4 +457,11 @@ struct MemoryProperties {
   size_t pagesizeStack{PAGESIZE_STACK};
 };
 
+struct TravellingWaveParameters {
+  std::array<double, 3> origin;
```
Thomas-Ulrich, line 461: uninformed question:
why double and not real?

src/Solver/Interoperability.cpp
```diff
@@ -369,6 +373,22 @@ seissol::Interoperability::~Interoperability()
...
 
+void seissol::Interoperability::setTravellingWaveInformation(const double* origin, const double* kVec, const double* ampField) {
+  assert(origin != nullptr);
+  assert(kVec != nullptr);
+  assert(ampField != nullptr);
+
+  m_travellingWaveParameters.origin = {origin[0], origin[1], origin[2]};
+  m_travellingWaveParameters.kVec = {kVec[0], kVec[1], kVec[2]};
+  constexpr double eps = 1e-15;
+  for (size_t i = 0; i < NUMBER_OF_QUANTITIES; i++) {
```
Thomas-Ulrich, line 384: this won't work with attenuation?

sebwolf-de: Indeed, that's why I did not enable it for attenuation: https://github.com/SeisSol/SeisSol/blob/d08b8eb4171e938514d35a9ca32c2756e16f3af4/src/Solver/Interoperability.cpp#L948

src/Initializer/typedefs.hpp
sebwolf-de: `InitialField::evaluate` takes `points` as double, so I made the origin also double....
https://github.com/SeisSol/SeisSol/blob/d08b8eb4171e938514d35a9ca32c2756e16f3af4/src/Physics/InitialField.h#L17

src/Physics/InitialField.cpp
krenzland: Hm, shouldn't you use consider the material of each given cell? Or doesn't this lead to an inconsistent init?

src/Solver/Interoperability.cpp
Thomas-Ulrich: ok!

src/Initializer/typedefs.hpp
Thomas-Ulrich: ok!

krenzland (approved):

PR #376, opened 2021-08-12
Stable gravitational bc/Changes for SC21 paper

krenzland:
This is a collection of changes from our SC21 paper. Sorry for the mess, you know how it gets.

The most important changes are:
- An entirely new, stable implementation of the gravitational free surface boundary condition
- Configurable weights that are used for mesh partitioning (cells with dynamic rupture/free surface are more expensive)
- Additionally, the free surface writer outputs a quantity called "LocationFlag", which has the values
0 and 1 when at the elastic or acoustic side of an elastic-acoustic interface.
It has the value 2 for an ordinary free surface boundary condition and the value 3 for a free surface with gravity
boundary condition. This value can be used to filter the output for example using Paraview's Threshold filter.
- Updated implementation of the ocean scenario
- Experimental likwid-markers in the proxy
- An option to deactivate MINI_SEISSOL

Feel free to already review it, there are these things missing:
- GPU support. Now the displacements are stored as face displacements (updated in Time.cpp). @ravil-mobile, I need your help here. It's essentially the same as before, but now we project the displacement update before adding it
- I've updated the free surface writer recently, I need to verify that I did not break anything.
- I need to add some documentation of the new flags.


src/Equations/viscoelastic2/Kernels/Time.cpp
```diff
@@ -140,6 +141,9 @@ void seissol::kernels::Time::computeAder( double                      i_timeStep
     intKrnl.power *= i_timeStepWidth / real(der+1);    
     intKrnl.execute(der);
   }
+
+  // TODO(Lukas) Implement!
+  // Compute integrated displacement over time step if needed.
```
sebwolf-de, line 146: Should we add a warning here?

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -263,6 +265,7 @@ void seissol::initializers::initializeBoundaryMappings(const MeshReader& i_meshR
 
   for (LTSTree::leaf_iterator it = io_ltsTree->beginLeaf(LayerMask(Ghost)); it != io_ltsTree->endLeaf(); ++it) {
     auto* cellInformation = it->var(i_lts->cellInformation);
+    auto* material = it->var(i_lts->material);
```
sebwolf-de, line 268: Seems to be an unused variable.

src/Initializer/MemoryManager.cpp
```diff
@@ -506,6 +506,7 @@ void seissol::initializers::MemoryManager::fixateBoundaryLtsTree() {
        layer != m_ltsTree.endLeaf() && boundaryLayer != m_boundaryTree.endLeaf();
        ++layer, ++boundaryLayer) {
     CellLocalInformation* cellInformation = layer->var(m_lts.cellInformation);
+    CellMaterialData* material = layer->var(m_lts.material);
```
sebwolf-de, line 509: Seems to be an unused variable.

src/Initializer/MemoryManager.cpp
```diff
@@ -532,6 +533,7 @@ void seissol::initializers::MemoryManager::fixateBoundaryLtsTree() {
     auto* cellInformation = layer->var(m_lts.cellInformation);
     auto* boundaryMapping = layer->var(m_lts.boundaryMapping);
     auto* faceInformation = boundaryLayer->var(m_boundary.faceInformation);
+    CellMaterialData* cellMaterialData = layer->var(m_lts.material);
```
sebwolf-de, line 536: Seems to be an unused variable.

src/Initializer/time_stepping/LtsWeights.cpp
```diff
@@ -194,15 +195,31 @@ void seissol::initializers::time_stepping::LtsWeights::computeWeights(PUML::TETP
...
+    PUML::Downward::faces(mesh, cells[cell], faceids);
+
     for (unsigned face = 0; face < 4; ++face) {
+      const auto faceType = static_cast<FaceType>(getBoundaryCondition(boundaryCond, cell, face));
       dynamicRupture += ( getBoundaryCondition(boundaryCond, cell, face) == 3) ? 1 : 0;
+
+      int cellIds[2];
+      PUML::Upward::cells(mesh, faces[faceids[face]], cellIds);
+
+      int neighbourCell = (cellIds[0] == static_cast<int>(cell)) ? cellIds[1] : cellIds[0];
```
sebwolf-de, line 212: Seems to be an unused variable.

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -44,6 +44,7 @@
 #include <cassert>
 
 #include <Initializer/ParameterDB.h>
+#include "Initializer/MemoryManager.h"
```
sebwolf-de, line 47: Seems to be an unnecessary include

src/Initializer/MemoryManager.h
```diff
@@ -201,6 +206,11 @@ class seissol::initializers::MemoryManager {
      */
     void initializeDisplacements();
 
+    /**
+      * Initializes the displacement accumulation buffer.
+    */
+  void initializeFaceDisplacements();
```
sebwolf-de, line 212: Indentation is wrong here

Documentation/parameters.par
```diff
@@ -77,6 +78,10 @@ PPFileName = 'tpv33_faultreceivers.dat'
 &MeshNml
 MeshFile = 'tpv33_gmsh'         ! Name of mesh file
 meshgenerator = 'PUML'          ! Name of meshgenerator (Gambit3D-fast, Netcdf or PUML)
+ertexWeightElement = 100 ! Base vertex weight for each element used as input to ParMETIS
+vertexWeightDynamicRupture = 200 ! Weight that's added for each DR face to element vertex weight
+vertexWeightDisplacement = 300 ! Weight that's added for each free surface with gravity face to element vertex weight
```
sebwolf-de, line 83: Maybe displacement is the wrong term here. At an elastic acoustic interface or at a normal free surface, we also compute displacements. Might confuse users with this naming

src/Initializer/time_stepping/LtsWeights.cpp
```diff
@@ -194,15 +195,31 @@ void seissol::initializers::time_stepping::LtsWeights::computeWeights(PUML::TETP
...
+
+      if (faceType == FaceType::freeSurfaceGravity) {
+        requiresDisplacement++;
```
sebwolf-de, line 215: Also here, displacement is not precisely, what we count here.

src/Physics/InitialField.cpp
```diff
@@ -197,32 +203,86 @@ void seissol::physics::Ocean::evaluate(double time,
     const auto z = points[i][2];
     const auto t = time;
 
-    const double g = 9.81; // m/s
+    const auto g = gravitationalAcceleration;
+    if (std::abs(g - 9.81e-3) > 10e-15) {
```
sebwolf-de, line 207: Why is there the `e-3`?

src/Physics/InitialField.cpp
```diff
@@ -186,6 +186,12 @@ void seissol::physics::SnellsLaw::evaluate(double time,
 #endif
 }
 
+seissol::physics::Ocean::Ocean(int mode, double gravitationalAcceleration)
```
sebwolf-de, line 189: Could you maybe add a reference to this test? Google does not give too much information, if I search for it ;-)

sebwolf-de (commented):
Hi, large, but good PR :sweat_smile: 
I think the functionality works. Most comments are things, which were not exactly clear to me while reading the code.
I did not check if all values are correct (e.g. the Butcher tableaus, or the how the PDE for eta is implemented). I guess you checked this.
Also I just had a brief look at the Free Surface Writer, as you said, it is not completely ready.

Documentation/free-surface-output.rst
```diff
@@ -29,3 +29,9 @@ variables
 
    | **u**, **v**, **w**: ground velocities, x y and z components
    | **U**, **V**, **W**: ground displacements, x y and z components
+
+Additionally, the writer outputs a quantity called "locationFlag", which has the values
+0 and 1 when at the elastic or acoustic side of an elastic-acoustic interface.
```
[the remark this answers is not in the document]
krenzland: Yes. The reason for this is that only the normal velocity need to be continuous - the other ones are discontinuous. So they actually have different values. This is also the reason for introducing the flags output. Otherwise we cannot distinguish between both sides.

auto_tuning/proxy/src/proxy_seissol_integrators.hpp
```diff
@@ -43,6 +53,7 @@ namespace proxy::cpu {
   #ifdef _OPENMP
     #pragma omp parallel
     {
+    LIKWID_MARKER_START("ader");
```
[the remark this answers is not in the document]
krenzland: I don't know. I really don't see any reason why one would want to run the proxy without openmp though?

Documentation/parameters.par
krenzland: True. It actually confused me. The original intention was that we weight each face which has an displacement. (They were originally updated with the same solver). I actually implemented it like you suggest. So slightly underreporting the FLOPS.

src/Physics/InitialField.cpp
krenzland: It's km/s not m/s for this scenario

src/Physics/InitialField.cpp
krenzland: Added. But the reference isn't really that helpful, because it's an AGU abstract - the results will be published by L. Abrahams et al. soonish

krenzland: Actually, I've added a reference to a very similar 2D scenario. Will update once other paper is published (at least as preprint)

Documentation/free-surface-output.rst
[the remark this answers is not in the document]
sebwolf-de: Can you add this explanation in the documentation? If I understand correctly, there will be several files then?

src/Physics/InitialField.cpp
sebwolf-de: :+1:

Documentation/free-surface-output.rst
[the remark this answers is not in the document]
krenzland: No, there is only one file which contains all outputs. I'll make documentation a bit clearer.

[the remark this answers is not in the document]
sebwolf-de: Thx

src/tests/Numerical_aux/ODEInt.t.h
```diff
@@ -0,0 +1,132 @@
...
+        double time) {
+      // A simple Lotka-Volterra model
+      // See: https://en.wikipedia.org/wiki/Lotka%E2%80%93Volterra_equations
+
+      // dx/dt = \alpha x - \beta x y
+      // dy/dt = \delta x y - \gamma y
+      auto& [alpha, beta, delta, gamma] = parameters;
+      du[0] = alpha * u[0] - beta * u[0] * u[1];
+      du[1] = delta * u[0] * u[1] - gamma * u[1];
+    };
```
sebwolf-de, line 105: Nice!

sebwolf-de (approved):

PR #384, opened 2021-08-31
precise tutorial with python3, precise paths

Thomas-Ulrich:
Small PR that precises the tutorials:

- use python3 instead of python when calling the cad scripts 
- precise path to Examples

(or maybe it would be preferable to add a shebang (`#!/usr/bin/env python3`) to all these scripts?)


krenzland:
I'd vote for the shebang, but we can merge this PR anyway.

krenzland (approved):

Thomas-Ulrich:
Ok, then let's do both!

PR #385, opened 2021-09-03
Fix multiple simulations

sebwolf-de:
This PR fixes the fused, multiple simulations. Before that, it was not possible to compile SeisSol with `NUMBER_OF_FUSED_SIMULATIONS > 1`.
*  Added support for Haswell; this is not optimized nor tuned, just some settings, such that SeisSol compiles.
*  Fix code generation; in particular `V3mTo2nFace` needs to be transposed such that the dimensions align.
* Add filter for multiple simulations to viewrec, to better analyse multiple simulations
* Map point sources to multiple simulations, i.e. if there are several point sources, source `i` is mapped to simulation `i % NUMBER_OF_FUSED_SIMULATIONS`. 
* Prepare the point sources for the fused GPU implementation.
* This will not work with Lukas' boundary conditions and might break some stuff there. Those will be disabled, if the simulation fusing is used.

This was tested with an LOH like test case with four different point sources.


sebwolf-de:
Todo: 

- [ ] Test LOH1 with 4 fused simulations
- [ ] Test TPV13 for energy output with 1 simulation
- [ ] Test Acoustic + Gravity BC setup

sebwolf-de:
https://github.com/SeisSol/SeisSol/blob/7afdc0ec29af4aa188fc18fd15d19f08b26aa74e/src/SourceTerm/Manager.cpp#L455 This line is broken in multi-node setups.

davschneller:
(one small comment (if it's not obvious to you already): you can probably get rid of some of the remaining `ifdef`s, if you utilize something like https://github.com/SeisSol/SeisSol/blob/0361e75d525ba358e607014bec17e9e10dc98a6e/src/Kernels/Receiver.cpp#L43-L57 more globally—basically everywhere, where you "call" a Yateto tensor in either a fused or not-fused fashion)

vikaskurapati:
> (one small comment (if it's not obvious to you already): you can probably get rid of some of the remaining `ifdef`s, if you utilize something like
> 
> https://github.com/SeisSol/SeisSol/blob/0361e75d525ba358e607014bec17e9e10dc98a6e/src/Kernels/Receiver.cpp#L43-L57
> 
> more globally—basically everywhere, where you "call" a Yateto tensor in either a fused or not-fused fashion)

To be honest, I don't intuitively understand how this method works when I look at it. In any case, in my experience with fused-simulations, I found the ifdefs a bit more useful during the debugging process. But once ready, if you think any of ifdefs are too obvious or unnecessary, point me towards them and I will look into it.

davschneller:
In essence, it does the following: you may always index a e.g. matrix by `multisimWrap(matrix, sim, index1, index2)`. If it's single-simulation, that'll be transformed to `matrix(index1, index2)`; in a multi-sim regime, it'll become `matrix(sim, index1, index2)`.
Thus, you can keep the loop, and remove the `ifdef`.

src/ResultWriter/AnalysisWriter.cpp
```diff
@@ -117,13 +118,7 @@ void AnalysisWriter::printAnalysis(double simulationTime) {
...
   seissol::quadrature::TetrahedronQuadrature(quadraturePoints, quadratureWeights, QuadPolyDegree);
 
-#ifdef MULTIPLE_SIMULATIONS
-  constexpr unsigned multipleSimulations = MULTIPLE_SIMULATIONS;
-#else
-  constexpr unsigned MultipleSimulations = 1;
-#endif
-
-  for (unsigned sim = 0; sim < MultipleSimulations; ++sim) {
+  for (unsigned sim = 0; sim < multipleSimulations::numberOfSimulations; ++sim) {
```
davschneller, line 121: Make sure to also apply something like `multisimWrap` (or a similar solution here)—there are some `ifdef`s in the loop body

src/Physics/InitialField.cpp
```diff
@@ -133,8 +134,12 @@ void seissol::physics::SuperimposedPlanarwave::evaluate(
     yateto::DenseTensorView<2, real, unsigned>& dofsQP) const {
   dofsQP.setZero();
 
-  real dofsPwData[tensor::dofsQP::size()];
-  yateto::DenseTensorView<2, real, unsigned> dofsPW = init::dofsQP::view::create(dofsPwData);
+  std::vector<real> dofsPwVector(dofsQP.size());
+  auto dofsPW = yateto::DenseTensorView<2, real, unsigned>(
```
davschneller, line 138: Actually, why can't we use `init::dofsQP::view::create` anymore here?

src/Kernels/PointSourceClusterOnHost.cpp
```diff
@@ -83,7 +84,10 @@ void PointSourceClusterOnHost::addTimeIntegratedPointSourceNRF(unsigned source,
   krnl.mArea = -sources_->A[source];
   krnl.momentToNRF = init::momentToNRF::Values;
 #ifdef MULTIPLE_SIMULATIONS
-  krnl.oneSimToMultSim = init::oneSimToMultSim::Values;
+  const auto originalIndex = sources_->originalIndex[source];
+  std::array<real, MULTIPLE_SIMULATIONS> sourceToMultSim{};
```
davschneller, line 88: Use the `constexpr` instead here

src/Kernels/PointSourceClusterOnHost.cpp
```diff
@@ -106,7 +110,10 @@ void PointSourceClusterOnHost::addTimeIntegratedPointSourceFSRM(unsigned source,
   krnl.momentFSRM = sources_->tensor[source].data();
   krnl.stfIntegral = slip;
 #ifdef MULTIPLE_SIMULATIONS
-  krnl.oneSimToMultSim = init::oneSimToMultSim::Values;
+  const auto originalIndex = sources_->originalIndex[source];
+  std::array<real, MULTIPLE_SIMULATIONS> sourceToMultSim{};
```
davschneller, line 114: Use the `constexpr` here

src/Kernels/Plasticity.cpp
```diff
@@ -45,6 +45,10 @@ unsigned Plasticity::computePlasticity(double oneMinusIntegratingFactor,
                                        const seissol::model::PlasticityData* plasticityData,
                                        real degreesOfFreedom[tensor::Q::size()],
                                        real* pstrain) {
+#ifdef MULTIPLE_SIMULATIONS
+  // Todo(SW) find a better solution here
+  logError() << "Plasticity does not work with multiple simulations";
+#else
```
davschneller, line 51: Is it possible to use an `endif` here? Just curious—if that breaks the code, then no need to forcibly change it now already.

src/Initializer/InitialFieldProjection.cpp
```diff
@@ -288,7 +293,13 @@ void projectEasiInitialField(const std::vector<std::string>& iniFields,
       // TODO: multisim loop
       for (std::size_t i = 0; i < NumQuadPoints; ++i) {
         for (std::size_t j = 0; j < quantityCount; ++j) {
-          iniCond(i, j) = data.at(meshId * dataStride + quantityCount * i + j);
+#ifdef MULTIPLE_SIMULATIONS
+          for (std::size_t s = 0; s < seissol::multipleSimulations::numberOfSimulations; s++) {
```
davschneller, line 297: The `ifdef` can be resolved with a `multisimWrap` or a variant of that; maybe.

src/ResultWriter/EnergyOutput.cpp
```diff
@@ -700,22 +751,42 @@ void EnergyOutput::checkAbortCriterion(real timeSinceThreshold, const std::strin
...
-        << time << ",acoustic_energy," << energiesStorage.acousticEnergy() << "\n"
-        << time << ",acoustic_kinetic_energy," << energiesStorage.acousticKineticEnergy() << "\n"
-        << time << ",elastic_energy," << energiesStorage.elasticEnergy() << "\n"
-        << time << ",elastic_kinetic_energy," << energiesStorage.elasticKineticEnergy() << "\n"
-        << time << ",momentumX," << energiesStorage.totalMomentumX() << "\n"
-        << time << ",momentumY," << energiesStorage.totalMomentumY() << "\n"
-        << time << ",momentumZ," << energiesStorage.totalMomentumZ() << "\n"
-        << time << ",plastic_moment," << energiesStorage.plasticMoment() << "\n";
+  for (size_t sim = 0; sim < multipleSimulations::numberOfSimulations; sim++) {
+#ifdef MULTIPLE_SIMULATIONS
```
davschneller, line 755: You could e.g. introduce a `constexpr` with `MultipleSimulations == true/false`. Then you can also shorten the string initialization here.

davschneller (commented):
Ok; first review pass done. I may need to check some code parts a bit more thoroughly for merge problems.

Generally, it looks quite good already; only some tiny remarks:

* it would be good to remove as much of the `ifdef`s as possible. You can use e.g. the `multisimWrap` here in most cases where you access a tensor with an optional dimension.
* Not sure if the `config/hsw_msN.xml` configs are even necessary—that'll depend on further testing. Note that here, PSpaMM (at least with the currently-open PR) may give a speedup over the vanilla LIBXSMM sparse version.
* The poroelasticity `Local.cpp` has been un-linked. By accident—or was that intended?

src/Equations/poroelastic/Kernels/Local.cpp
```diff
@@ -1 +0,0 @@
-../../elastic/Kernels/Local.cpp
```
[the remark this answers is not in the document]
vikaskurapati: I did not do it. I think it may be one of the things that happened during autoformatting.

src/DynamicRupture/Output/ReceiverBasedOutput.cpp
```diff
@@ -425,6 +425,15 @@ real ReceiverOutput::computeRuptureVelocity(Eigen::Matrix<real, 2, 2>& jacobiT2d
     auto chiTau2dPoints =
         init::quadpoints::view::create(const_cast<real*>(init::quadpoints::Values));
     auto weights = init::quadweights::view::create(const_cast<real*>(init::quadweights::Values));
+    // TODO: Understand why the dimension changes with MULTIPLE_SIMULATIONS. This is also probably
```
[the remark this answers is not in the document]
vikaskurapati: This was implemented just to make the code compile as fused simulations' implementation for Dynamic Rupture has not been implemented yet. This ifdef was to make sure it compiles and the code doesn't break for single simulations. IMO, we should let the DR parts be and modify them accordingly when we decide to use the fused simulations strategy for Dynamic Rupture too.

src/DynamicRupture/Output/OutputAux.cpp
```diff
@@ -106,11 +106,21 @@ TriangleQuadratureData generateTriangleQuadrature(unsigned polyDegree) {
   // Generate triangle quadrature points and weights (Factory Method)
   auto pointsView = init::quadpoints::view::create(const_cast<real*>(init::quadpoints::Values));
   auto weightsView = init::quadweights::view::create(const_cast<real*>(init::quadweights::Values));
-  auto* reshapedPoints = unsafe_reshape<2>(data.points.data());
+  // TODO: Understand why the dimension changes with MULTIPLE_SIMULATIONS. This is also probably
+  // wrong. Needs to be corrected when DR is implemented
+  auto getWeights = [&weightsView](size_t index) {
+#ifdef MULTIPLE_SIMULATIONS
```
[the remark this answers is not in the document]
vikaskurapati: This was implemented just to make the code compile as fused simulations' implementation for Dynamic Rupture has not been implemented yet. This ifdef was to make sure it compiles and the code doesn't break for single simulations. IMO, we should let the DR parts be and modify them accordingly when we decide to use the fused simulations strategy for Dynamic Rupture too.

gitlab-cpu-pipeline.yml
```diff
@@ -130,19 +159,20 @@ run_tpv:
...
             - precision: [double, single]
-              tpv: [5, 5-nuc, 5-poro, 5-visco, 6, 13, 16, 101, 101-slip, 104, 105, ahsp, yoffe, gaussian]
+              tpv: [5, 5-nuc, 5-poro, 5-visco, 6, 13, 16, 101, 101-slip, 104, 105, ahsp, loh1-fused, yoffe, gaussian]
     script: 
         - echo "run TPV${tpv}"
         - git clone https://github.com/SeisSol/precomputed-seissol.git
         - ls
         - cd precomputed-seissol
-        - git checkout 2ae63167d0c5e35cb01c56198dba94a2b7a224d9
+        - git checkout 414b914
```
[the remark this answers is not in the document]
vikaskurapati: Cf. https://github.com/SeisSol/precomputed-seissol/pull/8

generated-code/kernels/nodalbc.py
```diff
@@ -28,14 +28,20 @@ def addKernels(
...
-    easi_boundary_constant = Tensor(
+
+    easi_boundary_constant = OptionalDimTensor(
         "easiBoundaryConstant",
+        aderdg.Q.optName(),
+        aderdg.Q.optSize(),
+        aderdg.Q.optPos(),
         (aderdg.numberOfQuantities(), aderdg.numberOf2DBasisFunctions()),
-        alignStride=False,
+        alignStride=True,
```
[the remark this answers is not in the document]
vikaskurapati: I am not a 100% sure either. But I don't see why not.

generated-code/kernels/dynamic_rupture.py
```diff
@@ -145,13 +145,33 @@ def interpolateQGenerator(i, h):
...
+        aderdg.Q.optName(),
+        aderdg.Q.optSize(),
+        aderdg.Q.optPos(),
+        (1,),
+        alignStride=True,
     )
-    staticFrictionalWork = Tensor("staticFrictionalWork", ())
     minusSurfaceArea = Scalar("minusSurfaceArea")
-    spaceWeights = Tensor("spaceWeights", (numberOfPoints,), alignStride=True)
+    spaceWeights = Tensor("spaceWeights", (numberOfPoints, 1), alignStride=True)
```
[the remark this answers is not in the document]
vikaskurapati: Yateto complains otherwise and it lead to compilation failures

src/SourceTerm/Typedefs.h
```diff
@@ -41,6 +41,10 @@ struct PointSources {
       seissol::memory::AlignedArray<real, tensor::mInvJInvPhisAtSources::size()>>
       mInvJInvPhisAtSources;
 
+#ifdef MULTIPLE_SIMULATIONS
+  std::vector<unsigned> originalIndex;
```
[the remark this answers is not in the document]
vikaskurapati: No, I don't think so. We don't need it for serial runs. Here, number of point sources is read and allocated to different fused simulations. In serial cases, the point sources are just allotted to a single simulation. I think it is better to separate this to make it clear that it is only being used in fused simulations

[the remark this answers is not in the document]
davschneller: If it's not performance-detrimental or doesn't over-complicate things, then I'd vouch for always having it instead reducing the number of `ifdef`s over keeping an additional index array.
Maybe there are even more elegant ways (like having a wrapper type that annuls all calls if we're not in a multisim setup), but as long as the performance doesn't suffer and it doesn't clutter the code to much, then why not? Having the `ifdef`s instead is actually more painful IMO.

Generally, we should try to keep the number of new `ifdef`s as low as possible and gradually reduce all old `ifdef`s as well. Each `ifdef` complicates the code by a lot, and if we can avoid it, we should.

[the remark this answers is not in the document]
vikaskurapati: > If it's not performance-detrimental or doesn't over-complicate things, then I'd vouch for always having it instead reducing the number of `ifdef`s over keeping an additional index array. Maybe there are even more elegant ways (like having a wrapper type that annuls all calls if we're not in a multisim setup), but as long as the performance doesn't suffer and it doesn't clutter the code to much, then why not? Having the `ifdef`s instead is actually more painful IMO.
> 
I would generally agree with no having too many ifdefs is possible but in this case, I think this is needed in this place mainly because the originalIndex could be confused with and there is a higher chance of new bugs creeping in. This, I mention as one of the most recent issues was related to this. I think we should keep it here because at least that way it is clear for what purpose it is intended.
> Generally, we should try to keep the number of new `ifdef`s as low as possible and gradually reduce all old `ifdef`s as well. Each `ifdef` complicates the code by a lot, and if we can avoid it, we should.
Totally agree on this point. But I suggest we keep this one for absolute clarity

src/Physics/InitialField.cpp
vikaskurapati: The method has different dimensions for fused and non fused. It was in one of the older commits by Sebastian, IIRC.

src/Kernels/PointSourceClusterOnHost.cpp
vikaskurapati: I don't get what you mean here. Which one did you want to be used as a constexpr?

davschneller: For the `MULTIPLE_SIMULATIONS`

src/Kernels/PointSourceClusterOnHost.cpp
vikaskurapati: Where exactly did you want a constexpr? The originalIndex?

src/Kernels/Plasticity.cpp
vikaskurapati: It breaks compilation mainly due to tensor sizes underneath

src/SourceTerm/Typedefs.h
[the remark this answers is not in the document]
davschneller: If it's about the confusion—sure, a 100% guarantee won't exist. In fact, the `ifdef`s may increase the problems. E.g. that makes it also more difficult to work with GPUs, since we have a lot of preprocessor code inclusion/exclusion there.

Also e.g. the formatting etc. won't work with `ifdef`s enabled, and we effectively need to compile multiple versions to see that nothing broke.

But if it is about the name of the variable; then how about then renaming it to something like `multisimIndex`, `multisimOriginalIndex`, or similar? (and maybe leave a comment there) That should resolve the naming problems.

(and—I may have forgotten—which was the Issue that something was breaking due to the multisim structure or similar issues? Maybe it's just something that could have appear "out of the usual" otherwise anyways.)

Don't get me wrong—I fully agree otherwise to keep the code from unnecessary operations; really. However, the `ifdef`s are just as bad.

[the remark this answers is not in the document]
vikaskurapati: > If it's about the confusion—sure, a 100% guarantee won't exist. In fact, the `ifdef`s may increase the problems. E.g. that makes it also more difficult to work with GPUs, since we have a lot of preprocessor code inclusion/exclusion there.
> 
> Also e.g. the formatting etc. won't work with `ifdef`s enabled, and we effectively need to compile multiple versions to see that nothing broke.
> 
We will need to do that anyway with multiple simulations as we need to test with different binaries anyway. At least we will know with this which part is exactly breaking it. But we anyway need to compile multiple versions to see nothing broke like we already do it in the CI. We will just need to add another CI test eventually once we have the entire fused simulations strategy ready with Dynamic Rupture and Plasticity.

> But if it is about the name of the variable; then how about then renaming it to something like `multisimIndex`, `multisimOriginalIndex`, or similar? (and maybe leave a comment there) That should resolve the naming problems.
> 
I will do this even though I don't agree with removing the ifdef here just because we keep running in circles about this thing. 

> (and—I may have forgotten—which was the Issue that something was breaking due to the multisim structure or similar issues? Maybe it's just something that could have appear "out of the usual" otherwise anyways.)
> 
The point sources need to be mapped to the simulation number for multiple simulations from a source file. But if you use the same source file in a single simulation, you would read it as multiple sources in one simulation. This was the structure Sebastian used.

> Don't get me wrong—I fully agree otherwise to keep the code from unnecessary operations; really. However, the `ifdef`s are just as bad.
I think this is where the call needs to be made. I still believe the ifdef rather makes it easier to understand what exactly is happening better here from my personal experience. I would nevertheless make the change.

src/SourceTerm/Manager.cpp
```diff
@@ -344,6 +344,9 @@ auto loadSourcesFromFSRM(const char* fileName,
       sources.numberOfSources = numberOfSources;
       sources.mInvJInvPhisAtSources.resize(numberOfSources);
       sources.tensor.resize(numberOfSources);
+#ifdef MULTIPLE_SIMULATIONS
```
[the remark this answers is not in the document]
vikaskurapati: I don't like it but done.

.github/workflows/build-seissol-cpu.yml
```diff
@@ -156,29 +150,16 @@ jobs:
...
-          fi
+
+          ./SeisSol_proxy_${EXENAME} 100 1 ader
+          ./SeisSol_proxy_${EXENAME} 100 1 localwoader
+          ./SeisSol_proxy_${EXENAME} 100 1 local
+          ./SeisSol_proxy_${EXENAME} 100 1 neigh
+          ./SeisSol_proxy_${EXENAME} 100 1 all
+          ./SeisSol_proxy_${EXENAME} 100 1 neigh_dr
+          ./SeisSol_proxy_${EXENAME} 100 1 godunov_dr
+          ./SeisSol_proxy_${EXENAME} 100 1 all_dr
```
vikaskurapati, line 162: This will probably throw errors again in the CI for multiple simulations. DR is not yet implemented there. That is why there was an if...else above. Shifting the error to friction law is not a good solution, in my opinion, as the main changes in the multiple simulations to get DR working happen in the computeDynamicRupture and Neighbor kernels. So, the error is in the correct place. We need to omit running the DR tests for multiple simulations to keep it simple and sensible.

davschneller: It won't; those do work.

vikaskurapati: > (cf. [453f599](https://github.com/SeisSol/SeisSol/commit/453f599cca103448fb83df2e07036181c6e09eaa) )

Yeah, this is basically what I wanted to avoid shifting. But okay. In my opinion, the error there in the ComputeDynamicRupture made more sense. But just to prevent an "if...else" in the CI, if we want to do this, fine by me.

PR #388, opened 2021-09-07
Slightly Refactored LtsWeights class

ravil-mobile:
Hi all. I would like to add different implementations for LTS weights which we are using for mesh partitioning. For example, I need to add an extra constrain for balancing GPU memory. I am sure that it is not necessary for our CPU implementation because a decent HPC node has 100's of GB of main memory. So, I guess it has never been an issue; but it is an issue for GPUs.

The main idea to add an abstract class with a bunch of default methods and a couple of derived classes. A user will select an LTS weight strategy in his/her `parameters.par`. This option is going to get propagated from Fortran code to C++ where we select an implementation with a factory method. If the user doesn't specify the option then the default implementation (which is going to be the current one) is going to be selected in order to keep comparability with `parameters.par` files from examples and benchmarks.

Before, I start I did some slight refactoring  and `computeWeights` into a few steps. I supposed derived classes should implement their own `setVertexWeights` and `setImbalances` methods.

Btw, I had to change `PUML` interface a little bit. You can find the corresponding changes [here](https://github.com/TUM-I5/PUML2/blob/57bfa7df18ab3610e66a5827a64dca6b28bd168b/PartitionMetis.h#L59-L76).

This draft contains the following changes.

* split computeWeights into steps
* replace `new` with std::vector
* add checking errors after calling parmetis
* added computations of imbalances; it is a responsibility
of LtsWeights class but not PUML::PartitionMetis one
* updated reference to PUML

I would really appreciate if somebody can comment on this. Thanks in advance.


src/Reader/readpar.f90
```diff
@@ -2448,6 +2449,12 @@ SUBROUTINE readpar_discretisation(EQN,MESH,DISC,SOURCE,IO)
       logInfo(*) 'Using multi-rate clustered LTS:', disc%galerkin%clusteredLts
     endselect
 
+    disc%galerkin%ltsWeightTypeId = LtsWeightTypeId
+    if ((DISC%Galerkin%clusteredLts > 0) .and. (DISC%Galerkin%ltsWeightTypeId > 0)) then
+        logInfo(*) 'Using memory balancing for LTS scheme'
+    end if
+    logInfo(*) 'Using memory balancing for LTS scheme: ', clusteredLts
```
Thomas-Ulrich, line 2456: copy-paste error for this loginfo?

Thomas-Ulrich (commented):
I think your change make the implementation clearer.
Can you precise what `encoded` does?
Would a short page in the readthedocs about partition option help?
I can approve if nobody comments from TUM in the next days.

ravil-mobile:
@Thomas-Ulrich, thanks for your review.

> Would a short page in the readthedocs about partition option help?

Added an explanation into `memory-requirements.rst`

PR #391, opened 2021-09-23
remove patch needed for attenuation

Thomas-Ulrich:
I realized we have already a parameter to limit the max time step in SeiSsol.
Therefore, it should not be needed to patch SeiSsol for using attenaution.


uphoffc:
Fine in principle, just that for load balancing the FixTimeStep parameter is disregarded: https://github.com/SeisSol/SeisSol/blob/master/src/Initializer/time_stepping/LtsWeights.cpp#L70

Thomas-Ulrich:
But the clustering seems affected by the parameter.
Is it because of this line?
https://github.com/SeisSol/SeisSol/blame/master/src/Solver/calc_deltat.f90#L195
Then I would just need to bring the parameter to LtsWeights.cpp?

PR #394, opened 2021-09-30
print Mw: change logInfo to explicit logInfo0DR

Thomas-Ulrich:
What we need for printing Mw is a loginfo0 for the DR subgroup.
The current implementation (logInfo) does not print with the default:
```
LOG_LEVEL                        warning                                                                                                                                                                                                      
 LOG_LEVEL_MASTER                 info                                                                                                                                                                                                         
```
loginfo0 would not print either if rank 0 is not in the DR subgroup.
So I propose the following workaround.


krenzland:
Very good point...

PR #397, opened 2021-10-10
Documentation for fracture network

palgunadi1993:
This pull request is to add documentation for SeisSol enhancement to handle fault/fracture tagging more than 255.


palgunadi1993:
patches for SeisSol and PUML?

PR #414, opened 2021-11-05
Dr/tpv5 ci

ravil-mobile:
Hi @sebwolf-de , I fixed some bugs and CI pipeline seems to work (https://gitlab.lrz.de/ci_seissol/SeisSol/-/pipelines/735558). We need to discuss what to do with `cpu_tpv5_test` job. We can either mark it as `allowed to fail` or we can consider individual `eps`'s for each variables e.g., bigger for stress components and smaller for velocities. We can also compute the second error norm for each variable.

 This is my fix for `std::filesystem` and `gcc@8` (https://github.com/SeisSol/SeisSol/blob/dr/tpv5-ci/cmake/filesystem.cmake). I guess it is worth adding it to the master branch. In short, instead of just checking whether a sample source file is be able to be compiled, I just run the executable. If it results in a segfault then I try to another linking option.


sebwolf-de:
I would not mark it as `allowed to fail`. IIRC the gitlab CI is not mandatory for merging. Still we should see, if something is wrong before we merge. Individual epsilons still might be a good idea.
Unfortunately I can't download the artifacts to check, what was the problem for tpv5.

krenzland:
> This is my fix for `std::filesystem` and `gcc@8` (https://github.com/SeisSol/SeisSol/blob/dr/tpv5-ci/cmake/filesystem.cmake). I guess it is worth adding it to the master branch. In short, instead of just checking whether a sample source file is be able to be compiled, I just run the executable. If it results in a segfault then I try to another linking option.

Yes, please. In principle, It is a very strange bug not in gcc-8 but with how ubuntu handles gcc-8. Something similar to https://stackoverflow.com/a/56864842

7bits-register:
Hi, here is a working pipeline https://gitlab.lrz.de/ci_seissol/SeisSol/-/pipelines/737143
It turns out that a job doesn't create artifacts if fails. I split the original job into two: `run` and `check`. Now you can download artifacts from `cpu_tpv5_run`. You can find the "reference" solution (sorry for an incorrect term. Just recorded from the master branch) in `precomputed` sub-directory. 
I guess it is cool because one can easily download CI results and look at them in paraview.

@sebwolf-de, could I ask you to find good `eps`'s for each variable? It is more or less clear with the wave field but less obvious with the fault output

ravil-mobile:
Hi, it turned out that if a job fails then it gitlab-ci doesn't produce any artifacts. I split the job into 2 stages i.e., `run` and `check`. Now you can get the artifacts from `cpu_tpv5_run` job. There is `precomputed` sub-directory which contains a solution recorded from the master branch. I guess it is cool because one can download the artifacts and use paraview to make a visual inspection.

I also change the way how `h5diff` compares the output i.e., now it uses a relative difference. I set it to `0.1` but, of course, we can change it.

sebwolf-de:
Hi, I added a script to compute the (relative) L2 error between two quantities on a fault. Currently, I've set epsilon = 0.05, which is rather high. This is good for our current needs, since small changes don't break the CI, but if something goes _crazily wrong_, we see this. Once we approach a state, where we think about merging into master, we can decrease the epsilon.

Thomas-Ulrich:
I think it would make sense to combine with the script I proposed here: https://github.com/SeisSol/SeisSol/pull/360

PR #415, opened 2021-11-08
Fixed cmake files to handle std::filesystem

ravil-mobile:
Tested with gcc@8, gcc@9 and clang@11

Please, do not directly merge in case of approval. Let me test these changes for `dr/cpp` branch where problems were found.


cmake/FindFILESYSTEM.cmake
```diff
@@ -0,0 +1,55 @@
...
+
+namespace fs = std::filesystem;
+
+int main(int argc, char* argv[]) {
+    fs::path p{argv[0]};
+    std::cout << fs::canonical(p) << std::endl;
+    return 0;
+}")
+
+set(CMAKE_REQUIRED_DEFINITIONS -std=c++17)
```
uphoffc, line 27: That's not so great as some compilers might enable c++17 with other flags.

CMakeLists.txt
```diff
@@ -355,6 +353,8 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel")
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -fopenmp=libomp -Wall -Wextra -pedantic")
 endif()
+
+find_package(FILESYSTEM REQUIRED)
 target_link_libraries(SeisSol-lib PUBLIC ${FILESYSTEM_LIBRARIES})
```
uphoffc, line 358: Is it possible to create a target for filesystem?

uphoffc (changes requested):

cmake/FindFILESYSTEM.cmake
ravil-mobile: Hi, It turns out that there is no way how one can pass `CXX_STANDARD 17` for `check_cxx_source_runs` in a similar way how it is done for `try_compile` or `try_run`. The documentation says that `check_cxx_source_runs` internally calls `try_run`.

My solution is to save  `CXX_STANDARD` at the beginning, set it to 17 for the checks, and restore its original value at the end. 

Changes are pushed

CMakeLists.txt
ravil-mobile: What do you think about this one?
```
if (DEFINED FILESYSTEM_LIBRARIES)
    add_library(std::filesystem INTERFACE IMPORTED)
    set_property(TARGET std::filesystem APPEND PROPERTY INTERFACE_COMPILE_FEATURES cxx_std_17)

    if (FILESYSTEM_LIBRARIES)
        set_property(TARGET std::filesystem APPEND PROPERTY INTERFACE_LINK_LIBRARIES ${FILESYSTEM_LIBRARIES})
    endif()
endif()
```

ravil-mobile: What do you think about this one?

```
if (DEFINED FILESYSTEM_LIBRARIES)
    add_library(std::filesystem INTERFACE IMPORTED)
    set_property(TARGET std::filesystem APPEND PROPERTY INTERFACE_COMPILE_FEATURES cxx_std_17)

    if (FILESYSTEM_LIBRARIES)
        set_property(TARGET std::filesystem APPEND PROPERTY INTERFACE_LINK_LIBRARIES ${FILESYSTEM_LIBRARIES})
    endif()
endif()
```

[source](https://storage.bsc.es/gitlab/hpc/gekkofs/-/blob/c9ce879500713b0fb91cd284844af6bb1fd510c7/CMake/FindFilesystem.cmake)

uphoffc: looks good

cmake/FindFILESYSTEM.cmake
uphoffc: Btw. why not just use try_compile instead of cxx_source...

ravil-mobile: Ahh, that was the problem. It turned out that one can compile the sample program with `gcc@8` without any issue. For example, I checked that `FILESYSTEM_LIBRARIES` in case of `gcc@8` was empty with the original cmake files. But I had a segmentation fault when I tried to run SeisSol. A better solution is to use `try_run` which does the job of `try_compile`, plus it tries to run a sample program.

`check_cxx_source_runs` internally calls `try_run`. Additionally, the macro handles interpreting output as `success` of `failure` as everybody expects (true or false), whereas,  `try_run` just gives you a raw output.
For example, if you write `return 0` then you will write `if (${RESULT} STREQUAL "0")`

uphoffc: Ah ok

ravil-mobile: The drawback is `check_cxx_source_runs` doesn't take `CXX_STANDARD 17` as a parameter in a similar way how `try_run` or `try_compile` does. I had to store the global value of `CXX_STANDARD` at the beginning, set it to `17` for the checks, and then restore this back to its original value at the end

CMakeLists.txt
ravil-mobile: Ok. I tested with gcc@8, gcc@9 and clang@11 with  `std::filesystem` target and it worked for me

uphoffc (approved):

PR #417, opened 2021-11-08
Fix bug in Time.cpp related to receiver

krenzland:
Writing receivers lead to changes in the solution because receivers integrated displacement in time.
(Was correct for all published simulations, broke this before merging)


sebwolf-de (approved):
LGTM

PR #431, opened 2021-11-26
Adjusted OpenMP flags

ravil-mobile:
* bumped min. required CMake version to 3.10
to guarantee CMake support of OpenMP
* removed custom FindOpenMP.cmake from SeisSol
* removed hard-coded openmp compiler and linker flags

Tested with **gcc**, **intel** and **clang** and it worked well for both c++ and fortran compilers


CMakeLists.txt
```diff
@@ -195,9 +199,19 @@ endif()
...
 if (OPENMP)
   find_package(OpenMP REQUIRED)
-  target_link_libraries(SeisSol-lib PUBLIC OpenMP::OpenMP_CXX)
-  set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} ${OpenMP_Fortran_FLAGS}")
-  target_compile_definitions(SeisSol-lib PUBLIC OMP OMPI_SKIP_MPICXX)
+  target_link_libraries(SeisSol-gencode PRIVATE OpenMP::OpenMP_CXX)
+  target_link_libraries(SeisSol-lib PUBLIC OpenMP::OpenMP_Fortran)
+
+  find_package(OpenMP_Offloading QUIET)
+  # Note: there is a conflict between OpenMP offloading and libsxmm
```
krenzland, line 206: Maybe add a comment here why that is

CMakeLists.txt
```diff
@@ -309,19 +323,29 @@ if (ADDRESS_SANITIZER_DEBUG)
   )
 endif()
 
+
+# Note: we need to provide a support for Eigen backend
+#       for both gencode and SeisSol
+target_include_directories(SeisSol-gencode PUBLIC
```
krenzland, line 329: Hm, why is this PUBLIC? Shouldn't this be private?

CMakeLists.txt
```diff
@@ -309,19 +323,29 @@ if (ADDRESS_SANITIZER_DEBUG)
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/submodules/eigen3
+    ${CMAKE_CURRENT_SOURCE_DIR}/submodules/yateto/include
+)
+
+target_include_directories(SeisSol-gencode PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}/submodules/yateto/include
```
krenzland, line 335: Why is this here twice?

cmake/FindOpenMP_Offloading.cmake
```diff
@@ -0,0 +1,75 @@
...
+#  email: ravil.dorozhinskii@tum.de
+#
+
+include(FindPackageHandleStandardArgs)
+include(CheckCXXSourceRuns)
+include(CMakePushCheckState)
+
+cmake_push_check_state()
+
+set(_PARENT_CMAKE_CXX_STANDARD ${CMAKE_CXX_STANDARD})
```
krenzland, line 22: Why is this needed? We are using CMAKE_CXX_STANDARD 17 anyway?

krenzland (changes requested):
Cool! Looks mostly good, some comments

CMakeLists.txt
ravil-mobile: So, as the comment says we need eigen (and yateto) headers for the code generated by `yateto` (in case of Eigen backend) and `SeisSol-lib`. If it is `PRIVATE` then the includes will be applied only for `SeisSol-gencode` target.

CMakeLists.txt
ravil-mobile: you're right. I will remove it

cmake/FindOpenMP_Offloading.cmake
ravil-mobile: I guess I can remove `CMAKE_CXX_STANDARD` from this file. I started with a copy of `FindFILESYSTEM.cmake` file and forgot to remove it. 

In case of `FindFILESYSTEM.cmake`, _in general_, one doesn't know whether  `CMAKE_CXX_STANDARD 17` has already been set. It was just a generalization in case of somebody decides to re-use the file for other sub-projects.

CMakeLists.txt
ravil-mobile: Added
```
  # Note: there may be a conflict between OpenMP offloading and libsxmm.
  # A collision can happen due to -march flags for CPUs and OpenMP GPU target
  # (e.g. with clang -Xopenmp-target -march=sm_75). Therefore, we separate
  # compilations for the source and generated codes and apply different
  # imported OpenMP targets to them.
```

CMakeLists.txt
ravil-mobile: Yes, we are using. I mean the path to eigen will be applied to `SeisSol-gencode` and everything which depends on `SeisSol-gencode`. In our case, it is `SeisSol-lib` and so on (SeisSol-bin, SeisSol-proxy, tests, etc.).

krenzland (approved):

PR #432, opened 2021-11-29
Change testing framework to doctest & add code coverage

krenzland:
This PR changes all tests to use the testing framework [doctest](https://github.com/onqtam/doctest) because cxxtest is terrible and there isn't any support.

Furthermore, I've added code coverage, which can be visualized [here](https://app.codecov.io/gh/SeisSol/SeisSol). This is updated automatically by the CI.

Note that I've deleted a few tests but they have not been included for a long time (> 2 years). They also do not work anymore. I've deleted only one test that technically worked which tried reading a Gambit mesh in parallel. As we do not use/support this feature I've deleted the test.

@uphoffc I've deleted some license headers from the tests because they were not consistently used. I can re-add them if you wish but license-wise this is fine.


ravil-mobile:
In general, it looks good. Did you copied `external/doctest.h ` to the source code file tree?

krenzland:
Yup, that's the easiest way. Easier than adding an entire submodule for it imo

ravil-mobile (approved):
Looks good for me

PR #435, opened 2021-12-01
removed warnings of clang@11 compiler

ravil-mobile:
* added submodules/async via system system
* added -Wno-unused-parameter flag for clang

### DIFF
--- a/src/Initializer/MemoryAllocator.cpp
+++ b/src/Initializer/MemoryAllocator.cpp
@@ -79,13 +79,13 @@
 
 void* seissol::memory::allocate(size_t i_size, size_t i_alignment, enum Memkind i_memkind)
 {
-    void* l_ptrBuffer;
+    void* l_ptrBuffer{nullptr};
     bool error = false;
 
     /* handle zero allocation */
     if ( i_size == 0 ) {
-      //logWarning() << "allocation of size 0 requested, returning NULL; (alignment: " << i_alignment << ", memkind: " << i_memkind << ").";
-      l_ptrBuffer = NULL;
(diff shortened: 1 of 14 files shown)


ravil-mobile:
@krenzland , can you have a look at ?
```
ODEVector.h:86:14: warning: definition of implicit copy constructor for 'ODEVector' is deprecated because it has a user-provided copy assignment operator
```

src/Initializer/MemoryAllocator.cpp
```diff
@@ -79,7 +79,7 @@
 
 void* seissol::memory::allocate(size_t i_size, size_t i_alignment, enum Memkind i_memkind)
 {
-    void* l_ptrBuffer;
+    void* l_ptrBuffer{};
```
sebwolf-de, line 82: Maybe `=nullptr` would be more verbose?

krenzland: Imo that is clear enough, nulptr is obviously the default value

ravil-mobile:
Hi, I decided to remove all `NULL` from `seissol::memory::allocate` replace them with `nullptr`.

krenzland:
> @krenzland , can you have a look at ?
> 
> ```
> ODEVector.h:86:14: warning: definition of implicit copy constructor for 'ODEVector' is deprecated because it has a user-provided copy assignment operator
> ```

Hm, it probably means that we need to define a copy constructor manually for this class?

ravil-mobile:
Removed `ODEVector.h:86:14: warning: definition of implicit copy constructor for 'ODEVector' is deprecated because it has a user-provided copy assignment operator`

@krenzland, I just added
```
ODEVector(const ODEVector&) = default;
```
because all methods attributes have their defined copy-constructors

sebwolf-de (approved):

PR #436, opened 2021-12-02
remove yaml-cpp submodule

Thomas-Ulrich:
Hi,
I was trying to install some easi related scripts, and came to the idea that we could remove the yaml-cpp and ImpalaJIT (and maybe easi) submodule and download them with cmake.
Here is how we could do that for the yaml-cpp module.
I think the only problem would be if you are behind a firewall and you did not setup a ssh connection to git.
(any idea on how to fix it?).

Let me know what you think.

Thomas.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,9 +19,6 @@
 [submodule "submodules/easi"]
 	path = submodules/easi
 	url = ../../SeisSol/easi.git
-[submodule "submodules/yaml-cpp"]
-	path = submodules/yaml-cpp
-	url = ../../jbeder/yaml-cpp.git
 [submodule "submodules/ImpalaJIT"]
 	path = submodules/ImpalaJIT
 	url = ../../uphoffc/ImpalaJIT.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,17 @@ if(NOT CMAKE_BUILD_TYPE)
   message(STATUS "Set build type to Release as none was supplied.")
 endif()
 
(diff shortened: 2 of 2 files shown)


sebwolf-de:
I do not see, what is the benefit, we move complexity from git to cmake. By the way, there is a yaml-cpp module on supermuc and lot's of distros have yaml-cpp support. At the same time yaml is not performance critical. We could remove this completely from SeisSol and just require yaml-cpp to be available externally (like we already do with hdf5 e.g.)

Thomas-Ulrich:
Ok, thanks that's true. Actually, I think that yaml and IMPALAJIT installation could be moved to seissol/easi, as being not direct depencies of SeisSol.
@uphoffc ?

uphoffc:
Yes, yaml-cpp and impalajit are dependencies of easi, not of SeisSol.

Including dependencies as submodules is anyhow not good practice imho. I think it only makes sense if a project and its submodule are very tightly coupled. I originally included yaml-cpp as submodule for simplicity to encourage the adoption of easi :-)

But now easi has proper installation support, and also automatically finds yaml-cpp and impalajit once they are installed. Thus, I'd propose that one should install yaml-cpp / easi / impalajit using the standard cmake / make / make install procedure.

krenzland:
Yup, Impalajit should also be an optional dependency, because it is no longer supported by us and doesn't work on non x86 platform. I'll add Lua support to easi relatively soon because this works on basically all platforms and ImpalaJIT has no practical advantage. (Should of course still be left in the code for legacy reasons)

sebwolf-de:
This PR is now redundant, because of #454.

PR #438, opened 2021-12-09
Add documentation on how to use pip on supermuc

uphoffc:


git-taufiq (changes requested):
I think this looks good, just a typo in the Host name in line 212. It should be supermucNG to be consistent with a change in Host name in line 19.

git-taufiq (approved):

PR #440, opened 2021-12-10
change variable names of free surface output

Thomas-Ulrich:
Calling surface velocity uvw and displacements UVW might pose a problem for case insensitive operating systems.
(Windows, and maybe Mac. Note that Bo and Duo do not have problem with case sensitivity on their mac).


krenzland (approved):
:+1:

PR #451, opened 2022-01-17
Remove #pragma message

krenzland:
They just clutter the output

### DIFF
--- a/src/Equations/elastic/Kernels/Local.cpp
+++ b/src/Equations/elastic/Kernels/Local.cpp
@@ -41,10 +41,6 @@
 
 #include "Kernels/Local.h"
 
-#ifndef NDEBUG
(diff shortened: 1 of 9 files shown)


src/Equations/elastic/Kernels/Local.cpp
```diff
@@ -42,7 +42,6 @@
 #include "Kernels/Local.h"
 
 #ifndef NDEBUG
-#pragma message "compiling local kernel with assertions"
 #endif
```
ravil-mobile, line 45: I guess `#ifndef` and `#endif` can be also removed, right?

krenzland: Yup, forgot that one

ravil-mobile (approved):
Looks good

PR #453, opened 2022-01-18
update coastline script

Thomas-Ulrich:
Updates on the script that allows outputting the coastline in ParaView.
(e.g. I needed the geocent projection).


sebwolf-de (approved):
LGTM

PR #454, opened 2022-01-19
Update easi

krenzland:
Can only be merged after: https://github.com/SeisSol/easi/pull/18

I will write documentation and fix the CI soon (and then convert to a standard PR).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,56 +158,7 @@ endif()
 # Libs
 include(ExternalProject)
 
-find_package(YAML-CPP 0.5.3 QUIET)
-if (YAML-CPP_FOUND)
-
-else()
-  set(YAML_CPP_INSTALL_DIR  ${CMAKE_CURRENT_BINARY_DIR}/yaml-cpp-install)
-  ExternalProject_Add(yaml-cpp-build
-    SOURCE_DIR ${CMAKE_CURRENT_LIST_DIR}/submodules/yaml-cpp/
-    INSTALL_DIR  ${YAML_CPP_INSTALL_DIR}
-    PREFIX ${CMAKE_CURRENT_BINARY_DIR}/yaml-cpp
-    CMAKE_ARGS -DCMAKE_BUILD_TYPE=Release
-    -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
-    -DYAML_CPP_BUILD_TOOLS=OFF
-    -DYAML_CPP_BUILD_TESTS=OFF
-    -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
-    -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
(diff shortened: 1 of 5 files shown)


krenzland:
For a complete documentation, https://github.com/SeisSol/easi/pull/19 needs to be merged first.

Documentation/compilation.rst
```diff
@@ -143,6 +143,10 @@ See section :ref:`Installing ASAGI <installing_ASAGI>`.
 
 .. _compiling-seissol:
 
+Installing easi
+---------------------------
+Follow the `installation instructions <https://easyinit.readthedocs.io/en/latest/index.html>`.
```
Thomas-Ulrich, line 148: There is no installation instruction there.
It also says:
>easi is a header-only library written in C++11. Hence, one simply needs (Only unit tests must be compiled with CMake.)
Maybe you want to copy-paste some content from here:
https://gitlab.lrz.de/thomas.ulrich/TuSeisSolScripts/-/blob/master/EasiRelated/README.md

Thomas-Ulrich (approved):
Nice, but I think we need to better document the installation of easi in the readthedocs of easi.

Documentation/compilation.rst
krenzland: Hm, the corresponding PR is already merged but it doesn't seem to be building the documentation.

Thomas-Ulrich: ok, I see.
But don't you think that all users can figure out things like:
```
git clone git@github.com:uphoffc/ImpalaJIT
cd ImpalaJIT 
mkdir build && cd build
CC=mpicc CXX=mpiCC FC=mpif90 cmake ..  -DCMAKE_INSTALL_PREFIX=$HOME
make -j 4
make install
cd ../..
```
(I'm not sure, and detailing may prevent having to give further explanation).

krenzland: It is there, but you need to delete your cache - ctrl+F5 for most browsers.

CMakeLists.txt
```diff
@@ -269,7 +230,7 @@ endif()
 find_package(PkgConfig REQUIRED) 
 if (ASAGI)
   # todo warn if netcdf is off
-  pkg_check_modules(ASAGI REQUIRED asagi) # asagi_nompi?
+  pkg_check_modules(ASAGI REQUIRED IMPORTED_TARGET asagi) # asagi_nompi?
   target_compile_definitions(SeisSol-lib PUBLIC USE_ASAGI)
   target_link_libraries(SeisSol-lib PUBLIC ${ASAGI_STATIC_LDFLAGS})
   target_include_directories(SeisSol-lib PUBLIC ${ASAGI_INCLUDE_DIRS})
```
sebwolf-de, line 236: Is there still code, which uses ASAGI directly in SeisSol? I think everything should be behind easi now?

Thomas-Ulrich: I think so too.

krenzland: AsagiModule.h for MPI init. Could be at some point moved to easi I guess

PR #456, opened 2022-01-20
fix minor issue on LSW forced rupture

Thomas-Ulrich:
the strength decrease during nucleation should account for the substep time increment.


sebwolf-de (approved):
LGTM

PR #457, opened 2022-01-20
Fix name of vertexWeightFreeSurfaceWithGravity

krenzland:
See: https://github.com/SeisSol/SeisSol/blob/master/src/Reader/readpar.f90#L2218
Sorry!


sebwolf-de (approved):
LGTM

PR #459, opened 2022-01-24
Draft: clang-tidy for dr/cpp

sebwolf-de:
Let's use clang-tidy to improve the code quality on our dr development branch.
Merge strategy: 
- [ ] Agree on `.clang-tidy` content
- [ ] Add changes, such that the clang-tidy pipeline passes
- [ ] Squash all changes into 2 commits, one for `.clang-tidy` one for all the other commits
- [ ] Add the hash of the last commit to `.git-blame-ignore-revs`


krenzland:
Is there a reason why you do not include clang-tidy in CMake directly? There is direct support for it:
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_CLANG_TIDY.html

Sounds a bit easier than your approach

.clang-tidy
```diff
@@ -0,0 +1,299 @@
...
+#  - key:             readability-identifier-naming.ConstexprVariableSuffix
+#    value:           CamelCase
+  - key:             readability-identifier-naming.EnumCase
+    value:           CamelCase
+#  - key:             readability-identifier-naming.EnumPrefix
+#    value:           CamelCase
+#  - key:             readability-identifier-naming.EnumSuffix
+#    value:           CamelCase
+  - key:             readability-identifier-naming.EnumConstantCase
+    value:           lower_case
```
krenzland, line 83: This is different to current practice

sebwolf-de:
I tried the cmake approach. But it was actually harder. I wanted to restrict clang-tidy checks to the dynamic rupture subdirectory. Cmake can only do this, on the scope of targets. So I would need to split SeisSol into different targets (like e.g. `seissol-dr-lib`) and then only activate clang-tidy for this target. The python runner was way easier then.

krenzland:
Ah, okay. Using clang-tidy for everything would be quite smart but is probably a nightmare...

sebwolf-de:
We could easily run clang-tidy everywhere, but I guess we would be overwhelmed with issues...

.ci/run-clang-tidy.py
```diff
@@ -0,0 +1,327 @@
+#!/usr/bin/env python
```
ravil-mobile, line 1: Am I correct that this file was taken directly from LLVM? Does it really only work with python2 or it is possible to change it to pyhton3?

ravil-mobile: Sorry, I didn't notice
```
python3 ../.ci/run-clang-tidy.py -quiet -header-filter=../src/DynamicRupture/*
```

PR #467, opened 2022-01-24
Added async copies from GPU-to-CPU for point sources

ravil-mobile:
It is an attempt to improve data transfers from GPUs to CPUs while evaluating point sources. The implementation uses GPU streams and OpenMP tasks. In stead of waiting fault pages on CPUs which are currently on GPUs. We can copy only necessary data avoiding the page fault mechanism.

This needs to be tests with a scenario like 1994 Northridge.


ravil-mobile:
The PR doesn't make sense anymore b/c the kinematic point sources were ported to GPUs

PR #469, opened 2022-01-28
Fix link to easy install

krenzland:
1) Link was formatted incorrectly
2) Now link directly to installation section


Thomas-Ulrich:
Don't you think it is better to link to dependencies rather than compilation.

Thomas-Ulrich (approved):
LGTM

PR #475, opened 2022-02-07
Removed large stack allocations from dr/cpp

ravil-mobile:
Removed large stack allocations from dr/cpp

### DIFF
--- a/src/DynamicRupture/FrictionLaws/BaseFrictionLaw.h
+++ b/src/DynamicRupture/FrictionLaws/BaseFrictionLaw.h
@@ -43,6 +43,8 @@ class BaseFrictionLaw : public FrictionSolver {
   real (*dynStressTime)[misc::numPaddedPoints];
   bool (*dynStressTimePending)[misc::numPaddedPoints];
 
+  real (*qInterpolatedPlus)[CONVERGENCE_ORDER][tensor::QInterpolated::size()];
+  real (*qInterpolatedMinus)[CONVERGENCE_ORDER][tensor::QInterpolated::size()];
   /**
    * copies all parameters from the DynamicRupture LTS to the local attributes
    */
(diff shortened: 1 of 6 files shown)


Thomas-Ulrich (commented):
Hi,
Do I understand correctly that the main change of this PR is
```
  alignas(ALIGNMENT) real QInterpolatedPlus[layerData.getNumberOfCells()][CONVERGENCE_ORDER][tensor::QInterpolated::size()];
  ```
  by 
  ```
  auto *qInterpolatedPlus = layerData.var(m_dynRup->qInterpolatedPlus);
  ```
 Why does it require less memory allocated?.
  
  Also I m curious why this code was not linted by Sebastian previous PR?

**Comment by krenzland** [comment]
What exactly is the advantage of this? I don't think stack allocations actually make much of a difference on CPUs

**Comment by ravil-mobile** [comment]
> Why does it require less memory allocated?.

Hi Thomas,
I removed the allocation of `QInterpolatedPlus` from the stack and now it leaves on the heap.

> What exactly is the advantage of this? I don't think stack allocations actually make much of a difference on CPUs

Hi Lukas,

There is none from a perspective of performance. We removed a compiler warning which was complaining about `Variable Length Arrays`, caused by

```
[layerData.getNumberOfCells()]
```
C++ standard doesn’t support variable sized arrays. However, some compilers allows it but issue a warning which can be switched off. Anyway, setting `ulimit -Ss unlimited` was really necessary for `dr/cpp` because `QInterpolatedPlus` and `QInterpolatedMinus` were quite huge for the stack.

We came to this because we split the original loop:
```
for (unsigned face = 0; face < layerData.getNumberOfCells(); ++face) {
    unsigned prefetchFace = (face < layerData.getNumberOfCells()-1) ? face+1 : face;
    m_dynamicRuptureKernel.spaceTimeInterpolation(...);

    e_interoperability.evaluateFrictionLaw(...);
  }
```

to this one:
```
  for (unsigned face = 0; face < layerData.getNumberOfCells(); ++face) {
    unsigned prefetchFace = (face < layerData.getNumberOfCells()-1) ? face+1 : face;
    m_dynamicRuptureKernel.spaceTimeInterpolation(...);
  }

  m_FrictionSolver->evaluate(...);
```

krenzland:
Ah, yes, the variable-length stack allocations are really a problem - in this case, it makes perfect sense!

Thomas-Ulrich (approved):
Thank you

sebwolf-de (approved):

PR #476, opened 2022-02-07
Ravil/gitlab ci

ravil-mobile:


gitlab-gpu-pipeline.yml
```diff
@@ -0,0 +1,160 @@
...
+        - git clone https://github.com/uphoffc/ImpalaJIT.git impalajit &&
+          mkdir -p impalajit/build && cd impalajit/build &&
+          cmake .. -DCMAKE_INSTALL_PREFIX=../../self_usr -DCMAKE_CXX_FLAGS="-fPIC" &&
+          make -j $(nproc) install &&
+          cd ../..
+        - git clone --depth 1 --branch yaml-cpp-0.6.2 https://github.com/jbeder/yaml-cpp.git yaml-cpp &&
+          mkdir -p yaml-cpp/build && cd yaml-cpp/build &&
+          cmake .. -DCMAKE_INSTALL_PREFIX=../../self_usr -DCMAKE_CXX_FLAGS="-fPIC" &&
+          make -j $(nproc) install &&
+          cd ../..
```
krenzland, line 37: Why don't you just install the module from the package repo?

gitlab-cpu-pipeline.yml
```diff
@@ -0,0 +1,121 @@
...
+        - git submodule set-url submodules/cxxtest https://github.com/CxxTest/cxxtest.git
+        - git submodule set-url submodules/utils https://github.com/TUM-I5/utils.git
+        - git submodule set-url submodules/async https://github.com/TUM-I5/ASYNC.git
+        - git submodule set-url submodules/scons-tools https://github.com/TUM-I5/scons-tools.git
+        - git submodule set-url submodules/PUML https://github.com/TUM-I5/PUML2.git
+        - git submodule set-url submodules/yateto https://github.com/SeisSol/yateto.git
+        - git submodule set-url submodules/eigen3 https://github.com/eigenteam/eigen-git-mirror
+        - git submodule set-url submodules/Device https://github.com/SeisSol/Device.git
+        - git submodule --quiet update --init --recursive 
+        - wget -q https://syncandshare.lrz.de/dl/fiJNAokgbe2vNU66Ru17DAjT/netcdf-4.6.1.tar.gz
```
krenzland, line 29: By the way, is this link actually stable? Most syncandshare links only work for a few months now

krenzland (changes requested):

gitlab-gpu-pipeline.yml
ravil-mobile: Each job takes a fresh image. It turns out to be the default gitlab-ci policy. I specified a default image for the entire GPU pipeline hoping that gitlab-ci is going to re-use a previously used container but no. In short gitlab-ci doesn't hold an container state between jobs.

`yaml-cpp-0.6.2` is pretty stable package I can move the installation of it to the Dockerfile. What do you think?

gitlab-cpu-pipeline.yml
ravil-mobile: I don't know who added this and why it was done in this way. I guess @VMPW did it. It is better to ask her

gitlab-gpu-pipeline.yml
krenzland: Yes, that'd be better in my opinion. As long as it it 0.6.x it's going to work -0.7.x might break.

gitlab-cpu-pipeline.yml
krenzland: Let's just keep it like this, it's an easy fix in the worst case

gitlab-gpu-pipeline.yml
ravil-mobile: Ok. moved yaml-cpp to the docker test image

krenzland (approved):

PR #479, opened 2022-02-11
fix uninitialized ThermalPress variable

Thomas-Ulrich:
as noticed by Ravil


ravil-mobile:
Hi Thomas,

I solved this issue in a different way: 
https://github.com/SeisSol/SeisSol/blob/56841c201c4543aa3ab54b0e29de9daa288cf913/src/Numerical_aux/typesdef.f90#L691

Thomas-Ulrich:
ok fine!

PR #480, opened 2022-02-11
Add support for LibXSMM JIT

krenzland:
Undocumented because it's in beta and does not (yet) offer any advantages over old interface.
To use, just replace LIBXSMM with LIBXSMM_JIT

Draft because https://github.com/SeisSol/yateto/pull/38 has to be merged before this.


ravil-mobile:
Hi, is there a way to restrict version of `libsxmm`  to 1.17?

krenzland:
I don't see an elegant way. Maybe by using the defines from  libxsmm_version.h

```c++
#ifndef LIBXSMM_VERSION_H
#define LIBXSMM_VERSION_H

#define LIBXSMM_CONFIG_VERSION "1.17-2114"
#define LIBXSMM_CONFIG_BRANCH "master"
#define LIBXSMM_CONFIG_VERSION_MAJOR 1
#define LIBXSMM_CONFIG_VERSION_MINOR 17
#define LIBXSMM_CONFIG_VERSION_UPDATE 0
#define LIBXSMM_CONFIG_VERSION_PATCH 2114
#define LIBXSMM_CONFIG_BUILD_DATE 20220127

#endif
```

Do you have a better idea? We do not check the version for the generator either...

cmake/process_users_input.cmake
```diff
@@ -87,7 +87,8 @@ set_property(CACHE LOG_LEVEL_MASTER PROPERTY STRINGS ${LOG_LEVEL_MASTER_OPTIONS}
 
 
 set(GEMM_TOOLS_LIST "auto" CACHE STRING "choose a gemm tool(s) for the code generator")
-set(GEMM_TOOLS_OPTIONS "auto" "LIBXSMM,PSpaMM" "LIBXSMM" "MKL" "OpenBLAS" "BLIS" "PSpaMM" "Eigen" "LIBXSMM,PSpaMM,GemmForge" "Eigen,GemmForge")
+set(GEMM_TOOLS_OPTIONS "auto" "LIBXSMM,PSpaMM" "LIBXSMM" "MKL" "OpenBLAS" "BLIS" "PSpaMM" "Eigen" "LIBXSMM,PSpaMM,GemmForge" "Eigen,GemmForge"
+        "LIBXSMM_JIT,PSpaMM" "LIBXSMM_JIT" "MKL" "OpenBLAS" "BLIS" "PSpaMM" "Eigen" "LIBXSMM_JIT,PSpaMM,GemmForge" "Eigen,GemmForge" )
```
sebwolf-de, line 91: There are duplicate values.

sebwolf-de (commented):
How is the performance of the JIT version in comparison to the generator?

krenzland:
> How is the performance of the JIT version in comparison to the generator?

Typically about the same. Sometimes faster because it supports the case that B is transposed. Feel free to check for your own use case :)

sebwolf-de (approved):

PR #486, opened 2022-02-15
Optimize computation of gravitational boundary condition

krenzland:
Old way: Use RK to compute the ODE that defines the displacement
New way: Use ADER-magic to compute the Taylor series of the displacement directly and evaluate it at end of timestep and compute integral of displacement. Does this in-place without storing all coefficients. A lot faster, and of arbitrary order.

Draft because: Needs to be tested for realistic application, flop counter needs to be checked (@sebwolf-de if you're bored ;)).

I've already verified that it works for convergence tests. I get the exact same results but faster.

### DIFF
--- a/src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h
+++ b/src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h
@@ -17,21 +17,30 @@ double getGravitationalAcceleration();
 
 class GravitationalFreeSurfaceBc {
 public:
-  GravitationalFreeSurfaceBc() { }
+  GravitationalFreeSurfaceBc() = default;
 
   template <typename TimeKrnl>
   static std::pair<long long, long long> getFlopsDisplacementFace(unsigned face,
                                                                   [[maybe_unused]] FaceType faceType,
                                                                   TimeKrnl& timeKrnl) {
-    const auto numDofs = init::faceDisplacement::size() + init::averageNormalDisplacement::size();
     long long hardwareFlops = 0;
     long long nonZeroFlops = 0;
 
     // Note: This neglects roughly 10 * CONVERGENCE_ORDER * numNodes2D flops
     // Before adjusting the range of the loop, check range of loop in computation!
     for (int order = 1; order < CONVERGENCE_ORDER + 1; ++order) {
-      nonZeroFlops += kernel::projectDerivativeToNodalBoundaryRotated::nonZeroFlops(order - 1, face);
-      hardwareFlops += kernel::projectDerivativeToNodalBoundaryRotated::hardwareFlops(order - 1, face);
+#ifdef USE_ELASTIC
+      constexpr auto flopsPerQuadpoint =
+          4 + // Computing coefficient
+          6 + // Updating displacement
+          2; // Updating integral of displacement
+#else
+      constexpr auto flopsPerQuadpoint = 0;
+#endif
+      constexpr auto flopsUpdates = flopsPerQuadpoint * flopsPerQuadpoint;
+      
+      nonZeroFlops += kernel::projectDerivativeToNodalBoundaryRotated::nonZeroFlops(order - 1, face) + flopsUpdates;
+      hardwareFlops += kernel::projectDerivativeToNodalBoundaryRotated::hardwareFlops(order - 1, face) + flopsUpdates;
     }
 
     return {nonZeroFlops, hardwareFlops};
@@ -60,7 +69,7 @@ public:
     // This implementation sums up the Taylor series directly without storing
     // all coefficients.
 
-   // Prepare kernel that projects volume data to face and rotates it to face-nodal basis.
+    // Prepare kernel that projects volume data to face and rotates it to face-nodal basis.
     assert(boundaryMapping.nodes != nullptr);
     assert(boundaryMapping.TinvData != nullptr);
     assert(boundaryMapping.TData != nullptr);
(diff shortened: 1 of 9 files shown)


src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h
```diff
@@ -127,103 +112,82 @@ class GravitationalFreeSurfaceBc {
...
+    alignas(ALIGNMENT) std::array<real, nodal::tensor::nodes2D::Shape[0]> prevCoefficients;
+
+    const double deltaT = timeStepWidth;
+    const double deltaTInt = timeStepWidth;
+
+    // Initialize first component of Taylor series
+    for (unsigned int i = 0; i < nodal::tensor::nodes2D::Shape[0]; ++i) {
+      prevCoefficients[i] = rotatedFaceDisplacement(i, 0);
+      // This is clearly a zeroth order approximation of the integral!
+      integratedDisplacementNodal(i) = deltaTInt * rotatedFaceDisplacement(i, 0);
```
sebwolf-de, line 125: Did you count this flop?

src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h
```diff
@@ -127,103 +112,82 @@ class GravitationalFreeSurfaceBc {
...
+    projectKernel.INodal = dofsFaceNodal.data();
+    for (unsigned i = 0; i < yateto::numFamilyMembers<tensor::dQ>(); ++i) {
+      projectKernel.dQ(i) = derivatives + derivativesOffsets[i];
+    }
+
+#ifdef USE_ELASTIC
+    const double rho = materialData.local.rho;
+    const double g = getGravitationalAcceleration(); // [m/s^2]
+    const double Z = std::sqrt(materialData.local.lambda * rho) ;
+#endif
```
sebwolf-de, line 142: What happens else? Should abort, right?

src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h
```diff
@@ -127,103 +112,82 @@ class GravitationalFreeSurfaceBc {
...
-
-    auto curValue = ode::ODEVector{{integratedDisplacementNodal, rotatedFaceDisplacement},
-                                   {integratedEtaSize,           etaSize}};
 
-    // Apply initial condition to integrated displacement (start from 0 each PDE timestep)
-    std::fill_n(integratedDisplacementNodal, integratedEtaSize, 0.0);
+#ifdef USE_ELASTIC
+        const double curCoeff = uInside - (1.0/Z) * (rho * g * prevCoefficients[i] + pressureInside);
+        // Basically uInside - C_1 * (c_2 * prevCoeff[i] + pressureInside)
+        // 2 add, 2 mul = 4 flops
```
sebwolf-de, line 168: Here you rely on the compiler to precompute `1/Z` and `rho*g`, otherwise I count 6 flops.

src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h
```diff
@@ -127,103 +112,82 @@ class GravitationalFreeSurfaceBc {
...
+#else
+        const double curCoeff = uInside;
+#endif
+        prevCoefficients[i] = curCoeff;
 
-    // Setup ODE solver
-    ode::TimeSpan timeSpan = {startTime, startTime + timeStepWidth};
-    auto odeSolverConfig = ode::ODESolverConfig(timeStepWidth);
-    odeSolverConfig.initialDt = timeStepWidth;
+        // 2 * 3 = 6 flops for updating displacement
```
sebwolf-de, line 174: Correct!

src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h
```diff
@@ -127,103 +112,82 @@ class GravitationalFreeSurfaceBc {
...
+        rotatedFaceDisplacement(i, 0) += factorEvaluated * curCoeff;
+        rotatedFaceDisplacement(i, 1) += factorEvaluated * vInside;
+        rotatedFaceDisplacement(i, 2) += factorEvaluated * wInside;
 
-    odeSolver.setConfig(odeSolverConfig);
-    odeSolver.solve(f, curValue, timeSpan);
+        // 2 flops for updating integral of displacement
+        integratedDisplacementNodal(i) += factorInt * curCoeff;
```
sebwolf-de, line 180: Correct!

sebwolf-de (commented):
Hi looks good overall, I mostly focused on the flop counters.

src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h
krenzland: Yes, but the compiler figures this out. At least it should.

src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h
krenzland: It is a standard free surface boundary condition then - this is just s.t. it compiles when using other equations. I'll remove that at some point, but for now I blame our code structure.

krenzland:
I've compared the results for a small version of our Palu scenario; they look absolutely identical (in the eyeball norm). This, together with the working convergence tests means that the new implementation is working!
@sebwolf-de I've adjusted the flop counter (by addressing your comments mostly!). Could you have another look please?

sebwolf-de (approved):
LGTM

Issue #488, opened 2022-02-16
dr/cpp output


sebwolf-de:
**Describe the bug**
The output of dr/cpp branch, does not fit the master branch. Only the quantity `Ts0` is different. Therefore, I suspect that the computations are done correctly, we just miss the nucleation stress in the output.

**Expected behavior**
Output of dr/cpp branc and master branch should be the same.

**To Reproduce**
1. master: e4118559 (not so current, but latest master should not make a difference) dr/cpp: e4c45ff0
2. Which build settings do you use? Which compiler version do you use? 
Not relevant
3. On which machine does your problem occur? If on a cluster: Which modules are loaded?
Not relevant
4. Provide parameter/material files.
tpv104 from https://github.com/7bits-register/precomputed-seissol

**Screenshots/Console output**
dr/cpp: `Ts0` after 2s
![dr-cpp](https://user-images.githubusercontent.com/24592027/154255513-52d900e6-0d3b-44a6-aa04-52531c7af579.png)
master: `Ts0` after 2s
![master](https://user-images.githubusercontent.com/24592027/154255521-9094b60b-c556-4339-a22a-16517bf6c5bc.png)

**Additional context**
As the problem occurs only for `Ts0` and not `T_s` or `Td0`b, I suspect that the nucleation stress is the problem. Also the diff of both outputs, has the same shape as the nucleation function. Note: the relative error is too small to be detected by the pipeline.

Thomas-Ulrich:
It is indeed hardcoded here in the master branch:
https://github.com/SeisSol/SeisSol/blob/master/src/Physics/Evaluate_friction_law.f90#L884

PR #489, opened 2022-02-16
dr/fix output

sebwolf-de:
This commit should fix issue #488, using the fortran output. Note that in the cpp implementation, we accumulate the initial stress and the nucleationstress together. Still have to check the cpp output.


sebwolf-de:
Wrong base branch

PR #490, opened 2022-02-16
Fix #488 in fortran output

### LINKED ISSUE #488: dr/cpp output

sebwolf-de:
This commit fixes issue https://github.com/SeisSol/SeisSol/issues/488, using the fortran output. Note that in the cpp implementation, we accumulate the initial stress and the nucleationstress together. Still have to check the cpp output.

### DIFF
--- a/src/DynamicRupture/Output/Base.hpp
+++ b/src/DynamicRupture/Output/Base.hpp
@@ -28,16 +28,22 @@ class Base {
     }
 
     bool bothEnabled = generalParams.outputPointType == OutputType::AtPickpointAndElementwise;
-    if (generalParams.outputPointType == OutputType::AtPickpoint || bothEnabled) {
+    bool pointEnabled = generalParams.outputPointType == OutputType::AtPickpoint || bothEnabled;
+    bool elementwiseEnabled = generalParams.outputPointType == OutputType::Elementwise || bothEnabled;
+    if (pointEnabled) {
+      logInfo() << "Enable Dynamic Rupture Receiver Output";
       ppOutputBuilder = std::make_unique<PickPointBuilder>();
       ppOutputBuilder->setMeshReader(&mesher);
       ppOutputBuilder->setParams(reader.getPickPointParams());
-    } else if (generalParams.outputPointType == OutputType::Elementwise || bothEnabled) {
+    }
+    if (elementwiseEnabled) {
+      logInfo() << "Enable Dynamic Rupture 2D Fault Output";
       ewOutputBuilder = std::make_unique<ElementWiseBuilder>();
       ewOutputBuilder->setMeshReader(&mesher);
       ewOutputBuilder->setParams(reader.getElementwiseFaultParams());
-    } else {
-      logError() << "Unknown fault output type (not 3,4,5)";
+    }
+    if (!elementwiseEnabled && !pointEnabled) {
+      logInfo() << "No Dynamic Rupture Output enabled";
     }
   }
   void setDrData(seissol::initializers::LTSTree* userDrTree,
--- a/src/DynamicRupture/Output/Base.hpp
+++ b/src/DynamicRupture/Output/Base.hpp
@@ -29,21 +29,22 @@ class Base {
 
     bool bothEnabled = generalParams.outputPointType == OutputType::AtPickpointAndElementwise;
     bool pointEnabled = generalParams.outputPointType == OutputType::AtPickpoint || bothEnabled;
-    bool elementwiseEnabled = generalParams.outputPointType == OutputType::Elementwise || bothEnabled;
+    bool elementwiseEnabled =
+        generalParams.outputPointType == OutputType::Elementwise || bothEnabled;
     if (pointEnabled) {
-      logInfo() << "Enable Dynamic Rupture Receiver Output";
+      logInfo() << "Enabling on-fault receiver output";
       ppOutputBuilder = std::make_unique<PickPointBuilder>();
       ppOutputBuilder->setMeshReader(&mesher);
       ppOutputBuilder->setParams(reader.getPickPointParams());
     }
     if (elementwiseEnabled) {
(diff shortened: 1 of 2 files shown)


src/ResultWriter/faultoutput.f90
```diff
@@ -555,7 +555,7 @@ SUBROUTINE calc_FaultOutput( DynRup_output, DISC, EQN, MESH, MaterialVal, BND, t
 
               !add transient nucleation stress to fault output
               IF ((EQN%FL.EQ.103) .OR. (EQN%FL.EQ.3) .OR. (EQN%FL.EQ.4)) THEN
```
Thomas-Ulrich, line 557: Then if I understand correctly, you need to remove this if statement?
Else you won't see the initial stress on the fault output with other friction laws?

sebwolf-de: You're right. I've removed it and checked tpv5 and 104, both work.

src/ResultWriter/faultoutput.f90
```diff
@@ -554,21 +554,17 @@ SUBROUTINE calc_FaultOutput( DynRup_output, DISC, EQN, MESH, MaterialVal, BND, t
...
-                 Tnuc = DISC%DynRup%t_0
-                 Gnuc = Calc_SmoothStep(time, Tnuc)
-                 NucleationStressLocalCS = MATMUL(rotmat, NucleationStressXYZ)*Gnuc
-              ELSE
-                 NucleationStressLocalCS(:) = 0d0
-              ENDIF
+              NucleationStressXYZ = MATMUL(T(1:6,1:6), EQN%InitialStressInFaultCS(iBndGP,:,iFace))
+              Tnuc = DISC%DynRup%t_0
+              Gnuc = Calc_SmoothStep(time, Tnuc)
+              NucleationStressLocalCS = MATMUL(rotmat, NucleationStressXYZ)*Gnuc
```
Thomas-Ulrich, line 560: hum, I m not sure this works.
stress = transient + initial stress + Gnuc* nucleation_overstress.
but not stress = transient + Gnuc* (initial stress + nucleation_overstress).
So maybe, the idea of merging initial stress and nucleation overstress does not work?

src/DynamicRupture/Output/Base.hpp
```diff
@@ -28,16 +28,22 @@ class Base {
     }
 
     bool bothEnabled = generalParams.outputPointType == OutputType::AtPickpointAndElementwise;
-    if (generalParams.outputPointType == OutputType::AtPickpoint || bothEnabled) {
+    bool pointEnabled = generalParams.outputPointType == OutputType::AtPickpoint || bothEnabled;
+    bool elementwiseEnabled = generalParams.outputPointType == OutputType::Elementwise || bothEnabled;
+    if (pointEnabled) {
+      logInfo() << "Enable Dynamic Rupture Receiver Output";
```
Thomas-Ulrich, line 34: Enabling on-fault receivers output?

src/DynamicRupture/Output/Base.hpp
```diff
@@ -28,16 +28,22 @@ class Base {
...
       ppOutputBuilder = std::make_unique<PickPointBuilder>();
       ppOutputBuilder->setMeshReader(&mesher);
       ppOutputBuilder->setParams(reader.getPickPointParams());
-    } else if (generalParams.outputPointType == OutputType::Elementwise || bothEnabled) {
+    }
+    if (elementwiseEnabled) {
+      logInfo() << "Enable Dynamic Rupture 2D Fault Output";
```
Thomas-Ulrich, line 40: Enabling 3D fault output?

src/DynamicRupture/Output/Base.hpp
```diff
@@ -28,16 +28,22 @@ class Base {
...
       ewOutputBuilder = std::make_unique<ElementWiseBuilder>();
       ewOutputBuilder->setMeshReader(&mesher);
       ewOutputBuilder->setParams(reader.getElementwiseFaultParams());
-    } else {
-      logError() << "Unknown fault output type (not 3,4,5)";
+    }
+    if (!elementwiseEnabled && !pointEnabled) {
+      logInfo() << "No Dynamic Rupture Output enabled";
```
Thomas-Ulrich, line 46: remove capitalization.

src/ResultWriter/faultoutput.f90
sebwolf-de: This is how we do it: https://github.com/SeisSol/SeisSol/blob/dr/cpp/src/DynamicRupture/FrictionLaws/RateAndState.h#L146-L155, so with the Gnuc removed from this code  snippet, I think we are doing it correctly.

src/DynamicRupture/Output/Base.hpp
sebwolf-de: I would have said 2D since the faults are 2D manifolds (neglecting intersections) embedded in 3D space.

Thomas-Ulrich (approved):
ok, I think it should works!

PR #494, opened 2022-02-22
Computing plastic moment within SeisSol

Thomas-Ulrich:
Hi,

I've added a small extension to compute the contribution of plasticity to the seismic moment.
I tried with a small test based on Training/tpv13 and it kind of works, in the sense it gives a value which is close to the value (3.06e19) I get when computing from the plasticity wavefield output (less precise, as we only evaluate at a couple of locations).

Here are the outputs that I get:
```
1rank 1978459.NepM51.out:Tue Feb 22 13:59:39, Info:  Total plastic moment: 3.11013e+19 
2 ranks1978461.NepM51.out:Tue Feb 22 14:01:34, Info:  Total plastic moment: 3.11218e+19 
4 ranks 1978466.NepM51.out:Tue Feb 22 14:01:49, Info:  Total plastic moment: 3.12111e+19 
8 ranks 1978469.NepM51.out:Tue Feb 22 14:02:22, Info:  Total plastic moment: 3.14089e+19 
16 ranks 1978470.NepM51.out:Tue Feb 22 14:02:35, Info:  Total plastic moment: 3.17421e+19 
```
There are small variations depending on how many ranks I use. Maybe the consequence of the Ghost cells counted twice?
Todo:
- Try to fix the small differences.
- Try to get access to a usePlasticity variable to call the function only when plasticity is on.
- Make it work with anisotropy.

Any help is welcome!


krenzland:
To avoid duplicate mesh cells you have to iterate over them like this:
https://github.com/SeisSol/SeisSol/blob/master/src/ResultWriter/AnalysisWriter.cpp#L124
So iterate over the mesh and not the LTS data structure.

Thomas-Ulrich:
ok, thanks, I tried that, but still got inconsistencies
```
1 rank Tue Feb 22 16:52:06, Info:  Total plastic moment: 3.11013e+20``
2 ranks Tue Feb 22 16:46:58, Info:  Total plastic moment: 3.11013e+20
8 ranks:Tue Feb 22 16:52:55, Info:  Total plastic moment: 2.59159e+20
16 ranks Tue Feb 22 16:46:10, Info:  Total plastic moment: 2.53926e+20
```

Thomas-Ulrich:
Hi, the original code is from Carsten (probably based on a first Fortran code of Stephanie), that's why it looks great :-)
But the total work is only computed from the fault data (basically integrating traction * slip_rate), so its scope is not as broad as you may think. Here are some details about the mathematics behind, nicely summarized by Alice:
image

krenzland:
Ah, so it's only the fault energy. That's good, then I can extend it later with the other energies. It's more or less trivial to compute them as it's just an integral over all elements and over the surface (for grav.) The fault energy is the hard part that I didn't want to tackle, so it's great that you took care of it :)

generated_code/DynamicRupture.py
```diff
@@ -95,4 +97,39 @@ def interpolateQGenerator(i,h):
...
+  generator.add('computeSlipRateInterpolated', computeSlipRateInterpolated)
+
+  computeTractionInterpolated = tractionInterpolated['kp'] <= QInterpolatedMinus['kq'] * tractionMinusMatrix['qp'] + QInterpolatedPlus['kq'] * tractionPlusMatrix['qp']
+  generator.add('computeTractionInterpolated', computeTractionInterpolated)
+
+  addSlipRateInterpolated = slipInterpolated['kp'] <= slipInterpolated['kp'] + timeWeight * slipRateInterpolated['kp']
+  generator.add('addSlipRateInterpolated', addSlipRateInterpolated)
+
+  squareSlipRateInterpolated = absoluteSlipInterpolated['k'] <= slipRateInterpolated['kp'] * slipRateInterpolated['kp']
+  generator.add('squareSlipRateInterpolated', squareSlipRateInterpolated)
```
krenzland, line 130: Doesn't this also sum over index p?

generated_code/elastic.py
```diff
@@ -54,6 +55,10 @@ def __init__(self, order, multipleSimulations, matricesDir, memLayout, **kwargs)
 
     memoryLayoutFromFile(memLayout, self.db, clones)
 
+    selectVelocitySpp = np.zeros((self.numberOfQuantities(), 3))
```
krenzland, line 58: Same as:
https://github.com/SeisSol/SeisSol/blob/master/generated_code/SurfaceDisplacement.py#L56 

?

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -537,26 +539,48 @@ void seissol::initializers::initializeDynamicRuptureMatrices( MeshReader const&
           break;
         }
       }
+      /// Traction matrices for "average" traction
```
krenzland, line 542: Why "average"? The quotes are a bit scary

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -537,26 +539,48 @@ void seissol::initializers::initializeDynamicRuptureMatrices( MeshReader const&
...
+      auto tractionPlusMatrix = init::tractionPlusMatrix::view::create(godunovData[ltsFace].tractionPlusMatrix);
+      auto tractionMinusMatrix = init::tractionMinusMatrix::view::create(godunovData[ltsFace].tractionMinusMatrix);
+      double ZpP = plusMaterial->rho * waveSpeedsPlus[ltsFace].pWaveVelocity;
+      double ZsP = plusMaterial->rho * waveSpeedsPlus[ltsFace].sWaveVelocity;
+      double ZpM = minusMaterial->rho * waveSpeedsMinus[ltsFace].pWaveVelocity;
+      double ZsM = minusMaterial->rho * waveSpeedsMinus[ltsFace].sWaveVelocity;
+      double etaP = ZpP*ZpM / (ZpP + ZpM);
```
krenzland, line 549: There is probably no better name for the impedances, right?

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -537,26 +539,48 @@ void seissol::initializers::initializeDynamicRuptureMatrices( MeshReader const&
...
+      tractionMinusMatrix(5,2) = etaS / ZsM;
 
       /// Transpose Tinv
       dynamicRupture::kernel::transposeTinv ttKrnl;
       ttKrnl.Tinv = TinvData;
       ttKrnl.TinvT = godunovData[ltsFace].TinvT;
       ttKrnl.execute();
 
-      double plusSurfaceArea, plusVolume, minusSurfaceArea, minusVolume;
+      double plusSurfaceArea, plusVolume, minusSurfaceArea, minusVolume, surfaceArea = 1.e99;
```
krenzland, line 568: Why is this set to 1.e99?

src/Kernels/DynamicRupture.cpp
```diff
@@ -134,6 +137,31 @@ void seissol::kernels::DynamicRupture::spaceTimeInterpolation(  DRFaceInformatio
   real degreesOfFreedomPlus[tensor::Q::size()] __attribute__((aligned(PAGESIZE_STACK)));
   real degreesOfFreedomMinus[tensor::Q::size()] __attribute__((aligned(PAGESIZE_STACK)));
 
+  real slipRateInterpolated[tensor::slipRateInterpolated::size()] __attribute__((aligned(ALIGNMENT)));
```
krenzland, line 140: alignas(ALIGNMENT)

src/ResultWriter/EnergyOutput.cpp
```diff
@@ -0,0 +1,147 @@
...
+                                        real slip[seissol::tensor::slipInterpolated::size()]) {
+  real points[NUMBER_OF_SPACE_QUADRATURE_POINTS][2];
+  real spaceWeights[NUMBER_OF_SPACE_QUADRATURE_POINTS];
+  seissol::quadrature::TriangleQuadrature(points, spaceWeights, CONVERGENCE_ORDER + 1);
+
+  dynamicRupture::kernel::evaluateAndRotateQAtInterpolationPoints krnl;
+  krnl.V3mTo2n = global->faceToNodalMatrices;
+
+  real QInterpolatedPlus[tensor::QInterpolatedPlus::size()]
+      __attribute__((aligned(PAGESIZE_STACK)));
```
krenzland, line 45: alignas(PAGESIZE_STACK)

src/ResultWriter/EnergyOutput.cpp
```diff
@@ -0,0 +1,147 @@
...
+void seissol::writer::printEnergies(GlobalData const* global,
+                                    seissol::initializers::DynamicRupture* dynRup,
+                                    seissol::initializers::LTSTree* dynRupTree,
+                                    MeshReader const& i_meshReader,
+                                    seissol::initializers::LTSTree* i_ltsTree,
+                                    seissol::initializers::LTS* i_lts,
+                                    seissol::initializers::Lut* i_ltsLut,
+                                    bool usePlasticity)
+
+{
```
krenzland, line 92: Super weird spacing?

src/ResultWriter/EnergyOutput.cpp
```diff
@@ -0,0 +1,147 @@
...
+    real** timeDerivativeMinus = it->var(dynRup->timeDerivativeMinus);
+    DRGodunovData* godunovData = it->var(dynRup->godunovData);
+    DRFaceInformation* faceInformation = it->var(dynRup->faceInformation);
+    DROutput* drOutput = it->var(dynRup->drOutput);
+
+#ifdef _OPENMP
+#pragma omp parallel for reduction(+ : totalWorkLocal) reduction(+ : staticWorkLocal)
+#endif
+    for (unsigned i = 0; i < it->getNumberOfCells(); ++i) {
+      if (faceInformation[i].plusSideOnThisRank) {
```
krenzland, line 109: Is this to ensure that we do not count duplicate cells twice? Are you sure that this actually works?

krenzland (changes requested):
Looks mostly good, some small comments

src/ResultWriter/EnergyOutput.cpp
Thomas-Ulrich: done

src/ResultWriter/EnergyOutput.cpp
Thomas-Ulrich: done

generated_code/DynamicRupture.py
Thomas-Ulrich: Should I rename it to squaredNormSlipRateInterpolated?

src/Initializer/CellLocalMatrices.cpp
Thomas-Ulrich: I don't know. That's Carsten's comment. They are average in the sense the matrix works for all spatial points, as we do not have subcell material resolution?

src/ResultWriter/EnergyOutput.cpp
Thomas-Ulrich: I don't know. Sounds like a binary partition, right?

Thomas-Ulrich (commented):
Thanks for this review. I tried to address it but be aware that it is Carsten's code.

generated_code/DynamicRupture.py
krenzland: Yes, I think that's better

generated_code/elastic.py
krenzland: You could add this in the constructor of LinearADERDG (or whatever the name is) - so in the superclass of elastic, etc. And then remove the call in SurfaceDisplacement and use the one from the super class

ravil-mobile:
Hi,
Have you tested whether it works with GPUs? You have access to Heisenbug, right?. I can see that some changes are required for `batchedSpaceTimeInterpolation` in `Kernels/DynamicRupture.cpp`

src/Kernels/DynamicRupture.cpp
```diff
@@ -134,6 +137,31 @@ void seissol::kernels::DynamicRupture::spaceTimeInterpolation(  DRFaceInformatio
...
+  real slipRateNormSquared[tensor::absoluteSlipInterpolated::size()] __attribute__((aligned(ALIGNMENT)));
+  real tractionInterpolated[tensor::tractionInterpolated::size()] __attribute__((aligned(ALIGNMENT)));
```
ravil-mobile, line 142: Maybe `alignas(ALIGNMENT)` instead of `__attribute__((aligned(ALIGNMENT)));`

Thomas-Ulrich: ok, it was already fixed (I forgot to push the commit)

src/ResultWriter/EnergyOutput.cpp
```diff
@@ -0,0 +1,142 @@
...
+                                           seissol::initializers::LTS* i_lts,
+                                           seissol::initializers::Lut* i_ltsLut) {
+  real plasticMoment = 0.0;
+  std::vector<Element> const& elements = i_meshReader.getElements();
+  std::vector<Vertex> const& vertices = i_meshReader.getVertices();
+
+#ifdef _OPENMP
+#pragma omp parallel for schedule(static) reduction(+ : plasticMoment)
+#endif
+  for (std::size_t meshId = 0; meshId < elements.size(); ++meshId) {
```
sebwolf-de, line 17: I'd prefer `elementId` here, since we only have one mesh overall.

src/ResultWriter/EnergyOutput.cpp
sebwolf-de: We should easily be able to verify this with a small mesh and `std::cout` debugging. Or add a counter and count how many elements are processed on every node.

src/ResultWriter/EnergyOutput.cpp
```diff
@@ -0,0 +1,142 @@
...
+  MPI_Reduce(&plasticMomentLocal, &plasticMomentGlobal, 1, MPI_DOUBLE, MPI_SUM, 0, MPI::mpi.comm());
+#else
+  rank = 0;
+  totalWorkGlobal = totalWorkLocal;
+  staticWorkGlobal = staticWorkLocal;
+  plasticMomentGlobal = plasticMomentLocal;
+#endif
+
+  if (rank == 0) {
+    logInfo(rank) << "Total work:" << totalWorkGlobal;
```
sebwolf-de, line 135: IMHO total work is misleading, as total work would be static work + frictional energy + plastic moment.

sebwolf-de (changes requested):
Hi, most things look good. I just have some  minor remarks about naming things.
If we are not sure, whether we count elements twice, we have to find out, I propose to just check this with a small mesh. Take for example the setups in https://github.com/7bits-register/precomputed-seissol.

generated_code/DynamicRupture.py
```diff
@@ -95,4 +97,39 @@ def interpolateQGenerator(i,h):
...
+  tractionPlusMatrix = Tensor('tractionPlusMatrix', selectTractionSpp.shape, selectTractionSpp, CSCMemoryLayout)
+  tractionMinusMatrix = Tensor('tractionMinusMatrix', selectTractionSpp.shape, selectTractionSpp, CSCMemoryLayout)
+
+  computeSlipRateInterpolated = slipRateInterpolated['kp'] <= QInterpolatedMinus['kq'] * aderdg.selectVelocity['qp'] - QInterpolatedPlus['kq'] * aderdg.selectVelocity['qp']
+  generator.add('computeSlipRateInterpolated', computeSlipRateInterpolated)
+
+  computeTractionInterpolated = tractionInterpolated['kp'] <= QInterpolatedMinus['kq'] * tractionMinusMatrix['qp'] + QInterpolatedPlus['kq'] * tractionPlusMatrix['qp']
+  generator.add('computeTractionInterpolated', computeTractionInterpolated)
+
+  addSlipRateInterpolated = slipInterpolated['kp'] <= slipInterpolated['kp'] + timeWeight * slipRateInterpolated['kp']
```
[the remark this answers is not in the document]
Thomas-Ulrich: done

generated_code/elastic.py
Thomas-Ulrich: done

generated_code/poroelastic.py
```diff
@@ -26,6 +27,10 @@ def __init__(self, order, multipleSimulations, matricesDir, memLayout, numberOfM
 
     memoryLayoutFromFile(memLayout, self.db, clones)
 
+    selectVelocitySpp = np.zeros((self.numberOfQuantities(), 3))
+    selectVelocitySpp[6:9,0:3] = np.eye(3)
+    self.selectVelocity = Tensor('selectVelocity', selectVelocitySpp.shape, selectVelocitySpp, CSCMemoryLayout)
+
```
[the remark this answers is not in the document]
Thomas-Ulrich: done

src/ResultWriter/EnergyOutput.cpp
Thomas-Ulrich: ok changed to totalFrictionalWork

src/ResultWriter/EnergyOutput.h
```diff
@@ -0,0 +1,78 @@
+/**
+ * @file
+ * This file is part of SeisSol.
+ *
+ * @author Carsten Uphoff (c.uphoff AT tum.de,
+ * http://www5.in.tum.de/wiki/index.php/Carsten_Uphoff,_M.Sc.)
+ *
+ * @section LICENSE
+ * Copyright (c) 2018, SeisSol Group
```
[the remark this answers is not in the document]
Thomas-Ulrich: removed the header

src/Solver/Simulator.cpp
```diff
@@ -150,8 +156,18 @@ void seissol::Simulator::simulate() {
...
+  auto ltsTree = seissol::SeisSol::main.getMemoryManager().getLtsTree();
+  auto lts     = seissol::SeisSol::main.getMemoryManager().getLts();
+  auto* ltsLut = e_interoperability.getLtsLut();
+  seissol::writer::printEnergies(globalData.onHost, dynRup, dynRupTree, meshReader, ltsTree, lts, ltsLut, m_usePlasticity);
+
   seissol::SeisSol::main.analysisWriter().printAnalysis(m_currentTime);
 
   printFlops();
 
+
```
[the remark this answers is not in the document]
Thomas-Ulrich: done

generated_code/DynamicRupture.py
Thomas-Ulrich: @sebwolf-de yes you are right, this absoluteSlip is not used anywhere.  I just removed the kernel.

krenzland:
>  If we are not sure, whether we count elements twice, we have to find out, I propose to just check this with a small mesh. Take for example the setups in https://github.com/7bits-register/precomputed-seissol.
That won't necessarily work because elements are only duplicated in a few edge cases. The only one that I am aware of is one element has neighbors that are on different MPI ranks, then it is contained twice in the copy output. This can only happen if you have at least three ranks.

Iterating over the mesh (as I've mentioned in a comment) above definitely works and does not require the "plusSideOnThisRank" variable (assuming it's only needed for avoiding counting twice). It was probably implemented in this way because the mesh iteration approach requires the mesh reader which used to be destroyed directly after reading the mesh (until I changed it!).

Thomas-Ulrich:
tested plusSideOnThisRank on 2 ranks and 6 ranks, with this code:
```diff
--- a/src/ResultWriter/EnergyOutput.cpp
+++ b/src/ResultWriter/EnergyOutput.cpp
@@ -88,6 +88,7 @@ void seissol::writer::printEnergies(GlobalData const* global,
   double totalFrictionalWorkLocal = 0.0;
   double staticFrictionalWorkLocal = 0.0;
   double plasticMomentLocal = 0.0;
+  int nFacesLocal=0;
   for (auto it = dynRupTree->beginLeaf(); it != dynRupTree->endLeaf(); ++it) {
     /// \todo timeDerivativePlus and timeDerivativeMinus are missing the last timestep.
     /// (We'd need to send the dofs over the network in order to fix this.)
@@ -98,10 +99,11 @@ void seissol::writer::printEnergies(GlobalData const* global,
     DROutput* drOutput = it->var(dynRup->drOutput);
 
 #ifdef _OPENMP
-#pragma omp parallel for reduction(+ : totalFrictionalWorkLocal) reduction(+ : staticFrictionalWorkLocal)
+#pragma omp parallel for reduction(+ : totalFrictionalWorkLocal) reduction(+ : staticFrictionalWorkLocal) reduction(+ : nFacesLocal)
 #endif
     for (unsigned i = 0; i < it->getNumberOfCells(); ++i) {
       if (faceInformation[i].plusSideOnThisRank) {
+        nFacesLocal +=1;
         totalFrictionalWorkLocal += drOutput[i].frictionalEnergy;
         staticFrictionalWorkLocal += computeStaticWork(global,
                                                        timeDerivativePlus[i],
@@ -119,6 +121,7 @@ void seissol::writer::printEnergies(GlobalData const* global,
   double totalFrictionalWorkGlobal = 0.0;
   double staticFrictionalWorkGlobal = 0.0;
   double plasticMomentGlobal = 0.0;
+  int nFacesGlobal = 0;
 #ifdef USE_MPI
   MPI_Comm_rank(MPI::mpi.comm(), &rank);
   MPI_Reduce(&totalFrictionalWorkLocal,
@@ -136,11 +139,13 @@ void seissol::writer::printEnergies(GlobalData const* global,
              0,
              MPI::mpi.comm());
   MPI_Reduce(&plasticMomentLocal, &plasticMomentGlobal, 1, MPI_DOUBLE, MPI_SUM, 0, MPI::mpi.comm());
+  MPI_Reduce(&nFacesLocal, &nFacesGlobal, 1, MPI_INT, MPI_SUM, 0, MPI::mpi.comm());
 #else
   rank = 0;
   totalFrictionalWorkGlobal = totalFrictionalWorkLocal;
   staticFrictionalWorkGlobal = staticFrictionalWorkLocal;
   plasticMomentGlobal = plasticMomentLocal;
+  nFacesGlobal = nFacesLocal;
 #endif
 
   if (rank == 0) {
@@ -150,5 +155,6 @@ void seissol::writer::printEnergies(GlobalData const* global,
     if (usePlasticity) {
       logInfo(rank) << "Total plastic moment:" << plasticMomentGlobal;
     }
+    logInfo(rank) << "nFacesGlobal:" << nFacesGlobal;
   }
 }
```
and this works:
```
Thu Mar 17 11:55:02, Info:  maximum theoretical speedup (compared to GTS): 6.68862 per cell LTS, 4.54311 with the used clustering.
Thu Mar 17 11:55:02, Info:  Number of elements in dynamic rupture time clusters:
Thu Mar 17 11:55:02, Info:  0 (dr): 135
Thu Mar 17 11:55:02, Info:  1 (dr): 1384
Thu Mar 17 11:55:02, Info:  2 (dr): 45
Thu Mar 17 11:55:02, Info:  3 (dr): 0
Thu Mar 17 11:55:02, Info:  4 (dr): 0
--> 1564
Thu Mar 17 11:55:04, Info:  nFacesGlobal: 1564
```
Lukas said:
>I now understand the old version. It is safe to iterate over faces, as faces can by definition only be shared by two elements. And the plusSideOnThisRank basically checks that each face is not counted twice. All faces in the interior layer are only counted once without this, but faces that are on two different ranks could be counted twice. But by checking that the plus side is on this rank, you avoid counting them twice

So I think this problem is solved.

src/ResultWriter/EnergyOutput.cpp
Thomas-Ulrich: done

src/ResultWriter/EnergyOutput.cpp
```diff
@@ -0,0 +1,142 @@
...
+  for (std::size_t meshId = 0; meshId < elements.size(); ++meshId) {
+    real* pstrainCell = i_ltsLut->lookup(i_lts->pstrain, meshId);
+    real volume = MeshTools::volume(elements[meshId], vertices);
+    CellMaterialData& material = i_ltsLut->lookup(i_lts->material, meshId);
+#ifdef USE_ANISOTROPIC
+    real mu = (material.local.c44 + material.local.c55 + material.local.c66) / 3.0;
+#else
+    real mu = material.local.mu;
+#endif
+    plasticMoment += mu * volume * pstrainCell[6 * NUMBER_OF_ALIGNED_BASIS_FUNCTIONS];
```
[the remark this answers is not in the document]
sebwolf-de: Could you maybe add a comment explaining this datastructure. At this point in the code it comes a bit out of nothing and it would be great to have an explanation there.

krenzland:
Great! I would never trust anything I say about how DR works but I think that I might be correct :)

From side we can merge this.

krenzland (approved):

krenzland:
Hey, is there a reason that this isn't merged yet?

Thomas-Ulrich:
> Hey, is there a reason that this isn't merged yet?

yes, I needed to test if it works with GPU, but their was a bug with the GPU version which is now solved by #521.
(I will then test it again now).

Thomas-Ulrich:
Ok, so I tested the GPU version with tpv13, it runs but outputs 0 in energies:

```Wed Mar 30 09:50:52, Info:  Total time spent in compute kernels: 195.581 
Wed Mar 30 09:50:52, Info:  Total frictional work: 0 
Wed Mar 30 09:50:52, Info:  Static frictional work: 0 
Wed Mar 30 09:50:52, Info:  Radiated energy: 0 
Wed Mar 30 09:50:52, Info:  Total plastic moment: 3.08733e+19 
Wed Mar 30 09:50:52, Info:  Total   measured HW-GFLOP:  0 
Wed Mar 30 09:50:52, Info:  Total calculated HW-GFLOP:  3584.25 
Wed Mar 30 09:50:52, Info:  Total calculated NZ-GFLOP:  997.674 
Wed Mar 30 09:50:52, Info:  WP calculated HW-GFLOP:  3322.81 
Wed Mar 30 09:50:52, Info:  WP calculated NZ-GFLOP:  814.284 
Wed Mar 30 09:50:52, Info:  DR calculated HW-GFLOP:  175.838 
Wed Mar 30 09:50:52, Info:  DR calculated NZ-GFLOP:  135.41 
Wed Mar 30 09:50:52, Info:  PL calculated HW-GFLOP:  85.6084 
Wed Mar 30 09:50:52, Info:  PL calculated NZ-GFLOP:  47.9808 
Rank:        0 | Info    | total number of performed time steps:            0
Rank:        0 | Info    | final time of the simulation:    8.0000000000000000     
Rank:        0 | Info    | seismic moment   1.8029624024167783E+020 Mw   7.4339911135837715    
```

krenzland:
Maybe we can merge it and only enable CPU support? 
@ravil-mobile You probably do not have time to look at this right now anyway, right? And once we merge the CPP DR implementation, it could become easier to integrate this with GPUs.

PR #495, opened 2022-02-23
Use FindPython3 to find python3

krenzland:
Fixes compilation on some systems e.g. Isembard.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.13)
 
 # use <PackageName>_ROOT variables
 if(POLICY CMP0074)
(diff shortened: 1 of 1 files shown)


krenzland:
@ravil-mobile Any idea why this would break the GPU pipeline? There is some issue with pybind that I cannot explain- but it looks like it's genuinely broken. But the CPU pipeline works, which should use a similar setting for pybind, right?

Thomas-Ulrich:
Could it be because of ```python3_add_library`` instead of ```Python3_add_library```?
(https://cmake.org/cmake/help/latest/module/FindPython3.html)

krenzland:
Thanks, but CMake is case insensitive. The problem was rather that pybind needs the DEVELOPMENT components of Python3 but if we call find_package(Python3 ...) without specifying the development component, the find_package call in Pybind does not include DEVELOPMENT either

Thomas-Ulrich (approved):
LGTM

PR #497, opened 2022-02-24
Change Pspamm installation documentation

NicoSchlw:
Correct the PSpaMM symlink in the docs

### DIFF
--- a/Documentation/compilation.rst
+++ b/Documentation/compilation.rst
@@ -99,7 +99,7 @@ Installing PSpaMM
 .. code-block:: bash
 
    git clone https://github.com/SeisSol/PSpaMM.git
-   ln -s $(pwd)/PSpaMM/pspamm.py $HOME/bin
+   ln -s $(pwd)/PSpaMM/pspamm.py $HOME/bin/pspamm.py
 
 Installing GemmForge (for GPU)
 ------------------------------
--- a/Documentation/compilation.rst
+++ b/Documentation/compilation.rst
@@ -106,7 +106,7 @@ Instead of linking, you could also add the following line to your .bashrc:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


ravil-mobile:
Hi, could you also mention that `pspamm` can be installed with Spack
```
spack install py-pspamm
```
it will require Spack version >=0.17.2

Thomas-Ulrich:
This won't work if bin does not exist.
What you need is mkdir ~/bin before the symbolic link.

ravil-mobile:
In general, one needs to just clone pspamm to a directory and then append `PATH` and `PYTHONPATH` with the name of that directory

NicoSchlw:
So should we just provide the lines for the bashrc in the docs?

NicoSchlw:
Is appending to PYTHONPATH really necessary?

krenzland:
Nope, never done that.

Documentation/compilation.rst
```diff
@@ -99,7 +99,14 @@ Installing PSpaMM
...
    git clone https://github.com/SeisSol/PSpaMM.git
-   ln -s $(pwd)/PSpaMM/pspamm.py $HOME/bin
+   # make sure $HOME/bin exists or create it with "mkdir ~/bin"
+   ln -s $(pwd)/PSpaMM/pspamm.py $HOME/bin/pspamm.py
+   
+Instead of linking, you could also add the following line to your .bashrc:
+
+.. code-block:: bash
+
+   export PATH=<Your_Path_to_PSpaMM>/PSpaMM:$PATH
```
krenzland, line 109: This line is a bit confusing, shouldn't it be
export PATH=<Your_Path_to_PSpaMM>:$PATH
?

krenzland (changes requested):

krenzland (approved):

krenzland:
Thanks!

PR #498, opened 2022-02-24
dr/merge master

sebwolf-de:


sebwolf-de:
@ravil-mobile can we merge this? I think #509 is not directly caused by this PR and was already present before.
#509 should be fixed in another PR.

PR #506, opened 2022-03-02
Fix 492

sebwolf-de:
* Remove eigen3 as a submodule.
* Search eigen3 dependency with cmake.
* Add documentation on how to install eigen3.
Eigen3 is header only, so installation should not be a problem in general.
I tested this with tpv5...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -287,14 +287,17 @@ if (ADDRESS_SANITIZER_DEBUG)
 endif()
 
 
-# Note: it is better to include `eigen3` `async` as
+# Note: it is better to include `async` as
 # system headers because they emit lot's of warnings
 # from clang. Most of them are issues with respect
 # to overriden virtual methods
 target_include_directories(SeisSol-lib SYSTEM PUBLIC
-  ${CMAKE_CURRENT_SOURCE_DIR}/submodules/eigen3
   ${CMAKE_CURRENT_SOURCE_DIR}/submodules/async
 )
+
+
+find_package(Eigen3 3.4.0 REQUIRED)
+target_include_directories(SeisSol-lib SYSTEM PUBLIC ${EIGEN3_INCLUDE_DIRS})
 
 find_package(yaml-cpp REQUIRED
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -287,15 +287,18 @@ if (ADDRESS_SANITIZER_DEBUG)
...
 # from clang. Most of them are issues with respect
 # to overriden virtual methods
 target_include_directories(SeisSol-lib SYSTEM PUBLIC
-  ${CMAKE_CURRENT_SOURCE_DIR}/submodules/eigen3
   ${CMAKE_CURRENT_SOURCE_DIR}/submodules/async
 )
 
+
+find_package(Eigen3 3.4.0 REQUIRED)
+target_include_directories(SeisSol-lib SYSTEM PUBLIC ${EIGEN3_INCLUDE_DIRS})
```
krenzland, line 300: Wouldn't it be better to do it like this:
https://eigen.tuxfamily.org/dox/TopicCMakeGuide.html
`target_link_libraries (SeisSol-lib PUBLIC Eigen3::Eigen)`

CMakeLists.txt
```diff
@@ -287,15 +287,18 @@ if (ADDRESS_SANITIZER_DEBUG)
...
 # system headers because they emit lot's of warnings
 # from clang. Most of them are issues with respect
 # to overriden virtual methods
 target_include_directories(SeisSol-lib SYSTEM PUBLIC
-  ${CMAKE_CURRENT_SOURCE_DIR}/submodules/eigen3
   ${CMAKE_CURRENT_SOURCE_DIR}/submodules/async
 )
 
+
+find_package(Eigen3 3.4.0 REQUIRED)
```
krenzland, line 299: Isn't it enough to specify 3.4, assuming that eigen follow semver?

krenzland (changes requested):

krenzland (dismissed):

krenzland:
Could you please fix the gitlab ci GPU build? It's currently failing because it cannot find the Eigen3 lib

krenzland (approved):

PR #510, opened 2022-03-09
add CAD tutorial: mesh with water layer

Thomas-Ulrich:
Please review also:
https://github.com/SeisSol/Meshing/pull/31

Apart from the tutorial, I also changed the hdf5 version in compilation to 1.10 and provided an alternative file for the gif in the Documentation (to be able to create the pdf).


Documentation/fully-coupled-mesh-tutorial.rst
```diff
@@ -0,0 +1,127 @@
...
+Creating topography and boxes
+-------------------------------
+
+First, we download topography and bathymetry data from GEBCO
+(`http://www.gebco.net/ <http://www.gebco.net/>`__), and we triangulate it into a GoCad ts file.
+
+- Note that we downsample the topography data by a factor 2 for dealing with a reasonable size dataset in this tutorial.
+- Note also that we use a custom transverse Mercator roughly centered at the domain center.
+- With the option ``--change_zero_elevation 1.0``, we move the nodes with zero elevation to 1.0 m. This avoids having to intersect locally coplanar surfaces.
+- With the option ``--smooth 100``, we smooth the topography data which fall in the range :math:`\pm` 100 m. This facilitates the intersection of the sea surface with the topography and allows a smoother coastline (which can else have a saw-tooth shape due to rounded elevation data, stored as integers in Gebco files).
```
krenzland, line 25: What does it mean that the topography data that fall in the range of +/- 100 is smoothed?
Is the 100 a cutoff parameter or similar?

Documentation/fully-coupled-mesh-tutorial.rst
```diff
@@ -0,0 +1,127 @@
...
+We can double-check the quality of the intersections with ``Prepare->Remove Small features``.
+Here the shortest edge in the model is 139 m, which is large enough.
+If this was not the case, small edges can be identified, and the associated node of the not yet intersected topography moved (with ``Discrete->Deform Face``).
+Then the full workflow described above needs to be followed again with the updated topography (and hopefully the short edge has been suppressed).
+
+The next step consists of loading `domain_box.stl` and intersecting it with the current model.
+The union is done with SimModeler11.0-220403-dev (as it fails with SimModeler10.0).
+We then clean the model of the upper part of domain_box, the part of the topography outside the domain box, and the smaller water layer region.
+The sea floor below the removed smaller water layer region can be merged with ``Discrete->Combine Faces``).
+The model obtained has 2 regions.
```
krenzland, line 70: The obtained model

krenzland (changes requested):
Looks great, just two small comments

Documentation/fully-coupled-mesh-tutorial.rst
Thomas-Ulrich: Here is the code:
```python
    def smooth(self, zrange):
        "Smooth bathymetry in range +- zrange"
        "This is useful when preprocessing bathymetry data before intersection"
        import scipy.ndimage as ndimage

        z_smooth = ndimage.gaussian_filter(self.z, sigma=(1, 1), order=0)
        ids = np.where(abs(self.z) < zrange)
        self.z[ids] = z_smooth[ids]
```
The idea is to smooth the bathymetry (that is also degrade bathymetry data) only where it is necessary, that is at points with z close to 0, where the coastline should be.

Thomas-Ulrich: What about:
With the option ``--smooth 100``, we replace the elevation of topography nodes with z coordinates in the range :math:`\pm` 100 m by spatially smoothed (using a 2D Gaussian kernel) values.
This facilitates...

krenzland: Yup, sounds great. What happens with values > 100? They are not smoothed?

Thomas-Ulrich: no, they are not smoothed. The true topography data are not smooth. But smoothed data facilitate union (that's my experience at least).

Thomas-Ulrich:
When merging, please squash commits.

PR #516, opened 2022-03-17
Removing warnings from the Fortran section of SeisSol.

LukasvdWiel:
Compiling the code on GCC 10.2.0 resulted in six warnings in the Fortran section of SeisSol.
Five concerned variable conflicts on src/Physics/Evaluate_friction_law.f90
and one was an incorrect comment marker in calc_seissol.f90.


sebwolf-de (approved):
Hi Lukas,
thanks for taking care of that.
LGTM!

PR #519, opened 2022-03-22
Make desctructors virtual

### LINKED ISSUE #515: vtable for WeightsModels not created
**Describe the bug**
When compiling SeisSol, an undefined reference for the vtables is presented when linking SeisSol_Release_dhsw_6_elastic

libSeisSol-lib.a(MeshReaderFBinding.cpp.o): In function 

`seissol::initializers::time_stepping::getLtsWeightsImplementation(seissol::initializers::time_stepping::LtsWeightsTypes, seissol::initializers::time_stepping::LtsWeightsConfig const&)':

MeshReaderFBinding.cpp:(.text+0xbf0): undefined reference to `vtable for seissol::initializers::time_stepping::ExponentialBalancedWeights'

MeshReaderFBinding.cpp:(.text+0xd00): undefined reference to `vtable for seissol::initializers::time_stepping::ExponentialWeights'

MeshReaderFBinding.cpp:(.text+0xe10): undefined reference to `vtable for seissol::initializers::time_stepping::EncodedBalancedWeights'

collect2: error: ld returned 1 exit status

make[2]: *** [SeisSol_Release_dhsw_6_elastic] Error 1

make[1]: *** [CMakeFiles/SeisSol-bin.dir/all] Error 2

make: *** [all] Error 2

**Expected behavior**
Would expect SeisSol to compile without errors

**To Reproduce**
Install GCC 10.2.0.
Our installation has compiled countless programs already and we are confident that the compiler functions properly.
Compile SeisSol gives the errors almost at end, when linking SeisSol_Release_dhsw_6_elastic

**Additional context**
A well known programming mistake causing this error is apparently including the destructors in the header file in stead of in the C++ file. So I moved the destructors from header file to cpp file, but that did not change the compiling process.

sebwolf-de:
Fixes #515

### DIFF
--- a/src/Initializer/time_stepping/LtsWeights/WeightsModels.h
+++ b/src/Initializer/time_stepping/LtsWeights/WeightsModels.h
@@ -9,7 +9,7 @@ namespace seissol::initializers::time_stepping {
 class ExponentialWeights : public LtsWeights {
 public:
   explicit ExponentialWeights(const LtsWeightsConfig &config) : LtsWeights(config) {}
-  ~ExponentialWeights() override = default;
+  virtual ~ExponentialWeights() override = default;
 
 protected:
   int evaluateNumberOfConstraints() final { return 1; }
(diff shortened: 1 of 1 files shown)


Thomas-Ulrich:
Actually when reading https://stackoverflow.com/a/15903538/9892060, I wonder if it is not the base class destructor that should be made virtual?

sebwolf-de:
https://github.com/SeisSol/SeisSol/blob/c82e6c6c66a272b8fe86afe265cc59965a3e7b36/src/Initializer/time_stepping/LtsWeights/LtsWeights.h#L72 The base destructor is already virtual.

PR #529, opened 2022-04-12
Update SNG modules.

krenzland:
These modules work with the new spack environment on supermuc ng. Note that you need to recompile everything (at least ASAGI, easi, SeisSol).

Not sure why it works with this yaml-cpp module: It's version 0.7 and we search for 0.6.3 explicitly. However, easi finds it and links with it correctly.

https://github.com/TUM-I5/ASAGI/pull/12 needs to be merged beforehand, otherwise ASAGI does not find netcdf.

### DIFF
--- a/Documentation/compilation.rst
+++ b/Documentation/compilation.rst
@@ -139,7 +139,10 @@ Additionally, one can install *chainforge* GEMM generator which can result in be
    pip3 install https://github.com/ravil-mobile/chainforge.git
 
 
-Installing ParMetis (Optional: PUML mesh format)
+
+.. _installing_parmetis:
+
+Installing ParMetis
 ------------------------------------------------
 
 .. code-block:: bash
--- a/Documentation/supermuc.rst
+++ b/Documentation/supermuc.rst
@@ -143,10 +143,11 @@ For that modules and compiler need to be switched:
 
 ::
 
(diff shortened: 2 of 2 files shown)


Thomas-Ulrich:
Thx, do you want also to update the modules for the GCC build, a few lines below? Just after

>Note that to use sanitzer (https://en.wikipedia.org/wiki/AddressSanitizer), SeisSol needs to be compiled with gcc. For that modules and compiler need to be switched:

krenzland:
Not really at the moment - I don't have time to re-compile all dependencies this week (and I'm on vacation next week).
I can only guess the correct modules.
Probably the following would work:
```
module load parmetis/4.0.3-gcc11-impi-i64-r32 metis/5.1.0-gcc11-i64-r64
module load netcdf-hdf5-all/4.7_hdf5-1.10-gcc11-impi
module load numactl/2.0.14-gcc11
module load yaml-cpp/0.7.0
```

krenzland:
I've added them, they should work. I've only tested that the modules exists but not verified that they actually work. As they are probably automatically generated via spack, I see no reason why they wouldn't work though.

Thomas-Ulrich (approved):

krenzland:
Do not merge yet, the (par)metis module is broken... :(

PR #536, opened 2022-04-26
use nucleation stress with FL2

Thomas-Ulrich:
Add the possibility to add nucleation stress or traction in FL=2 (linear slip weakening), as it is currently done with RS friction.
(feature needed by Elisa Tinti and Emanuelle Casarotti).
So the idea would be FL2: nucleation as in RS, FL16: nucleation with forced rupture.
(I know that you might think it can wait dr/cpp but it seems they need the feature urgently).

### DIFF
--- a/Documentation/asagi.rst
+++ b/Documentation/asagi.rst
@@ -38,53 +38,34 @@ Installing ASAGI
 
 Be careful that the python and gcc package is the same as for the
 compilation of SeisSol in a later step!
-
-example on SuperMuc
-~~~~~~~~~~~~~~~~~~~
-
--  load the following modules 
+First clone ASAGI with:
 
 .. code-block:: bash
 
-   module load intel intel-mpi
-   module load netcdf-hdf5-all/4.7_hdf5-1.10-intel19-impi
-   module load gcc/9
-   module load cmake/3.16.5
+  git clone git@github.com:TUM-I5/ASAGI
+  # git clone https://github.com/TUM-I5/ASAGI.git
+  cd ASAGI
+  git submodule update --init
 
--  get the repository
-
-On a cluster without restricted access to outside sources, you could then clone the repository using:
+Set compiler options, e.g. for intel compilers on SuperMuc:
 
 .. code-block:: bash
 
-   git clone --recursive https://github.com/TUM-I5/ASAGI.git
-   
-On supermuc, you have to set up port forwarding as described in :ref:`compile_run_supermuc`.
+  export FC=mpif90
+  export CXX=mpiCC
+  export CC=mpicc
 
-Then you can clone the project with 
-
-.. code-block:: bash
-
-   git clone git@github.com:TUM-I5/ASAGI.git
-   git submodule update --init
-
-
--  install:
+Run cmake, and compile with:
 
 .. code-block:: bash
 
-   mkdir build && cd build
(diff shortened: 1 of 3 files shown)


src/Physics/Evaluate_friction_law.f90
```diff
@@ -403,21 +404,36 @@ SUBROUTINE Linear_slip_weakening_TPV1617(TractionGP_XY,TractionGP_XZ,       & !
...
+            !DISC%DynRup%NucBulk_** is already in fault coordinate system
+            EQN%InitialStressInFaultCS(:,1,iFace)=EQN%InitialStressInFaultCS(:,1,iFace)+EQN%NucleationStressInFaultCS(:,1,iFace)*Gnuc
+            EQN%InitialStressInFaultCS(:,2,iFace)=EQN%InitialStressInFaultCS(:,2,iFace)+EQN%NucleationStressInFaultCS(:,2,iFace)*Gnuc
+            EQN%InitialStressInFaultCS(:,3,iFace)=EQN%InitialStressInFaultCS(:,3,iFace)+EQN%NucleationStressInFaultCS(:,3,iFace)*Gnuc
+            EQN%InitialStressInFaultCS(:,4,iFace)=EQN%InitialStressInFaultCS(:,4,iFace)+EQN%NucleationStressInFaultCS(:,4,iFace)*Gnuc
+            EQN%InitialStressInFaultCS(:,5,iFace)=EQN%InitialStressInFaultCS(:,5,iFace)+EQN%NucleationStressInFaultCS(:,5,iFace)*Gnuc
+            EQN%InitialStressInFaultCS(:,6,iFace)=EQN%InitialStressInFaultCS(:,6,iFace)+EQN%NucleationStressInFaultCS(:,6,iFace)*Gnuc
+        ENDIF ! t_0
+    ENDIF ! FL
+
```
sebwolf-de, line 436: Shouldn't we move this whole block in to the loop over `iTimeGP` and add NucleationStress with the sub time steps? I know this is done the same way for `rate_and_state`, but I think with the sub time steps, we would be more accurate?

sebwolf-de (commented):
Yes, we can merge this fast, `dr/cpp` will take a little longer to be merged into master... Still, I have one question about the nucleation function.

src/Physics/Evaluate_friction_law.f90
Thomas-Ulrich: Yes!

sebwolf-de (approved):

sebwolf-de:
Does this feature actually work?
https://github.com/SeisSol/SeisSol/blob/master/src/Reader/readpar.f90#L1013-1017 <- We set `t_0` only for FL16, otherwise, it defaults to 0.
https://github.com/SeisSol/SeisSol/blob/master/src/Physics/Evaluate_friction_law.f90#L425-L433 <- The nucleation routine, is only executed for FL2, then implicitely `t_0 = 0` and we never execute lines 427-431.

Thomas-Ulrich:
It seems it does not work indeed.
But t_0 is typically set to 0 in benchmarks.
E.g. I checked tpv12_13. The parameter file states t_0=0.5 but it turns out that the benchmark does not specify any time dependence so both problems were self-compensated.
If you fix t_0 on master (that is what I would do), then you need to update the corresponding example files.

PR #539, opened 2022-04-28
another round of edits on supermuc install...

Thomas-Ulrich:
Hopefully, this will fix all compilation issues on supermuc.
(not that I did not need to manually fix hdf5 in CMakelist, as I was reporting before).
Please test.


sebwolf-de (approved):
LGTM, if Fabian approves :D

PR #540, opened 2022-04-28
Updated CMake version for SeisSol in docu

ravil-mobile:


krenzland (approved):
Makes sense. Especially CUDA support is a critical issue.

PR #542, opened 2022-05-02
dr/better error message

sebwolf-de:
Previously the error message, just stated `bad conversion`, but not more information


ravil-mobile (approved):

ravil-mobile:
:+1:

PR #550, opened 2022-05-17
add reproducible research page (for Cocorecs)

Thomas-Ulrich:
- add reproducible research page 
- remove 2 outdated pages (and move last item of UNSORTED from index)
- refactor cookbook table
- reorder publications and add doi as a link

Here is how it look like:

image

image


AliceGabriel (approved):
looks good! Thank you Thomas!

PR #554, opened 2022-05-20
Implement edge weights for the partitioning 

ThrudPrimrose:
Why? Approximating the communication with edge weights usually improves the performance of the code.

The main changes are located in src/Initializer/time_stepping.
LtsWeights class was changed to accomodate to sub classes, respectively NodeWeightModel and EdgeWeightModel. Where NodeWeightModel implements the already existing node partitioning strategies (exponential, exponential balanced, encoded balanced) while adding new node weight partitioning strategies, (expnential balanced with balanced messaging, exponential balanced with minimum messaging).

Further edge weight strategies are implemented in the EdgeWeightModel (naive, communication approximation, communication approximation with balanced messaging, communication approximation with minimum messaging.

The details of what each strategy does can be found in the report for my guided research. (In the chapter IV: Strategies to Partition Meshes)
[Guided_Research_SeisSol_Partitioning.pdf](https://github.com/SeisSol/SeisSol/files/8743277/Guided_Research_SeisSol_Partitioning.pdf)

Also for the implementation I have preferred composition over inheritance. The NodeWeight and EdgeWeight models are now members of the LtsWeights class. (Concrete implementations used to inherit an abstract class before)

In the parameters.par file, LtsWeights input is no longer valid, and the user has to provide new variables NodeWeightModelTypeId and EdgeWeightModelTypeId that govern the used node and edge weight model. Brief explanation of implemented strategies are available there.

If edgeweight model is set to naive (0, default value) then the result is the same as before, as if there was no edge weight strategy implemented.

This PR also requires the PR I submitted to PUML2  to be accepted, as computing edge weights requires metis to return the dual graph of  the mesh, and we need to compute the partitioning after we set the edge weights, so the partitiong function also takes edge weights as arguments (it is defaulted to nullptr, so it should be backwards compatible) and computes the edge weights on the dual graph.


ThrudPrimrose:
( I will fix the problem with the formatted files, the submodule error, and create another pull request)

PR #555, opened 2022-05-21
add edge weight cost models and new vertex weight cost models to allo…

ThrudPrimrose:
Adding edge weights improve the performance in most cases. 

To allow this the old LtsWeights class is changed to be a class that contains NodeWeightModel and EdgeWeightModel classes as members, which respectively compute the vertex and edge cost models, depending on the given input for the strategy.

The old LtsWeights parameter that would be defined in .par file is not allowed anymore, user has to define NodeWeightModelTypeId and EdgeWeightModelTypeId.

This change also needs my PR request on PUML2, as we need to get the dual graph representation of the mesh to be able to compute edges. And we propagate the edge weights to metis, and the function to partition the graph will need the accept edgeweights as an additional parameter.


ThrudPrimrose:
Hi, I totally oversaw the documentation part in my pull request. I will
adapt the documentation for the new strategies and update the PR.

PR #557, opened 2022-05-23
Fix viscoelastic compilation

sebwolf-de:
I just found out that the `viscoelastic` compilation (without the `2`) failed. Here is an easy fix.


krenzland (approved):
Thanks!

In the future, we could think about removing this configuration. No one is using it anyway and its superseded by viscoelastic2

Issue #562, opened 2022-06-01
possible mistake in `saveDynamicStressOutput` in `dr/cpp`


ravil-mobile:
https://github.com/SeisSol/SeisSol/blob/81179a747b4d6d33aaf17906d41e52c6051b3e70/src/DynamicRupture/FrictionLaws/LinearSlipWeakening.h#L136-L146

why do we use `this->dynStressTimePending[pointIndex]` in line 139 instead of
 `this->dynStressTimePending[ltsFace][pointIndex]`

sebwolf-de:
Closed with #566

PR #566, opened 2022-06-02
dr/some fixes

sebwolf-de:
Two fixes:
* `slip1/2` were declared twice for the imposed slip rates FL
* #562


src/DynamicRupture/FrictionLaws/ImposedSlipRates.h
```diff
@@ -12,17 +12,15 @@ class ImposedSlipRates : public BaseFrictionLaw<ImposedSlipRates<STF>> {
   public:
   using BaseFrictionLaw<ImposedSlipRates>::BaseFrictionLaw;
 
-  real (*slip1)[misc::numPaddedPoints];
-  real (*slip2)[misc::numPaddedPoints];
+  real (*imposedSlipDirections)[misc::numPaddedPoints][2];
```
ravil-mobile, line 15: I like previous solution. Why couldn't we make `imposedSlipDirections1` and `imposedSlipDirections2`

Here are some comments in case if a simple renaming is not going to work. `real (*imposedSlipDirections)[misc::numPaddedPoints][2]` is going to reside in a non-unit stride access.
Maybe,  we could make `real (*imposedSlipDirections)[2][misc::numPaddedPoints];`

I would also propose to add some enum to access data in each direction 
```
enum Direction {
  One = 0,
  Two,
  Count
}
```

src/DynamicRupture/FrictionLaws/ImposedSlipRates.h
```diff
@@ -41,15 +39,19 @@ class ImposedSlipRates : public BaseFrictionLaw<ImposedSlipRates<STF>> {
...
-      tractionResults.traction2[timeIndex][pointIndex] =
-          faultStresses.traction2[timeIndex][pointIndex] -
-          this->impAndEta[ltsFace].etaS * slip2[ltsFace][pointIndex] * stfEvaluated;
+      this->traction1[ltsFace][pointIndex] = faultStresses.traction1[timeIndex][pointIndex] -
+                                             this->impAndEta[ltsFace].etaS *
+                                                 imposedSlipDirections[ltsFace][pointIndex][0] *
+                                                 stfEvaluated;
+      this->traction2[ltsFace][pointIndex] = faultStresses.traction2[timeIndex][pointIndex] -
+                                             this->impAndEta[ltsFace].etaS *
+                                                 imposedSlipDirections[ltsFace][pointIndex][1] *
```
ravil-mobile, line 48: I guess if we re-write it as follows  `imposedSlipDirections[ltsFace][0][pointIndex]` and `imposedSlipDirections[ltsFace][1][pointIndex]` it will result in a better vectorization. Moreover, it will become GPU-friendly i.e., we would share more data structures

src/DynamicRupture/Initializers/ImposedSlipRatesInitializer.cpp
```diff
@@ -85,10 +89,10 @@ void ImposedSlipRatesInitializer::rotateSlipToFaultCS(
...
     real sin = std::sqrt(1 - cos * cos) * std::copysign(1.0, scalarProduct);
     for (size_t pointIndex = 0; pointIndex < misc::numPaddedPoints; ++pointIndex) {
-      real strikeSlip = slip1[ltsFace][pointIndex];
-      real dipSlip = slip2[ltsFace][pointIndex];
-      slip1[ltsFace][pointIndex] = cos * strikeSlip + sin * dipSlip;
-      slip2[ltsFace][pointIndex] = -sin * strikeSlip + cos * dipSlip;
+      imposedSlipDirections[ltsFace][pointIndex][0] =
+          cos * strikeSlip[ltsFace][pointIndex] + sin * dipSlip[ltsFace][pointIndex];
+      imposedSlipDirections[ltsFace][pointIndex][1] =
+          -sin * strikeSlip[ltsFace][pointIndex] + cos * dipSlip[ltsFace][pointIndex];
```
ravil-mobile, line 95: the same comments about vectorization. I guess the compiler will find a way to vectorize this code and it will be easy in case of a unit-stride access with respect to `imposedSlipDirections`

Thomas-Ulrich (approved):
Just a suggestion: why not simply `imposedSlip1, imposedSlip2`?

PR #580, opened 2022-06-21
Dr/check new output

ravil-mobile:


src/DynamicRupture/Output/ReceiverBasedOutput.cpp
```diff
@@ -75,6 +79,13 @@ void ReceiverBasedOutput::calcFaultOutput(const OutputType type,
     auto* initStress = initStresses[local.ltsId][local.nearestGpIndex];
 
     local.frictionCoefficient = (local.layer->var(drDescr->mu))[local.ltsId][local.nearestGpIndex];
+    if (containsStateVariable) {
+      auto* descr = reinterpret_cast<seissol::initializers::LTS_RateAndState*>(drDescr);
+      local.stateVariable =
+          (local.layer->var(descr->stateVariable))[local.ltsId][local.nearestGpIndex];
+    } else {
+      local.stateVariable = 0.0;
+    }
```
ravil-mobile, line 88: Maybe it is better to use virtual methods

ravil-mobile:
All tests didn't fail. I guess I can completely switch the Fortran output. What do you think?

src/DynamicRupture/Output/ReceiverBasedOutput.cpp
ravil-mobile: done

sebwolf-de (approved):

src/ResultWriter/ini_faultoutput.f90
```diff
@@ -344,6 +344,8 @@ SUBROUTINE ini_fault_receiver(EQN,MESH,BND,DISC,IO,MPI)
     INTENT(INOUT)           :: DISC
     !-----------------------------------------------------------!
 
+    ! DEPRECATED: can be removed
+#if 0
```
Thomas-Ulrich, line 348: Why don't you remove it now?

PR #582, opened 2022-06-22
Remove empty files

sebwolf-de:


ravil-mobile:
Thanks for cleaning the code up. Could you also have a look here

https://github.com/SeisSol/SeisSol/blob/a4fe0b0e48f003694b536c51aab53829e67633b8/src/DynamicRupture/Initializers/LinearSlipWeakeningInitializer.h#L17-L19

I guess we don't need it any more. If yes, could you please remove it

sebwolf-de:
I am pretty sure, this function is still used.

ravil-mobile:
> I am pretty sure, this function is still used.

But why do we need it if we completely switched to the DR impl? Could you check,  please, whether `seissol::Interoperability* eInteroperability` parameter is still needed?

Thomas-Ulrich (approved):
LGTM

PR #585, opened 2022-06-23
Added a flush method to `OutputManager`. Clean up the code

ravil-mobile:


sebwolf-de:
Unfortunately, the last few data points are still not written to the receiver file :/

src/ResultWriter/FaultWriterC.cpp
```diff
@@ -36,6 +36,8 @@
  *
  * @section DESCRIPTION
  */
+#if 0
```
Thomas-Ulrich, line 39: just remove then?

src/ResultWriter/faultoutput.f90
```diff
@@ -766,7 +766,7 @@ END SUBROUTINE calc_FaultOutput
   SUBROUTINE write_FaultOutput_elementwise(EQN, DISC, MESH, IO, MPI, MaterialVal, BND, time, dt)
       !-------------------------------------------------------------------------!
       USE JacobiNormal_mod
-      use FaultWriter
+      ! use FaultWriter ! DEPRECATED: can be removed
```
Thomas-Ulrich, line 769: do not commet but remove?

ravil-mobile: it will be removed later. I am keeping this in my mind

src/ResultWriter/FaultWriterC.cpp
ravil-mobile: It will be removed later. I gave my reasoning below

sebwolf-de:
Okay from my side, but let's see what @Thomas-Ulrich says...

Thomas-Ulrich (approved):

PR #589, opened 2022-06-28
Update parameter-file.rst

Yonatan298:
I believe the normal sign should be changed to (0,+1,0) instead of (0,-1,0).


Thomas-Ulrich:
Hi, 
Thank you for spotting that.
Honestly, I think this page is more confusing than helpful.
Anyway, I tried to do the math (applying the isPlus formula on this configuration), and get to the same conclusion as you.

PR #595, opened 2022-06-30
dr/fix-forced-rupture-time

sebwolf-de:
Rather big refactoring, ~will add tpv16 as a test to the CI in another PR~.

### DIFF
--- a/src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
+++ b/src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
@@ -1,37 +1,6 @@
 #include "LinearSlipWeakening.h"
 namespace seissol::dr::friction_law {
 
-void ForcedRuptureTime::copyLtsTreeToLocal(seissol::initializers::Layer& layerData,
-                                           seissol::initializers::DynamicRupture* dynRup,
-                                           real fullUpdateTime) {
-  // maybe change later to const_cast?
-  auto* concreteLts =
-      dynamic_cast<seissol::initializers::LTS_LinearSlipWeakeningForcedRuptureTime*>(dynRup);
-  forcedRuptureTime = layerData.var(concreteLts->forcedRuptureTime);
-}
-
-void ForcedRuptureTime::stateVariableHook(std::array<real, misc::numPaddedPoints>& stateVariable,
-                                          real time,
-                                          unsigned int ltsFace) {
-  for (unsigned pointIndex = 0; pointIndex < misc::numPaddedPoints; pointIndex++) {
-    real f2 = 0.0;
-    if (drParameters.t0 == 0) {
-      if (time >= forcedRuptureTime[ltsFace][pointIndex]) {
-        f2 = 1.0;
-      } else {
-        f2 = 0.0;
-      }
-    } else {
-      f2 = std::max(static_cast<real>(0.0),
-                    std::min(static_cast<real>(1.0),
-                             // Note: In the fortran implementation on the master branch, this is
-                             // m_fullUpdateTime, but this implementation is correct.
-                             (time - forcedRuptureTime[ltsFace][pointIndex]) / drParameters.t0));
-    }
-    stateVariable[pointIndex] = std::max(stateVariable[pointIndex], f2);
-  }
-}
-
 void BiMaterialFault::copyLtsTreeToLocal(seissol::initializers::Layer& layerData,
                                          seissol::initializers::DynamicRupture* dynRup,
                                          real fullUpdateTime) {
--- a/src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
+++ b/src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
@@ -1,142 +1,64 @@
 #include "LinearSlipWeakening.h"
 namespace seissol::dr::friction_law {
-void LinearSlipWeakeningLaw::calcStrengthHook(FaultStresses& faultStresses,
-                                              std::array<real, misc::numPaddedPoints>& strength,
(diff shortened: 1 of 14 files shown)


ravil-mobile:
I guess we don't need `tn` anymore, right?
https://github.com/SeisSol/SeisSol/blob/f688be461f3bf710aabe1768f33998d15212ca58/src/Initializer/DynamicRupture.h#L175-L177

src/Initializer/DynamicRupture.h
```diff
@@ -174,13 +174,11 @@ struct seissol::initializers::LTS_LinearSlipWeakening : public seissol::initiali
 
 struct seissol::initializers::LTS_LinearSlipWeakeningForcedRuptureTime : public seissol::initializers::LTS_LinearSlipWeakening {
   Variable<real[dr::misc::numPaddedPoints]> forcedRuptureTime;
-  Variable<real> tn;
```
sebwolf-de, line 177: @ravil-mobile I removed this parameter here.

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
```diff
@@ -1,142 +1,64 @@
...
+  // modify strength according to prakash clifton
+  // see e.g.: Pelties - Verification of an ADER-DG method for complex dynamic rupture problems
+  prak_clif_mod(strength, sigma, localSlipRate, mu, deltaT);
+  // save for output
+  regularisedStrength[ltsFace][pointIndex] = strength;
+}
+
 /*
- * calculates strength
+ * calculates prakash clifton regularization
```
Thomas-Ulrich, line 59: Prakash-Clifton

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
```diff
@@ -1,142 +1,64 @@
...
+                                   real& sigma,
+                                   real& mu,
+                                   real& deltaT,
+                                   unsigned int ltsFace,
+                                   unsigned int pointIndex) {
+  // modify strength according to prakash clifton
+  // see e.g.: Pelties - Verification of an ADER-DG method for complex dynamic rupture problems
+  prak_clif_mod(strength, sigma, localSlipRate, mu, deltaT);
+  // save for output
+  regularisedStrength[ltsFace][pointIndex] = strength;
```
Thomas-Ulrich, line 55: why not returning the regularized strength in prak_clif_mod?

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
```diff
@@ -1,142 +1,64 @@
...
 }
 
+void BiMaterialFault::strengthHook(real& strength,
+                                   real& localSlipRate,
+                                   real& sigma,
+                                   real& mu,
+                                   real& deltaT,
+                                   unsigned int ltsFace,
+                                   unsigned int pointIndex) {
+  // modify strength according to prakash clifton
```
Thomas-Ulrich, line 51: Prakash-Clifton

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.h
```diff
@@ -139,20 +148,53 @@ class LinearSlipWeakeningBase : public BaseFrictionLaw<LinearSlipWeakeningBase<D
...
+    // the reference triangle with degree less or equal than CONVERGENCE_ORDER-1, and then evaluates
+    // the polynomial at the quadrature points
+    resampleKrnl.execute();
+
+    for (unsigned pointIndex = 0; pointIndex < misc::numPaddedPoints; pointIndex++) {
+      // integrate slip rate to get slip = state variable
+      this->accumulatedSlipMagnitude[ltsFace][pointIndex] +=
+          resampledSlipRate[pointIndex] * this->deltaT[timeIndex];
+
+      // Modif T. Ulrich-> generalisation of tpv16/17 to 30/31
```
Thomas-Ulrich, line 189: you can remove this line

Thomas-Ulrich (approved):
I wonder if having a specialiazation for ForcedRuptureTime is really worth it, I mean, can't we just keep this feature for all LinearSlipWeakeningLaws?
tpv6 benchmark does not have ForcedRuptureTime, but it would make sense to be able to use bimaterial with ForcedRuptureTime.
(then we would initialize ForcedRuptureTime with a very large number to disable it).

sebwolf-de:
> I wonder if having a specialiazation for ForcedRuptureTime is really worth it, I mean, can't we just keep this feature for all LinearSlipWeakeningLaws?

Sure, we could do that. At the same time, we have to include the nucleation stress option, which was recently added to the master branch. How does ForcedRuptureTime interact with the NucleationStress option?
Now, I see four options: `{LSW with ForcedRuptureTime, LSW with NucleationStress} X {LSW with Prakash-Clifton, LSW without Prakash-Clifton}`. Is that right? Standard behaviour (e.g. for tpv5) can be achieved by using `(ForcedRuptureTime, No Prakash-Clifton)` and `ForcedRuptureTime = +inf`.

Thomas-Ulrich:
>Now, I see four options: {LSW with ForcedRuptureTime, LSW with NucleationStress} X {LSW with Prakash-Clifton, LSW without Prakash-Clifton}.
I think so too.

sebwolf-de:
I made Forced Rupture Time the default now and deprecated FL16. Still, if no `forced_rupture_time` is given in `fault.yaml`, I set it to a big number, so we have the same behaviour as for FL2.
I would be very happy, if we could leave out Prakash-Clifton for this PR and take care of that in another PR.

Thomas-Ulrich (approved):
Thanks for implementing that!

Issue #597, opened 2022-07-03
SeisSol crash during installation on Shaheen II


palgunadi1993:
**Describe the bug**
SeisSol crashes during installation with the following error:
```bash
Solver/time_stepping/ActorState.cpp(109): internal error: assertion failed at: "shared/cfe/edgcpfe/lower_name.c", line 9585


compilation aborted for /project/k1587/kadek/myLibs/installSeisSol/SeisSol/src/Solver/time_stepping/ActorState.cpp (code 4)
make[2]: *** [CMakeFiles/SeisSol-lib.dir/build.make:366: CMakeFiles/SeisSol-lib.dir/src/Solver/time_stepping/ActorState.cpp.o] Error 4
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:196: CMakeFiles/SeisSol-lib.dir/all] Error 2
```

**Expected behavior**
Finish compilation

**To Reproduce**
1. Which version do you use? master branch 4357f5e07c021df3c78909b14d6978f2899ada96
2. Which build settings do you use? 
```bash
 ADDRESS_SANITIZER_DEBUG          OFF                                                                                                   
 ASAGI                            ON                                                                                                    
 CMAKE_BUILD_TYPE                 Release                                                                                               
 CMAKE_INSTALL_PREFIX             /usr/local                                                                                            
 COMMTHREAD                       OFF                                                                                                   
 COVERAGE                         OFF                                                                                                   
 DEVICE_ARCH                      none                                                                                                  
 DEVICE_BACKEND                   none                                                                                                  
 DYNAMIC_RUPTURE_METHOD           quadrature                                                                                            
 EQUATIONS                        elastic                                                                                               
 Eigen3_DIR                       /project/k1587/kadek/myLibs/cmakeSeisSolNew/share/eigen3/cmake                                        
 GEMM_TOOLS_LIST                  LIBXSMM,PSpaMM                                                                                        
 HDF5                             ON                                                                                                    
 HDF5_C_LIBRARY_dl                /usr/lib64/libdl.so                                                                                   
 HDF5_C_LIBRARY_hdf5              /project/k1587/kadek/myLibs/cmakeSeisSolNew/lib/libhdf5.a                                             
 HDF5_C_LIBRARY_hdf5_hl           /project/k1587/kadek/myLibs/cmakeSeisSolNew/lib/libhdf5_hl.a                                          
[...]

PR #598, opened 2022-07-04
Bump needed intel compiler

krenzland:
As dicussed in  #597


sebwolf-de:
Should we also include a check in the `CMakeLists.txt`? Then a user immediately sees, what is going wrong...

krenzland:
Maybe, but in a separate PR.

Thomas-Ulrich (approved):

PR #600, opened 2022-07-05
Make planar wave structure for acoustic expl

krenzland:
Note: This leads to a small difference in the error (and is conceptually cleaner)

### DIFF
--- a/src/Solver/Interoperability.cpp
+++ b/src/Solver/Interoperability.cpp
@@ -988,42 +988,14 @@ void seissol::Interoperability::initInitialConditions()
   auto initialConditionDescription = m_initialConditionType;
   if (m_initialConditionType == "Planarwave") {
     auto materialData = m_ltsLut.lookup(m_lts->material, 0);
-    std::array<double, 3> kVec = {M_PI, M_PI, M_PI};
-    std::vector<int> varField = {};
-    std::vector<std::complex<double>> ampField = {};
-#ifndef USE_POROELASTIC
-    if (materialData.local.mu <= 1e-15) {
-      //Elastic materials has the following wave modes:
-      //-P, N, N, N, N, N, N, N, P
-      //Here we impose the P mode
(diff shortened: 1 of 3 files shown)


src/Solver/Interoperability.cpp
```diff
@@ -987,12 +987,43 @@ void seissol::Interoperability::initInitialConditions()
...
+      varField = {1, 8};
+      ampField = {1.0, 1.0};
+    }
+#else
+    //Poroelastic materials have the following wave modes:
+    //-P, -S2, -S1, -Ps, N, N, N, N, N, Ps, S1, S2, P
+    //Here we impose -S1, -Ps and P
+    std::vector<int> varField = {2,3,12};
+    std::vector<std::complex<double>> ampField = {1.0, 1.0, 1.0};
+#endif
```
sebwolf-de, line 1014: Why did you move this initializer here? It is already implemented here (without the check for mu): https://github.com/SeisSol/SeisSol/blob/master/src/Physics/InitialField.h#L39-L51
IMHO this change spaghettifies the code.
I'd prefer, if you add you changes regard `mu` in `InitialField.h`.

sebwolf-de (changes requested):

src/Solver/Interoperability.cpp
krenzland: Because this doesn't work like you're suggesting. It doesn't make sense to pass the variable field explicitly and then change the fields for the acoustic wave equation implicitly. Maybe it'd make sense to add methods that return default parameters?

sebwolf-de: Yep, like a PlanarWaveFactory ;-)

sebwolf-de (approved):

sebwolf-de:
`double muBar = (c44 + c55 + c66) / 3.0;` ;-)

PR #604, opened 2022-07-08
Copy neighbor material also in case of dynamic rupture BC

sebwolf-de:
With this fix, we set the neighbor material correctly for all internal faces. In particular, this fixes the problem that at dynamic rupture interfaces, we used the wrong impedances to compute the Godunov state.
Verification: See attached picture of receiver 3 for the TPV6 benchmark. Blue: master, orange: this fix.
Estimated arrival time of the wave is ~3.8s (see SCEC screenshot)
Screenshot from 2022-07-08 13-55-03
Screenshot from 2022-07-08 13-54-40


Thomas-Ulrich (approved):

AliceGabriel:
Thanks!

PR #606, opened 2022-07-08
dr/fix tpv6

sebwolf-de:
Implement Prakash-Clifton regularization in C++.

### DIFF
--- a/Documentation/dynamic-rupture.rst
+++ b/Documentation/dynamic-rupture.rst
@@ -178,7 +178,7 @@ Friction law :code:`16` implements linear slip-weakening with a forced rupture t
 If you are only interested in linear slip weakening friction without forced rupture time, do not supply the parameter `forced_rupture_time` in the fault `yaml` file.
 Friction law :code:`6` uses Prakash-Clifton regularization for bimaterial faults.
 For friction law :code:`16`, we resample the slip rate in every step to suppress spurious oscillations.
-In the case of Prakash-Clifton regularization, we do not resample the slilp rate.
+In the case of Prakash-Clifton regularization, we do not resample the slip rate.
 
 
 
--- a/Documentation/dynamic-rupture.rst
+++ b/Documentation/dynamic-rupture.rst
@@ -177,8 +177,12 @@ Friction parameters:
 Friction law :code:`16` implements linear slip-weakening with a forced rupture time.
 If you are only interested in linear slip weakening friction without forced rupture time, do not supply the parameter `forced_rupture_time` in the fault `yaml` file.
 Friction law :code:`6` uses Prakash-Clifton regularization for bimaterial faults.
+For friction law :code:`16`, we resample the slip rate in every step to suppress spurious oscillations.
+In the case of Prakash-Clifton regularization, we do not resample the slilp rate.
 
-Examples of input files for the friction laws :code:`2` and :code:`16` are availbable in the :ref:`cookbook<cookbook overview>`.
+
+
+Examples of input files for the friction laws :code:`6` and :code:`16` are availbable in the :ref:`cookbook<cookbook overview>`.
 
 Linear slip weakening can be seen as a special case of rate-and-state friction with
 
--- a/src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
+++ b/src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
@@ -19,13 +19,9 @@ real BiMaterialFault::strengthHook(real& faultStrength,
   // see e.g.: Pelties - Verification of an ADER-DG method for complex dynamic rupture problems
   real newStrength =
       prak_clif_mod(regularisedStrength[ltsFace][pointIndex], faultStrength, localSlipRate, deltaT);
-  // regularisedStrength[ltsFace][pointIndex] = newStrength;
-  return newStrength;
-}
+  regularisedStrength[ltsFace][pointIndex] = newStrength;
 
-void BiMaterialFault::postHook(std::array<real, misc::numPaddedPoints>& strengthBuffer,
-                               unsigned int ltsFace) {
-  std::copy(strengthBuffer.begin(), strengthBuffer.end(), regularisedStrength[ltsFace]);
+  return newStrength;
 }
 
 real BiMaterialFault::prak_clif_mod(real& regularisedStrength,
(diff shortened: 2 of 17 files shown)


src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
```diff
@@ -1,6 +1,13 @@
 #include "LinearSlipWeakening.h"
 namespace seissol::dr::friction_law {
 
+void NoSpecialization::resampleSlipRates(
+    real (&resampledSlipRate)[dr::misc::numPaddedPoints],
+    real const (&slipRateMagnitude)[dr::misc::numPaddedPoints]) {
+  resampleKrnl.originalQ = slipRateMagnitude;
+  resampleKrnl.resampledQ = resampledSlipRate;
+  resampleKrnl.execute();
+}
```
ravil-mobile, line 10: Why did you move it here?

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
```diff
@@ -10,27 +17,19 @@ void BiMaterialFault::copyLtsTreeToLocal(seissol::initializers::Layer& layerData
   regularisedStrength = layerData.var(concreteLts->regularisedStrength);
 }
 
-void BiMaterialFault::strengthHook(real& strength,
+real BiMaterialFault::strengthHook(real& faultStrength,
                                    real& localSlipRate,
-                                   real& sigma,
-                                   real& mu,
                                    real& deltaT,
```
ravil-mobile, line 22: Why do we need to pass it by-reference? It is going to be the same if we do it by-value

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
sebwolf-de: `FL=6` only works nicely without resampling. Otherwise, we get some artefacts in the solution (totally unrelated to #604). So I need `FL=6` without resampling and `FL=16` with resampling, so I decided to move the resampling to the specialization.
`NoSpecialization`: leaves the strength as it was, and resamples the slip rate.
`BiMaterialFault`: regularizes the strength and leaves the slip rate as it was.
Awful, I know :D

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.h
```diff
@@ -157,30 +157,20 @@ class LinearSlipWeakeningLaw : public BaseFrictionLaw<LinearSlipWeakeningLaw<Spe
...
-    resampleKrnl.resample = init::resample::Values;
-    resampleKrnl.originalQ = this->slipRateMagnitude[ltsFace];
-    resampleKrnl.resampledQ = resampledSlipRate;
-
-    // Resample slip-rate, such that the state increment (slip) lies in the same polynomial space as
-    // the degrees of freedom resampleMatrix first projects LocSR on the two-dimensional basis on
-    // the reference triangle with degree less or equal than CONVERGENCE_ORDER-1, and then evaluates
-    // the polynomial at the quadrature points
-    resampleKrnl.execute();
+    specialization.resampleSlipRates(resampledSlipRate, this->slipRateMagnitude[ltsFace]);
```
Thomas-Ulrich, line 173: About this (L178):

```
// integrate slip rate to get slip = state variable
      this->accumulatedSlipMagnitude[ltsFace][pointIndex] +=
          resampledSlipRate[pointIndex] * this->deltaT[timeIndex];
```
Do you think resampledSlipRate could get negative because of the resampling?
Shouldn't we add a max(0, ) somewhere?

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
```diff
@@ -10,27 +17,19 @@ void BiMaterialFault::copyLtsTreeToLocal(seissol::initializers::Layer& layerData
...
                                    unsigned int ltsFace,
                                    unsigned int pointIndex) {
   // modify strength according to Prakash-Clifton
   // see e.g.: Pelties - Verification of an ADER-DG method for complex dynamic rupture problems
-  prak_clif_mod(strength, sigma, localSlipRate, mu, deltaT);
-  // save for output
-  regularisedStrength[ltsFace][pointIndex] = strength;
+  real expterm = std::exp(-(std::abs(localSlipRate) + drParameters.vStar) * deltaT /
```
Thomas-Ulrich, line 27: localSlipRate should not be negative. But maybe it is because of the resampling?
But in case it gets negative I would rather use max(0, localSlipRate) in the formula.

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.h
```diff
@@ -253,18 +252,24 @@ class BiMaterialFault {
   void copyLtsTreeToLocal(seissol::initializers::Layer& layerData,
                           seissol::initializers::DynamicRupture* dynRup,
                           real fullUpdateTime);
-  void strengthHook(real& strength,
+  /**
+   * The bimaterial Fault FL has been implemented without resampling on the master branch.
+   * Resampling of the sliprate introduces artificial oszillations into the solution, if we use it
```
Thomas-Ulrich, line 257: typo oszillations

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
Thomas-Ulrich: That's strange. It might be related to negative SR induced by the resampling combined with the abs(SR) in the PC routine.

Thomas-Ulrich (changes requested):

Documentation/dynamic-rupture.rst
```diff
@@ -177,8 +177,12 @@ Friction parameters:
 Friction law :code:`16` implements linear slip-weakening with a forced rupture time.
 If you are only interested in linear slip weakening friction without forced rupture time, do not supply the parameter `forced_rupture_time` in the fault `yaml` file.
 Friction law :code:`6` uses Prakash-Clifton regularization for bimaterial faults.
+For friction law :code:`16`, we resample the slip rate in every step to suppress spurious oscillations.
+In the case of Prakash-Clifton regularization, we do not resample the slilp rate.
```
Thomas-Ulrich, line 181: typo: slilp

Thomas-Ulrich (approved):

src/DynamicRupture/FrictionLaws/LinearSlipWeakening.cpp
Thomas-Ulrich: ok, then let's keep it as it is.

Thomas-Ulrich:
Yes, that is expected as max(0, resampledSlipRate[pointIndex])  is not applied on the master branch.
https://github.com/SeisSol/SeisSol/blob/master/src/Physics/Evaluate_friction_law.f90#L453
Ok then maybe in the future, the best would be to evaluate max(0, resampledSlipRate[pointIndex]) with the master  branch (which could be dr/cpp in the future), see if it produces better results on a simple benchmark, and use it or not.

ravil-mobile (approved):

PR #615, opened 2022-07-14
dr/merge master

sebwolf-de:
* tpv5 works
* CI probably broken
* Need to check, if new energy output still works.


ravil-mobile:
> ahaaa @ravil-mobile ;-) image

But, at that time there was not `actor` model :smile:

PR #629, opened 2022-07-22
adjust receiver variable names for better consistency

fabian-kutschera:
Within the free surface output of SeisSol we have the following output variables:

    v1, v2, v3: ground velocities, x y and z components
    u1, u2, u3: ground displacements, x y and z components

So far the three particle velocities within the off-fault receiver output files were named u, v, w. I would like to rename them to v1, v2, v3 for better consistency.


krenzland:
I am in favor of this change. Is there anyone who has a different opinion? This is a breaking change after all.

Thomas-Ulrich (approved):

PR #633, opened 2022-07-27
dr/comments 1

sebwolf-de:
Address the first round of comments from #625 
Note that `dr/cpp` is protected, so we need to work through PRs. @krenzland, can you have a look at this PR and either you or I resolve all comments in #625 after this PR has been merged into `dr/cpp`.


krenzland (approved):
LGTM

Thomas-Ulrich (commented):
late comments but might be yet useful

PR #635, opened 2022-07-29
dr/const

sebwolf-de:
Make everything const...
@ravil-mobile I did not touch the GPU implementation. Can you have another look?


krenzland (approved):
Thanks for the const(ant) good work!

PR #636, opened 2022-08-01
Added a common interface cmake target

ravil-mobile:
This PR removes the following code in CMakeLists.txt

https://github.com/SeisSol/SeisSol/blob/f009cf3479fce56b48dc4d3730fd4db4885ae572/CMakeLists.txt#L350-L361

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,7 @@ set(GENERATED_FILES_FOR_SEISSOL src/generated_code/subroutine.h
                                 src/generated_code/kernel.h
                                 src/generated_code/kernel.cpp)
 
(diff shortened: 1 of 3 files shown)


krenzland (commented):
I guess that looks reasonable. Does it have any downsides?

ravil-mobile:
> I guess that looks reasonable. Does it have any downsides?

In general, no. One thing, which I learnt, is that one cannot include non-existing directory to an imported target. So, CMake doesn't allow it. But, in general we shouldn't do it anyway.

sebwolf-de (approved):
LGTM. Is it ready to be merged?

PR #637, opened 2022-08-03
dr/fix merge chaos

sebwolf-de:
The gitlab pipeline is offline!
Unfortunately, during the last few merges, we introduced some bugs, which are fixed by this PR.

### DIFF
--- a/src/DynamicRupture/Misc.h
+++ b/src/DynamicRupture/Misc.h
@@ -92,6 +92,23 @@ inline double asinh(double x) { return std::log(x + std::sqrt(x * x + 1.0)); }
  * @param dip
  */
 void computeStrikeAndDipVectors(const VrtxCoords normal, VrtxCoords strike, VrtxCoords dip);
+
+/**
+ * Defines the indices under which one can find a specific quantity
+ * U, V, W: Velocities in x, y, z direction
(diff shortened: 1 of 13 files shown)


src/DynamicRupture/Misc.h
```diff
@@ -92,6 +92,23 @@ inline double asinh(double x) { return std::log(x + std::sqrt(x * x + 1.0)); }
...
+
+/**
+ * Defines the indices under which one can find a specific quantity
+ * U, V, W: Velocities in x, y, z direction
+ * N, T1, T2: traction in normal and fault aligned directions
+ * Note: can not be an enum class, otherwise we would need to use
+ * static_cast<size_t>(dr::misc::QuanityIndices::X), whenever we want to use
+ * one of the indices as an array subscript.
+ */
+enum QuantityIndices : size_t {
```
krenzland, line 104: Maybe 
namespace quantity_indices {
enum QuantityIndices
}
Then it's not typesafe but the indices are at least not in the global namespace. And you can do using namespace, so it's quite convenient. 

Here enum class is definitely the wrong choice, because this uses enum to define a selection of constants and not a, well, enum

ravil-mobile (approved):
everything is good from my side

krenzland (approved):

PR #641, opened 2022-08-12
dr/performance

sebwolf-de:
Sulawesi testcase from Training repository
214690 elements
SeisSol elastic O6
dr/cpp (9901e229) and master (c952a4a9) with the same quadrature rule (Stroud, 49 points per face)
4 nodes of SuperMUC-NG
```
Metric                           | LSW dr/cpp  | LSW master  | R&S dr/cpp  | R&S master
---------------------------------+-------------+-------------+-------------+------------
elapsed time (s)                 | 345.822     | 494.091     | 415.139     | 610.418
TFLOP/s after 20s                | 4.5365      | 3.17504     | 3.77853     | 2.57075
Load imbalance (%)               | 1.79832     | 3.04396     | 5.53968     | 6.3358
computeDynamicRupture (constant) | 7.49916e-05 | 1.49072e-04 | 1.57612e-04 | 2.55247e-04
```
LSW as well as R&S friction are faster with dr/cpp now.
@ravil-mobile should I wrap all new pragmas in `#ifndef ACL_DEVICE`?
Should we keep all the likwid and score-p instrumentation?


krenzland:
> @ravil-mobile should I wrap all new pragmas in `#ifndef ACL_DEVICE`?

Please don't... Better to define a macro that does declare simd for cpu but not for gpus?

src/DynamicRupture/FrictionLaws/BaseFrictionLaw.h
```diff
@@ -34,17 +36,31 @@ class BaseFrictionLaw : public FrictionSolver {
...
       common::precomputeStressFromQInterpolated(faultStresses,
                                                 impAndEta[ltsFace],
                                                 qInterpolatedPlus[ltsFace],
                                                 qInterpolatedMinus[ltsFace]);
+      LIKWID_MARKER_STOP("PrecomputeStress");
+      SCOREP_USER_REGION_END(myRegionHandle)
 
+      SCOREP_USER_REGION_BEGIN(
+          myRegionHandle, "computeDynamicRupturePreHook", SCOREP_USER_REGION_TYPE_COMMON)
+      LIKWID_MARKER_START("PreHook");
```
Thomas-Ulrich, line 51: why don't you use the same names ;-)
It is not a bit of overkill to have the profiling instructions for both LIKWID and Scorep?

src/DynamicRupture/FrictionLaws/FrictionSolverCommon.h
```diff
@@ -6,13 +6,41 @@
...
  * Contains common functions required both for CPU and GPU impl.
  * of Dynamic Rupture solvers. The functions placed in
  * this class definition (of the header file) result
  * in the function inlining required for GPU impl.
  */
+namespace seissol::dr::friction_law::common {
+/**
+ * Asserts whether all relevant arrays are properly aligned
+ */
+inline void checkAlignmentPreCompute(
```
Thomas-Ulrich, line 19: is it called anywhere?

Thomas-Ulrich (commented):
great that it is now faster, and I expect it will be even faster when changing the quadrature rule (and maybe also decrease the order of the quadrature rule).

sebwolf-de:
I checked Thomas' Palu setup (without water layer). Visually, master, dr-cpp(dunavant) and dr-cpp(stroud) produce the same results. Attached, you can find a scaling plot. Indeed, the new implementation is faster and the dunavant quadrature rule gives a little speedup for the higher node numbers. Indeed, there might be a little room for improvement with other quadrature rules, but I wouldn't expect too much of a difference at this point.
palu_scaling

src/DynamicRupture/FrictionLaws/FrictionSolverCommon.h
sebwolf-de: Good point, I forgot to add that :see_no_evil:

Thomas-Ulrich (approved):

PR #650, opened 2022-08-25
Update documentation

sebwolf-de:
The first-example section in the documentation was very outdated.

### DIFF
--- a/Documentation/environment-variables.rst
+++ b/Documentation/environment-variables.rst
@@ -28,10 +28,10 @@ Some environment variables related to checkpointing are described in the :ref:`C
 
 .. _optimal_environment_variables_on_supermuc_ng:
 
-Optimal environment variables on SuperMuc-NG
+Optimal environment variables on SuperMUC-NG
 --------------------------------------------
 
-On SuperMuc-NG, we recommend using SeisSol with async output in thread mode.
+On SuperMUC-NG, we recommend using SeisSol with async output in thread mode.
 That is SeisSol should be compiled with :code:`-DCOMMTHREAD=ON`, and then run with the environment variables proposed below.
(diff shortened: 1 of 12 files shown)


Thomas-Ulrich:
Thanks.
3D 2D and 2D could be also highlighted in the link:
image
missing is the energy output file.

><threads> is the number of OpenMP threads per MPI rank (we usually use the number of CPUs - 1, to reserve one CPU for communication).

--> <threads> is the number of OpenMP threads per MPI rank (typically the number of CPUs if seissol was compiled `-DCOMMTHREAD=OFF`, otherwise the number of CPUs-1).

We could also point to this file:
environment-variables.html?highlight=environment#optimal-environment-variables-on-supermuc

(which also deserves some cleaning)

Documentation/environment-variables.rst
```diff
@@ -21,47 +21,24 @@ In addition to the variables in SeisSol, the
...
+That is SeisSol should be compiled with :code:`-DCOMMTHREAD=ON`, and then run with the environment variables proposed below.
 Also, we recommend using hyperthreading capabilities (that is using 96 CPUs instead of 48. 2 threads out of 96 are used as communication threads).
 Here are some proposed environment variables, to be added prior to invoking SeisSol in your batch file:
 
 .. code:: bash
 
-   #SBATCH --nodes=
+   #SBATCH --nodes=<number of nodes>
    #SBATCH --ntasks-per-node=1 
    #SBATCH --cpus-per-task=96
```
Thomas-Ulrich, line 43: ```#SBATCH --cpus-per-task=96``` could (should?) be removed.
You could also add a link to https://seissol.readthedocs.io/en/latest/supermuc.html#running-seissol

Thomas-Ulrich (approved):

PR #652, opened 2022-08-25
Adding animation visualization tools

jrekoske:
This PR adds two Python utilities for visualization. These are particularly useful for quickly visualizing the output of small-scale simulations without having to open the output in ParaView.

1) `make_animation` creates an animation (GIF) of the free surface wave-field using a free surface output XDMF file, ex:

anim

2) `plot_imt_map` creates a map of a ground motion intensity metric using the output from `ComputeGroundMotionParametersFromSurfaceOutput_Hybrid.py`, ex:

pgv


jrekoske:
`make_animation` should probably work for 2D fault output as well, though I haven't used that yet.

postprocessing/visualization/tools/make_animation.py
```diff
@@ -0,0 +1,62 @@
...
+    parser.add_argument(
+        'fname', help='SeisSol surface (XDMF) output file.')
+    parser.add_argument(
+        'dataid', help='Which data component to use', default='v1')
+    parser.add_argument(
+        '--cmap', help='Matplotlib colormap', default='bwr')
+    parser.add_argument(
+        '--dpi', default=200, type=int)
+    parser.add_argument(
+        '--bounds', nargs='+', help='xmin xmax ymin ymax', type=float)
```
sebwolf-de, line 24: Maybe add an option to define the output filename.

postprocessing/visualization/tools/plot_imt_map.py
```diff
@@ -0,0 +1,67 @@
...
+    if args.contour:
+        plt.tricontourf(x, y, imt, vmin=vmin, vmax=vmax)
+    else:
+        plt.tripcolor(geo.T[0], geo.T[1], connect, imt, vmin=vmin, vmax=vmax)
+    if args.bounds:
+        plt.xlim(args.bounds[0], args.bounds[1])
+        plt.ylim(args.bounds[2], args.bounds[3])
+    plt.colorbar(label=clabel)
+    plt.xlabel('Easting (m)')
+    plt.ylabel('Northing (m)')
```
sebwolf-de, line 60: This only holds for ENU meshes. I'd prefer {x, y} here, since some user might use a different coordinate system.

postprocessing/visualization/tools/make_animation.py
```diff
@@ -0,0 +1,62 @@
...
+    tarr = np.arange(0.0, nt * dt, dt)
+
+    fig = plt.figure(figsize=(8, 7))
+    ax = plt.subplot(111)
+    dmax = abs(data).max()
+    im = ax.tripcolor(
+        geo.T[0], geo.T[1], connect, data[0], cmap=args.cmap, vmin=-dmax,
+        vmax=dmax)
+    ax.set_xlabel('Easting (m)')
+    ax.set_ylabel('Northing (m)')
```
sebwolf-de, line 42: This only holds for ENU meshes. I'd prefer {x, y} here, since some user might use a different coordinate system.

sebwolf-de (changes requested):
Hi, thanks for these scripts. I think they are particularly useful, since you can easily execute them on some cluster and then just download the final image. I have a few remarks, but nothing serious.

jrekoske:
Thanks I made the requested changes

sebwolf-de (approved):

PR #666, opened 2022-09-02
Update spectral_corner_frequency_from_surface_xdmf.py

NicoSchlw:
Fix eventDuration for constantly slipping faults and add a minimum frequency depending on the length of the time series

### DIFF
--- a/postprocessing/science/spectral_corner_frequency_from_surface_xdmf.py
+++ b/postprocessing/science/spectral_corner_frequency_from_surface_xdmf.py
@@ -128,7 +128,8 @@ def ComputeCornerFrequency(tuples):
     """Preproccesing of a single waveform before calling fc_Qzero_gridsearch to determine the corner frequency"""
     
     fc_arr = np.zeros(tuples[0].shape[0])
-    endfreq=args.maxFreq[0]
+    endfreq = args.maxFreq[0]
+    startfreq = np.max(minFreq, 0.01)
     
     for i in range(0, fc_arr.size):
         
@@ -152,7 +153,7 @@ def ComputeCornerFrequency(tuples):
         # integrate within the frequency domain and convert to amplitude spectrum
         data = np.abs(data*(1/(1j*2*np.pi*freqs)))
 
-        newfreqs = np.arange(0.01, endfreq+0.005, 0.005)
+        newfreqs = np.arange(startfreq, endfreq+0.005, 0.005)
         data = np.interp(newfreqs, freqs, data)
 
         fc_arr[i] = fc_Qzero_gridsearch(data, newfreqs, misfit="SSM", n=2, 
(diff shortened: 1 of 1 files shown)


postprocessing/science/spectral_corner_frequency_from_surface_xdmf.py
```diff
@@ -243,7 +244,15 @@ def ApproximateEventDurationAndHypocenter(faultxdmf, faultxyz):
         ASl = LoadSingleComponentParallel("ASl", surfaceData=False)
     ASR = np.abs(np.gradient(ASl, dtFault, axis=1))
     slippingElement = np.where(np.amax(ASR, axis=0) > slipRateThreshold, 1, 0)
-    duration = slippingElement[np.argmax(slippingElement):-np.argmax(np.flip(slippingElement))].size * dtFault
+    if np.all(slippingElement == 1):
+        print("Warning: at least one element is constantly slipping")
```
Thomas-Ulrich, line 248: you mean 'there is always one element slipping during the whole simulation'?
(not necessarily always the same?).

postprocessing/science/spectral_corner_frequency_from_surface_xdmf.py
```diff
@@ -243,7 +244,15 @@ def ApproximateEventDurationAndHypocenter(faultxdmf, faultxyz):
...
+        print("eventDuration is set to the entire length of the time series")
+        duration = (slippingElement.size - 1) * dtFault
+    elif np.argmax(np.flip(slippingElement)) == 0:
+        print("Warning: at least one element is still slipping at the last time step")
+        duration = slippingElement[np.argmax(slippingElement):-1].size * dtFault
+    else:
+        duration = slippingElement[np.argmax(slippingElement):-np.argmax(np.flip(slippingElement))].size * dtFault
```
Thomas-Ulrich, line 255: why is this formula not valid for all cases?

Thomas-Ulrich (approved):

postprocessing/science/spectral_corner_frequency_from_surface_xdmf.py
NicoSchlw: When there is still an element slipping at the last time step `-np.argmax(np.flip(slippingElement))` is zero, which means duration is also zero

postprocessing/science/spectral_corner_frequency_from_surface_xdmf.py
NicoSchlw: Yes, `"Warning: at least one element is always slipping"` is probably better phrased

postprocessing/science/spectral_corner_frequency_from_surface_xdmf.py
Thomas-Ulrich: What about np.trim_zeros(slippingElement).size?

NicoSchlw: Good idea, I changed it to np.trim_zeros

PR #676, opened 2022-09-26
Remove incorrect error

krenzland:
Remove check for group filter

This checked whether all elements are filtered out.
However, it only checks that the current rank has more than zero elements.
Hence, in MPI scenarios, this may crash.

Includes changes in #675, hence only merged after #675 has been merged.


Thomas-Ulrich (approved):
LGTM (I guess if numElems==0, the code does not crash?).

PR #686, opened 2022-10-06
Fix supermuc example job script

krenzland:
Formatting was broken by #683

### DIFF
--- a/Documentation/supermuc.rst
+++ b/Documentation/supermuc.rst
@@ -114,8 +114,10 @@ Running SeisSol
 ---------------
 
 This is an example job submission script for SeisSol on SuperMUC-NG. For your applications, change 
-:: #SBATCH --nodes= 
+#SBATCH --nodes= 
 to the amount of nodes you want to run on. A rule of thumb for opmital performance is to distribute your jobs to 1 node per 100k elements. This rule of thumb does not account for potentially shorter queue times, for example when using the test queue or when asking for a large amount of nodes. 
+
+::
 
   #!/bin/bash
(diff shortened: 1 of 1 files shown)


Documentation/supermuc.rst
```diff
@@ -114,9 +114,11 @@ Running SeisSol
 ---------------
 
 This is an example job submission script for SeisSol on SuperMUC-NG. For your applications, change 
-:: #SBATCH --nodes= 
+#SBATCH --nodes= 
 to the amount of nodes you want to run on. A rule of thumb for opmital performance is to distribute your jobs to 1 node per 100k elements. This rule of thumb does not account for potentially shorter queue times, for example when using the test queue or when asking for a large amount of nodes.
```
Thomas-Ulrich, line 118: typo : opmital
"is to distribute your jobs to 1 node per AT LEAST 100k elements"

Thomas-Ulrich (approved):

Documentation/supermuc.rst
krenzland: I'll fix the typo but the other statement is incorrect. You can get good performance with far fewer elements, depending on the LTS clustering (and poly order, etc)

Thomas-Ulrich: ok. I proposed to change because it reads as if we need 1 node per 500k elements would not lead to optimal performances.

krenzland: Yeah, that's true. 100k/elem per node is safer because then people don't try running with far more and then complain that they run out of memory...

PR #709, opened 2022-10-24
seissol-env on Frontera 

Thomas-Ulrich:
Document seissol-env installation on Frontera.
You will say, it starts to be a catalogue of supercomputer installation.
But there are always tricks (e.g. here precising the module in the compilers.yaml), which were not detailed before.


Thomas-Ulrich:
oh wait, there is the double nucleation in this commit

PR #710, opened 2022-10-24
Document Frontera seissol-env

Thomas-Ulrich:
Document seissol-env installation on Frontera.
You will say, it starts to be a catalog of supercomputer installations.
But there are always tricks (e.g., the module in the compilers.yaml) that were not detailed before.

### DIFF
--- a/Documentation/frontera.rst
+++ b/Documentation/frontera.rst
@@ -12,9 +12,9 @@ Add these lines to ``~/.bashrc``:
     module switch python3 python3/3.9.2
     module use /work2/09160/ulrich/frontera/spack/share/spack/modules/linux-centos7-cascadelake
     module load seissol-env-develop-intel-19.1.1.217-x52n3zf
-    export cc=mpiicc    
+    export cc=mpiicc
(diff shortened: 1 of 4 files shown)


krenzland (approved):

krenzland:
I don't see the problem with this. However, if we add more of these we should add a disclaimer that support is not guaranteed.

Documentation/frontera.rst
```diff
@@ -0,0 +1,83 @@
...
+Add these lines to ``~/.bashrc``:
+
+::
+
+    module switch python3 python3/3.9.2
+    module use /work2/09160/ulrich/frontera/spack/share/spack/modules/linux-centos7-cascadelake
+    module load seissol-env-develop-intel-19.1.1.217-x52n3zf
+    export cc=mpiicc    
+    export CXX=mpiicpc
+    export FN=mpiifort
```
NicoSchlw, line 17: FC?

Thomas-Ulrich: yes, thank you!

PR #711, opened 2022-10-25
Fix typo in Frontera doc

sebwolf-de:
`cc` in small letters does not have an effect, which is not a big problem, since we do not use the c compiler, but for completeness, I'd correct it.


Thomas-Ulrich:
thx, you can do the same change in shaheen.rst

Thomas-Ulrich (approved):

PR #712, opened 2022-10-25
fixes check on MPI thread support

Thomas-Ulrich:


Documentation/shaheen.rst
```diff
@@ -88,4 +88,53 @@ To access the module at start up, add to your ``~/.bashrc``:
...
+  export OMP_NUM_THREADS=31
+  # you could also consider OMP_NUM_THREADS=62 for high order
+  export OMP_PLACES="cores(31)"
+  #Prevents errors such as experience in Issue #691
+  export I_MPI_SHM_HEAP_VSIZE=8192
+
+  export XDMFWRITER_ALIGNMENT=8388608
+  export XDMFWRITER_BLOCK_SIZE=8388608
+
+  # required for using with commthread
```
krenzland, line 136: AFAIK this is now required in all cases. (The case distinction was fairly complex)

krenzland (changes requested):
Looks good! Two small things

Documentation/shaheen.rst
Thomas-Ulrich: on Shaheen, it was working fine without when not using commthread.

Thomas-Ulrich: So I've tested more:
if not specifying `MPICH_MAX_THREAD_SAFETY=multiple`
```
commthread + THREAD: crash
commthread + SYNC: crash
no commthread and THREAD: crash
no commthread + SYNC: ok
```

src/Parallel/MPI.h
```diff
@@ -111,9 +111,13 @@ class MPI : public MPIBasic
 		setComm(MPI_COMM_WORLD);
 
 		// Test this after setComm() to get the correct m_rank
-		if (required < provided)
+		if (provided < required) {
 			logWarning(m_rank) << utils::nospace << "Required MPI thread support (" << required
 				<< ") is smaller than provided thread support (" << provided << ").";
+#ifdef USE_COMM_THREAD
```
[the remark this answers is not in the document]
Thomas-Ulrich: I don't think so, see above.

krenzland:
To be honest I wouldn't want to re-introduce the complicated case distinction when we need async MPI. We basically need it for any setting used for production runs. Shaheen is the first system that I've ever seen where that configuration leads to a problem.

krenzland (approved):

PR #720, opened 2022-10-31
Switched CPU and GPU pipelines to gitlab-kubernetes-runners

ravil-mobile:
- Created a kubernetes cluster on LRZ compute cloud with two nodes: master (10 cores) & worker (20 cores)
- Set `static` CPU Management Policies to enforce `Guaranteed` Quality of Service Class which sets `cpu-set` for each pod (container)
- installed gitlab-runner with `helm chart`: 4 cores, 10 gb of RAM for each pod, concurrency level 5
- switched our GPU pipeline to a GPU kubernetes gitlab runner provided by TUMI05

Notes for the next CI update:
- let's use `make install` to copy all executables and libraries to a specific folder which we will copy to the next stage as `artifacts`
- @sebwolf-de, I guess we need to use checkpoints to drastically reduce run-times for `tpv*` tests
- move recorded solutions from `https://github.com/7bits-register/precomputed-seissol.git` to a sub-project of `SeisSol organization`

### DIFF
--- a/gitlab-cpu-pipeline.yml
+++ b/gitlab-cpu-pipeline.yml
@@ -1,6 +1,6 @@
 default:
     tags:
-        - 10core-docker-runner
+        - lrzcc-k8s-4cores-16gb 
     image:
         name: $CI_REGISTRY_USER/$cpu_image_name:$cpu_image_version
         entrypoint: [""]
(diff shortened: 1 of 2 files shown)


krenzland (approved):
LGTM

PR #726, opened 2022-11-07
Fixing typo in TPV24 example

JeenaYun:
There was a typo in the table showing the static and dynamic friction coefficient values: static was 0.12 and dynamic was 0.18, which should be the opposite. Fixed the values in the table accordingly.


Thomas-Ulrich (approved):
LGTM.
Thank you for fixing it!

PR #733, opened 2022-11-17
negative state variable check (Fix 668) + updated space weights initialization

sebwolf-de:
* Found a way to easier initialize the space weights
* Throw an error, if the state variable is negative


Thomas-Ulrich (approved):
I think that should work, while I m not 100%  (a few % not sure) that the state variable must be positive.

PR #747, opened 2022-12-01
Add install target to rconv

sebwolf-de:
Now you can use `make install` to install rconv and use it everywhere ;-)


krenzland (approved):
:+1: 

Is there a reason you're not simply using https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_RPATH_USE_LINK_PATH.html? Should do the same thing as you're doing manually

sebwolf-de:
I tried it, but it didn't work as expected (no idea why, though)

PR #748, opened 2022-12-02
Compute rotation for receiver output

sebwolf-de:
* Compute the rotation of the velocity fields at the receivers.
* Tested with planar wave
* Configurable in `parameters.par`, standard is off.


Thomas-Ulrich (approved):
Looks good to me (Well I did not understand what is  STP to be honest, and why there is this option in the receivers).
What about adding the formula of the rotation in the off-fault receiver documentation?

sebwolf-de:
STP, is the SpaceTimePredictor used for poroelastic materials.

Thomas-Ulrich (approved):
thx!

Issue #668, opened 2022-09-06
NaN detected with RS with Dunavant quadrature rule, not with Strout


Thomas-Ulrich:
**Describe the bug**
I'm using the sulawesi RS Training with a new mesh (fault a bit more segmented).
on v1.0.0-rc with Dunavant quadrature rule, I get:
```
Tue Sep 06 10:52:55, Info:  Writing faultoutput at time 0.
Tue Sep 06 10:52:55, Info:  Writing faultoutput at time 0. Done.
Tue Sep 06 10:52:55, Info:  Writing free surface at time 0.
Tue Sep 06 10:52:55, Info:  Writing free surface at time 0. Done.
Tue Sep 06 10:52:55, Info:  Writing energy output at time 0 
Tue Sep 06 10:52:55, Info:  Writing energy output at time 0 Done. 
Tue Sep 06 10:53:01, Error: Detected Inf/NaN in free surface output. Aborting. 
Tue Sep 06 10:53:01, Error: Detected Inf/NaN in free surface output. Aborting.
```

and no problem with Strout:
```
ue Sep 06 10:51:35, Info:  Writing free surface at time 0.
Tue Sep 06 10:51:35, Info:  Writing free surface at time 0. Done.
Tue Sep 06 10:51:35, Info:  Writing energy output at time 0
Tue Sep 06 10:51:36, Info:  Writing energy output at time 0 Done.
Tue Sep 06 10:51:36, Info:  Writing faultoutput at time 0.
Tue Sep 06 10:51:36, Info:  Writing faultoutput at time 0. Done.
Tue Sep 06 10:51:40, Info:  Writing free surface at time 0.25.
Tue Sep 06 10:51:40, Info:  Writing free surface at time 0.25. Done.
```

**Expected behavior**
no nan with Dunavant
**To Reproduce**

compiled with  ```cmake -DCOMMTHREAD=ON -DNUMA_AWARE_PINNING=ON -DASAGI=ON -DCMAKE_BUILD_TYPE=Release -DHOST_ARCH=skx -DPRECISION=double -DORDER=4 -DGEMM_TOOLS_LIST=LIBXSMM,PSpaMM ..```
with seissol-env/develop-intel21-impi-x2b on supermucNG.

setup: /hppfs/work/pr63qo/di73yeq4/Training_bug/sulawesi_bugRS

note: v1.0.0-rc setup with LSW and Dunavant works.

Thomas-Ulrich:
[...]

PR #752, opened 2022-12-07
remove positive state enforcement

Thomas-Ulrich:
Hi,
Following https://github.com/SeisSol/SeisSol/discussions/746 (while the discussion did not really come to a conclusion)
I propose removing the max on the state variable and documenting this change.
Note that sem2dpack does not enforce positivity of the state variable either:
https://github.com/jpampuero/sem2dpack/blob/master/SRC/bc_dynflt_rsf.f90
I would create v1.1.0 SeisSol afterward (tagging the version).

It should not require recomputing a solution of tpv103, as the state never gets negative in this benchmark.


sebwolf-de:
How does this influence #668?

Thomas-Ulrich:
> How does this influence #668?

this does not fix it.

Thomas-Ulrich:
But I think it then makes sense to allow negative initial states (But then flood the user with warnings :-))

sebwolf-de:
Okay, for #668 we also came to the conclusion that a zero traction is unphyiscal in the R&S framework, that's why we needed the omega > 0...

Documentation/dynamic-rupture.rst
```diff
@@ -282,7 +282,9 @@ with
   \end{aligned}.
 
 
-
+Note that between commits b014d1180d53a5231e92bb5062a45711fea9aade (merging #306) of March 17th, 2021, and v1.1.0, the state variable was enforced positive in this friction law.
```
sebwolf-de, line 285: I wouldn't hard code the versions here, just mention the PRs (#306 and #752).

sebwolf-de (commented):
LGTM, but I wouldn't add a v1.1.0 yet. After all we haven't really released v1.0.0 yet, so this would be v1.0.0-rc2 probably.

Thomas-Ulrich:
> v1.0.0-rc2

So what feature is needed to release v1.0.0?

sebwolf-de (approved):

sebwolf-de:
Push new references to the `precomputed-seissol` repository.
* Checkout https://github.com/7bits-register/precomputed-seissol
* Run TPV105 again
* Check if output is reasonable.
* Update https://github.com/7bits-register/precomputed-seissol/tree/master/tpv105/precomputed/double (+ single directory)
* Push
* Add commit id https://github.com/SeisSol/SeisSol/blob/master/gitlab-cpu-pipeline.yml#L113 and https://github.com/SeisSol/SeisSol/blob/master/gitlab-gpu-pipeline.yml#L149

Thomas-Ulrich:
I just check tpv105-3D, and the state indeed gets negative but on a very narrow zone at shallow depth (e.g. state variable, saturated colormap, t=9s)
image

sebwolf-de:
Huh, that's interesting. But the CI only runs for 2s, so this should not be the problem..

sebwolf-de:
I compared the CI output:
With positive state variable enforced:
Screenshot from 2022-12-15 10-48-09
With negative state variable allowed:
Screenshot from 2022-12-15 10-48-05
Note that both images share the same colorbar, in the upper picture, the blue circle is exactly at zero, in the lower picture, the blue circle is slightly below zero.

Thomas-Ulrich:
Yes, I ve compared the folder, and I only found this problem:
https://github.com/SeisSol/Examples/pull/57
Else the inputs are similar (well SlipRateOutputType=0 in the CI, but I checked that it is not the problem).
So at the end, I think the state is negative because of the coarse resolution.

PR #753, opened 2022-12-07
Fix fault output scheduling

### LINKED ISSUE #693: Fault receivers are not printed (SeisSol GPU)
**Describe the bug**
When running with `ClusteredLTS = 2` the code doesn't print the output for the faultreceivers file, except for the 0s time.

Schermata 2022-10-10 alle 15 25 58

**Expected behavior**
Expect to print the faultreceivers output for every time step

**To Reproduce**
1. Used version ` 202103_Sumatra-1090-g7ae55cfa  `   The branch is ravil/energy-output 
2. Build setting for GPU with CUDA. SeisSol_RelWithDebInfo_ssm_70_cuda_6_elastic. With MPI compilers.
   Marconi100.   

Schermata 2022-10-10 alle 15 40 10

[parameters.txt](https://github.com/SeisSol/SeisSol/files/9746405/parameters.txt)

Many thanks in advance.

krenzland:
Write fault output in first cluster which actually has dr faces.
Fix #693.


src/Solver/time_stepping/ActorState.cpp
```diff
@@ -105,5 +105,10 @@ void DynamicRuptureScheduler::setLastFaultOutput(long steps) {
 bool DynamicRuptureScheduler::hasDynamicRuptureFaces() const {
   return numberOfDynamicRuptureFaces > 0;
 }
+
+
+bool DynamicRuptureScheduler::getIsFirstDynamicRuptureCluster() const {
```
ravil-mobile, line 110: Maybe it is better to rename `getIsFirstDynamicRuptureCluster` to `isFirstDynamicRuptureCluster`

src/Solver/time_stepping/TimeCluster.cpp
```diff
@@ -695,7 +695,7 @@ void TimeCluster::correct() {
   // First cluster calls fault receiver output
   // Call fault output only if both interior and copy parts of DR were computed
   // TODO: Change from iteration based to time based
-  if (m_clusterId == 0
+  if (dynamicRuptureScheduler->getIsFirstDynamicRuptureCluster()
```
ravil-mobile, line 698: Another idea is to name the method as `isFirstNotEmptyDynamicRuptureCluster`. The name will give a hint to the others (in the future) why `m_clusterId == 0` was incorrect

ravil-mobile (approved):

PR #754, opened 2022-12-07
Remove asserts for zero updates - broken for empty clusters

krenzland:
Zero flops is a valid amount of flops for operations on empty clusters


sebwolf-de:
Indeed this is true. Can you add a `>=`? After all, the flops should not decrease...

PR #755, opened 2022-12-07
Optimizations for Fugaku: Pinning, Compile flags

krenzland:
Draft: Needs to be merged with master, new pinning routines need to be checked on other clusters.

This PR adds changes needed to run on Fugaku. This includes adding support for NUMA aware pinning in cases when CPU cores are deactivated using the hot-plugging system of Linux: https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu


Thomas-Ulrich (commented):
LGTM

Thomas-Ulrich (approved):

PR #769, opened 2022-12-23
increase size of OutputFile, add nan checker

Thomas-Ulrich:
- 60 is sometimes not enough, in particular because scratch (/hppfs/scratch/0A/di73yeq4/) takes already half of it.
- nan checker on energies make sense as energies traverse the all mesh (and we got them for free).
(I also applied clang-format on the file)


sebwolf-de (approved):
LGTM

PR #772, opened 2023-01-10
Fix averaging for acoustic material

krenzland:


krenzland:
This works for purely acoustic and purely elastic material. IMO the way I'm doing it is also reasonable for the mixed-case (which should always be meshed explicitly anyway!)

sebwolf-de (approved):
LGTM, thanks

PR #779, opened 2023-01-19
Fix constants for ocean wavemode 0

krenzland:


Thomas-Ulrich (approved):
I suppose you know what you are doing ^^

Issue #160, opened 2019-11-11
Implement a warning if not all boundary conditions are set


krenzland:
Currently, if a user forgets to specify a boundary condition, SeisSol silently crashes after computing the LTS weights. 
We should add a warning, e.g. in https://github.com/SeisSol/SeisSol/blob/master/src/Geometry/PUMLReader.cpp#L298

daisy20170101:
Good idea! it helps to clarify errors.

Thomas-Ulrich:
#43

Thomas-Ulrich:
note also that if your model has a boundary tagged DR (e.g 3) and you use a purely kinematic source
``` 
&Boundaries
BC_fs = 1
BC_dr = 0
BC_of = 1
/
``` 
and not &DynamicRupture namelist, you get a Segmentation fault:
```
Thu Mar 05 14:17:03, Info:  Finding meshIds for receivers... 
Thu Mar 05 14:17:03, Warn:  "../output_Ridgecrest/test-surface_cell.h5" already exists. Creating backup. 
Thu Mar 05 14:17:03, Warn:  "../output_Ridgecrest/test-surface_vertex.h5" already exists. Creating backup. 
Thu Mar 05 14:17:03, Info:  Cleaning possible double occurring receivers for MPI... 
Thu Mar 05 14:17:04, Info:  Initializing free surface output. Done. 
Thu Mar 05 14:17:04, Info:  Mapping receivers to LTS cells... 
Rank:        0 | Info    | <--------------------------------------------------------->
Rank:        0 | Info    | <     Start calc_SeisSol ...                              >
Rank:        0 | Info    | <--------------------------------------------------------->
Thu Mar 05 14:17:07, Info:  Waiting for last wave field. 
Thu Mar 05 14:17:07, Info:  Writing wave field at time 0.
Thu Mar 05 14:17:07, Info:  Writing wave field at time 0. Done.
Thu Mar 05 14:17:07, Info:  Writing free surface at time 0.
Thu Mar 05 14:17:07, Info:  Writing free surface at time 0. Done.
Thu Mar 05 14:17:07, Info:  #max-updates since sync:  0  @  0 
srun: error: i01r01c04s07: task 15: Segmentation fault (core dumped)
srun: Terminating job step 516482.0
srun: error: i01r01c01s01: task 0: Segmentation fault (core dumped)
srun: error: i01r01c01s03: task 2: Segmentation fault (core dumped)
[...]

Issue #778, opened 2023-01-19
Crash using WL (fully coupled DR-tsunami) without Error Message


AnikoWirp:
**Describe the bug**
I run a fully coupled model (with waterlayer) with the latest SeisSol (O2, single precision), and a mesh size of 243 million elements on NG (700 nodes).

It fails without any error message.
The *.out file gives: 

```
700
Wed Jan 18 18:20:20, Info:  Welcome to SeisSol
Wed Jan 18 18:20:20, Info:  Copyright (c) 2012-2021, SeisSol Group
Wed Jan 18 18:20:20, Info:  Built on: Jan 18 2023 16:07:00
Wed Jan 18 18:20:20, Info:  Version: 201807_Kaikoura-2164-gcc370a6a
Wed Jan 18 18:20:20, Info:  Running on: i03r01c01s02
Wed Jan 18 18:20:20, Info:  Using MPI with #ranks: 700
Wed Jan 18 18:20:20, Info:  Using OMP with #threads/rank: 94
Wed Jan 18 18:20:20, Info:  OpenMP worker affinity (this process): "0123456789|0123456789|0123456789|0123456789|0123456-89|0123456789|0123456789|0123456789|0123456789|01234-"
Wed Jan 18 18:20:20, Info:  OpenMP worker affinity (this node)   : "0123456789|0123456789|0123456789|0123456789|0123456-89|0123456789|0123456789|0123456789|0123456789|01234-"
Wed Jan 18 18:20:20, Info:  Communication thread affinity        : "----------|----------|----------|----------|-------7--|----------|----------|----------|----------|-----5"
Wed Jan 18 18:20:20, Info:  The stack size ulimit is  8192 [kb].
Wed Jan 18 18:20:21, Info:  Input Parameters:
 0x1d046f0
Rank:        0 | Info    | <--------------------------------------------------------->
Rank:        0 | Info    | <                SeisSol MPI initialization               >
Rank:        0 | Info    | <--------------------------------------------------------->
Rank:        0 | Info    |  Double precision used for real.
Rank:        0 | Info    | <--------------------------------------------------------->
 INFORMATION: The assumed unit number is           6 for stdout and           0
 for stderr.
              If no information follows, please change the value.
Rank:        0 | Info    | <--------------------------------------------------------->
Rank:        0 | Info    | <     Start ini_SeisSol ...                               >
Rank:        0 | Info    | <--------------------------------------------------------->
[...]

Issue #751, opened 2022-12-06
Segmentation Fault Error during Supermuc seissol run at  Initialize Friction Model step


Thomas-Ulrich:
### Discussed in https://github.com/SeisSol/SeisSol/discussions/742

<div type='discussions-op-text'>

<sup>Originally posted by **jwjeremy** November 25, 2022</sup>
**Describe the bug:**
Segmentation fault error encountered with running seissol in supermuc. The stout shows the segmentation error and the sterr is empty about the error. Please see the attached stout file.

The error is probably due to the mesh as previous mesh files work well with the parameter files and the rest of the settings. 

**Expected behavior:**
To be able to initialize the fault interface

**To Reproduce:**
The seissol setup and mesh files are in /hppfs/work/pn49ha/di75yik/TOHOKU_FL33
The Mesh can also be seen here. 
https://drive.google.com/drive/folders/1B0VY-Hh00t7bIJwGJN7zJ_Dd2Lv2IXdA?usp=sharing

[2461877.Tohoku_docker_test1.err.txt](https://github.com/SeisSol/SeisSol/files/10094820/2461877.Tohoku_docker_test1.err.txt)
[2461877.Tohoku_docker_test1.out.txt](https://github.com/SeisSol/SeisSol/files/10094821/2461877.Tohoku_docker_test1.out.txt)
</div>

sebwolf-de:
```
cp -r /hppfs/work/pn49ha/di75yik/TOHOKU_FL33 .
cp: cannot stat '/hppfs/work/pn49ha/di75yik/TOHOKU_FL33': Permission denied
``` 
Unfortunately, I don't have access to the folder...

Thomas-Ulrich:
He changed the permission and should have access now.
Here is what I get when running in Debug mode:
```
Wed Dec 07 11:53:59, Info:  Initialize Friction Model 
SeisSol_Debug_dskx_4_elastic: /hppfs/work/pr45fi/di73yeq4/SeisSol/src/Initializer/time_stepping/LtsLayout.h:213: unsigned int seissol::initializers::time_stepping::LtsLayout::getLocalClusterId(unsigned int): Assertion `l_clusterId < m_localClusters.size()' failed.
```

sebwolf-de:
I get the same error. It is not a problem with the dynamic rupture parts, but with the LTS setup. Just the first node is already ahead and prints the `Initialize Friction Model` message early. It doesn't work with `v0.9.0` either.

[...]

PR #781, opened 2023-01-23
Include check for mesh correctness

sebwolf-de:
Detect errors like in #751 or #778 early and give a proper error message to the user.
Merge after https://github.com/TUM-I5/PUML2/pull/7

### DIFF
--- a/src/Geometry/PUMLReader.h
+++ b/src/Geometry/PUMLReader.h
@@ -92,40 +92,40 @@ private:
  * Possible types of boundary conditions for SeisSol.
  */
 enum class BCType {
-  internal, external, unknown
+	internal, external, unknown
 };
 
-/*
+/**
  * Decodes the boundary condition tag into a BCType.
  */
 inline BCType bcToType(int id) {
-  if (id == 0 || id == 3 || id > 64) {
-    return BCType::internal;
-  } else if (id == 1 || id == 2 || id == 5 || id == 6) {
-    return BCType::external;
-  } else {
-    return BCType::unknown;
-  }
+	if (id == 0 || id == 3 || id > 64) {
+		return BCType::internal;
+	} else if (id == 1 || id == 2 || id == 5 || id == 6) {
+		return BCType::external;
+	} else {
+		return BCType::unknown;
+	}
 }
 
-/*
+/**
  * Decodes the boundary condition tag into a string representation.
  */
 inline std::string bcToString(int id) {
-  if (id == 0) { return std::string("regular"); }
-  else if (id == 1) { return std::string("free surface"); }
-  else if (id == 2) { return std::string("free surface with gravity"); }
-  else if (id == 3) { return std::string("dynamic rupture"); }
-  else if (id == 5) { return std::string("absorbing"); }
-  else if (id == 6) { return std::string("periodic"); }
-  else if (id > 64) {
-    std::stringstream s;
-    s << "fault-tagging (" << id << ")";
-    return s.str();
-  } else {return std::string(""); }
+	if (id == 0) { return std::string("regular"); }
+	else if (id == 1) { return std::string("free surface"); }
+	else if (id == 2) { return std::string("free surface with gravity"); }
+	else if (id == 3) { return std::string("dynamic rupture"); }
+	else if (id == 5) { return std::string("absorbing"); }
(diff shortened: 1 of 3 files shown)


src/Geometry/PUMLReader.h
```diff
@@ -94,6 +88,81 @@ class PUMLReader : public MeshReader
...
+ * Check if the mesh is locally correct:
+ * - if a face is an internal face, the neighbor has to exist.
+ * - if a face is an external face, the neighbor must not exist.
+ * @param face: face to check
+ * @param cellNeighbors: ids of the neighboring cells
+ * @param side: side of the tetrahedron to check
+ * @param sideBC: boundary condition tag at the side to check
+ * @param cellIdAsInFile: Original cell id as it is given in the h5 file
+ */
+inline bool checkMeshCorrectnessLocally(PUML::TETPUML::face_t face, int* cellNeighbors, int side, int sideBC, int cellIdAsInFile) {
```
Thomas-Ulrich, line 138: checkBoundaryConditionsOfCell?

src/Geometry/PUMLReader.h
```diff
@@ -94,6 +88,81 @@ class PUMLReader : public MeshReader
 	static int FIRST_FACE_VERTEX[4];
 };
 
+/*
+ * Possible types of boundary conditions for SeisSol.
+ */
+enum class BCType {
+  internal, external, unknown
```
Thomas-Ulrich, line 95: consider (or not) using the same format (with Tabs) as the initial file.

Thomas-Ulrich (approved):
Looks good, see proposed change yet.

src/Geometry/PUMLReader.h
sebwolf-de: Hmm, I don't really agree, IMHO `checkBoundaryConditionsOfCell` could also imply that we only check that some BC is set. I think Correctness is better here, since it implies that there are some correct and some incorrect meshes.

Thomas-Ulrich: ok.
my point is that this function only checks one face of the mesh (but I guess this is what is meant by Locally).
Let's keep it as it is then.

Thomas-Ulrich (approved):

Thomas-Ulrich:
fixes #160

src/Geometry/PUMLReader.h
```diff
@@ -94,6 +88,81 @@ class PUMLReader : public MeshReader
...
+	else if (id == 1) { return std::string("free surface"); }
+	else if (id == 2) { return std::string("free surface with gravity"); }
+	else if (id == 3) { return std::string("dynamic rupture"); }
+	else if (id == 5) { return std::string("absorbing"); }
+	else if (id == 6) { return std::string("periodic"); }
+	else if (id > 64) {
+		std::stringstream s;
+		s << "fault-tagging (" << id << ")";
+		return s.str();
+	} else { return std::string(""); }
```
krenzland, line 125: This should include the id of the cell - right now you're missing e.g. bc 7 (analytical/dirichlet). I also have some branches with internal boundary conditions.

src/Geometry/PUMLReader.h
```diff
@@ -94,6 +88,81 @@ class PUMLReader : public MeshReader
...
+};
+
+/**
+ * Decodes the boundary condition tag into a BCType.
+ */
+inline BCType bcToType(int id) {
+	if (id == 0 || id == 3 || id > 64) {
+		return BCType::internal;
+	} else if (id == 1 || id == 2 || id == 5 || id == 6) {
+		return BCType::external;
```
krenzland, line 105: Misses 4/7: https://github.com/SeisSol/SeisSol/blob/master/src/Initializer/BasicTypedefs.hpp#L50

krenzland (changes requested):
Looks good, but you missed bcs 4/7

sebwolf-de:
> I also have some branches with internal boundary conditions.
Which branches? `BasicTypedefs.hpp` already includes a hint to update the PUMLReader ;-)

krenzland (approved):

krenzland:
> > I also have some branches with internal boundary conditions.
> 
> Which branches? `BasicTypedefs.hpp` already includes a hint to update the PUMLReader ;-)

Good question - I'll dig through them at some later date...

PR #788, opened 2023-02-06
Ravil/mini-seissol-gpu

ravil-mobile:
- added mini-seissol for GPUs
- added mini-seissol writer
- added threads-pinning writer
- updated parallel::MPI


src/Solver/time_stepping/MiniSeisSol.cpp
```diff
@@ -43,10 +43,49 @@
...
+    if (config.numElements < 1) {
+      throw std::runtime_error("expecting a positive integer number");
+    }
+  }
+  catch (std::runtime_error& err) {
+    logWarning(rank) << "failed to read `SEISSOL_MINI_NUM_ELEMENTS`," << err.what();
+    config.numElements = numElements;
+  }
+  return config;
+}
```
sebwolf-de, line 83: Should we really use environment variables here? IMHO everything should be in one place, so I'd prefer to put these values in `parameters.par` if there is not a big argument for having them as environment variables.

sebwolf-de (commented):
Hi, LGTM, except for one comment.

src/Solver/time_stepping/MiniSeisSol.cpp
ravil-mobile: We can do it but I would do it in a separate PR. 

There are 2 options in general, namely: 1) read parameters from fortran and then pass it from it (ugly); 2) read from YAML::Node.

The second is better. However, it would required to write a dedicated function (e.g., `readSeiSolMiniFromYaml`). I can see that that we have a couple of such functions. The best way is to do refactoring and put them in one place.    

I've already started prototyping the refactoring but the size of the PR is growing.

I would prefer to separate PRs. What do you think?

src/Parallel/MPIBasic.h
```diff
@@ -54,9 +54,15 @@ class MPIBasic
 	/** This rank */
 	int m_rank;
 
+	/** This rank */
+	int m_sharedMemMpiRank;
+
 	/** Number of processors */
 	int m_size;
 
+	/** Number of processors */
```
Thomas-Ulrich, line 63: Consider updating:
number of ranks in the shared memory sub-communicator

src/Parallel/MPIBasic.h
```diff
@@ -54,9 +54,15 @@ class MPIBasic
 	/** This rank */
 	int m_rank;
 
+	/** This rank */
```
Thomas-Ulrich, line 57: Consider updating:
rank in the shared memory sub-communicator

src/Parallel/MPIBasic.h
```diff
@@ -81,6 +87,14 @@ class MPIBasic
 		return m_rank;
 	}
 
+	/**
+	 * @return The rank of this process within this node
```
Thomas-Ulrich, line 91: Consider updating:
return the rank within the shared memory sub-communicator

src/Parallel/MPIBasic.h
```diff
@@ -89,6 +103,13 @@ class MPIBasic
 		return m_size;
 	}
 
+	/**
+	 * @return The rank of this process within this node
```
Thomas-Ulrich, line 107: Consider updating:
return the number of ranks within the shared memory sub-communicator

ravil-mobile: done

src/Parallel/MPIBasic.h
ravil-mobile: done

src/Parallel/MPIBasic.h
ravil-mobile: done

src/Parallel/MPIBasic.h
ravil-mobile: done

ravil-mobile:
Hi @Thomas-Ulrich, thanks for your suggestions. Comments are updated

Thomas-Ulrich (approved):
LGTM

PR #793, opened 2023-02-10
Added changes required to compile with gcc@11.2

ravil-mobile:


krenzland (approved):

krenzland:
Look good. Why doesn't the enum with size_t as basetype work anymore? Or was this just a case of a std::size_t instead of size_t?

ravil-mobile:
> Look good. Why doesn't the enum with size_t as basetype work anymore? Or was this just a case of a std::size_t instead of size_t?

I have no idea. It is ok for some compilers and for some reasons prohibited in others. it `turned out that `amdclang` had the same issue.

PR #802, opened 2023-02-22
Remove exclusive_scan due to bad complier support

krenzland:
Exclusive_scan has pretty bad compiler support, e.g. does not work on Frontera (Icc 19.1)


ravil-mobile (approved):

ravil-mobile:
> Exclusive_scan has pretty bad compiler support, e.g. does not work on Frontera (Icc 19.1)

Agree. I also experienced it with the amdclang compiler

PR #809, opened 2023-02-28
Fixes strength formula in documentation, and code-block

Thomas-Ulrich:
(to be consistence with the code:
https://github.com/SeisSol/SeisSol/blob/master/src/DynamicRupture/FrictionLaws/LinearSlipWeakening.h#L146)


AliceGabriel (approved):
important fix

PR #814, opened 2023-03-08
Fix profiling

krenzland:
- Ignore time spent in actor states in summary
- Remove empty/unused states
- Fix time spent in DR kernels


ravil-mobile (approved):

krenzland:
The pipeline fails but this is not caused by the PR but rather by a bad environment. Trying again...

PR #815, opened 2023-03-10
Add clustering writer

krenzland:
This PR adds a writer that outputs information about the (local) clusters, including their id, size, number of dynamic rupture faces, etc.
Furthermore, it uses a "profilingid" for the loop information, which allows us to differentiate between the costs of computing copy and interior clusters.
This is useful, as copy clusters are typically quite small and thus have larger loops which do not gain that much from shared memory parallelization.

### DIFF
--- a/src/Initializer/MemoryManager.h
+++ b/src/Initializer/MemoryManager.h
@@ -83,6 +83,7 @@
 #include <Initializer/LTS.h>
 #include <Initializer/tree/LTSTree.hpp>
 #include <Initializer/DynamicRupture.h>
+#include <Initializer/InputAux.hpp>
 #include <Initializer/Boundary.h>
 #include <Initializer/ParameterDB.h>
 #include <Initializer/time_stepping/LtsParameters.h>
@@ -358,11 +359,7 @@ class seissol::initializers::MemoryManager {
     }
 
     std::string getOutputPrefix() const {
-      const auto outputDirectory = (*m_inputParams)["output"]["outputfile"];
-      if (!outputDirectory) {
-          logError() << "No outputfile given.";
-      }
-      return outputDirectory.as<std::string>();
+      return getUnsafe<std::string>((*m_inputParams)["output"], "outputfile");
     }
 
 #ifdef ACL_DEVICE
--- a/src/Initializer/MemoryManager.h
+++ b/src/Initializer/MemoryManager.h
@@ -357,6 +357,14 @@ class seissol::initializers::MemoryManager {
       ltsParameters = std::make_shared<time_stepping::LtsParameters>(time_stepping::readLtsParametersFromYaml(m_inputParams));
     }
 
+    std::string getOutputPrefix() const {
+      const auto outputDirectory = (*m_inputParams)["output"]["outputfile"];
+      if (!outputDirectory) {
+          logError() << "No outputfile given.";
+      }
+      return outputDirectory.as<std::string>();
+    }
+
 #ifdef ACL_DEVICE
   void recordExecutionPaths(bool usePlasticity);
 
--- a/src/ResultWriter/ClusteringWriter.cpp
+++ b/src/ResultWriter/ClusteringWriter.cpp
@@ -1,0 +1,73 @@
+#include "ClusteringWriter.h"
+
+#include <filesystem>
+#include <fstream>
+
+#include <Parallel/MPI.h>
+namespace seissol::writer {
+
+ClusteringWriter::ClusteringWriter(const std::string& outputPrefix) : outputPrefix(outputPrefix) {
+
+}
+
+void ClusteringWriter::addCluster(unsigned profilingId,
+                                  unsigned localClusterId,
+                                  LayerType layerType,
+                                  unsigned size,
+                                  unsigned dynRupSize) {
+  clusteringInformation.profilingIds.push_back(profilingId);
+  clusteringInformation.localClusterIds.push_back(localClusterId);
+  clusteringInformation.layerTypes.push_back(layerType);
+  clusteringInformation.sizes.push_back(size);
+  clusteringInformation.dynamicRuptureSizes.push_back(dynRupSize);
+}
+
+void ClusteringWriter::write() const {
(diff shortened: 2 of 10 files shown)


src/Solver/time_stepping/TimeCluster.cpp
```diff
@@ -281,7 +283,7 @@ void seissol::time_stepping::TimeCluster::computeDynamicRupture( seissol::initia
   LIKWID_MARKER_STOP("computeDynamicRuptureFrictionLaw");
   }
 
-  m_loopStatistics->end(m_regionComputeDynamicRupture, layerData.getNumberOfCells(), m_globalClusterId);
+  m_loopStatistics->end(m_regionComputeDynamicRupture, layerData.getNumberOfCells(), m_profilingId);
```
Thomas-Ulrich, line 286: I just did a quickly read this PR which reads well (and I have no idea about the interest of the profiling data (and I don't want to hear more :-) ).
But here you change m_globalClusterId by m_profilingId = l_globalClusterId + offsetMonitoring?
if offsetMonitoring is non zero, this is not just a variable change?

krenzland: `const auto offsetMonitoring = type == Interior ? 0 : m_timeStepping.numberOfGlobalClusters;`
This allows us to have statistics for interior and copy clusters.

(One of the interesting possible results of this profiling is checking if statements such as "the cost of the dynamic rupture changes significantly over time" are true. The other, main interest for me, is creating nice plots for my thesis ;))

src/Initializer/MemoryManager.h
```diff
@@ -357,6 +357,14 @@ class seissol::initializers::MemoryManager {
       ltsParameters = std::make_shared<time_stepping::LtsParameters>(time_stepping::readLtsParametersFromYaml(m_inputParams));
     }
 
+    std::string getOutputPrefix() const {
+      const auto outputDirectory = (*m_inputParams)["output"]["outputfile"];
+      if (!outputDirectory) {
+          logError() << "No outputfile given.";
+      }
+      return outputDirectory.as<std::string>();
+    }
```
sebwolf-de, line 366: Could use: https://github.com/SeisSol/SeisSol/blob/731f2b56375e0d7d782bd8064e3da1bc0c3a88a6/src/Initializer/InputAux.hpp#L39 e.g: 
```
const auto outputCandidate = getWithDefault(m_inputParams["output"], "outputfile", "file-not-present");
if (outputCandidate == "file-not-present") {
  logError() << ...
}
return outputCandidate;
```

src/ResultWriter/ClusteringWriter.cpp
```diff
@@ -0,0 +1,73 @@
...
+      const auto localRank = localRanks[rank];
+      const auto& curProfilingIds = profilingIds[rank];
+      const auto& curLocalClusterIds = localClusterIds[rank];
+      const auto& curLayerTypes = layerTypes[rank];
+      const auto& curSizes = sizes[rank];
+      const auto& curDynamicRuptureSizes = dynamicRuptureSizes[rank];
+
+      for (unsigned i = 0; i < curProfilingIds.size(); ++i) {
+        const auto layerType = static_cast<LayerType>(curLayerTypes[i]);
+        const auto layerTypeStr = layerType == Interior ? "Interior" : "Copy";
```
sebwolf-de, line 55: Are we sure `Ghost` is strictly not possible?

sebwolf-de (commented):
Just a short question and a possible improvement.

src/ResultWriter/ClusteringWriter.cpp
```diff
@@ -0,0 +1,73 @@
...
+  const auto layerTypes = mpi.collectContainer(clusteringInformation.layerTypes);
+  const auto sizes = mpi.collectContainer(clusteringInformation.sizes);
+  const auto dynamicRuptureSizes = mpi.collectContainer(clusteringInformation.dynamicRuptureSizes);
+
+  if (mpi.rank() == 0) {
+
+    auto filepath = path(outputPrefix);
+    filepath += path("-clustering.csv");
+
+    auto fileStream = std::fstream(filepath, std::ios::out);
```
davschneller, line 41: Maybe `std::ofstream`?

src/ResultWriter/ClusteringWriter.h
```diff
@@ -0,0 +1,39 @@
...
+    void write() const;
+
+  // SoA that contains info about clusters
+  struct ClusteringInformation {
+    std::vector<int> ranks;
+    std::vector<int> localRanks;
+    std::vector<int> profilingIds;
+    std::vector<int> localClusterIds;
+    std::vector<std::underlying_type<LayerType>::type> layerTypes;
+    std::vector<unsigned> sizes;
```
davschneller, line 28: Just a side comment: will we stay within 32 Bit size mesh/cluster sizes for the foreseeable future? Just wondering about the size of `unsigned` here again...

davschneller (commented):
Thanks for the addition! I've added some _really_ minor comments about some typing. And a general question about the types for specifying cluster sizes and the likes (which, if it is handled, will most likely be content of some future PR instead).

src/Initializer/MemoryManager.h
sebwolf-de: I see your points. To avoid rewriting everything again, I'd favour `getOrFail` in `InputAux.hpp`. Otherwise, parameter reading will become spaghettified too soon.

src/ResultWriter/ClusteringWriter.h
krenzland: Probably not. In this case it's definitely fine, as we're only dealing with local sizes (and they can't be much more than ~10 million due to memory.

src/ResultWriter/ClusteringWriter.cpp
krenzland: Yes. Ghost clusters are created seperately and are basically an abstraction for communication. (They don't have any data!)

krenzland:
Thanks, should all be fixed.

PR #821, opened 2023-03-17
Improve profiling

krenzland:
This PR has the following improvements for the profiling:

- Skip loops of size zero, arising from clusters with size 0 (may improve performance slightly, handled differently for predict/correct due DR scheduling, not done for some GPU kernels)
- Fix time spent in DR (last PR did not work, missed an MPI_Reduce)
-  Use int64 as offset for nc files, required for long, large-scale simulations
- Add config option in parameters.par, use better output name

The profiling is not documented on purpose, as it should not used by non-developers. I'll add some analysis scripts later (with documentation).

@ravil-mobile I didn't add skipping of zero-sized loops for all kernels for GPUs. The profiling works without this but you might add this later to improve performance slightly.

### DIFF
--- a/Documentation/parameters.par
+++ b/Documentation/parameters.par
@@ -149,7 +149,9 @@ xdmfWriterBackend = 'posix' ! (optional) The backend used in fault, wavefield,
 EnergyOutput = 1 ! Computation of energy, written in csv file
 EnergyTerminalOutput = 1 ! Write energy to standard output
 EnergyOutputInterval = 0.05
-ComputeVolumeEnergiesEveryOutput = 4 ! Compute volume energies only once every ComputeVolumeEnergiesEveryOutput * EnergyOutputInterval 
+ComputeVolumeEnergiesEveryOutput = 4 ! Compute volume energies only once every ComputeVolumeEnergiesEveryOutput * EnergyOutputInterval
+
+LoopStatisticsNetcdfOutput = 0 ! Writes detailed loop statistics. Warning: Produces terabytes of data!
 /
            
 &AbortCriteria
--- a/src/Initializer/MemoryManager.h
+++ b/src/Initializer/MemoryManager.h
@@ -362,6 +362,10 @@ class seissol::initializers::MemoryManager {
       return getUnsafe<std::string>((*m_inputParams)["output"], "outputfile");
     }
 
+    bool isLoopStatisticsNetcdfOutputOn() const {
+      return getWithDefault((*m_inputParams)["output"], "loopstatisticsnetcdfoutput", false);
+    }
+
 #ifdef ACL_DEVICE
   void recordExecutionPaths(bool usePlasticity);
 
--- a/src/Solver/time_stepping/TimeCluster.cpp
+++ b/src/Solver/time_stepping/TimeCluster.cpp
@@ -229,6 +229,7 @@ void seissol::time_stepping::TimeCluster::computeSources() {
 
 #ifndef ACL_DEVICE
 void seissol::time_stepping::TimeCluster::computeDynamicRupture( seissol::initializers::Layer&  layerData ) {
+  if (layerData.getNumberOfCells() == 0) return;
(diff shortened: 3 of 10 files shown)


ravil-mobile:
Hi @krenzland, maybe you can add a couple of words for developers to our doxygen documentation (https://ci_seissol.pages.gitlab.lrz.de/SeisSol/master/)? I mean nobody will know (except you how to use it) in the future. I am personally going to add something about gitlab-ci to help people to maintain it after me.  

I mean Sebastian and I though the doxygen documentation can be a great place to share knowledge between developers

krenzland:
Yup, once I've got the analysis scripts ready :)

Documentation/parameters.par
```diff
@@ -149,7 +149,9 @@ xdmfWriterBackend = 'posix' ! (optional) The backend used in fault, wavefield,
 EnergyOutput = 1 ! Computation of energy, written in csv file
 EnergyTerminalOutput = 1 ! Write energy to standard output
 EnergyOutputInterval = 0.05
-ComputeVolumeEnergiesEveryOutput = 4 ! Compute volume energies only once every ComputeVolumeEnergiesEveryOutput * EnergyOutputInterval 
+ComputeVolumeEnergiesEveryOutput = 4 ! Compute volume energies only once every ComputeVolumeEnergiesEveryOutput * EnergyOutputInterval
+
+LoopStatisticsNetcdfOutput = 0 ! Writes detailed loop statistics. Warning: Produces terabytes of data!
```
davschneller, line 154: Considering the warning: when does it produce terabytes of data? On which parameters do the statistics depend?

src/Initializer/MemoryManager.h
```diff
@@ -362,6 +362,10 @@ class seissol::initializers::MemoryManager {
       return getUnsafe<std::string>((*m_inputParams)["output"], "outputfile");
     }
 
+    bool isLoopStatisticsNetcdfOutputOn() const {
```
davschneller, line 365: Just to put it out here: we're slowly building a parameter reader inside the `MemoryManager` class, and I am not sure if it is a good place to loosely put it in there like that. However, discussing/changing this is probably off-topic for this PR. I'd say it's good like that for now.

(still, just as a side note: there is some rather complete parameter reader on the branch `davschneller/no-fortran`, right now in `Initializer/SeisSolParameters.hpp`; but the branch is not ready to merge so far unfortunately ...)

src/Solver/time_stepping/TimeCluster.cpp
```diff
@@ -229,6 +229,7 @@ void seissol::time_stepping::TimeCluster::computeSources() {
 
 #ifndef ACL_DEVICE
 void seissol::time_stepping::TimeCluster::computeDynamicRupture( seissol::initializers::Layer&  layerData ) {
+  if (layerData.getNumberOfCells() == 0) return;
   SCOREP_USER_REGION_DEFINE(myRegionHandle)
```
davschneller, line 233: Just to make really sure: are we good by placing the SCOREP region define after the skip for 0 cells?

davschneller (commented):
Thanks alot! From my side, there are only a few comments/questions, and that's it.

Documentation/parameters.par
krenzland: I'd like to address this later, when adding documentation. However, for most large runs it will lead to huge amounts of data. Depends linearly on the number of total timesteps (=cluster updates). So scales with number of elements, time step size and also number of MPI ranks.

src/Solver/time_stepping/TimeCluster.cpp
krenzland: I think so. It's abit hard to test because the SCOREP stuff is more-or-less broken currently.
However, if we always match START/STOP region, we should be fine?

sebwolf-de: Yes should be fine. SCOREP is not particularly broken, it is just very cumbersome to get it working...

PR #825, opened 2023-03-21
Fix broken link in cookbook_overview.rst

mredenti:
The link https://seissol.readthedocs.io/en/latest/compilation.html is broken, should it perhaps point to https://seissol.readthedocs.io/en/latest/compiling-seissol.html ?


krenzland (approved):
Thanks!

PR #837, opened 2023-04-13
Fix off by one error in cluster merging

krenzland:


krenzland:
Thanks Thomas, before merging, @davschneller wanted to have a quick look as well.

davschneller (approved):
Looked over it a bit again; I see no other overt problems. The naming is consistent as well.

PR #847, opened 2023-04-25
Added an option to change locations of the MPI buffers for the communication layers

ravil-mobile:
This PR adds an option to perform MPI p2p non-blocking communication using either managed (unified), or host, or device memory. Some MPI GPU-aware libraries have problems with sending data over the unified memory (either too slow or just not capable of doing so e.g., CrayMPI v8.1.23)

Note: this PR depends on: https://github.com/SeisSol/Device/pull/23


uphoffc:
Hi Ravil,

this looks great. Two general comments:
1) I would not call this a "prefetch". Here, you introduce MPI-through-host via explicit copies. So the name of the feature should suggest that you work-around a bad or non-existent GPU-aware MPI implementation. Other codes such a flag "have GPU-aware mpi" or similar.
2) From a software design perspective, I would not introduce yet another ifdef, because ifdefs tend to code that does not compile, because it is not regularly checked. (And if you test it in CI then you increase the cost of CI by yet another build.) Here, you already have most of the things you need in order to make this a parameter or an environment variable. The AbstractCommunicationManager already offers an abstract interface, so you when you create the ghost cluster you can choose at run-time whether you want to create a "generic" ghost time cluster or one with "prefetch". The same holds true for the additional memory that is allocated: You can check at run-time whether it needs to be allocated or whether it can be omitted. I think it might be best to allocate the additional memory inside of GhostTimeClusterWithPrefetch because I don't think you need it anywhere else.

Best regards,
Carsten

ravil-mobile:
Hi @uphoffc , thanks for your comments.

>I would not call this a "prefetch". Here, you introduce MPI-through-host via explicit copies. So the name of the feature should suggest that you work-around a bad or non-existent GPU-aware MPI implementation. Other codes such a flag "have GPU-aware mpi" or similar.

It is a good point. However, in stead of just copying the data I am trying to copy them asynchronously, potentially overlapping with other tasks (for example, computations of the interior LTS layers).

Taking into account the MPI user-buffers are allocated with the unified/managed memory, there are two possible cases, namely: 1) copy them to the device memory (should be fast) and 2) copy to the host memory (usually slow).

The first case can be handled with `USE_GPU_AWARE_MPI=OFF`. What would you suggest for the second case? How shall I name it?

uphoffc:
Maybe in that case an enum would be self-explaining. E.g.
```c++
enum class GpuMPIMode {
  Direct,                            /// Direct MPI
  CopyInCopyOutDevice, /// Copy-in / copy-out to device memory
  CopyInCopyOutHost     /// Copy-in / copy-out to host memory
};
```

ravil-mobile:
> Maybe in that case an enum would be self-explaining. E.g.
> 
> ```
> enum class GpuMPIMode {
>   Direct,                            /// Direct MPI
>   CopyInCopyOutDevice, /// Copy-in / copy-out to device memory
>   CopyInCopyOutHost     /// Copy-in / copy-out to host memory
> };
> ```

Thanks! God you.

Btw, do you know by any chance whether the following lines of code can result in sending zero-size messages?

https://github.com/SeisSol/SeisSol/blob/f8f5a948ba1978cd6ee8388756a64bf477538421/src/Solver/time_stepping/GhostTimeCluster.cpp#L13-L20
https://github.com/SeisSol/SeisSol/blob/f8f5a948ba1978cd6ee8388756a64bf477538421/src/Solver/time_stepping/GhostTimeCluster.cpp#L29-L36

I mean that `meshStructure->neighboringClusters[region][0]` could be equal to zero for some reason? or we filter such cases during pre-processing?

uphoffc:
A partition might have empty time clusters. E.g. if the minimum time-step is 4*dt and the max timestep is 16*dt, then there might be no elements with 8*dt. Therefore, there might be empty messages. I don't recall any piece of code that would filter out these messages, but I am not 100% sure.

krenzland:
Yup, I think it probably does send empty messages. Be aware that changing this could lead to bugs in the synchronization between clusters as it could lead to unmatched receives...

ravil-mobile:
Thanks @uphoffc  and @krenzland. I will keep in my mind.

@uphoffc , could you, please, have a look at this PR one more time? I removed if-defs as you suggested.

Currently, I added the env. variable. The ideal case would be to add this option to the parameter-file. I think it is better to do once PR #829 is merged.

src/Solver/time_stepping/GenericGhostTimeCluster.h
```diff
@@ -6,23 +6,22 @@
 #include "AbstractTimeCluster.h"
 
 namespace seissol::time_stepping {
-
-class GhostTimeCluster : public AbstractTimeCluster {
- private:
+class GenericGhostTimeCluster : public AbstractTimeCluster {
```
krenzland, line 9: IMO the design should be different. Both GhostTimeClusters should inherit from AbstractGhostTimeCluster, which defines the interface.
This here should be called DirectGhostTimeCluster or simular.
testQueue can be a non-class function (poor design on my part there!)

src/Solver/time_stepping/GhostTimeClusterWithCopy.h
```diff
@@ -0,0 +1,41 @@
...
+
+  private:
+  unsigned int numberOfRegions{};
+  std::vector<real*> duplicatedCopyRegions{};
+  std::vector<real*> duplicatedGhostRegions{};
+
+  std::vector<void*> prefetchCopyRegionsStreams{};
+  std::vector<void*> prefetchGhostRegionsStreams{};
+
+  enum ReceiveState { RequiresMpiTesting, RequiresPrefetchTesting, Ready };
```
krenzland, line 36: should be enum class

src/Solver/time_stepping/GhostTimeClusterWithCopy.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+template <MPI::DataTransferMode CommType>
+void GhostTimeClusterWithCopy<CommType>::sendCopyLayer() {
+  SCOREP_USER_REGION("sendCopyLayer", SCOREP_USER_REGION_TYPE_FUNCTION)
+  assert(ct.correctionTime > lastSendTime);
+
+  lastSendTime = ct.correctionTime;
+  auto prefetchedRegions = prefetchCopyLayer();
+
+  while (not prefetchedRegions.empty()) {
```
krenzland, line 73: I know that this is legal C++ but it's kinda weird :D

src/Solver/time_stepping/GhostTimeClusterWithCopy.cpp
```diff
@@ -0,0 +1,199 @@
...
+        ++region;
+      }
+    }
+  }
+}
+
+template <MPI::DataTransferMode CommType>
+void GhostTimeClusterWithCopy<CommType>::receiveGhostLayer() {
+  SCOREP_USER_REGION("receiveGhostLayer", SCOREP_USER_REGION_TYPE_FUNCTION)
+  assert(ct.predictionTime > lastSendTime);
```
krenzland, line 96: You can change this to >=, there's an open issue somewhere that this assert is buggy

src/Solver/time_stepping/GhostTimeClusterWithCopy.cpp
```diff
@@ -0,0 +1,199 @@
...
+        receiveRegionsStates[*region] = RequiresPrefetchTesting;
+      }
+      ++region;
+      break;
+    }
+    case ReceiveState::RequiresPrefetchTesting: {
+      auto* stream = prefetchGhostRegionsStreams[*region];
+      if (device.api->isStreamWorkDone(stream)) {
+        receiveRegionsStates[*region] = Ready;
+        // Note: fall to `Ready` is intentional
```
krenzland, line 133: I don't get this comment

krenzland: Ah, you mean fallthrough to ready is intentional? Yeah, makes sense.

krenzland (changes requested):
Looks mostly good! The design overall is nice, however I think it may be a good idea to introduce an AbstractGhostTimeCluster (see comment in GenericGhostTimeCluster.h

src/Solver/time_stepping/GhostTimeClusterWithCopy.cpp
ravil-mobile: yes, `[[fallthrough]]` only works inside `switch`-statements. I just edited the comment

src/Solver/time_stepping/GhostTimeClusterWithCopy.cpp
ravil-mobile: Ok. I will change this to `!prefetchedRegions.empty()`

src/Solver/time_stepping/GhostTimeClusterWithCopy.h
ravil-mobile: done

src/Solver/time_stepping/TimeManager.cpp
```diff
@@ -378,3 +382,7 @@ void seissol::time_stepping::TimeManager::setTv(double tv) {
   }
 }
 
+void seissol::time_stepping::TimeManager::releaseCommunicationManager() {
```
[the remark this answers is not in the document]
ravil-mobile: ok. done

src/Parallel/MPI.cpp
```diff
@@ -90,4 +92,36 @@ void seissol::MPI::bindAcceleratorDevice() {
 }
 #endif
 
+void seissol::MPI::checkEnvVariables() {
```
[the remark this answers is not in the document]
ravil-mobile: done

src/Solver/time_stepping/GhostTimeClusterWithCopy.cpp
ravil-mobile: Ok. I changed in both places, namely: here and in the parent class.

src/Solver/time_stepping/GenericGhostTimeCluster.h
ravil-mobile: > Both GhostTimeClusters should inherit from AbstractGhostTimeCluster, which defines the interface.

Done. I moved all common parts to `AbstractGhostTimeCluster`. Both `DirectGhostTimeCluster` and `DirectGhostTimeCluster` add their own behavior i.e., `sendCopyLayer`, `receiveGhostLayer` and `testForGhostLayerReceives`

> testQueue can be a non-class function
I don't know what to do with it. As for me, it is not super critical here.

src/SeisSol.cpp
```diff
@@ -184,6 +186,8 @@ void seissol::SeisSol::finalize()
 
 	const int rank = MPI::mpi.rank();
 
+	m_timeManager.resetCommunicationManager();
+
```
uphoffc, line 190: Not sure why this is necessary.

src/Solver/time_stepping/AbstractGhostTimeCluster.h
```diff
@@ -1,28 +1,27 @@
-#ifndef SEISSOL_GHOSTTIMECLUSTER_H
-#define SEISSOL_GHOSTTIMECLUSTER_H
+#pragma once
```
uphoffc, line 1: pragma once is non-standard

src/Solver/time_stepping/DirectGhostTimeCluster.h
```diff
@@ -0,0 +1,25 @@
+#ifndef SEISSOL_GHOSTTIMECLUSTER_H
+#define SEISSOL_GHOSTTIMECLUSTER_H
```
uphoffc, line 2: Headerguard should have same name as file

src/Solver/time_stepping/GhostTimeClusterFactory.h
```diff
@@ -0,0 +1,52 @@
+#pragma once
```
uphoffc, line 1: pragma once is non-standard

src/Solver/time_stepping/GhostTimeClusterWithCopy.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
```
uphoffc, line 1: pragma once is non-standard

src/Solver/time_stepping/GhostTimeClusterWithCopy.cpp
```diff
@@ -0,0 +1,199 @@
...
+    if constexpr (CommType == MPI::DataTransferMode::CopyInCopyOutDevice) {
+      device.api->freeMem(duplicatedCopyRegions[region]);
+      device.api->freeMem(duplicatedGhostRegions[region]);
+    } else {
+      device.api->freePinnedMem(duplicatedCopyRegions[region]);
+      device.api->freePinnedMem(duplicatedGhostRegions[region]);
+    }
+  }
+}
+
```
uphoffc, line 64: If you have a custom dtor you should also have custom copy / move operators and copy-/move-initializers. If you do not plan on copying / moving objects of this class then delete the 4 operators.

src/Solver/time_stepping/TimeManager.cpp
```diff
@@ -378,3 +382,7 @@ void seissol::time_stepping::TimeManager::setTv(double tv) {
   }
 }
 
+void seissol::time_stepping::TimeManager::resetCommunicationManager() {
+  communicationManager.reset(nullptr);
+}
```
uphoffc, line 387: reset accepts a double and not a pointer. Also not sure why this function is neccessary.

uphoffc (changes requested):

src/Solver/time_stepping/GhostTimeClusterFactory.h
krenzland: Yes but supported by nearly everything: https://en.wikipedia.org/wiki/Pragma_once#Portability

src/Solver/time_stepping/TimeManager.cpp
ravil-mobile: Hi @uphoffc , I didn't get why it should accept `double`. `communicationManager` is a unique pointer. Calling reset will trigger the destructor of the underlying object.

Note, the `SeisSol` object is static and thus it is going to be deleted only at the very very end. At this point, the connection with the GPU driver will be lost. Omitting a call to the dtor of `communicationManager` will trigger a segfault because `WithCopy` implementations uses device memory which is freed by the dtor

src/SeisSol.cpp
```diff
@@ -95,9 +95,11 @@ bool seissol::SeisSol::init(int argc, char* argv[]) {
 
 #ifdef USE_MPI
   logInfo(rank) << "Using MPI with #ranks:" << MPI::mpi.size();
+  // TODO (Ravil, David): switch to reading MPI options from the parameter-file.
+  MPI::mpi.checkEnvVariables();
 #endif
 #ifdef _OPENMP
-  pinning.checkEnvVariables();
+  pinning.setDataTransferModeFromEnv();
```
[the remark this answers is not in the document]
ravil-mobile: @uphoffc, thanks for noticing. Done

src/Solver/time_stepping/GhostTimeClusterFactory.h
ravil-mobile: @uphoffc, @krenzland, Shall I leave `#pragma once` or replace it with the guards? 

I personally agree with @krenzland, this pragma is supported by almost all compilers and there will be no issues with it.

Documentation/gpus.rst
```diff
@@ -118,3 +118,15 @@ the following will force SeisSol to allocate 1.5GB of stack GPU memory for tempo
     
     export DEVICE_STACK_MEM_SIZE=1.5
     mpirun -n <M x N> ./SeisSol_dsm70_cuda_* ./parameters.par
+
+
+Currently, SeisSol allocates MPI user-buffers using the unified/managed memory
+type. Some MPI implementations perform poorly during non-blocking
+point-to-point communication. SeisSol provides *SEISSOL_PREFERRED_MPI_DATA_TRANSFER_MODE*
```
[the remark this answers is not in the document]
ravil-mobile: Done

src/Solver/time_stepping/GhostTimeClusterWithCopy.cpp
ravil-mobile: I deleted them

src/Solver/time_stepping/DirectGhostTimeCluster.h
ravil-mobile: I replaced to `#pragma once`. If you and @krenzland make a decision to keep using header guards then I will revert and rename it

Documentation/gpus.rst
```diff
@@ -118,3 +118,15 @@ the following will force SeisSol to allocate 1.5GB of stack GPU memory for tempo
...
+environment variable which can be used to select the memory type for the user-buffers.
+The *host* value means that the data will be copied to/from the host memory
+before/after each *MPI_Isend* / *MPI_Irecv*. Setting the variable to *device*
+will result in utilizing the regular device memory for non-blocking
+communication. The default value is *direct* which means that the communication
+goes over the unified/managed memory and thus does not involve explicit data
+copies.
```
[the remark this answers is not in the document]
ravil-mobile: Added. See here: https://github.com/SeisSol/SeisSol/blob/ravil/comm-layer/Documentation/LatexFigures/gpu-comm-layer-data-flow.png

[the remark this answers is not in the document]
uphoffc: Great figure, thanks!

cmake/process_users_input.cmake
```diff
@@ -146,9 +146,9 @@ if (NOT ${DEVICE_ARCH} STREQUAL "none")
     endif()
 
     if (${DEVICE_ARCH} MATCHES "sm_*")
-        set(ALIGNMENT  64)
-    elseif(${DEVICE_ARCH} MATCHES "gfx*")
         set(ALIGNMENT  128)
+    elseif(${DEVICE_ARCH} MATCHES "gfx*")
+        set(ALIGNMENT  256)
```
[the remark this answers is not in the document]
uphoffc: ok

src/SeisSol.cpp
uphoffc: Ah that makes sense. But why don't you call it freeCommunicationManager if that's the purpose?

src/Solver/time_stepping/GhostTimeClusterFactory.h
uphoffc: I don't mind, that's for you guys to decide.

src/Solver/time_stepping/TimeManager.cpp
uphoffc: Ok sorry about that. I saw that AbstractCommunicationManager has a reset function and did not notice that it is stored as unique_ptr in TimeManager.

uphoffc (approved):
I suggest to rename resetCommunicationManager to freeCommunicationManager in TimeManager. LGTM otherwise.

krenzland:
Yup, I agree with Carsten here, otherwise it looks great!

PR #848, opened 2023-04-25
Added `SEISSOL_FREE_CPUS_MASK` env. variable

ravil-mobile:
* Note, it an option for developers
* it can be useful for systems like LUMI, Fugaku
* tests are added
* docu has been updated

@krenzland said that it was ok to implement this feature using an env. variable

### DIFF
--- a/src/Common/IntegerMaskParser.h
+++ b/src/Common/IntegerMaskParser.h
@@ -2,7 +2,6 @@
 
 #include <string>
 #include <optional>
-#include <regex>
 #include <vector>
 
 
@@ -10,75 +9,11 @@ namespace seissol {
 class IntegerMaskParser {
   public:
   using MaskType = std::vector<std::vector<int>>;
-  static std::optional<MaskType> parse(std::string mask) {
-    MaskType resultMask{};
-
-    auto const regex = std::regex("^(\\d+|\\d+-\\d+|\\{(\\d+,?)+\\})(,|$)");
-    std::smatch match;
-
-    while (std::regex_search(mask, match, regex)) {
-      std::string subString = match[1].str();
-      mask = match.suffix();
-
-      auto intRange = IntegerMaskParser::parseIntRange(subString);
-      if (intRange) {
-        resultMask.push_back(*intRange);
-        continue;
-      }
-
-      auto intList = IntegerMaskParser::parseIntList(subString);
-      if (intList) {
-        resultMask.push_back(*intList);
-        continue;
-      }
-
-      resultMask.push_back({std::stoi(subString)});
-    }
-
-    if (resultMask.empty()) {
-      return std::nullopt;
-    }
-    return resultMask;
-  }
+  static std::optional<MaskType> parse(std::string mask);
 
   private:
-  using OptionalIntVecotorType = std::optional<std::vector<int>>;
-  static OptionalIntVecotorType parseIntRange(const std::string& str) {
-    OptionalIntVecotorType result{std::nullopt};
-    std::size_t separator = str.find('-');
-    if (separator != std::string::npos) {
-      auto start = std::stoi(std::string(str, 0, separator));
-      auto end = std::stoi(std::string(str, ++separator, str.size()));
-
-      if (start > end)
-        std::swap(start, end);
-      std::vector<int> integers;
-      for (int i{start}; i <= end; ++i) {
-        integers.push_back(i);
-      }
-      result = integers;
-    }
-    return result;
-  }
-
-  static OptionalIntVecotorType parseIntList(const std::string& str) {
-    OptionalIntVecotorType result{std::nullopt};
-    if (str[0] == '{') {
-      std::string subStr{};
-      std::vector<int> integers;
-
-      const auto strCopy = std::string(str, 1, str.size());
-      for (size_t i = 0; i < strCopy.size(); ++i) {
-        if ((strCopy[i] == ',') || (strCopy[i] == '}')) {
-          integers.push_back(std::stoi(subStr));
-          subStr = std::string();
-        } else {
-          subStr.push_back(strCopy[i]);
-        }
-      }
-      result = integers;
-    }
-    return result;
-  }
+  using OptionalIntVectorType = std::optional<std::vector<int>>;
+  static OptionalIntVectorType parseIntRange(const std::string& str);
+  static OptionalIntVectorType parseIntList(const std::string& str);
 };
 } // namespace seissol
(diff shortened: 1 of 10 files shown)


src/Common/IntegerMaskParser.h
```diff
@@ -0,0 +1,84 @@
+#pragma once
+
+#include <string>
+#include <optional>
+#include <regex>
+#include <vector>
+
+
+namespace seissol {
+class IntegerMaskParser {
```
uphoffc, line 10: Separate code in .h and .cpp. No point for header-only here as the class does not use templates.

src/Common/IntegerMaskParser.h
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    if (resultMask.empty()) {
+      return std::nullopt;
+    }
+    return resultMask;
+  }
+
+  private:
+  using OptionalIntVecotorType = std::optional<std::vector<int>>;
```
uphoffc, line 45: OptionalIntVecotorType -> OptionalIntVectorType

src/Common/IntegerMaskParser.h
```diff
@@ -0,0 +1,84 @@
...
+  static OptionalIntVecotorType parseIntRange(const std::string& str) {
+    OptionalIntVecotorType result{std::nullopt};
+    std::size_t separator = str.find('-');
+    if (separator != std::string::npos) {
+      auto start = std::stoi(std::string(str, 0, separator));
+      auto end = std::stoi(std::string(str, ++separator, str.size()));
+
+      if (start > end)
```
uphoffc, line 53: Please use parentheses {}

src/Parallel/Pin.cpp
```diff
@@ -55,6 +57,52 @@ seissol::parallel::Pinning::Pinning() {
...
+    auto parsedResult = seissol::IntegerMaskParser::parse(std::string(envVariable));
+    if (parsedResult) {
+      parsedFreeCPUsMask = parsedResult.value();
+
+      bool isMaskGood{true};
+      const auto numLocalProcesses = MPI::mpi.sharedMemMpiSize();
+      if (numLocalProcesses > parsedFreeCPUsMask.size()) {
+        logInfo(rank) << "Num. local processes is greater than free cpus mask";
+        isMaskGood = false;
+      }
```
uphoffc, line 72: Must be an == here. Otherwise the code will break later if numLocalProcesses < parsedFreeCPUsMask.size()

src/Parallel/Pin.cpp
```diff
@@ -82,6 +130,14 @@ cpu_set_t seissol::parallel::Pinning::getFreeCPUsMask() const {
   cpu_set_t freeMask{};
   CPU_ZERO(&freeMask);
 
+  if (not parsedFreeCPUsMask.empty()) {
+    const auto localProcessor = MPI::mpi.sharedMemMpiRank();
+    for (auto& cpu : parsedFreeCPUsMask[localProcessor]) {
```
uphoffc, line 135: Breaks here if parsedFreeCPUsMask.size() <= localProcessor

uphoffc: You could also allow round-robin assignment. E.g. if the free list is "39-40" and you have two processes then you could pin both threads to CPUs 39 and 40 and let the OS scheduler decide the rest.

uphoffc (changes requested):

src/Common/IntegerMaskParser.h
ravil-mobile: done

src/Parallel/Pin.cpp
ravil-mobile: Hi, this part of the code checks whether a number of local processes is greater than the provided mask. If it is true it sets `isMaskGood` to `false`.

https://github.com/SeisSol/SeisSol/blob/b550796b6555d3e77b3dc5b1f2e5934d98251b19/src/Parallel/Pin.cpp#L90-L98

If the mask is bad (for some reason) the mask will become empty (see Line 97).

Line `133` is supposed to ensure that we are going to operate on a mask which does not have side-effects e.g., the local MPI rank is greater or equal to the mask size 

https://github.com/SeisSol/SeisSol/blob/b550796b6555d3e77b3dc5b1f2e5934d98251b19/src/Parallel/Pin.cpp#L133-L139

uphoffc: Ah, yes, sorry read the if the wrong way... nevermind then.

src/Parallel/Pin.cpp
ravil-mobile: > You could also allow round-robin assignment.

I can see the point but I would like to leave the implementation as it is if nobody is against. Here is my reasoning. 

I think that people (mostly developers) are going to use this option when the automatic free cpus pinning failed for some reason. In this case, a person is desperate and wants to describe pinning  as clear as possible. In such a scenario, he/she is going to be ok to spend a bit of time on typing but be 100% sure about the locations of free cpus.

uphoffc: Sure, maybe better that way.

uphoffc (approved):

src/Parallel/Pin.cpp
```diff
@@ -55,6 +57,52 @@ seissol::parallel::Pinning::Pinning() {
...
+      }
+      else {
+        logWarning(rank) << "Ignoring `SEISSOL_FREE_CPUS_MASK` env. variable.";
+        logWarning(rank) << "`SEISSOL_FREE_CPUS_MASK` Format:"
+                         << "(<int>|<range: int-int>|<list: {int,+}>),+";
+        parsedFreeCPUsMask = IntegerMaskParser::MaskType{};
+      }
+    }
+    else {
+      logWarning(rank) << "Failed to parse `SEISSOL_FREE_CPUS_MASK` env. variable";
```
Thomas-Ulrich, line 101: I would write it as a logInfo, e.g. 
logInfo(rank) << 'no custom binding is considered as SEISSOL_FREE_CPUS_MASK is empty.";

ravil-mobile: This warning corresponds to this code

https://github.com/SeisSol/SeisSol/blob/a66459f4629e377035b80d59b966adef2650ac69/src/Parallel/Pin.cpp#L63-L64

It means that `SEISSOL_FREE_CPUS_MASK` was provided but we failed to parse it for some reason  (most probably due to formatting). In this case, it makes sense to say that we failed to parse`. `SEISSOL_FREE_CPUS_MASK is empty` will be ambiguous and thus confusing in this case.

I would leave it as warning and preserve the message

Thomas-Ulrich (approved):

PR #849, opened 2023-04-28
refactor code calculating fault energies and fix associated documentation

Thomas-Ulrich:
In the documentation, there were 2 factors 1/2 in the seismic moment and total frictional work, that I fixed.
I also now precise that the total and static frictional works are only the contribution of the stress change (total frictional work would be the integral over the absolute stress).
Here is how the documentation now looks like:
image


daisy20170101:
Should it be "d \Sigma" as used in Ma and Archuleta [2006], instead of "dx",  in the integral?

Thomas-Ulrich:
> Should it be "d \Sigma" as used in Ma and Archuleta [2006], instead of "dx", in the integral?

dx here means dS (or d \Sigma) or dV (for the volume integrals), I'm just consistent with the notations from above.

Thomas-Ulrich:
after a discussion with Nicolas, and checking again Ma and Archuleta (2006), we found that sigma(t) was flipped in the current document, which I fixed.
And we added a line precise how to compute fracture energy from these quantities.

AliceGabriel:
@Thomas-Ulrich , this should not be integrated across the full fault surface but only across the slipped area?

Thomas-Ulrich:
At any points outside the rupture area, slip and slip-rate are always zero. Therefore their contribution to the integral is 0, so integrating over the fault or the rupture area should lead to the same result.

Thomas-Ulrich:
I've tried to calculate the total_frictional_work and static_frictional_work from the 0.1s fault output from tpv5, with e.g.:
```
for i in range(n):
    total_frictional_work -= area[i] * np.trapz(VT[:, i], dx=dt)
    static_frictional_work -= 0.5 * area[i] * TS_tf[i]
```
(full code attached)
and I get 
```
total_frictional_work from fault output : 9.757027e+15
static_frictional_work from fault output : 5.530416e+15
```
Which is about the double I get with the energy output (incorporating the 0.5 factor)
```
25,total_frictional_work,4.83827e+15
25,static_frictional_work,3.55024e+15
````
This suggests that the 0.5 factors I wanted to add are not needed and the current implementation of seissol is correct.
I'm trying to make sense of all these pieces of information.
E.g. the seismic moment is calculated with 0.5*doubleSurfaceArea
https://github.com/SeisSol/SeisSol/blob/master/src/ResultWriter/EnergyOutput.cpp#L179
but mu is given by
https://github.com/SeisSol/SeisSol/blob/master/src/ResultWriter/EnergyOutput.cpp#L174
(why the factor 2? My understanding is that there should be no factor 2, nor a factor 0.5
At least that would make things coherent).

Finally,
[compute_frictional_work_from_fault_output.py.txt](https://github.com/SeisSol/SeisSol/files/13365612/compute_frictional_work_from_fault_output.py.txt)
 can we use 2Wstatic-Wtotal for estimating the fracture energy? It does not seem to work practically, and it is because (at least in tpv5), the stress drop below Tr=mu_d*sigma_n, so the integral do not only compute the triangular area (integral from 0 to Dc) but also the area between Tr and Tau_final.
image

Thomas-Ulrich:
I asked Carsten about this doubledSurface area:
"About the doubled surface area: Check eq. (35) in the tandem paper (https://academic.oup.com/gji/article/233/1/586/6847163?login=false#390396038)
e is the physical element that we map to the reference element \hat{e}. Now if you do that map you need to add the “surface jacobian” |n^e(\chi)|. (See also generic description on Wikipedia: https://en.wikipedia.org/wiki/Surface_integral)
In SeisSol n^e(\chi) is constant and computed as t0 x t1, where t0 and t1 are two tangent vectors of the physical triangle, hence |n^e| is equivalent to twice the surface area."
image

All in all, everything is now clear and the PR is ready for review..

davschneller (approved):
One final question: `aPlus` and `aMinus` are not needed anymore?

Thomas-Ulrich:
no, and clang-tidy picked it too.

PR #853, opened 2023-05-08
Fix read the docs build

uphoffc:


krenzland:
This makes sense but is caused by a missing dependecy in sphinx: https://github.com/sphinx-doc/sphinx/issues/10927
Wouldn't it make more sense to upgrade to a more recent sphinx version or would this break our documentation?

uphoffc:
No, the issue is the release of urllib3 on May 4th:

"ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2n  7 Dec 2017. See: https://github.com/urllib3/urllib3/issues/2168"

krenzland:
Okay, cool. Then, I think, we could remove the urllib dependency as it should be included by the requirements of sphinx (via requests). We can also leave it in, it doesn't hurt probably

krenzland (approved):

uphoffc:
It might still be necessary as the requests package could install urllib3 >= 2 and then the problem reappears.

PR #864, opened 2023-05-19
Improve 'Pick components' filter

sebwolf-de:
Update quantity list, if different receivers have different quantities, e.g. fault receivers and off-fault receivers.

### DIFF
--- a/postprocessing/visualization/receiver/src/Filters.py
+++ b/postprocessing/visualization/receiver/src/Filters.py
@@ -223,7 +223,7 @@ class Pick(Filter):
       if not w.text() in required_widgets:
         widgets_to_remove.append(w)
     for w in widgets_to_remove:
-      w.setParent(None)
+      w.deleteLater()
       self.cb_widget_list.remove(w)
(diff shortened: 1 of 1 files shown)


postprocessing/visualization/receiver/src/Filters.py
```diff
@@ -187,23 +187,58 @@ class Pick(Filter):
...
+    required_widgets = list(wf.waveforms.keys())
+
+    # remove not needed checkboxes
+    # can't remove from list, while iterating over list
+    widgets_to_remove = []
+    for w in self.cb_widget_list:
+      if not w.text() in required_widgets:
+        widgets_to_remove.append(w)
+    for w in widgets_to_remove:
+      w.setParent(None)
```
Thomas-Ulrich, line 226: it seems deleteLater() should be better.
https://stackoverflow.com/questions/30241684/pyqt4-setparent-vs-deletelater

Thomas-Ulrich (approved):
LGTM

PR #866, opened 2023-05-22
use seissol::MPI to avoid ambiguity

uphoffc:


davschneller (approved):

davschneller:
The patch still needs a clang-format run for the tests to pass (e.g. `sh ./.ci/format.sh `which clang-format` ./` in the SeisSol root)

uphoffc:
> The PR still needs a clang-format run for the tests to pass (e.g. `sh ./.ci/format.sh `which clang-format` ./` in the SeisSol root)

Ah nice. What's the policy for clang-format? I saw in the format script that some files are in the white list and some are not?

Thomas-Ulrich:
> > The PR still needs a clang-format run for the tests to pass (e.g. `sh ./.ci/format.sh `which clang-format` ./` in the SeisSol root)
> 
> Ah nice. What's the policy for clang-format? I saw in the format script that some files are in the white list and some are not?

I guess there is no clear rule. It was introduced by the dynamic rupture refactoring. At first it was only for the files in the DynamicRupture folder. 
But in the meanwhile, new files have been clang-formated, and David added this while list filter also for individual files  (before it was on full folder only), so that some of these new files could be moved to the list.
Also, some old files that were at +50% refactored by David also moved there.

PR #871, opened 2023-05-31
Fix some input parameters (incl. RFileName)

davschneller:
This PR makes `RFilename` in the parameters file optional again (and not mandatory). Also, the vertex weight parameters are now read correctly.

A legacy FORTRAN definition was also removed.


krenzland (approved):

krenzland:
IMO, readWithDefault / similar methods should lower-case all strings automatically. The cost is pretty much irrelevant IMO.

PR #872, opened 2023-05-31
Clean Up pre/postprocessing

davschneller:
This PR shall clean up the pre/postprocessing folders (if possible, move anything still needed into their own repositories), and remove some old scons files.

The following is done:

* `site_scons` has been removed
* The submodules `scons-tools` and `cxxtest` have been removed
* `Documentation/UserManual` was cleaned up, except for `RD_sign_conventions.*`.
* For `preprocessing`:
  * `meshing`: moved. See https://github.com/SeisSol/Meshing/pull/48 .
  * `partitioning`: removed (due to age). Only `gambit.py` was used by `meshing/cube`, so it has been copied to SeisSol/Meshing.
  * `science`: **unsure what to do...** (keep? Although extracting rconv would be good, I'd say. But where to put it?)
  * `workflow`: removed (due to age)
* For `postprocessing`:
  * `performance`: removed (due to age)
  * `science`: **unsure what to do...** (or where to put)
  * `validation`: **unsure what to do...** (or where to put)
  * `visualization`: moved. See https://github.com/SeisSol/Visualization/pull/10 .
Since there are some points for which I'm not sure what to do, this PR is still a draft.


sebwolf-de:
We could leave `rconv` in the SeisSol repository for now, since it does not fit into neither meshing nor visualization.
I'd keep validation here or move it to `precomputed-seissol`. We need the scripts there for the tests.

davschneller:
Personally, I think it would be good to have all compilable code but SeisSol itself out of the repo—at least long-term. More thoughts:

* In https://github.com/SeisSol/Meshing/pull/48 , Thomas suggested to turn the repositories into preprocessing (formerly Meshing and maybe Geodata) and postprocessing (formerly Visualization). ... Or should we maybe merge all of them into one? (we could also introduce a repository with submodules)
* I agree that `rconv` doesn't really fit into any of the existing repositories. Maybe, it would do best in a repository on its own, since the `SeisSol/Examples` repository uses it for e.g. WP2_LOH1. So why not make it like `pumgen`?
* That would leave only `cube_c` without its own repo (which is needed for the convergence tests).
* Also, it seems like `gmsh2gambit` is sort of obsolete at this point? The `SeisSol/Examples` repository goes directly from `gmsh` to `pumgen`. TODO: remove it from the cookbook then, I'd guess.

Or we can really leave the validation and science folders in here for now (since the latter have PRs for them). Maybe move `rconv` out, or not. Moving the validation to precomputed-seissol also seems good, although the scripts are not tied to the data there—so it's only maybe the right place for them.

davschneller:
(closing; at this point in time—the most relevant parts have been integrated into master, and it's probably more useful to open a new PR instead for the rest)

PR #875, opened 2023-06-05
Fix Receiver Output Length Error

### LINKED ISSUE #873: SeisSol fails if receiveroutputinterval is not set
**Describe the bug**
If receiveroutputinterval is omitted then SeisSol crashes with

```bash
terminate called after throwing an instance of 'std::length_error'
  what():  vector::reserve
```

The reason for this failure is that `receiverParameters.interval` is set to 1e+100 by default such that the receiver writer tries to pre-allocate a huge vector.

**Expected behavior**
Either make receiveroutputinterval non-optional or set the default interval to a sensible value (e.g. to EndTime, I think that used to be the previous default behaviour when readpar.F90 was still alive).

davschneller:
Fixes #873.


davschneller:
Sorry, I unfortunately messed up the other PR #874 ... Therefore #875 it is.

ravil-mobile (approved):

PR #876, opened 2023-06-06
Point source GPU

uphoffc:
This PR makes the following contributions:

* Refactor PointSources struct to work with std::vector only instead of raw pointers (relying on C++17 for correct memory alignment)
* If ACL_DEVICE is defined, PiecewiseLinearFunction1D and PointSources use sycl::usm_allocator instead of std::allocator for allocating shared memory.
* Point source data now belongs to TimeCluster instead to sourceterm::Manager
* Introduce PointSourceCluster class that takes care of point source calculation (moved out of TimeCluster)
* Introduce SYCL implementation for computing point sources on the device (class PointSourceClusterOnDevice). Major benefit is that h2d and d2h transfers become unnecessary.

### DIFF
--- a/src/SourceTerm/typedefs.hpp
+++ b/src/SourceTerm/typedefs.hpp
@@ -6,6 +6,7 @@
  *
  * @section LICENSE
  * Copyright (c) 2015, SeisSol Group
+ * Copyright (c) 2023, Intel corporation
  * All rights reserved.
  * 
  * Redistribution and use in source and binary forms, with or without
@@ -46,9 +47,23 @@
 #include <vector>
 #include <Initializer/typedefs.hpp>
 #include <generated_code/tensor.h>
+#include <cstdint>
 
 namespace seissol {
   namespace sourceterm {    
+    template<typename T, std::size_t N>
+    class AlignedArray {
+    public:
+        inline T* data() { return data_; }
+        inline T const* data() const { return data_; }
+        constexpr T& operator[](std::size_t pos) { return data_[pos]; }
+        constexpr T const& operator[](std::size_t pos) const { return data_[pos]; }
+        constexpr std::size_t size() const noexcept { return N; }
+
+    private:
+        alignas(ALIGNMENT) T data_[N];
+    };
+
     /** Models point sources of the form
      *    S(xi, eta, zeta, t) := (1 / |J|) * S(t) * M * delta(xi-xi_s, eta-eta_s, zeta-zeta_s),
      * where S(t) : t -> \mathbb R is the moment time history,
@@ -68,7 +83,7 @@ namespace seissol {
       /** mInvJInvPhisAtSources[][k] := M_{kl}^-1 * |J|^-1 * phi_l(xi_s, eta_s, zeta_s), where phi_l is the l-th
        *  basis function and xi_s, eta_s, and zeta_s are the space position
        *  of the point source in the reference tetrahedron. */
-      real (*mInvJInvPhisAtSources)[tensor::mInvJInvPhisAtSources::size()];
+      std::vector<AlignedArray<real, tensor::mInvJInvPhisAtSources::size()>> mInvJInvPhisAtSources;
 
       /** NRF: Basis vectors of the fault.
        * 0-2: Tan1X-Z   = first fault tangent (main slip direction in most cases)
@@ -76,7 +91,7 @@ namespace seissol {
        * 6-8: NormalX-Z = fault normal
        * 
        * FSRM: Moment tensor */
-      real (*tensor)[TensorSize];
+      std::vector<AlignedArray<real, TensorSize>> tensor;
 
       /// Area
       std::vector<real> A;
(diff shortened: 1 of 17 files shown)


src/SourceTerm/typedefs.hpp
```diff
@@ -46,9 +47,73 @@
...
+    template <typename T>
+    using AllocatorT = sycl::usm_allocator<T, sycl::usm::alloc::shared>;
+    class AllocatorFactory {
+    public:
+      AllocatorFactory(sycl::queue q) : q_(std::move(q)) {}
+      template <typename T> AllocatorT<T> make() const { return AllocatorT<T>(q_); }
+    private:
+      sycl::queue q_;
+    };
+#else
```
ravil-mobile, line 68: Hi, this is defined in the header file. It means that all files which explicitly or implicitly include it must to be compiler with a `sycl` compiler. It is totally ok for DPC++ compiler but I think it is not going to work for OpenSYCL (we are using separate compilation for the host and device codes).

For example, I haven't managed to install DPC++ on LUMI (Cray Machine) and thus I have only one alternative - i.e., OpenSYCL.

What do you think if we use our `device` wrapper in this part of the code?  

```
device.api->allocUnifiedMem(size * sizeof(T))
```

I think we just need to create `AllocatorT` for it.

davschneller:
Just a tiny suggestion from my side—to add some of the files to the clang formatting list. That is, all new files which were added (to the `Kernels` folder), and I'd say we can add the whole `SourceTerm` folder to the formatting by now as well.

src/SourceTerm/typedefs.hpp
uphoffc: I don't mind using the device wrapper but usm_allocator is part of the SYCL spec and also included in opensycl: https://github.com/OpenSYCL/OpenSYCL/blob/12fdcaedfa990ab58ddf8bce304fa8cf917e6182/include/hipSYCL/sycl/usm.hpp#L270

Let me try whether I can compile with opensycl as well

uphoffc:
> Just a tiny suggestion from my side—to add some of the files to the clang formatting list. That is, all new files which were added (to the `Kernels` folder), and I'd say we can add the whole `SourceTerm` folder to the formatting by now as well.

Yes, adding the SourceTerm folder is a good idea. I think I'll wait until the reviews are finished, though, then it is easier to see what actually changed.

ravil-mobile:
Hi Carsten (@uphoffc), Could you, please, review PR #881? I would like to push it to this branch. 

In #881 all CI pipelines are fixed.

ravil-mobile (approved):
LGTM. One thing that I want to ask is whether we should replace all `new/delete` with `std::vector` in this PR or whether it is better to make it in a dedicated PR. What do you think?

uphoffc:
> One thing that I want to ask is whether we should replace all `new/delete` with `std::vector` in this PR or whether it is better to make it within a dedicated PR. What do you think?

Which new/delete?

ravil-mobile:
> > One thing that I want to ask is whether we should replace all `new/delete` with `std::vector` in this PR or whether it is better to make it within a dedicated PR. What do you think?
> 
> Which new/delete?

For example,

https://github.com/SeisSol/SeisSol/blob/38b1c7d7855ab8086520f2ac6e3506e402f6572d/src/SourceTerm/Manager.cpp#L329-L330
https://github.com/SeisSol/SeisSol/blob/38b1c7d7855ab8086520f2ac6e3506e402f6572d/src/SourceTerm/Manager.cpp#L339

https://github.com/SeisSol/SeisSol/blob/38b1c7d7855ab8086520f2ac6e3506e402f6572d/src/SourceTerm/Manager.cpp#L438-L439
https://github.com/SeisSol/SeisSol/blob/38b1c7d7855ab8086520f2ac6e3506e402f6572d/src/SourceTerm/Manager.cpp#L449


https://github.com/SeisSol/SeisSol/blob/38b1c7d7855ab8086520f2ac6e3506e402f6572d/src/SourceTerm/Manager.cpp#L346
https://github.com/SeisSol/SeisSol/blob/38b1c7d7855ab8086520f2ac6e3506e402f6572d/src/SourceTerm/Manager.cpp#L411-L412

https://github.com/SeisSol/SeisSol/blob/38b1c7d7855ab8086520f2ac6e3506e402f6572d/src/SourceTerm/Manager.cpp#L456
https://github.com/SeisSol/SeisSol/blob/38b1c7d7855ab8086520f2ac6e3506e402f6572d/src/SourceTerm/Manager.cpp#L515-L516

https://github.com/SeisSol/SeisSol/blob/38b1c7d7855ab8086520f2ac6e3506e402f6572d/src/SourceTerm/NRF.h#L83-L89

uphoffc:
Ah ok, let me quickly update those.

ravil-mobile:
@uphoffc, shall I apply formatting to all files in `src/SourceTerm`, merge with the master and push it?

uphoffc:
> @uphoffc, shall I apply formatting to all files in `src/SourceTerm`, merge with the master and push it?

Yes, please do!

PR #859, opened 2023-05-17
A bit more descriptive Git version display

davschneller:
This PR fixes/prettifies two things:

* The year in the first line after `Welcome to SeisSol` is now taken from the last commit. In our case, we get to 2023 for now. But if we tag a version in 2024, it'll also have 2024 in, all years after 2024, without us doing anything.
* The latest version is now taken from all tags, thus we print v1.0.1 finally. Also, I've put in the current branch, since it is from time to time even more descriptive. Maybe we could also add the exact commit in the future...

PR #877, opened 2023-06-06
Fix CMake for Not Being in a Git Repository

davschneller:
Addresses the latest comments in #859 .

If we clone without a Git repository (e.g. by downloading the ZIP file of master or one of the tagged versions), the Git version cannot be determined. On master, this causes CMake to fail, since we then extract the year for the copyright from an empty string. For now, we add some dummy values to fix that; other suggestions are welcome.

Thus, we will now show `Copyright (c) 2012 - 9999 SeisSol Group` in such case—for now at least with this PR.


uphoffc:
Why would you show the Copyright at all in the log? The code is copyrighted, and therefore you have license headers in the source file, but the data the code generates is not copyrighted.

krenzland:
> but the data the code generates is not copyrighted

Sure it is?

> and therefore you have license headers in the source file

That's not a requirement for the license to hold. A LICENSE file is enough.

uphoffc:
> > but the data the code generates is not copyrighted
> 
> Sure it is?
> 
Pretty sure, yes.
Cf. https://intellectual-property-helpdesk.ec.europa.eu/system/files/2021-01/european-ipr-helpdesk-your-guide-to-ip-in-europe.pdf
https://www.lexology.com/library/detail.aspx?g=eb52567a-2104-4034-8daf-bf6c5cb9cd45

davschneller:
Maybe we could merge this PR soon-ish?

(again, the copyright discussions etc. are a bit out of scope for it)

krenzland (approved):

PR #884, opened 2023-06-13
Ravil/fs

ravil-mobile:
- The automatic creation of the output folder if it does not exist
- Enforced the same time stamp for all backup files

Depends on [this xdmfWriter PR](https://github.com/TUM-I5/XdmfWriter/pull/6)


src/Common/filesystem.h
```diff
@@ -26,9 +29,32 @@ namespace filesystem = std::filesystem;
...
+inline void generateBackupFileIfNecessary(std::string fileName,
+                                          std::string fileExtension,
+                                          std::optional<std::string> timeStamp = {}) {
+  std::stringstream fullName;
+  fullName << fileName << '.' << fileExtension;
+  seissol::filesystem::path path(fullName.str());
+  seissol::filesystem::directory_entry entry(path);
+
+  if (seissol::directoryExists(entry)) {
+    if (!timeStamp.has_value()) {
```
Thomas-Ulrich, line 47: I don't understand why this is tested: timeStamp should exist because it is initialized in src/main.cpp, right?

ravil-mobile:
I am curious. Why couldn't we just backup the entire output folder?

Thomas-Ulrich:
because there is not necessary an output folder?
(you can write directly in the input folder, default writes e.g. data-fault.xdmf where parameters.par is).

ravil-mobile:
> because there is not necessary an output folder?

ravil-mobile:
I closed the PR by mistake. Reopening...

sebwolf-de (commented):
Hi, thanks for your work. Most things look good except for one remark.
I know, adding new work to PRs is bad, but I have a problem with the backups since essentially always.
We backup the data, but we don't backup the xdmf file. If we change the mesh (size), we can't read old data anymore (without additional work). How hard would it be to also backup the xdmf files?

Thomas-Ulrich:
> Hi, thanks for your work. Most things look good except for one remark. I know, adding new work to PRs is bad, but I have a problem with the backups since essentially always. We backup the data, but we don't backup the xdmf file. If we change the mesh (size), we can't read old data anymore (without additional work). How hard would it be to also backup the xdmf files?

I think the problem is you need to update all the file links in the xdmf file.

krenzland:
An alternative solution would be to append a counter to the filenames linked in the xdmf file. Then you don't have to change the names.

sebwolf-de:
I see, it's not so easy. This was just a stupid idea. I will take care of that otherwise.

sebwolf-de:
I just had a conversation with a user, who had problems restoring a backup. That's why I would like to merge https://github.com/SeisSol/SeisSol/tree/sebastian/backup-xdmf-files soon, but after this PR. Are there any major revision necessary?

davschneller:
Ok, I've added a warning and re-ordered the includes.

sebwolf-de (approved):

PR #887, opened 2023-06-15
Update first example

sebwolf-de:
There were a lot of inconsistencies....

### DIFF
--- a/Documentation/a-first-example.rst
+++ b/Documentation/a-first-example.rst
@@ -11,39 +11,50 @@ in this tutorial. We assume that you have successfully compiled SeisSol.
 Setup
 -----
 
--  Download the `parameter file and additional setup
-   files <https://github.com/SeisSol/Examples/tree/master/tpv33>`__ and
-   put them in your launch directory, hereafter named
-   ``launch_SeisSol``.
--  Download the mesh `binary
-   file <https://syncandshare.lrz.de/getlink/fi72mQiszp6vSs7qN8tdZJf9/tpv33_gmsh>`__
-   and the associated `xml
-   file <https://syncandshare.lrz.de/getlink/fiEi52Xiwwqkf2sNpTrCHjhw/tpv33_gmsh.xdmf>`__
-   and store them in ``launch_SeisSol``.
-   **Optional:** For performance reasons, we suggest that you store the
+*  Clone our examples repository: https://github.com/SeisSol/Examples/.
+
+*  Navigate to the folder ``tpv33``.
+
+*  To create the mesh, execute ``./generating_the_mesh.sh``. To do so,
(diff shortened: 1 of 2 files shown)


Documentation/a-first-example.rst
```diff
@@ -11,39 +11,50 @@ in this tutorial. We assume that you have successfully compiled SeisSol.
...
-   file <https://syncandshare.lrz.de/getlink/fi72mQiszp6vSs7qN8tdZJf9/tpv33_gmsh>`__
-   and the associated `xml
-   file <https://syncandshare.lrz.de/getlink/fiEi52Xiwwqkf2sNpTrCHjhw/tpv33_gmsh.xdmf>`__
-   and store them in ``launch_SeisSol``.
-   **Optional:** For performance reasons, we suggest that you store the
+*  Clone our examples repository: https://github.com/SeisSol/Examples/.
+
+*  Navigate to the folder ``tpv33``.
+
+*  To create the mesh, execute ``./generating_the_mesh.sh``. To do so,
```
Thomas-Ulrich, line 18: When this page was created, it was argued (e.g. Alice) that it would be better if the mesh was already generated and ready to download. Could we (=you :-)) put the mesh on zenodo and remove the details of its generation?

Documentation/a-first-example.rst
```diff
@@ -11,39 +11,50 @@ in this tutorial. We assume that you have successfully compiled SeisSol.
...
+   The first one is a binary file, which contains all the data (e.g. 
+   coordinates, connectivity) and the ``xdmf`` file contains information 
+   on how to read that data for visualization software such as paraview.
+
+*  **Optional:** For performance reasons, we suggest that you store the
    mesh file in a scratch file system (if one is available at your
    cluster) and create symbolic links in your launch directory (e.g.
-   ``ln -s <path/to/tpv33_gmsh> launch_SeisSol/tpv33_gmsh;``). You may
-   not see a huge difference in this small test case but for larger
+   ``ln -s <path/to/tpv33_half_sym> tpv33_half_sym``).
```
Thomas-Ulrich, line 31: wouldn't it make sense to put all this stuff with `ln -s` at the end of the page.
I think it is not very important for a first example.

Thomas-Ulrich (commented):
Thanks, I've proposed a few suggestions.

Documentation/a-first-example.rst
Thomas-Ulrich: I suppose it would be easier for the users, up to you then, as it is extra work that you ask for ;-)

sebwolf-de: Not gonna do it. Just a good idea :D

Thomas-Ulrich (approved):

PR #893, opened 2023-06-22
Added `showEdgeCutStatistics` option

ravil-mobile:
Hi all, 

This PR adds an option to display the edge-cut statistics. I personally need it for one experiment. If you think it can be a useful feature then we may consider adding it to the master.

What do you think?

### DIFF
--- a/src/Initializer/InitProcedure/InitMesh.cpp
+++ b/src/Initializer/InitProcedure/InitMesh.cpp
@@ -122,32 +122,42 @@ static void readMeshPUML(const seissol::initializer::parameters::SeisSolParamete
 #endif // defined(USE_HDF) && defined(USE_MPI)
 }
 
+size_t getNumOutgoingEdges(seissol::geometry::MeshReader& meshReader) {
+  auto& mpiNeighbors = meshReader.getMPINeighbors();
+  size_t numEdges{0};
+  for (auto& [_, neighborInfo] : mpiNeighbors) {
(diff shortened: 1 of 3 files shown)


Thomas-Ulrich:
Interesting for comparing different mesh or clustering strategies?
Might be difficult to interpret the value, don't you think?

src/Initializer/InitProcedure/InitMesh.cpp
```diff
@@ -175,5 +186,14 @@ void seissol::initializer::initprocedure::initMesh() {
   // Call the post mesh initialization hook
   seissol::Modules::callHook<seissol::POST_MESH>();
 
-  logInfo(seissol::MPI::mpi.rank()) << "End init mesh.";
+  logInfo(commRank) << "End init mesh.";
+
+  if ((seissolParams.mesh.showEdgeCutStatistics) && (commSize > 0)) {
```
krenzland, line 191: Shouldn't this be > 1 or am I missing something?

krenzland (changes requested):
Looks great, just one small question. (Not sure if it's a bug or whether I am missing something)

src/Initializer/InitProcedure/InitMesh.cpp
ravil-mobile: You're right

ravil-mobile: done

PR #895, opened 2023-06-27
dr/poro

sebwolf-de:
This PR enables dynamic rupture in poroelastic materials. 
Note that the pressure in normal direction is coupled to the fluid pressure when computing the Godunov state. That's why we need the new kernel. For (visco)elastic nothing changes.


postprocessing/validation/compare-energies.py
```diff
@@ -38,8 +38,8 @@
     print("Relative difference")
     print(relative_difference)
 
-    relative_difference_smaller_eps = (
-        relative_difference.iloc[1:, :] < args.epsilon
+    relative_difference_larger_eps = (
+        relative_difference.iloc[1:, :] > args.epsilon
     ).values
-    if not np.all(relative_difference_smaller_eps):
+    if np.any(relative_difference_larger_eps):
```
davschneller, line 44: Would be good to see where the problems lie (and how big they are)

src/Initializer/InitProcedure/InitSideConditions.cpp
```diff
@@ -94,6 +94,10 @@ static std::vector<std::unique_ptr<physics::InitialField>> buildInitialCondition
         "Ocean, an uncoupled ocean test case for acoustic equations (mode 2)";
     const auto g = seissol::SeisSol::main.getGravitationSetup().acceleration;
     initConditions.emplace_back(new physics::Ocean(2, g));
+  } else if (initConditionParams.type ==
+             seissol::initializer::parameters::InitializationType::PressureInjection) {
+    initialConditionDescription = "Pressure Injection (only viable for poroelastic materials)";
```
davschneller, line 99: It would be good to let it explicitly fail if we don't use poroelastic material

davschneller (commented):
Hi, just some comments for in between. I cannot comment that much anyways on the exact nature of the equations right now.

Can we add some files to the formatting or tidying workflow? (e.g. `Equations/poroelastic`?)

postprocessing/validation/compare-energies.py
sebwolf-de: If both energies are zero and we compute the relative difference it is (0 - 0) / 0 = NaN.
`NaN < eps` evaluates to false, while `! (NaN > eps)` evaluates to true ;-)

src/tests/Reader/NRFReader.t.h
```diff
@@ -31,7 +31,7 @@ TEST_CASE("NRF Reader") {
   REQUIRE(nrf.subfaults[0].tinit == AbsApprox(0.0));
   REQUIRE(nrf.subfaults[0].timestep == AbsApprox(0.0002));
   REQUIRE(nrf.subfaults[0].mu == AbsApprox(0.0));
-  REQUIRE(nrf.source == 1);
```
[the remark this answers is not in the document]
sebwolf-de: Sure, but I wanted to wait until after a first review...

src/DynamicRupture/FrictionLaws/FrictionSolverCommon.h
```diff
@@ -45,6 +45,23 @@ struct QInterpolated {
   using Range = typename std::conditional<Type == RangeType::CPU, CpuRange, GpuRange>::type;
 };
 
+/**
+ * Copies an eigen3 matrix to a 2D yateto tensor
+ */
+template <typename T, int dim1, int dim2>
+void copyEigenToYateto(Eigen::Matrix<T, dim1, dim2> const& matrix,
```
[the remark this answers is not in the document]
sebwolf-de: Indeed, I've moved it to CellLocalMatrices.cpp, but forgot to remove it here.

sebwolf-de:
> Can we add some files to the formatting or tidying workflow? (e.g. Equations/poroelastic?)

I'd do that separately, after merging this PR.

gitlab-cpu-pipeline.yml
```diff
@@ -39,37 +39,35 @@ build_seissol:
...
-            -DEQUATIONS=$equation
-            -DNUMBER_OF_MECHANISMS=$mechanisms
-            -DDR_QUAD_RULE=stroud
-            -DGEMM_TOOLS_LIST=LIBXSMM
-            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON;
-            make -j $(nproc);
-            cd .. ;
-          done;
+        - dirname="build_${equation}_${precision}_${build_type}";
+          echo "mkdir $dirname";
```
davschneller, line 47: Effective duplicate info to line 50 (`pwd`) below?

src/DynamicRupture/Typedefs.hpp
```diff
@@ -40,6 +51,7 @@ struct FaultStresses {
   alignas(ALIGNMENT) real normalStress[CONVERGENCE_ORDER][misc::numPaddedPoints] = {{}};
   alignas(ALIGNMENT) real traction1[CONVERGENCE_ORDER][misc::numPaddedPoints] = {{}};
   alignas(ALIGNMENT) real traction2[CONVERGENCE_ORDER][misc::numPaddedPoints] = {{}};
+  alignas(ALIGNMENT) real fluidPressure[CONVERGENCE_ORDER][misc::numPaddedPoints] = {{}};
```
davschneller, line 54: Maybe `#ifdef USE_POROELASTIC`? Though I'm not sure if it's really necessary here...

src/Model/common.hpp
```diff
@@ -173,6 +177,58 @@ void setBlocks(T QgodLocal, Tmatrix S, Tarray1 traction_indices, Tarray2 velocit
...
+  int evPos = 0;
+  for (int i = 0; i < NUMBER_OF_QUANTITIES; ++i) {
+    assert(std::abs(eigenvalues(i).imag()) < zeroThreshold);
+    if (eigenvalues(i).real() < -zeroThreshold) {
+      ++evNeg;
+    } else if (eigenvalues(i).real() > zeroThreshold) {
+      ++evPos;
+    }
+  }
+  assert(evNeg == 3);
```
davschneller, line 211: Is it good to hard-code that here?

src/Model/common.hpp
```diff
@@ -173,6 +177,58 @@ void setBlocks(T QgodLocal, Tmatrix S, Tarray1 traction_indices, Tarray2 velocit
...
+  assert(evPos == 3);
```
davschneller, line 212: Same here

davschneller (commented):
Hi, some more comments. I don't think I can come up with many more, i.e. from my side we'd be good then. Of course, reducing the number of `USE_STP` ifdefs would be good, but that can be something for the future. Maybe we'll have to subclass by solver in the end...

src/Equations/poroelastic/Model/PoroelasticSetup.h
```diff
@@ -253,6 +253,59 @@ namespace seissol {
       setBlocks(QgodLocal, S, tractionIndices, velocityIndices);
     }
 
+    template<>
+    inline seissol::eigenvalues::Eigenpair<std::complex<double>, NUMBER_OF_QUANTITIES> getEigenDecomposition (PoroElasticMaterial const& material, double zeroThreshold) {
```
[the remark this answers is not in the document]
sebwolf-de: The matrices are too bad in the poroelastic case, eigen3 can't handle them. The numerical flux term will become unstable. Somehow LAPACK works way better. That's why poroelastic materials _only_ work with LAPACK and don't work with eigen3.
When I first implemented the poroelastic material model, we decided to have LAPACK as a required dependency only for poroelastic materials and use eigen3 only, whenever possible.

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -531,8 +574,25 @@ void seissol::initializers::initializeDynamicRuptureMatrices( seissol::geometry:
...
+          seissol::model::getTransposedCoefficientMatrix(*dynamic_cast<seissol::model::PoroElasticMaterial*>(minusMaterial), 0, AMinus);
+
+          auto plusEigenpair = seissol::model::getEigenDecomposition(*dynamic_cast<seissol::model::PoroElasticMaterial*>(plusMaterial));
+          auto minusEigenpair = seissol::model::getEigenDecomposition(*dynamic_cast<seissol::model::PoroElasticMaterial*>(minusMaterial));
+
+          Eigen::Matrix<real, N, N> impedanceMatrix = extractMatrix(plusEigenpair);
+          Eigen::Matrix<real, N, N> impedanceNeigMatrix = extractMatrix(minusEigenpair);
+          Eigen::Matrix<real, N, N> etaMatrix = (impedanceMatrix + impedanceNeigMatrix).inverse();
+
+          auto impedanceView = init::Zplus::view::create(impedanceMatrices[ltsFace].impedance);
```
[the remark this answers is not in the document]
sebwolf-de: Yes, this is correct. In the (visco)elastic case, the `impedanceMatrix` is diagonal, with entries zp, zs.
I'll add a comment to make this more clear. The values are default initialized to zero.

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -161,6 +161,12 @@ void seissol::initializers::initializeCellLocalMatrices( seissol::geometry::Mesh
       double volume = MeshTools::volume(elements[meshId], vertices);
 
       for (unsigned side = 0; side < 4; ++side) {
+        seissol::model::getTransposedGodunovState(  material[cell].local,
```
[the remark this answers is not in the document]
sebwolf-de: Indeed, I guess the call in line 164 ff is obsolete. But that's year old code ;-)

src/Initializer/DynamicRupture.h
```diff
@@ -127,8 +131,11 @@ struct seissol::initializers::DynamicRupture {
     tree.addVar(         waveSpeedsMinus,             mask,                 1,      MEMKIND_STANDARD );
     tree.addVar(          drEnergyOutput,             mask,         ALIGNMENT,      MEMKIND_STANDARD );
     tree.addVar(      impAndEta,                      mask,                 1,      MEMKIND_STANDARD );
+    tree.addVar(      impedanceMatrices,              mask,                 1,      MEMKIND_STANDARD );
```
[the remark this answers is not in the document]
sebwolf-de: No, we don't always need them, but the overall agreement is to reduce ifdefs wherever possible.
If I add an ifdef here, we would need more ifdefs in CellLocalMatrices.cpp
Otherwise I could create a more complicated class structure with a poroelastic DR LTS tree and a non-poroelastic DR LTS tree, which I would like to avoid. If we add the matrices here and just leave them in the tree, they don't do any harm, IMHO.

src/DynamicRupture/FrictionLaws/FrictionSolverCommon.h
```diff
@@ -136,6 +140,33 @@ inline void precomputeStressFromQInterpolated(
                   qIMinus[o][T2][i] * invZsNeig);
     }
   }
+#else
+  seissol::dynamicRupture::kernel::computeTheta krnl;
```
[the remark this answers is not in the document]
sebwolf-de: We could, but I don't see a benefit. The impedance matrix is diagonal in the elastic case. Also this would break the GPU implementation of this method.

src/DynamicRupture/FrictionLaws/FrictionSolverCommon.h
```diff
@@ -272,6 +305,48 @@ inline void postcomputeImposedStateFromNewStress(
       imposedStateP[W][i] += weight * (qIPlus[o][W][i] + invZs * (traction2 - qIPlus[o][T2][i]));
     }
   }
+#else
+  // setup kernel
+  seissol::dynamicRupture::kernel::computeImposedStateM krnlM;
```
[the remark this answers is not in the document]
sebwolf-de: Same answer as above ;-)

src/DynamicRupture/Typedefs.hpp
sebwolf-de: Same thing with ifdefs, would like to reduce their usage as much as possible. One ifdef in one place, needs five more ifdefs in other places, to make compilation always possible.

src/Model/common.hpp
sebwolf-de: IMHO yes. After all it's only executed in debug mode and might help, if a scenario fails.

davschneller (approved):

PR #903, opened 2023-07-13
Synchronize LTSTree after Setup of the Initial Conditions

davschneller:
Fixes another bug caused by the initialization re-ordering by #829: the initial conditions were behaving a bit off. To be exact, duplicates in the LTS tree were not initialized for initial conditions, since an Lut lookup was used for that (and to my knowledge, it's the last place where that was the case). Now, we just synchronize the LTSTree after initializing the data. (if someone wants to, we could transform it into a loop over the LTSTree eventually)

### DIFF
--- a/src/Initializer/InitProcedure/InitModel.cpp
+++ b/src/Initializer/InitProcedure/InitModel.cpp
@@ -9,6 +9,7 @@
 #include "Physics/Attenuation.hpp"
 #include "Equations/datastructures.hpp"
 #include "Initializer/tree/Lut.hpp"
+#include "Initializer/tree/LTSSync.hpp"
 #include "Initializer/typedefs.hpp"
 
 #include "SeisSol.h"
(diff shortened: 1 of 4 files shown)


davschneller:
Most likely, the `synchronize` operations in `initModel` are not necessary anymore—we go over the LTSTree directly there in all cases.

src/Initializer/InitialFieldProjection.cpp
```diff
@@ -120,4 +122,9 @@ void seissol::initializers::projectInitialField(std::vector<std::unique_ptr<phys
 #ifdef _OPENMP
   }
 #endif
+
+  seissol::initializer::synchronize(lts.dofs);
```
Thomas-Ulrich, line 126: Isn't it a bit strange to have this call in a file called InitialFieldProjection.cpp?
(I mean this has nothing to do with field projection?

src/Initializer/InitProcedure/InitModel.cpp
```diff
@@ -25,66 +26,6 @@ using namespace seissol::initializer;
...
-      // now call new here.
-      new (&target) T(value);
-    }
-  }
-  // (these two methods cannot be combined, unless we have some way for C-style arrays, i.e. S[N]
-  // for <typename S, size_t N>, to use a copy constructor as well)
-}
-
-template <typename T>
-static void synchronize(const seissol::initializers::Variable<T>& handle) {
```
Thomas-Ulrich, line 65: so this function was never called? Or did you forget to remove the previous call?

src/Initializer/tree/LTSSync.hpp
```diff
@@ -0,0 +1,77 @@
...
+      // now call new here.
+      new (&target) T(value);
+    }
+  }
+  // (these two methods cannot be combined, unless we have some way for C-style arrays, i.e. S[N]
+  // for <typename S, size_t N>, to use a copy constructor as well)
+}
+
+template <typename T>
+static void synchronize(const seissol::initializers::Variable<T>& handle) {
```
Thomas-Ulrich, line 52: synchronizeDuplicatesInTheLtsTree ?

src/Initializer/InitProcedure/InitModel.cpp
davschneller: It was, in the same file.

I kept the calls from FORTRAN, but changed the iteration method—before it iterated over all mesh cells, but now it iterates over the whole LTS tree (thus including duplicates). Hence, we may probably remove it.

src/Initializer/InitialFieldProjection.cpp
davschneller: It kind of has something to do with it—as the purpose of the method is to set the DOFs to the initial condition data. That should also includes the duplicates in the LTS tree.

(it's the same in the end as if we were to just iterate over all non-ghost layers in the LTS tree—except that we do not compute the initial condition multiple times for each duplicate)

src/Initializer/tree/LTSSync.hpp
davschneller: Ok, renamed it to something more descriptive

src/Initializer/InitialFieldProjection.cpp
davschneller: I've removed the calls in `InitModel`; the duplicates were really initialized before that already, because we went through the whole LTS tree there.

src/Initializer/InitProcedure/InitModel.cpp
davschneller: I've removed the calls in InitModel; the duplicates were really initialized before that already, because we went through the whole LTS tree there.

sebwolf-de (approved):

PR #904, opened 2023-07-24
Fix NRF reader test

sebwolf-de:
Was broken by https://github.com/SeisSol/SeisSol/commit/29ce470d0e5bb216e645203c6028deb5d44a83db

### DIFF
--- a/src/tests/Reader/NRFReader.t.h
+++ b/src/tests/Reader/NRFReader.t.h
@@ -32,6 +32,9 @@ TEST_CASE("NRF Reader") {
   REQUIRE(nrf.subfaults[0].timestep == AbsApprox(0.0002));
   REQUIRE(nrf.subfaults[0].mu == AbsApprox(0.0));
 
+  REQUIRE(nrf.size() == 1);
+  REQURIE(nrf.subfaults.size() == nrf.centres.size() && nrf.centres.size() == nrf.sroffsets.size())
+
   for (size_t dim = 0; dim < 3; dim++) {
     for (unsigned i = 0;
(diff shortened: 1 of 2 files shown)


src/tests/Reader/NRFReader.t.h
```diff
@@ -31,7 +31,6 @@ TEST_CASE("NRF Reader") {
   REQUIRE(nrf.subfaults[0].tinit == AbsApprox(0.0));
   REQUIRE(nrf.subfaults[0].timestep == AbsApprox(0.0002));
   REQUIRE(nrf.subfaults[0].mu == AbsApprox(0.0));
-  REQUIRE(nrf.source == 1);
```
davschneller, line 34: `REQUIRE(nrf.size() == 1);`

Maybe also test that `nrf.subfaults.size()==nrf.centres.size()==nrf.sroffsets.size()` (the syntax used in this comment here is of course not correct)

davschneller (commented):
Thanks! However, I'd not remove the assert, but rather update it as well.

That brings up the question in more general ... why is Netcdf disabled for the tests right now?

sebwolf-de:
That's a good question. The CI docker container explicitely contains netcdf. I just enabled it for the tests.

davschneller:
I fear that's the wrong pipeline still (it's the `Testing_SeisSol_CPU` one); the reader tests are still not executed.

For the unit tests, cf. https://github.com/SeisSol/SeisSol/blob/master/azure-pipelines-seissol-template.yml#L64

davschneller (approved):
Ok, approved.

Though I guess long-term (and if feasible), it would be best to integrate all unit tests in either the CPU runner or Azure.

PR #909, opened 2023-08-01
Update filesystem.h

ravil-mobile:
Bug fix in std::experimental::filesystem


davschneller (commented):
If we _really_ want to check if a directory exists, then we may have to check if it's a directory in the first place as well. Cf. https://stackoverflow.com/a/24544128
(if that doesn't matter here—then maybe `directoryExists` is not the best-descriptive name for the method)

davschneller (approved):

ravil-mobile:
Thanks for merging! In general, your idea is correct.

PR #911, opened 2023-08-03
Make the Output/FaultOutput Flag Work

davschneller:
The output section by now had a `faultoutput` field which did not do anything so far.

Now it enables/disables the fault outputs, as if a dynamic rupture didn't exist. However: we'll have a duplicate with `outputpointtype` in the dynamic rupture part as well (i.e. if we say `outputpointtype = 0`, we also disable DR output).

So instead, we could also just remove the `faultoutput` flag.


Thomas-Ulrich (approved):
I would opt for removing the faultoutputflag.

davschneller:
Ok, sounds good to me. Cf. a new PR #924 .

PR #914, opened 2023-08-04
Prepare to Apply `clang-tidy` to More Source Files

davschneller:
This PR paves the way for running more files through the `clang-tidy` tool.

That is, just like for `clang-format`, we now have whitelists for folders and files.

Right now, we only have the DR folder in there, and I've added the SeisSol `main.cpp` as a starter.


.ci/tidy.sh
```diff
@@ -0,0 +1,48 @@
...
+format() {
+    # don't use a directory with whitespace
+    local whitelist_dir="
+        src/DynamicRupture
+        "
+    
+    # NOTE: once the files of a directory are (almost) fully covered, consider moving it to whitelist_dir instead
+    local whitelist_file="
+        src/main.cpp
+        "
```
sebwolf-de, line 19: Try to avoid white/blacklist and use allow/blocklist instead.

.ci/tidy.sh
```diff
@@ -0,0 +1,48 @@
...
+    local SEISSOL_BUILD_DIR="${2}"
+    shift 2
+
+    # check for self
+    if [ ! -f "${SEISSOL_SOURCE_DIR}/.ci/tidy.sh" ]; then
+        echo "Please ensure that SEISSOL_SOURCE_DIR is passed as the first argument"
+        exit 177
+    fi
+
+    # we'll treat all files as headers... Any code files are never included somewhere anyways (hopefully, at least).
```
sebwolf-de, line 31: Should be always, instead of never, right?

sebwolf-de (changes requested):
Looks good, does `main.cpp` pass without problems, already?

.ci/tidy.sh
davschneller: Ok, changed it in the clang-format infrastructure as well

.ci/tidy.sh
davschneller: Comment clarified

davschneller:
`main.cpp` passes. It's included in the `SeisSol.SeisSol` CI pipeline already. (though locally on my personal machine, `clang-tidy` throws several errors for the DR folder, since we could make some variables `const` without a problem)

sebwolf-de:
Are you sure about the constness? https://github.com/llvm/llvm-project/issues/60163

sebwolf-de (approved):

davschneller:
There was one false positive when implementing all `const` suggestions (that is to make the `strengthBuffer` const); for the rest it seems to pass. E.g. these two here: https://github.com/SeisSol/SeisSol/blob/e68052171d847b7baf887027a54e8d16724b7c27/src/DynamicRupture/Initializers/ImposedSlipRatesInitializer.cpp#L92-L93
It also was `misc-const-correctness` that gave the error; not `readability-non-const-parameter`. 

(the difference may be to be due to me using version 15 locally, versus version 14 in the CI—I just tested it with a version 14 clang-tidy on our local GPU cluster, and it passed without errors)

PR #915, opened 2023-08-04
Remove FaultMPI

davschneller:
The FaultMPI class is seemingly unused by now, so this PR removes it.

Unless, of course, the plan is to bring back the ASAGI-based `StressReader` for the dynamic ruptures.


sebwolf-de:
I don't think we want to bring back any ASAGI functionality directly into SeisSol. The DR parameter ready is entirely based on easi.

sebwolf-de (approved):

PR #917, opened 2023-08-05
Fix `ZinvData` being used after going out of scope

davschneller:
Just as the title says, in the poroelastic `Time.cpp`, we have under rare circumstances (such as before a sync) that some temporary array `ZinvData` is stack-allocated and pointed to for the kernel. But before the kernel gets executed, the array goes out of scope, since it is defined inside an `if`. This PR fixes that by moving the execution inside the `if` as well.


sebwolf-de (approved):
LGTM

PR #918, opened 2023-08-10
Allow cubeGenerator to be used as a mesh generator in SeisSol

montrie:
This PR allows specifying the cubeGenerator to be used as a mesh generator by setting the values in the parameter file as follows:
```
&MeshNml
MeshFile = <fileName>
meshgenerator = 'CubeGenerator'
/
```

I also extended the markUnused() function in InpuParameters.cpp such that it takes a vector of fields as input. This should declutter some of the markUnused() calls in that file.


src/Initializer/InitProcedure/InitMesh.cpp
```diff
@@ -130,6 +131,97 @@ size_t getNumOutgoingEdges(seissol::geometry::MeshReader& meshReader) {
...
+  numBndElements[2] = 2*numCubesPerPart[0]*numCubesPerPart[1];
+
+  // output file name
+  auto cubeOutput = seissolParams.mesh.meshFileName + ".nc";  // mesh file name doesnt include file type ending
+  std::string fileName = seissolParams.mesh.meshFileName + ".nc";
+
+  // TODO: I think passing output into the cubeGenerator function breaks the call? there was a case where the output name's string representation was suddenly empty when passing output to the funciton
+  // instead, replace the outut variable with its value in the function call
+  logInfo() << "Start generating a mesh using the CubeGenerator";
+  cubeGenerator(numCubes, numPartitions, \
```
davschneller, line 219: Send a `struct` or `class` as parameter here (instead of passing all arguments individually)

davschneller (changes requested):
Thanks for your work! Some general comments on it (some I've made also in the code itself):

* I'd suggest to keep the `cube_c` repository untouched, just copy the code to `src/Geometry` as needed and adjust it (also, apply `clang-format` etc to it). All code in SeisSol should be under the `src` repository, but having `cube_c` as a standalone is still a good thing to have.
* Infer all parameters that you can from SeisSol directly. For example, we shouldn't need to specify partition information at all.
* Make sure to apply `clang-format` and `clang-tidy` to your files. Otherwise, the CI will fail. Especially to any new files you add to the `src` folder. The respective bash scripts are located in the `.ci` folder.
* Once you're done, adjust the CI to avoid using `cube_c` entirely.
* Also, it would be best to avoid generating a Netcdf file in between entirely. That is, make the `CubeGenerator` a subclass of `MeshReader`. Though I'd understand that it may take a bit more work.

davschneller:
One thing I should maybe add:

Some weeks ago I wrote some file which generates a PUML mesh for cube generation, but it still misses a proper boundary implementation (which would be needed to be included into PUML2 itself).

The code is in this PR: https://github.com/SeisSol/Meshing/pull/49 . Maybe it is helpful when integrating the cube generator, especially when avoiding the NetCDF intermediate file. Feel free to take any code from there if you want to.

src/Initializer/InitProcedure/InitMesh.cpp
```diff
@@ -130,6 +131,97 @@ size_t getNumOutgoingEdges(seissol::geometry::MeshReader& meshReader) {
...
+  // create additional variables necessary for cubeGenerator()
+  unsigned int numCubes[4];
+  unsigned int numPartitions[4];
+
+  // check input arguments
+  numCubes[0] = cubeX;
+  numCubes[1] = cubeY;
+  numCubes[2] = cubeZ;
+
+  numPartitions[0] = cubePx;
```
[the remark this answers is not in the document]
krenzland: https://github.com/SeisSol/SeisSol/blob/a6c4deea8dab2732bdee11c70deab30ea4a497c2/src/Parallel/MPIBasic.h#L89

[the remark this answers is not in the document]
krenzland: (There's a global singleton somewhere just grep through the code for "mpi" to find it

montrie:
Thanks for your feedback! I'll restructure the changes and implement your suggestions in the next few days.

Do you want me to keep pushing commits into this PR or should I simply open a new one to keep the number of commits low?

krenzland:
> Do you want me to keep pushing commits into this PR or should I simply open a new one to keep the number of commits low?

One PR is better here in my opinion because the changes belong together.

davschneller:
If you do want to, you could squash some of the commits. But you don't have to do it in my opinion, we don't require that right now as far as I'm aware.

montrie:
I created separate files for the CubeGenerator in src/Geometry and implemented most of the feedback. Please let me know if i missed something.
Right now the CubeGenerator.cpp file still generates and writes a new mesh every time it's called + I still use the NetCDFReader to actually read the generated mesh. I just wanted to push some progress into this PR. I'll start adjusting the CubeGenerator into a MeshReader tomorrow.

I tried rebasing the commit history since it would allow discarding the first two commits (which are now obsolete as we want the cube_c directory to be untouched). Due to internet issues my terminal keeps crashing and ending the rebase. I'll try again tomorrow when my internet is (hopefully) more stable.

montrie:
I couldn't get the clang tidy script to run locally without crashing, but the checks now pass anyway. 
I'll create a seperate PR when I adjusted the CubeGenerator to read a mesh file without generating it first.

CMakeLists.txt
```diff
@@ -242,7 +242,7 @@ endif()
 #set(HDF5_PREFER_PARALLEL True)
 if (NETCDF)
   find_package(NetCDF REQUIRED)
-  target_include_directories(SeisSol-common-properties INTERFACE ${NetCDF_INCLUDE_DIRS})
+  target_include_directories(SeisSol-common-properties INTERFACE ${NetCDF_INCLUDE_DIRS} ${CMAKE_SOURCE_DIR}/src/Geometry/)
```
davschneller, line 245: Obsolete?

src/Geometry/CubeGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+#include <map>
+#include <utility>
+#include <vector>
+
+#include <iostream>
+
+typedef int t_vertex[3];
+
+// changed Vertex to CubeVertex due to same struct naming here and in another part of SeisSol
+struct CubeVertex {
```
davschneller, line 60: Is this needed anywhere in the header or in some other places of SeisSol? Otherwise, better to to the cpp file. And enclose with an anonymous namespace then.

src/Geometry/CubeGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+
+  bool operator<(const CubeVertex& other) const {
+    return (v[0] < other.v[0]) || ((v[0] == other.v[0]) && (v[1] < other.v[1])) ||
+           ((v[0] == other.v[0]) && (v[1] == other.v[1]) && (v[2] < other.v[2]));
+  }
+};
+
+// Index of the vertices of a tetraedra in a cube
+// even/odd, index of the tetrahedra, index of vertex, offset of the vertices in x/y/z
+static const t_vertex TET_VERTICES[2][5][4] = {{{{0, 0, 0}, {1, 0, 0}, {0, 1, 0}, {0, 0, 1}},
```
davschneller, line 71: Same here; better to move into the cpp (or define it there, at least; and only have the forward declaration in the header), especially if it's not needed anywhere else.

src/Geometry/CubeGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+
+#include <algorithm>
+#include <cstring>
+#include <map>
+#include <utility>
+#include <vector>
+
+#include <iostream>
+
+typedef int t_vertex[3];
```
davschneller, line 57: Move to cpp?

src/Geometry/CubeGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+                                                {{1, 0, 0}, {1, 1, 1}, {0, 0, 1}, {1, 0, 1}},
+                                                {{0, 1, 0}, {0, 1, 1}, {0, 0, 1}, {1, 1, 1}},
+                                                {{1, 0, 0}, {0, 1, 0}, {0, 0, 1}, {1, 1, 1}}},
+                                               {{{0, 0, 0}, {0, 1, 0}, {0, 1, 1}, {1, 1, 0}},
+                                                {{0, 0, 0}, {1, 1, 0}, {1, 0, 1}, {1, 0, 0}},
+                                                {{0, 0, 0}, {1, 0, 1}, {0, 1, 1}, {0, 0, 1}},
+                                                {{1, 1, 0}, {1, 0, 1}, {1, 1, 1}, {0, 1, 1}},
+                                                {{0, 0, 0}, {1, 1, 0}, {0, 1, 1}, {1, 0, 1}}}};
+
+static const int TET_SIDE_NEIGHBORS[2][5 * 4] = {
```
davschneller, line 82: Move to cpp?

src/Geometry/CubeGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+    {3, 3, 3, 0, 1, 3, 0, 2, 2, 2, 2, 1, 0, 1, 3, 1, 3, 0, 0, 2},
+    {2, 3, 0, 1, 1, 3, 3, 2, 2, 1, 1, 0, 0, 3, 2, 1, 2, 0, 0, 1}};
+
+static const int TET_SIDE_ORIENTATIONS[2][5 * 4] = {
```
davschneller, line 86: Move to cpp?

src/Geometry/CubeGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+  double cubeTx;
+  double cubeTy;
+  double cubeTz;
+};
+
+class CubeGenerator : public seissol::geometry::MeshReader {
+  public:
+  CubeGenerator(int rank,
+                int nProcs,
+                const char* meshFile,
```
davschneller, line 118: `const std::string& meshFile`

src/Geometry/CubeGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+  double cubeSy;
+  double cubeSz;
+  double cubeTx;
+  double cubeTy;
+  double cubeTz;
+};
+
+class CubeGenerator : public seissol::geometry::MeshReader {
+  public:
+  CubeGenerator(int rank,
```
davschneller, line 116: As a suggestion, how about passing a SeisSol MPI object here directly?

src/Geometry/CubeGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+                int nProcs,
+                const char* meshFile,
+                const seissol::geometry::CubeGeneratorParameters& cubeParams);
+
+  inline void loadBar(int x, int n, int r = 100, int w = 50);
+  const char* dim2str(unsigned int dim);
+  template <typename A, typename B>
+  std::pair<B, A> flip_pair(const std::pair<A, B>& p);
```
davschneller, line 124: Needed in the class, or dependent on a member? If yes, then good. Otherwise, it's better to remove it from the class, and make static.

src/Geometry/CubeGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+  void checkNcError(int error);
```
davschneller, line 125: Cf. comment one line above.

src/Geometry/CubeGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+
+class CubeGenerator : public seissol::geometry::MeshReader {
+  public:
+  CubeGenerator(int rank,
+                int nProcs,
+                const char* meshFile,
+                const seissol::geometry::CubeGeneratorParameters& cubeParams);
+
+  inline void loadBar(int x, int n, int r = 100, int w = 50);
+  const char* dim2str(unsigned int dim);
```
davschneller, line 122: Cf. comment one line below

CMakeLists.txt
```diff
@@ -235,7 +235,7 @@ endif()
 #set(HDF5_PREFER_PARALLEL True)
 if (NETCDF)
   find_package(NetCDF REQUIRED)
-  target_include_directories(SeisSol-common-properties INTERFACE ${NetCDF_INCLUDE_DIRS})
+  target_include_directories(SeisSol-common-properties INTERFACE ${NetCDF_INCLUDE_DIRS} ${CMAKE_SOURCE_DIR}/preprocessing/meshing/cube_c/include/)
```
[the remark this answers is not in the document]
montrie: Implemented requested changes except for avoiding generating a temporary Netcdf mesh. I'll mark the rest of the review comments as resolved without answering them.

src/Initializer/InputParameters.cpp
```diff
@@ -185,6 +188,38 @@ static void readModel(ParameterReader& baseReader, SeisSolParameters& seissolPar
...
+  seissolParams.cubeGenerator.parameters.cubeMaxY = reader.readWithDefault("cubemaxy", 6);
+  seissolParams.cubeGenerator.parameters.cubeMinZ = reader.readWithDefault("cubeminz", 6);
+  seissolParams.cubeGenerator.parameters.cubeMaxZ = reader.readWithDefault("cubemaxz", 6);
+
+  seissolParams.cubeGenerator.parameters.cubeX = reader.readWithDefault("cubex", 2);
+  seissolParams.cubeGenerator.parameters.cubeY = reader.readWithDefault("cubey", 2);
+  seissolParams.cubeGenerator.parameters.cubeZ = reader.readWithDefault("cubez", 2);
+
+  // only x dimension has its number of partitions set to number of MPI processes
+  seissolParams.cubeGenerator.parameters.cubePx = seissol::MPI::mpi.size();
```
davschneller, line 206: Just to be clear—we slice in the X direction, and keep the rest as it is? That may not really be the most surface-reducing method—but anything beyond is good in a another PR, if needed.

src/Initializer/InputParameters.hpp
```diff
@@ -104,6 +105,11 @@ struct MeshParameters {
   bool showEdgeCutStatistics;
 };
 
+struct CubeGeneratorParameters {
```
davschneller, line 108: Why enclose this into a struct on its own? Maybe `seissol::geometry::CubeGeneratorParameters` works already here?

src/Geometry/CubeGenerator.cpp
```diff
@@ -0,0 +1,1582 @@
...
+  checkNcError(nc_close(ncFile));
+
+  logInfo() << "Finished";
+}
+
+// Process had done i out of n rounds,
+// and we want a bar of width w and resolution r.
+// TODO put this i an library
+// default args from header file: r = 100, w = 50
+inline void seissol::geometry::CubeGenerator::loadBar(int x, int n, int r, int w) {
```
davschneller, line 1526: Make static, remove from class (cf. comments there).

src/Geometry/CubeGenerator.cpp
```diff
@@ -0,0 +1,1582 @@
...
+  // Only update r times.
+  if (x % (n / r) != 0)
+    return;
+
+  // Calculuate the ratio of complete-to-incomplete.
+  float ratio = x / (float)n;
+  int c = ratio * w;
+
+  // Show the percentage complete.
+  std::cout << std::setw(3) << (int)(ratio * 100) << "% [";
```
davschneller, line 1538: Why `std::cout` here, and not e.g. `logInfo`?

(other than that, I like having a progress bar :) )

src/Geometry/CubeGenerator.cpp
```diff
@@ -0,0 +1,1582 @@
...
+  //&<namespace>::<class>::<member_function> to pass pointer to member
+
+  int* vrtxSize = new int[numPartitions[3]];
+  std::fill(vrtxSize, vrtxSize + numPartitions[3], uniqueVertices.size());
+  checkNcError(nc_put_var_int(ncFile, ncVarVrtxSize, vrtxSize));
+  delete[] vrtxSize;
+  writes_done++;
+  loadBar(writes_done, netcdf_writes);
+
+  double* vrtxCoords = new double[uniqueVertices.size() * 3];
```
davschneller, line 1480: vector? (though that's probably something more long-term) Same several times above.

src/Geometry/CubeGenerator.cpp
```diff
@@ -0,0 +1,1582 @@
...
+              bndLocalIds[nextMPIIndex] =
+                  (((numCubesPerPart[2] - 1) * numCubesPerPart[1] + yy) * numCubesPerPart[0] + xx) *
+                      5 +
+                  3;
+              elemMPIIndices[(((numCubesPerPart[2] - 1) * numCubesPerPart[1] + yy) *
+                                  numCubesPerPart[0] +
+                              xx) *
+                                 20 +
+                             15] = nextMPIIndex++;
+              //                                                      } else {
```
davschneller, line 1411: Comment still needed?

src/Geometry/CubeGenerator.cpp
```diff
@@ -0,0 +1,1582 @@
...
+
+  unsigned int numBndElements[3];
+  numBndElements[0] = 2 * numCubesPerPart[1] * numCubesPerPart[2];
+  numBndElements[1] = 2 * numCubesPerPart[0] * numCubesPerPart[2];
+  numBndElements[2] = 2 * numCubesPerPart[0] * numCubesPerPart[1];
+
+  // output file name
+  std::string fileName = meshFile;
+
+  logInfo() << "Start generating a mesh using the CubeGenerator";
```
davschneller, line 101: logInfo(rank)

montrie:
Thanks for the feedback! I'll resolve the conversations as I work through them.

CMakeLists.txt
montrie: You're right, specifying that directory is redundant.

src/Geometry/CubeGenerator.h
montrie: Fixed

src/Initializer/InputParameters.cpp
montrie: That's correct. I'd delay finding a better solution until the next PR

src/Geometry/CubeGenerator.cpp
montrie: A leftover from the original CubeGenerator. I'll try to figure out if the commented code could be necessary for adjusting the CubeGenerator to read an existing mesh. In the meantime, I will remove the comment for this PR.

src/Geometry/CubeGenerator.cpp
montrie: Fixed

src/Geometry/CubeGenerator.cpp
montrie: That's a leftover from the original CubeGenerator code from `cube_c/`. I tried substituting the last `std::cout` call with `logInfo()` , but that results in a disjointed progress bar, e.g.
```
Thu Oct 05 15:20:53, Info:                              ]
Thu Oct 05 15:20:53, Info:  ===                         ]
Thu Oct 05 15:20:53, Info:  ========                    ]
Thu Oct 05 15:20:53, Info:  =============               ]
Thu Oct 05 15:20:53, Info:  ==================          ]
Thu Oct 05 15:20:53, Info:  =======================     ]
```
I could call `logInfo()` only when the progress bar is at 100% and write the progress into a `std::osstream`, but then for bigger files we would see no progress bar until the `CubeGenerator` is finished.
I can also try to create a logger object specifically for the progress bar to try to replicate the current behaviour, since AFAIK calling `logInfo()` always ends with a newline (\n).

src/Geometry/CubeGenerator.h
montrie: The `CubeVertex` struct is only used in the `CubeGenerator`. There was a naming clash with other parts of SeisSol which is why I renamed it to `CubeVertex`. 
I'll move it to the source file and enclose it in an anonymous namespace.

src/Geometry/CubeGenerator.h
montrie: I'll move it into the source file and add it to the anonymous namespace (cf. comment above)

src/Geometry/CubeGenerator.h
montrie: cf. comment above

src/Geometry/CubeGenerator.h
montrie: cf. comment above

src/Geometry/CubeGenerator.h
montrie: I removed the function declaration from the header file, pulled it out of the class and made it static. Afaik it's neither dependant on a member or explicitly needed in the class.

src/Geometry/CubeGenerator.h
montrie: cf. comment above

src/Geometry/CubeGenerator.h
montrie: cf. comment above

src/Geometry/CubeGenerator.cpp
montrie: cf. comment regarding `flip_pair()`

src/Geometry/CubeGenerator.h
montrie: I used the existing `Geometry/NetcdfReader.cpp` and `Geometry/PUMLReader.cpp` files as a rough 'template' when I created the `CubeGenerator` class. There, we also don't pass a SeisSol MPI object.
If an MPI object is the preferred implementation choice, I can adjust the constructor such that we pass a SeisSol MPI object directly.

src/Initializer/InputParameters.hpp
montrie: I enclosed it into a seperate struct to preserve the design symmetry of the struct `SeisSolParameters`. If that doesn't matter, I can adjust `SeisSolParameters` to use `seissol::geometry::CubeGeneratorParameters` directly.

src/Geometry/CubeGenerator.cpp
montrie: Since you noted that this change would be something more long-term, I'd delay it to another PR for the time being.

davschneller (approved):

PR #919, opened 2023-08-16
Add A64FX support

krenzland:
Minimal PR to add A64FX support


davschneller:
How does it compare to #755 (except that PR being rather old by now)? Here and there, the alignment for a64fx is different.

davschneller (approved):

krenzland:
Two things:
- The alignment is now to the correct cache line size of 256 bytes
- The other PR contains a lot of features that I'm unable to test at the moment as I no longer have access to Fugaku. E.g. the pinning stuff is not necessary on the clusters that I currently have access to as the CPUs are slightly differently configures.

PR #921, opened 2023-08-19
GPU Flux Merging

davschneller:
Based on the idea of Ravil's GPU flux matrix merge branch (https://github.com/SeisSol/SeisSol/tree/ravil/gpu-flux ).

From this branch of, this PR adds a CMake option that is enabled by default for Nvidia and AMD GPUs; also we now use the tensor simplification expression from YATeTo as far as it allows.

I've compared the two implementations in a sample scenario, and they yield the same results up to minuscule numerical errors (which are to be expected when changing multiplication orders etc.).


ravil-mobile:
It looks good to me.

In general, the merged flux matrices stem from the original ADER-DG. The split variant is an advanced feature implemented in one of Carsten's papers. In theory, it would make more sense to use `-DSEISSOL_SPLIT_FLUX` which is the default for CPUs and Intel GPUs. I mean if you look at the original ADER-DG papers you will find  $\mathcal{F}^{+,jih}$ and $\mathcal{F}^{-,j}$. 

I remember I was struggling with understanding of `self.db.rDivM[i][self.t('km')]`, `self.db.fP[h][self.t('mn')]`, `self.db.rT[j][self.t('nl')] ` at the beginning of my thesis.

If you decide to stay with the `flux merge` option then there are a few comments.

cmake/process_users_input.cmake
```diff
@@ -140,6 +139,23 @@ else()
...
+    set(DEVICE_VENDOR "amd")
+    set(DEVICE_FLUXMERGE_DEFAULT ON)
+else()
+    # TODO(David): adjust as soon as we add support for more vendors
+    set(DEVICE_VENDOR "intel")
+    set(DEVICE_FLUXMERGE_DEFAULT OFF)
+endif()
+
+if (WITH_GPU)
+    option(DEVICE_FLUXMERGE "Merge device flux matrices (recommended for AMD and Nvidia GPUs)" ${DEVICE_FLUXMERGE_DEFAULT})
```
ravil-mobile, line 156: Could you rename `DEVICE_FLUXMERGE` to something like `MERGE_FLUX` or `MERGED_FLUX`. Yes, it affects GPU performance. But, technically speaking, it belongs to numerics. I mean we don't need to be super specific and add `DEVICE` prefix. 

When it says `DEVICE` I immediately think that it belongs to the Device API (device submodule). However, the changes affect mainly YATeTo high level description of the numerics.

If you decide to go for the split variant you can name it `SPLIT_FLUX`

ravil-mobile:
Thanks a lot for taking care of it. I would suggest to `squash and merge` at the end.

davschneller:
Thanks for the feedback as well! I think I'll stay with the "flux merge" option—even though it may seem more advanced I'd personally consider Carsten's description the more "basic" variant.

Or rather, it's now called `PREMULTIPLY_FLUX`. And the commits are squashed now.

cmake/process_users_input.cmake
davschneller: Done

generated_code/generate.py
```diff
@@ -69,6 +69,7 @@
 cmdLineParser.add_argument('--PlasticityMethod')
 cmdLineParser.add_argument('--gemm_tools')
 cmdLineParser.add_argument('--drQuadRule')
+cmdLineParser.add_argument('--use_premultiply_flux', default='OFF')
```
krenzland, line 72: This should be a boolean flag like:
```
parser.add_argument('--feature', action='store_true')
parser.add_argument('--no-feature', dest='feature', action='store_false')
parser.set_defaults(feature=True)

```

krenzland (changes requested):
Looks good with one minor issue

krenzland:
By the way, the name "premultiply flux" isn't great imo. Mathematically it should be --use-volume-basis-surface-integral or simiar

(What the "flux decomposition" does is switching to a modal basis defined on the faces)

krenzland (approved):

PR #832, opened 2023-03-31
Rename hipSYCL/Open SYCL target to AdaptiveCpp

ravil-mobile:
Related PRs:

yateto: https://github.com/SeisSol/yateto/pull/46
Device: https://github.com/SeisSol/Device/pull/21

PR #922, opened 2023-08-19
Make SYCL choosable in CMake, and make NVHPC choosable for HIPSYCL/Open SYCL + Nvidia GPUs

davschneller:
Just what the title says.

However: it does break with the usual procedure that HIPsycl is looked for silently, and if not found, DPCPP is expected. Now it has to be chosen explicitly instead.

Also, we now force-override the `HIPSYCL_TARGETS` variable. Thus, if we first compile for Nvidia and then for AMD; the variable will update accordingly. And it will be corrected if in some setup, if has not been setup correctly.


CMakeLists.txt
```diff
@@ -379,36 +379,45 @@ if (WITH_GPU)
...
+    message(STATUS "Using HipSYCL for general SYCL offloading")
+    # add GPU dynamic rupture target
+    # always override HIPSYCL_TARGETS, as we may need to override it when the architecture changes
+    if (DEVICE_ARCH MATCHES "sm_*")
+      if (CMAKE_CXX_COMPILER_ID MATCHES "NVHPC|PGI")
+        set(SYCL_USE_NVHPC_DEFAULT ON)
+      else()
+        set(SYCL_USE_NVHPC_DEFAULT OFF)
+      endif()
+      option(SYCL_USE_NVHPC "For Nvidia GPUs, use nvhpc instead of CUDA/nvcc." ${SYCL_USE_NVHPC_DEFAULT})
```
ravil-mobile, line 396: I would suggest to you the following script:

https://github.com/SeisSol/Device/blob/ravil/opensycl/cmake/find-opensycl.py

and the corresponding cmake file

https://github.com/SeisSol/Device/blob/ravil/opensycl/cmake/FindOpenSYCLSettings.cmake

It is from the following PR: https://github.com/SeisSol/Device/pull/21

The idea is that one may want to configure SeisSol with just SYCL and use `hipsycl`+`nvhpc` as a backend.

Currently, the change in this PR need to be applied in three places: DR, Device and `src/sycl.cmake`.

However, the solution that I am proposing will result in something like
```
  find_package(OpenSYCLSettings REQUIRED)
  find_package(OpenSYCL REQUIRED)
  target_link_libraries(SeisSol-device-lib PRIVATE opensycl-settings::device_flags)
```
in three parts of the code.

You can see all required changes here: https://github.com/SeisSol/SeisSol/pull/832

krenzland (approved):

krenzland:
I've approved it because the change looks reasonable to me. However, I haven't merged it (yet) because I'm not sure whether you finished your discussion of this PR.

PR #923, opened 2023-08-19
Optionally Serialize Device Initialization

davschneller:
This PR serializes Device initialization on a node. (the patch was originally by Ravil, I merely made the serialization operation a bit more general and placed it into `Parallel/MPI.h`)

Needed for some AMD GPU drivers.

### DIFF
--- a/src/Initializer/MemoryManager.cpp
+++ b/src/Initializer/MemoryManager.cpp
@@ -95,7 +95,18 @@ void seissol::initializers::MemoryManager::initialize()
   // initialize global matrices
   GlobalDataInitializerOnHost::init(m_globalDataOnHost, m_memoryAllocator, MEMKIND_GLOBAL);
   if constexpr (seissol::isDeviceOn()) {
-    GlobalDataInitializerOnDevice::init(m_globalDataOnDevice, m_memoryAllocator, memory::DeviceGlobalMemory);
+    bool serialize = false;
+    if (const char* value = std::getenv("SEISSOL_SERIALIZE_NODE_DEVICE_INIT")) {
+      serialize = strcmp(value, "1") == 0;
+    }
+    if (serialize) {
+      MPI::mpi.serializeOperation([&]() {
+        GlobalDataInitializerOnDevice::init(m_globalDataOnDevice, m_memoryAllocator, memory::DeviceGlobalMemory);
+      }, MPI::mpi.sharedMemComm());
+    }
+    else {
+      GlobalDataInitializerOnDevice::init(m_globalDataOnDevice, m_memoryAllocator, memory::DeviceGlobalMemory);
+    }
   }
 }
 
--- a/src/Initializer/MemoryManager.cpp
+++ b/src/Initializer/MemoryManager.cpp
@@ -100,6 +100,7 @@ void seissol::initializers::MemoryManager::initialize()
       serialize = strcmp(value, "1") == 0;
     }
     if (serialize) {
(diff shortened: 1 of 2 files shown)


src/Initializer/MemoryManager.cpp
```diff
@@ -95,7 +95,19 @@ void seissol::initializers::MemoryManager::initialize()
   // initialize global matrices
   GlobalDataInitializerOnHost::init(m_globalDataOnHost, m_memoryAllocator, MEMKIND_GLOBAL);
   if constexpr (seissol::isDeviceOn()) {
-    GlobalDataInitializerOnDevice::init(m_globalDataOnDevice, m_memoryAllocator, memory::DeviceGlobalMemory);
+    bool serialize = false;
+    if (const char* value = std::getenv("SEISSOL_SERIALIZE_NODE_DEVICE_INIT")) {
+      serialize = strcmp(value, "1") == 0;
+    }
```
krenzland, line 101: This isn't really robust as it doesn't handle incorrect values for the env flag. 
IMO there should be one check if the flag exists and one that the value is correct (and if it isn't, errors)

src/Initializer/MemoryManager.cpp
```diff
@@ -95,7 +95,19 @@ void seissol::initializers::MemoryManager::initialize()
...
+    if (serialize) {
+      logInfo(MPI::mpi.rank()) << "Serializing device global data initialization.";
```
krenzland, line 103: I would use a different term as serializing for me rather means "serializing the data". Something like "Using serial mode for device global data initialization."?

src/Parallel/MPI.h
```diff
@@ -180,6 +181,30 @@ class MPI : public MPIBasic {
     return collected;
   }
 
+  template<typename F>
```
krenzland, line 184: Add a comment that explains what this function is serializing

src/Initializer/MemoryManager.cpp
```diff
@@ -95,7 +95,19 @@ void seissol::initializers::MemoryManager::initialize()
...
+      MPI::mpi.serializeOperation([&]() {
```
krenzland, line 104: Add a comment why this is needed

krenzland (changes requested):
Logic looks good, but needs more documentation (please also in our readthedocs).

src/Initializer/MemoryManager.cpp
davschneller: That is already done by the `if`: `value == nullptr` if it does not exist. I'll make it more explicitly nonetheless.

src/Initializer/MemoryManager.cpp
davschneller: Done

src/Parallel/MPI.h
```diff
@@ -180,6 +181,30 @@ class MPI : public MPIBasic {
...
+  void serializeOperation(F&& operation, std::optional<MPI_Comm> comm = {}) {
+    if (!comm.has_value()) {
+      comm = std::optional<MPI_Comm>(m_comm);
+    }
+
+    int rank, size;
+    MPI_Comm_rank(comm.value(), &rank);
+    MPI_Comm_size(comm.value(), &size);
+
+    const int tag = 15; // TODO(David, or someone else): replace by a tag allocation system one day
```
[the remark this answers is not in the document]
davschneller: Done

davschneller:
Thanks for the review. Some documentation is also included now.

krenzland (approved):

PR #928, opened 2023-08-22
Merge YATeTo Taylor Expansion Kernels

davschneller:
(For now a draft PR, until properly benchmarked and integrated with chainforge)

This PR uses the latest YATeTo features (scalar families) to combine the `derivative` and `derivativeTaylorExpansion` kernel families into single kernels. As a result, we obtain slightly cleaner code. Also, once propagated properly to gemmforge/chainforge, we should also see less kernels being pushed to the GPU as a result (that is just beginning to be the case—but the copy-scale-adds are still to be combined).


davschneller:
Note: even though the tests pass, this PR not quite ready yet: chainforge still fails due to hitting the recursion limit in a deepcopy (the GPU builds were done with gemmforge only, there it works). Also, the viscoelastic2 kernel, simplified like that, is still a bit whacky; maybe there's a way to get Yateto to allocate intermediate arrays on the fly. Maybe we'll also only do the kernel merge for GPU kernels in the end.

But, as soon as chainforge works here, we'll have one big GPU kernel for the whole Cauchy-Kovalevski procedure, minus the Taylor series computation (which'll stay a bunch of copy-scale-add calls for now). It would be highly desirable to merge the latter kernels as well though in there—maybe soon.

Needless to say, the kernels will also need to be benchmarked before merge ... Or they'll stay here as a draft for some time.

vikaskurapati (approved):
Approved per my knowledge.

PR #934, opened 2023-08-30
Update Gitlab pipelines

davschneller:
Makes the Gitlab pipelines work.

Internally, we moved the CPU pipeline to the SCCS CI node now.

Also, the single stages are now using different node configurations.

But still, right now there are dependencies of the CPU version to be build with the respective CPU features (due to the doctest CMake script requiring them—I might file an issue for that; and the GPU version needs a CUDA runtime available which may become dificult, for now)

### DIFF
--- a/gitlab-cpu-pipeline.yml
+++ b/gitlab-cpu-pipeline.yml
@@ -1,6 +1,7 @@
 default:
     tags:
-        - lrzcc-k8s-4cores-10gb 
+        - sccs
+        - helper
     image:
         name: $CI_REGISTRY_USER/$cpu_image_name:$cpu_image_version
         entrypoint: [""]
@@ -33,6 +34,9 @@ fetch_submodules:
         
 build_seissol:
     stage: build
+    tags:
+        - sccs
+        - build
     needs:
         - job: fetch_submodules
     parallel:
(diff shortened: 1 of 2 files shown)


krenzland (approved):

krenzland:
> But still, right now there are dependencies of the CPU version to be build with the respective CPU features (due to the doctest CMake script requiring them—I might file an issue for that; and the GPU version needs a CUDA runtime available which may become dificult, for now)

What features exactly?

davschneller:
> > But still, right now there are dependencies of the CPU version to be build with the respective CPU features (due to the doctest CMake script requiring them—I might file an issue for that; and the GPU version needs a CUDA runtime available which may become dificult, for now)
> 
> What features exactly?

AVX or AVX2—when compiling on a Westmere CPU, they cause a Sigill quickly. Doctest executes the serial test file once—which may be compiled for the target architectures.

davschneller:
Cf. here: https://github.com/SeisSol/SeisSol/blob/e6ef66194c9b4aecc62bd8d5e808427f1376a8d3/cmake/doctestAddTests.cmake#L36-L40

krenzland:
Ah, but I think that is the intended behavior... :(

PR #942, opened 2023-09-04
Adding `fixtimestep` Restriction for Viscoelastic

### LINKED ISSUE #939: Recent changes break viscoelastic2
**Describe the bug**
Viscoelastic2 in single precision results in `Inf/NaN` error. Double precision works well.

**Expected behavior**
Viscoelastic2 should not result in an `Inf/Nan` error, when using single precision.

**To Reproduce**
1. Master branch (7fde47786f10ae)
2. gcc, viscoelastic2, single precision build
3. local
4. https://github.com/SeisSol/precomputed-seissol/tree/sebastian/more-ci/tpv5-visco

Slice through x==0, in the middle you see the correct solution, towards the boundaries you see numerical artifacts.
image

**Additional context**
It works with `v1.0.1`

davschneller:
(from pre #829 `readpar.f90`)

Fixes #939 ; see details for the bug there.

For now, I've just added it to the parameter reading again. In the future (especially with regard to multimaterial simulations), it may be ultimately more useful to instead have a template which gives us an allowed maximum timestep per parameters and material—but right now that would probably over-complicate things I'd guess...


sebwolf-de (approved):
LGTM

PR #951, opened 2023-09-11
Allow retry for GPU build

sebwolf-de:
The GPU pipeline fails quite often due to these errors, e.g. https://gitlab.lrz.de/ci_seissol/SeisSol/-/jobs/6640476
image
This PR makes the pipeline more robust, as it allow retries. Maybe 3 tries would be even better.


davschneller (approved):
We may want to also increase the runner waiting time—but why not have some retries in case something goes wrong anyways?

PR #954, opened 2023-09-13
Add "Dynamic Rupture" to the Proxy

davschneller:
As the title says, this PR adds the dynamic rupture computation to the proxy. For that, we also add a new parameter specifying the fault type. For now, no FLOP/s nor bytes numbers are reported, but merely the computational time.


davschneller:
(superseded due to the recent refactor; may be re-written)

PR #961, opened 2023-09-21
Non-Temporal Memory Transfers for ARM

davschneller:
This PR adds non-temporal loads/stores for ARM processors—both for an AARCH64 and SVE flavor.


krenzland:
Works on my M2. I tried a convergence test (Snell's law) and achieved the same error for both this branch and master. Performance difference is hart to quantify. As this is a laptop, the difference between runs with the same settings is smaller than the difference between the settings.

sebwolf-de:
@jwjeremy do you have capacity to test this branch and the master branch on your macbook before the training next week?

davschneller:
Just checked on a Grace CPU—the energy results (up to 15 digits) are the same for the noarch, the SVE and the AARCH64 codepath.

vikaskurapati (approved):
LGTM to as far as I can understand

Issue #960, opened 2023-09-21
Undef behavior in plasticity kernel?


krenzland:
**Describe the bug**
I get the following warnings when compiling (current master with slight modifications NOT in plasticity, but with an ALIGNMENT=128bytes):
```
/Users/lukaskrenz/src/SeisSol/src/Kernels/Plasticity.cpp:198:25: warning: iteration 336 invokes undefined behavior [-Waggressive-loop-optimizations]
  198 |         dudt_pstrain[q] = factor * (prev_degreesOfFreedom[q] - degreesOfFreedom[q]);
      |         ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/lukaskrenz/src/SeisSol/src/Kernels/Plasticity.cpp:173:30: note: within this loop
  173 |       for (unsigned q = 0; q < 6 * NUMBER_OF_ALIGNED_BASIS_FUNCTIONS; ++q) {
      |                            ~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/lukaskrenz/src/SeisSol/src/Kernels/Plasticity.cpp:210:22: warning: iteration 56 invokes undefined behavior [-Waggressive-loop-optimizations]
  210 |         QEtaModal[q] = pstrain[6 * NUMBER_OF_ALIGNED_BASIS_FUNCTIONS + q];
      |         ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/lukaskrenz/src/SeisSol/src/Kernels/Plasticity.cpp:209:30: note: within this loop
  209 |       for (unsigned q = 0; q < NUMBER_OF_ALIGNED_BASIS_FUNCTIONS; ++q) {
      |                            ~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/lukaskrenz/src/SeisSol/src/Kernels/Plasticity.cpp:238:73: warning: iteration 56 invokes undefined behavior [-Waggressive-loop-optimizations]
  238 |         pstrain[6 * NUMBER_OF_ALIGNED_BASIS_FUNCTIONS + q] = QEtaModal[q];
      |                                                              ~~~~~~~~~~~^
[...]

PR #962, opened 2023-09-22
Update CPU Architectures, Update YATeTo

davschneller:
* Adds several new CPU architectures, as shown in YATeTo right now (but not yet `apple-m2`). Also adds dummy architectures for SVE and NEON.
* Splits between `ALIGNMENT` and `VECTORSIZE`. The latter, `VECTORSIZE` should equal the YaTeTo value; it's from now on used in calculating any aligned reals or basis functions. The former value, `ALIGNMENT` may be larger (e.g. as large as a cache line), but mostly, it's still the same value.
* Updates YATeTo. That also includes https://github.com/SeisSol/yateto/pull/66 , and by that a significant speed improvement for the kernel generation.
* Reworks the `GEMM_TOOLS_LIST=auto` option to include LIBXSMM_JIT before LIBXSMM, and the new architectures. Also, all of these programs need to be found by CMake before they are considered for `auto`. Eigen is now always added at the end in this case.
* Redzone is now disabled by default (especially for noarch), and enabled for most X86_64 CPUs again.


davschneller:
Maybe also related to #960 . (i.e. the split between `ALIGNMENT` and `VECTORSIZE`)

src/Kernels/common.hpp
```diff
@@ -126,7 +126,7 @@ constexpr unsigned int
  * @return aligned number of reals.
  **/
 constexpr unsigned int getNumberOfAlignedReals(unsigned int i_numberOfReals,
-                                               unsigned int i_alignment = ALIGNMENT) {
+                                               unsigned int i_alignment = VECTORSIZE) {
```
sebwolf-de, line 129: Should we then also change the variable name to `i_vectorsize`?

src/Kernels/common.hpp
```diff
@@ -145,7 +145,7 @@ constexpr unsigned int getNumberOfAlignedReals(unsigned int i_numberOfReals,
  **/
 constexpr unsigned int
     getNumberOfAlignedBasisFunctions(unsigned int i_convergenceOrder = CONVERGENCE_ORDER,
-                                     unsigned int i_alignment = ALIGNMENT) {
+                                     unsigned int i_alignment = VECTORSIZE) {
```
sebwolf-de, line 148: Change variable name?

src/Kernels/common.hpp
```diff
@@ -161,7 +161,7 @@ constexpr unsigned int
  **/
 constexpr unsigned
     getNumberOfAlignedDerivativeBasisFunctions(unsigned int i_convergenceOrder = CONVERGENCE_ORDER,
-                                               unsigned int i_alignment = ALIGNMENT) {
+                                               unsigned int i_alignment = VECTORSIZE) {
```
sebwolf-de, line 164: Change variable name?

src/Kernels/common.hpp
krenzland: IMO no, because it is an alignment (to vectorsize)

krenzland: (same below)

davschneller: Ok, I've left it that way for now—since even though it's `VECTORSIZE`, we're aligning everything. Instead, we may have to rename the `ALIGNMENT` variable soon, I'd guess. And also, make both of them a C++ constexpr maybe.

I've also cleaned up the file a bit (i.e. remove `i_` etc.) and templated the functions with the size of the reals used. ... That could also become a parameter instead.

sebwolf-de (approved):

PR #963, opened 2023-09-25
Add support for MacOS & M1/M2

krenzland:
Requires https://github.com/TUM-I5/ASYNC/pull/1 before merging.

Most ifdefs are in Checkpointing (which gets a rewrite quite soon). Other than that, some changes needed to be done in the pinning part.

Can also be used as a template for other non POSIX system. Technically, it doesn't add anything Mac specific, it just removed non-POSIX features. (However, there's

### DIFF
--- a/cmake/cpu_arch_flags.cmake
+++ b/cmake/cpu_arch_flags.cmake
@@ -22,7 +22,7 @@ function(get_arch_flags architecture compiler)
 
     # Knights Corner (Xeon Phi)
     elseif ("${HOST_ARCH}" STREQUAL "knc")
-        set(HAS_REDZONE ON PARENT_SCOPE)    
+        set(HAS_REDZONE ON PARENT_SCOPE)
         set(CPU_ARCH_FLAGS "-mmic" "-fma" PARENT_SCOPE)
     
     # Knight Landing (Xeon Phi)
@@ -101,8 +101,16 @@ function(get_arch_flags architecture compiler)
         set(CPU_ARCH_FLAGS "" PARENT_SCOPE)
     elseif ("${HOST_ARCH}" STREQUAL "sve2048")
         set(CPU_ARCH_FLAGS "" PARENT_SCOPE)
+
     elseif ("${HOST_ARCH}" STREQUAL "apple-m1")
-        set(CPU_ARCH_FLAGS "-mcpu=apple-m1" PARENT_SCOPE)
+        if (compiler MATCHES "GNU")
+            set(CPU_ARCH_FLAGS "-march=armv8.4-a -mtune=generic" PARENT_SCOPE)
+        endif()
+
+    elseif ("${HOST_ARCH}" STREQUAL "apple-m2")
+        if (compiler MATCHES "GNU")
+            set(CPU_ARCH_FLAGS "-march=armv8.5-a" "-mtune=generic" PARENT_SCOPE)
+        endif()
     endif()
 
     if (compiler MATCHES "NVHPC|PGI")
--- a/cmake/process_users_input.cmake
+++ b/cmake/process_users_input.cmake
@@ -33,10 +33,10 @@ set_property(CACHE EQUATIONS PROPERTY STRINGS ${EQUATIONS_OPTIONS})
 
 
 set(HOST_ARCH "hsw" CACHE STRING "Type of host architecture")
-set(HOST_ARCH_OPTIONS noarch wsm snb hsw knc knl skx naples rome milan bergamo thunderx2t99 power9 a64fx neon sve128 sve256 sve512 sve1024 sve2048 apple-m1)
+set(HOST_ARCH_OPTIONS noarch wsm snb hsw knc knl skx naples rome milan bergamo thunderx2t99 power9 a64fx neon sve128 sve256 sve512 sve1024 sve2048 apple-m1 apple-m2)
 # size of a vector registers in bytes for a given architecture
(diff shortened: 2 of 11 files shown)


davschneller:
Effectively depends on #962 due to #960 .

krenzland:
Not really, the bug in #960 isn't new.

davschneller:
In what sense? Where has it already occurred?

krenzland:
Well, it's a compiler warning. I'm not even sure it's actually broken in practice :D

davschneller:
I'd guess it's the mismatch between tensor size of `QEtaModal` and `NUMBER_OF_ALIGNED_BASIS_FUNCTIONS`, in this case caused by the difference in alignment between Yateto (where it's 16 B) and SeisSol proper (128 B). While `QEtaModal` has 56 entries at 8 B each, we get to 448 B which will give a different result whether it's rounded up against 16 B (nothing happens) or 128 B (we get 512 B). It may not have appeared so far, since we've mostly had Yateto alignment == SeisSol alignment.

...

Now I'm wondering... Did these warnings also occur for A64FX on double precision? (there, we should have the same problem as well)

krenzland:
I honestly haven't checked. I didn't run anything involving plasticity on A64FX. I'll fix it in this PR.

krenzland:
Should be fixed now, if it isn't, the CI should catch it, right? (Assuming that we have a plasticity test?)

krenzland:
Not really, but I've used a more modern GCC. The warnings definitely come from the difference in alignments of the tensors.

cmake/process_users_input.cmake
```diff
@@ -139,23 +139,6 @@ else()
 endif()
 message(STATUS "GEMM TOOLS are: ${GEMM_TOOLS_LIST}")
 
-
-if (DEVICE_ARCH MATCHES "sm_*")
```
davschneller, line 143: Has been incorrectly removed (until line 158)?

src/Parallel/Pin.cpp
```diff
@@ -164,26 +178,37 @@ cpu_set_t seissol::parallel::Pinning::getFreeCPUsMask() const {
...
+#endif // __APPLE__
 }
 
-bool seissol::parallel::Pinning::freeCPUsMaskEmpty(cpu_set_t const& set) {
-  return CPU_COUNT(&set) == 0;
+bool Pinning::freeCPUsMaskEmpty(const CpuMask& mask) {
+#ifndef __APPLE__
+  return CPU_COUNT(&(mask.set)) == 0;
+#else
+  return false;
```
davschneller, line 198: For the commthread? Wouldn't `true` make here more sense otherwise?

davschneller (commented):
One thing vanished during some merge (?), rest looks good. Although subclassing the pinning would have been even cleaner IMO

cmake/cpu_arch_flags.cmake
```diff
@@ -63,8 +63,20 @@ function(get_arch_flags architecture compiler)
 
     elseif ("${HOST_ARCH}" STREQUAL "a64fx")
         set(HAS_REDZONE OFF PARENT_SCOPE)
-	set(CPU_ARCH_FLAGS "-mcpu=a64fx" PARENT_SCOPE)
-     endif()
+        set(CPU_ARCH_FLAGS "-mcpu=a64fx" PARENT_SCOPE)
+
+    elseif ("${HOST_ARCH}" STREQUAL "apple-m1")
+        set(HAS_REDZONE OFF PARENT_SCOPE)
+        if (compiler MATCHES "GNU")
```
davschneller, line 70: What about non-GNU?

krenzland: Haven't checked yet ;)
The only real alternative is clang but for that OpenMP becomes rather painful because Apple doesn't include the runtime lib per default...

davschneller: According to godbolt, they may also have the `-mcpu=apple-m1` or `-mcpu=apple-m2` flags the last time I checked (cf. also #962)

krenzland: Clang has, but I have no way of testing it without changing how OpenMP is linked - definitely not worth the effort for now
GCC doesn't have these flags.

src/Monitoring/LoopStatistics.cpp
```diff
@@ -197,6 +197,14 @@ template <> struct type2nc<int32_t> {
 template <> struct type2nc<uint32_t> {
     static constexpr auto type = NC_UINT;
 };
+
+// Long int is guaranteed to be of size 64 on Unix/Linux:
+// https://en.cppreference.com/w/cpp/language/types
+// This specialization is required on some architectures as it is later used for the type of
+// timespec::tv_sec, which is not fixed in the C++ standard.
+template <> struct type2nc<long int> {
```
davschneller, line 205: However, if `std::is_same_v<long int, int64_t>` holds, we got a re-definition. Since `int64_t` is merely a type redefinition. For some suggestions, we could either

* `#ifdef __APPLE__` that one.
* choose the type generically based on type traits and size (`std::numeric_limits` etc may help as well)
* if NetCDF has a HDF5-like `H5T_NATIVE_LONG`, then that would be also a way to go: do it for `int`, `long`, `long long` and you should be good.

krenzland: Huh, that's actually a good point. The CI catched it anyway. I'll look for a nice solution :)

davschneller (approved):

PR #967, opened 2023-10-10
Install SeisSol into bin and lib dirs

davschneller:
This PR puts the SeisSol binaries in the bin dir, and libraries like the general offloading etc. into the lib dir. The libraries are labeled appropriately to match potential different SeisSol builds.

The launch script is placed into share.


davschneller:
Maybe we could also think of either merging the libraries into the binary longer term (if that's feasible), or do the opposite: extract all code into a library and have slim binaries for proxy and SeisSol proper.

krenzland (approved):
LGTM
Is there a reason why we do not link statically?

davschneller:
@ravil-mobile were there any concerns with keeping the SYCL and GPU-compiled parts separate from the SeisSol executable like that?

PR #971, opened 2023-10-12
Add (shorter) CI expiration time span for artifacts

davschneller:
Default time span is 30 days; seemingly accumulating too much memory in total.

Thus, we now set the duration of all artifacts to 2 days.


davschneller:
Maybe, we could also merge the results checking (i.e. energy, receiver confirming etc.) with the test execution—unless we may need such data for finding errors.

krenzland (approved):

PR #972, opened 2023-10-13
Remove new sharedmem comm allocation for node mask

davschneller:
Title says it all. We have communicators allocated each time the communication thread gets re-started, but they never get deleted.

### DIFF
--- a/src/Parallel/Pin.cpp
+++ b/src/Parallel/Pin.cpp
@@ -237,9 +237,7 @@ CpuMask Pinning::getNodeMask() const {
     workerMaskArray[cpu] = CPU_ISSET(cpu, &workerMask);
   }
 
-  MPI_Comm commNode;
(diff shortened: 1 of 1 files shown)


davschneller:
We could think of caching the node and free CPU mask maybe? Or are they bound to change from thread to thread or over time in some extreme cases?

krenzland:
> We could think of caching the node and free CPU mask maybe? Or are they bound to change from thread to thread or over time in some extreme cases?

Why would we? Optimizations like this are a waste of time as they don't contribute much (if at all) to a measurable increase of the time-to-solution.

davschneller:
So that this PR wouldn't have to exist.

Other than that I agree, extremely low priority—unless more errors appear.

krenzland (approved):

krenzland:
This PR here makes some sense but caching the mask doesn't really. It's a simple all-to-all communication. If that fails, the MPI implementation on the cluster is defective.

PR #975, opened 2023-10-13
Update find libxsmm

uphoffc:
SeisSol calls find_package(LIBXSMM) twice if GEMM_TOOLS_LIST is omitted, but that is illegal as add_library is called twice. This PR fixes the LIBXSMM find script such that add_library is only called once should the find script be invoked multiple times.

### DIFF
--- a/cmake/FindLIBXSMM.cmake
+++ b/cmake/FindLIBXSMM.cmake
@@ -1,59 +1,58 @@
-include(FindPackageHandleStandardArgs)
+if(LIBXSMM_INCLUDE_DIRS AND LIBXSMM_INCLUDE_DIRS AND LIBXSMM_LIBRARIES AND LIBXSMM_VERSION)
+    set(LIBXSMM_FOUND TRUE)
+else()
+    find_path(LIBXSMM_INCLUDE_DIRS libxsmm.h
+            PATH_SUFFIXES include
+            DOC "Directory where libxsmm header files are located")
+    find_library(LIBXSMM_LIBRARIES
+            NAMES xsmm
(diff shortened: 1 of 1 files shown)


krenzland (approved):
LGTM

cmake/FindLIBXSMM.cmake
```diff
@@ -1,59 +1,58 @@
...
-list(APPEND LIBXSMM_LIBRARIES ${LIBDL_LIBRARIES})
-
-if (NOT(
-        LIBXSMM_INCLUDE_DIRS STREQUAL "LIBXSMM_INCLUDE_DIRS-NOTFOUND" OR
-        LIBXSMM_LIBARIES STREQUAL "LIBXSMM_LIBRARIES-NOTFOUND"))
-    add_library(LIBXSMM::LIBXSMM INTERFACE IMPORTED)
-    set_target_properties(LIBXSMM::LIBXSMM PROPERTIES
-            INTERFACE_LINK_LIBRARIES "${LIBXSMM_LIBRARIES}"
-            INTERFACE_INCLUDE_DIRECTORIES "${LIBXSMM_INCLUDE_DIRS}")
+if(LIBXSMM_INCLUDE_DIRS AND LIBXSMM_INCLUDE_DIRS AND LIBXSMM_LIBRARIES AND LIBXSMM_VERSION)
```
davschneller, line 1: What happens if all of them are `-NOTFOUND`?

uphoffc: That case should be handled by find_package_handle_standard_args

PR #978, opened 2023-10-19
Filtering

krenzland:
This PR adds filtering to SeisSol (compare with the filtering section in Hesthaven's Nodal DG book).
The idea is to decrease the coefficients of the high-order terms of the modal basis.

For matrix computation, refer to: https://github.com/SeisSol/Matrices/pull/10

Filtering works (and is tested to some extent):
- [x] Volume kernel
- [x] Local flux
- [x] Neighbor flux
- [x] Initial condition
- [x] Point sources 
- [x] Using a filtering matrix instead of the resample matrix for the slip rate
- [ ] Dynamic rupture flux
- [ ] Nodal flux (used for some boundary conditions
- [ ] Source terms
- [ ] Viscoelasticity

Works on the following architectures:
- [x] CPUs
- [ ] GPUs

You can use it like this in the config file:

```Fortran
&DynamicRupture
filtertype = 'none' ! none or exponential, exponential activates it
filterorder = 32 ! 32 is a good starting value, 16 increases dissipation

&Discretization
filtertype = 'exponential'
filterorder = 32
```


vikaskurapati:
Testing the current version of filtering with point source simulations. Slices are taken through the source and there is a lot of noise when u-velocity is shown near the source.
without_filter
After applying an exponential filter with order 32
with_filter
We notice that there is lesser noise around the source now, even though not perfect, the noise decreases

krenzland:
I've made some changes and now the PR only filters the residuals. This is more correct and also results in 0% overhead compared to no filtering. No support for GPUs and some updates required by some boundary conditons. Only tested for elasticity but should work for other materials (except visoelasticity).

As this is my last day at TUM, @vikaskurapati is going to take this over.

PR #979, opened 2023-10-20
Carsten/comm proxy

uphoffc:
I was looking for a bug related to communication and I thought it would be useful to have a reproducer for the communication part only. I added a reproducer that currently only checks whether communication "works", but I think one could easily extend it to analyze communication performance on a large number of nodes.

In order to use it one has to run SeisSol with "DumpLocalTimeSteppingStructure=1" in the Output section. SeisSol then writes a JSON dump of the MeshStructure and TimeStepping structs. The JSON file is then read into the communication proxy and the MeshStructure thingy is reconstructed. I then re-use the Actor classes in order to orchestrate the communication.


davschneller (commented):
Hi, sorry for the delay; partial review for now.

Also, why not use the SeisSol Device module instead of "just" SYCL allocation? It should have all functions available there already—if not, then we can add them there.

PR #982, opened 2023-10-23
[GPU] Move buffersDerivatives bucket to device memory instead of shared memory

uphoffc:
MPI on shared / unified memory is tricky to handle for an MPI implementation and might lead to increased latency and reduced bandwidth even for an advanced MPI implemention (cf. https://sc19.supercomputing.org/proceedings/workshops/workshop_files/ws_pmbsf116s2-file1.pdf).

The only relevant place where MPI on shared memory is needed is the buffersDerivatives bucket. That bucket is only used by kernels ported to the GPU already, when setting the initial value (first touch), and by an energy calculation routine. I here propose to move buffersDerivatives to global memory. Necessary changes are to port "first touch" on the GPU and to explicitly copy time derivatives to the host in the energy output routine.

This PR makes the SEISSOL_PREFERRED_MPI_DATA_TRANSFER_MODE=device mode obsolete and is thus removed. I left SEISSOL_PREFERRED_MPI_DATA_TRANSFER_MODE=host as that may be used as work-around if the MPI implementation is not GPU-aware.


krenzland:
Okay, this seems to make sense. Some caveats may apply:
- The paper is from a 2019 conference. Are the results still true for current gen MPI implementations/architectures?
- On some modern architectures the explicit copy can be actually slower. As this is only required for output here, I see no problem.

Do you have initial performance results?

Documentation/LatexFigures/gpu-comm-layer-data-flow.png
krenzland, line 1: Do you have this figure also in a vector graphics format? Would make further editing easier.

src/Solver/time_stepping/GhostTimeClusterFactory.h
```diff
@@ -26,15 +26,6 @@ struct GhostTimeClusterFactory {
                                               otherGlobalTimeClusterId,
                                               meshStructure);
     }
-    case MPI::DataTransferMode::CopyInCopyOutDevice: {
```
krenzland, line 29: I'm wondering here whether it would make sense to keep this copy mode in place - maybe it is useful in the future for prototyping?

krenzland (commented):
Looks good. I have two remarks/questions which may improve the PR slightly. I did not approve yet to avoid accidentally merging this PR but it is IMO good enough without changes to merge.

davschneller:
Hi, sorry for the delay with my review; I'll try to finish it soon-ish.

One thing to start with however: it would be good to not introduce new SYCL kernels (or reduce them to a minimum); the SeisSol device module should have respective kernels already—or if it does not have it, then it would be better to add it there, IMO; to keep consistency for the CUDA and ROCm implementations.

Documentation/LatexFigures/gpu-comm-layer-data-flow.png
davschneller: Probably, we'd have to ask @ravil-mobile  for that—he's the original author.

uphoffc:
> Okay, this seems to make sense. Some caveats may apply:
> 
> * The paper is from a 2019 conference. Are the results still true for current gen MPI implementations/architectures?
> * On some modern architectures the explicit copy can be actually slower. As this is only required for output here, I see no problem.
> 
> Do you have initial performance results?

Re MPI implementations: I don't have a more recent reference, but the case MPI-on-device-memory is certainly simpler than MPI-on-shared-memory, therefore there is no disadvantage introduced for anybody.

Re explicit copy: The only places where I introduced copies is either in initialization or at sync points. Both occur rarely in comparison to time-steps.

Re performance results: Similar performance as with SEISSOL_PREFERRED_MPI_DATA_TRANSFER_MODE=device for the test cases I looked at, but without the necessity to set an environment variable. (Every env variable less is a win imho...)

src/Solver/time_stepping/GhostTimeClusterFactory.h
uphoffc: I don't think so. Device2device copy makes little sense and if one wants to prototype a new feature one can base that on CopyInCopyOutHost

uphoffc:
> Hi, sorry for the delay with my review; I'll try to finish it soon-ish.
> 
> One thing to start with however: it would be good to not introduce new SYCL kernels (or reduce them to a minimum); the SeisSol device module should have respective kernels already—or if it does not have it, then it would be better to add it there, IMO; to keep consistency for the CUDA and ROCm implementations.

Why? The entire dynamic rupture implementation is SYCL-only and does not use the device submodule anymore. SYCL is already a mandatory dependency for the GPU build so I don't see why one would continue to have one's own abstraction layer for simple things (?).

davschneller:
But we (should) already have all the functionality available sort of; e.g. https://github.com/SeisSol/Device/blob/1d3eddef605c1bbccc9f4fe61af7149fc27bc9d6/Algorithms.h#L23 and so on. So why not use it?

More generally... Do we even need to touch the data on the GPUs? It's overridden otherwise anyways, in all cases.

Maybe the initialization kernels are only needed for the CPU in the end? Or is a copy enough? (cf. e.g. something like https://github.com/SeisSol/SeisSol/tree/davschneller/semi-transparent-buffers does at the moment)

uphoffc:
Re "why not use it": So why would you roll your own C++ device abstraction layer (that someone has to implement and maintain) if there's a Khronos-standardized C++ device abstraction layer?
The obvious answer to this question is: "Because I do not want to rely on feature X being supported in SYCL for vendor Y, therefore I rather roll my own device abstraction layer."
That answer is completely fine and valid. However, as the dynamic rupture part is completely written in SeisSol and you cannot compile the GPU version without SYCL, the Device sub-module is obsolete for the "simple" stuff, because there's really no need to have two different device abstraction layers.
So either you throw out the SYCL stuff from the main code and implement everything in the Device sub-module OR you start throwing out the Device sub-module. Due to the existence of the dynamic rupture SYCL code I understood that SeisSol follows the first strategy - eventually throw out the Device sub-module. Hence, I have not added new kernels or used existing kernels in the Device sub-module.

Re "first touch": It might not be necessary for the GPU version, but it ensures that the GPU version behaves the same way as the CPU version and it really does not take long.

ravil-mobile:
Hi, I agree. Initially, the dynamic rupture part was implemented with OpenMP offloading, which resulted in a very low performance (especially regarding GCC). Moreover, OpenMP implementations are inconsistent with each other; thus, it forces one to use ugly C-macros. Therefore, I switched to SYCL, which actually showed good results.

The main problem is mixing CUDA/HIP and SYCL in one codebase. The SYCL queue creates and maintains its own CUDA/HIP stream, which is not exposed to the user. Therefore, it is impossible to make a proper synchronization when switching from the wave propagation solver to the dynamic rupture one. The only thing that we can do is to synchronize the entire device at the beginning and at the end of the dynamic rupture code. Once it is done, we can safely use SYCL. But, device synchronization can be expensive if it is used too frequently.

I implemented 3 versions of the off-fault plasticity model and the gravitational free surface boundary condition -i.e., one for CUDA, one for HIP, and one for SYCL. I  immediately regretted that because I had to maintain 3 versions of the same code, but it helped me avoid frequent device synchronizations.

Can we use just SYCL? So, the performance of the generated kernels is better with nvcc, and amdclang compilers. Secondly, both CUDA and HIP provide the graph-based execution model and graph capturing. I find it quite useful for the LTS scheme during strong scaling. Unfortunately, as far as I know, there is no such feature in SYCL. By and large, the answer depends on whether we are okay with sacrificing performance over portability. 

I wish SYCL had a feature to construct a queue from a CUDA/HIP device stream (maybe through some standard extensions). In this case, we would solve the problem with device synchronization. In general, it would really simplify the GPU programming in SeisSol; it would allow us to move many auxiliary algorithms to GPUs through SYCL. In theory, it is possible to extend and patch the SYCL runtime to make it happen, but the standard committee will probably reject it immediately.

krenzland:
> There's SYCL graphs. Currently an experimental extension but it might make into the standard in the future:

Do you know whether these extensions are implemented in DPC++/OpenSYCL and what the typical process of that is? I'm asking just out of curiosity

davschneller:
Hi, sorry to be the guy; but all in all, I'm still not fully convinced... I'm sorry. I offer to replace the kernels myself.

In total, the stability for SYCL with CUDA/HIP(ROCm) was not that convincing in the encounters I've had so far—it seems like we may still a bit of time to fully mature in all details. And the build process was made rather intricate that way. Feel free to contest me here, but that was my experience so far.

This all may be my humble (or rather un-informed) opinion as a pretty new PhD candidate; but I'd like to keep my options open for the coming years. Especially before we start now locking in into a direction that may turn out to be the wrong one (I do know about the various SYCL implementations etc.).

Maybe I'll change the DR implementation again, maybe SYCL will be gone or can be made optional then. E.g. when extending Yateto (or a potential successor project) with non-linearities which are enough to describe the plasticity kernels, and most of the DR part. It was by now not my plan to eventually deprecate the device module at this moment—we have it, and it works well.

In a few years, I'm probably on your side with regards to SYCL, but not right now. Once again, sorry for holding the odd opinion out here.

uphoffc:
> > There's SYCL graphs. Currently an experimental extension but it might make into the standard in the future:
> 
> Do you know whether these extensions are implemented in DPC++/OpenSYCL and what the typical process of that is? I'm asking just out of curiosity

I think it is already implemented, the headers are there:
https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/ext/oneapi/experimental/graph.hpp

Looks like it is available for the codeplay CUDA plugin as well:
https://developer.codeplay.com/products/oneapi/nvidia/2024.0.0/guides/changelog-nvidia

I don't know how the typical process is.

krenzland:
> This all may be my humble (or rather un-informed) opinion as a pretty new PhD candidate; but I'd like to keep my options open for the coming years. Especially before we start now locking in into a direction that may turn out to be the wrong one (I do know about the various SYCL implementations etc.).

Keeping options open isn't a valid strategy here because we already committed to SYCL. 
Just because you *maybe* want to rewrite the dynamic rupture implementation (why?) doesn't mean that we cannot use SYCL now. SeisSol isn't usable without SYCL on GPUs - it's a dependency.

So I see no valid reason (except the CI failing currently) not to merge this PR.

davschneller:
Anyways, there is unfortunately still something that this branch will need to do so that we can merge it: the test will need to pass. For that, the branch will need to cover the fault output—there we also need the derivatives.
Due to the long delay (sorry for that again); and to throw a suggestion into the ring, I had started to prepare a branch which should do that (minus bugs), under https://github.com/SeisSol/SeisSol/tree/davschneller/mpi-on-device-memory-with-device-submodule . Of course, again not completely error-free.

Feel free to take from it what you want; it builds upon this branch here. Or leave it, if you know a better way or want to do it with pure SYCL instead (and not rely on the device module)—that is good for me as well. It still yields the wrong values at the moment, but the structure should be good already. I had originally intended it as a pseudo-"showcase" for what we can do without SYCL, but ok.

To say, generally I prefer _not_ using SYCL (due to reasons mentioned before), but I can understand using it. In most cases, it seems to be simple to replace anyways if we want to switch to something different in some years. I apologize for causing any inconvenience; I thought that the device module would capture all simple use cases (which it mostly/sort of does), hence I was confused by the non-usage of it. For more complicated kernels, I would not have contested a SYCL implementation.

Also thanks for the hints regarding the oneAPI / SYCL graphs. I'll try to get them in soon-ish.

davschneller (approved):
Let's merge for now—I'll open another PR in just a second which fixes the fault output issue (and also plays back a bit of the SYCL things, so that they stay contained within point sources and DR for the time being; until we get a better kernel generator out).

PR #983, opened 2023-10-23
Fix bug in free surface integrator that occurs if SeisSol alignment != yateto alignment

uphoffc:
Alignment in yateto does not match alignment in cmake/process_users_input.cmake. Therefore, the code in initializeProjectionMatrices had an out-of-bounds write. Fix in this PR.

### DIFF
--- a/src/Solver/FreeSurfaceIntegrator.cpp
+++ b/src/Solver/FreeSurfaceIntegrator.cpp
@@ -162,9 +162,12 @@ void seissol::solver::FreeSurfaceIntegrator::initializeProjectionMatrices(unsign
   // Sub triangles
   triRefiner.refine(maxRefinementDepth);
 
-  numberOfSubTriangles = triRefiner.subTris.size();
-  numberOfAlignedSubTriangles = seissol::kernels::getNumberOfAlignedReals(numberOfSubTriangles);
+  const auto projectionMatrixNCols = tensor::subTriangleProjection::Shape[tensor::subTriangleProjection::index(maxRefinementDepth)][1];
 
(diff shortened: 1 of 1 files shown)


davschneller:
Note that we also have `VECTORSIZE` as new parameter which should have fixed that issue. However—seems like the Intel GPUs has been forgotten here (`VECTORSIZE` should be 32 here, but it's still 128).

src/Solver/FreeSurfaceIntegrator.cpp
```diff
@@ -162,9 +162,12 @@ void seissol::solver::FreeSurfaceIntegrator::initializeProjectionMatrices(unsign
   // Sub triangles
   triRefiner.refine(maxRefinementDepth);
 
+  const auto projectionMatrixNCols = tensor::subTriangleProjection::Shape[tensor::subTriangleProjection::index(maxRefinementDepth)][1];
```
davschneller, line 165: `constexpr`?

davschneller: Ok; thanks, missed that one.

uphoffc:
> Note that we also have `VECTORSIZE` as new variable which should have fixed that issue. However—seems like the Intel GPUs have been forgotten here (`VECTORSIZE` should be 32 here, but it's still 128).

Yes that's a good idea. Here, I compute the value directly from yateto, so that should be quite robust.

davschneller (approved):

src/Solver/FreeSurfaceIntegrator.cpp
uphoffc: Tried it briefly but compiler said NO :-D

Issue #427, opened 2021-11-18
Deterministic mesh partitioning


jrekoske:
When running identical jobs (same mesh and number of MPI ranks), I noticed that the partitions are different. This is undesirable for certain post-processing activities. 

I read the parmetis manual (http://glaros.dtc.umn.edu/gkhome/fetch/sw/parmetis/manual.pdf) and all the parmetis API’s accept an option[] array. It appears that option[2] specifies the seed used to generate a partition.

From page 24:
 options This is an array of integers that is used to pass additional parameters for the routine. The first element
(i.e., options[0]) can take either the value of 0 or 1. If it is 0, then the default values are used,
otherwise the remaining three elements of options are interpreted as follows:

options[2] This is the random number seed for the routine.

So it would appear that fixing the value of option[2] = XXX would solve the non-deterministic partition problem.

Does anyone foresee any problems with fixing the seed for all simulations?

jrekoske:
It seems like a simple fix to change this line 

https://github.com/TUM-I5/PUML2/blob/57bfa7df18ab3610e66a5827a64dca6b28bd168b/PartitionMetis.h#L139

to idx_t options[3] = {1, 1, 0}

krenzland:
METIS_RANDOM_SEED is a constant already (defined in the same file), so the metis call itself should be deterministic.
There is also the compile option -DMINI_SEISSOL=OFF which sets the weights of all nodes to 1 (instead of varying weights). Note that this will likely lead to a loss of performance for large runs.

Why is the inconsistent mesh partitioning a problem for your use case?

AliceGabriel:
Hi, we here really need to have consistent element ids across many simulations using the same mesh. The simulations are not large.

krenzland:
Sure, but it is quite easy to reorder elements in post-processing, especially for small simulations. You can steal the logic from here:  https://github.com/SeisSol/SeisSol/blob/master/postprocessing/science/compute_diff_seissol_data.py

In any case, I believe that -DMINI_SEISSOL=OFF should lead to deterministic partitioning.

ravil-mobile:
I can confirm that by configuring SeisSol with `-DMINI_SEISSOL=OFF` is going to solve the problem.

Thomas-Ulrich:
good to know. Maybe we can document this idea and close the issue?

krenzland:
[...]

PR #986, opened 2023-10-30
Always compile with comm thread and mini SeisSol enabled

davschneller:
As the title says, both the communication thread and mini SeisSol currently have compile-time variables to switch them on/off—but there's nothing from always supporting them. Thus, we do the following:

* introduce environment variables `SEISSOL_MINISEISSOL` and `SEISSOL_COMMTHREAD` (both boolean) which control if the two features are enabled, respectively
* also, introduce an additional parameter variable for mini SeisSol, in case we want to encode the determinism there.


src/Parallel/Helper.hpp
```diff
@@ -0,0 +1,25 @@
...
+namespace seissol {
+template <typename T>
+void printCommThreadInfo(const T& mpiBasic) {
+  bool useThread = utils::Env::get<bool>("SEISSOL_COMMTHREAD", true);
+  if (mpiBasic.isSingleProcess()) {
+    logInfo(mpiBasic.rank()) << "Using polling for advancing MPI communication, due to having only "
+                                "one MPI rank running.";
+  }
+  if (useThread) {
+    logInfo(mpiBasic.rank()) << "Using a communication thread for advancing MPI communication.";
```
Thomas-Ulrich, line 14: so if SEISSOL_COMMTHREAD=True and one node, we will get:
```
Using polling for advancing MPI communication, due to having only one MPI rank running.
Using a communication thread for advancing MPI communication.
```
Which is not what expected.

cmake/process_users_input.cmake
```diff
@@ -6,11 +6,9 @@ set(GRAPH_PARTITIONING_LIB_OPTIONS parmetis parhip ptscotch)
 set_property(CACHE GRAPH_PARTITIONING_LIBS PROPERTY STRINGS ${GRAPH_PARTITIONING_LIB_OPTIONS})
 
 option(MPI "Use MPI parallelization" ON)
-option(MINI_SEISSOL "Use MiniSeisSol to compute node weights for load balancing" ON)
 option(OPENMP "Use OpenMP parallelization" ON)
 option(ASAGI "Use asagi for material input" OFF)
 option(MEMKIND "Use memkind library for hbw memory support" OFF)
-option(USE_IMPALA_JIT_LLVM "Use llvm version of impalajit" OFF)
```
Thomas-Ulrich, line 13: this is not really related to the PR?

src/SeisSol.cpp
```diff
@@ -98,14 +99,19 @@ bool seissol::SeisSol::init(int argc, char* argv[]) {
...
+  if (seissol::useCommThread(MPI::mpi)) {
+    auto freeCpus = pinning.getFreeCPUsMask();
+    logInfo(rank) << "Communication thread affinity        :"
+                  << parallel::Pinning::maskToString(freeCpus);
+    if (parallel::Pinning::freeCPUsMaskEmpty(freeCpus)) {
+      logError()
+          << "There are no free CPUs left. Make sure to leave one for the communication thread. If "
+             "you want to run SeisSol without a communication thread (and instead use polling), "
+             "then try running with the environment variable \"SEISSOL_COMMTHREAD=false\". "
+             "However, for more balanced performance, we recommend using the communication thread.";
```
Thomas-Ulrich, line 113: what means more balanced performance? higher performance?

src/Initializer/InitProcedure/InitMesh.cpp
```diff
@@ -58,24 +59,26 @@ static void readMeshPUML(const seissol::initializer::parameters::SeisSolParamete
...
-    nodeWeight = 1.0 / elapsedTime;
-
-    const auto summary = seissol::statistics::parallelSummary(nodeWeight);
-    logInfo(rank) << "Node weights: mean =" << summary.mean << " std =" << summary.std
-                  << " min =" << summary.min << " median =" << summary.median
-                  << " max =" << summary.max;
-
-    writer::MiniSeisSolWriter writer(seissolParams.output.prefix.c_str());
-    writer.write(elapsedTime, nodeWeight);
+  if (utils::Env::get<bool>("SEISSOL_MINISEISSOL", true) && seissolParams.mesh.miniseissol) {
```
Thomas-Ulrich, line 62: why to read the environment variable, if miniseissol is defined in the parameter file?
I would only keep the parameter file.
(by the way, commthread could also be defined in the parameter file).

Thomas-Ulrich (commented):
I think it is a good idea. I ve noted a few things

Thomas-Ulrich:
you also need to update the doc.
(e.g. supermucNG page, frontera, etc, parameters.par, do a search, there are a lot of occurences of COMMTHREAD)

src/Initializer/InitProcedure/InitMesh.cpp
davschneller: Personally, I'd tend to keep them an environment variable at the moment, because they don't influence the simulation per se.
The mini SeisSol got into the parameter file as well, because it seemed to be needed in #427 IMO.

davschneller:
> you also need to update the doc. (e.g. supermucNG page, frontera, etc, parameters.par, do a search, there are a lot of occurences of COMMTHREAD)

Yup, you're right. Probably, it's good if we enable versioning for the readthedocs before that; I'll look into it.

src/Parallel/Helper.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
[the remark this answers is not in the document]
davschneller: Yup, I think so. I've replaced it.

cmake/process_users_input.cmake
davschneller: Not really; it's only an unused option at the moment IIRC—or at least, since easi was removed as a submodule.

src/SeisSol.cpp
davschneller: Removed that sentence.

src/Initializer/InitProcedure/InitMesh.cpp
Thomas-Ulrich: I don't understand. I would only keep the environement variable, and remove seissolParams.mesh.miniseissol.
#427 can be fixed by using the environement variable, right?

davschneller: Yup, it can. Ok, I'll remove the parameter from the parameter file.

davschneller:
Ok, the docs should be updated and the new environment variables documented.

Documentation/environment-variables.rst
```diff
@@ -1,7 +1,25 @@
...
+By default, any SeisSol run with more than one MPI rank will use a communication thread to advance the MPI progress engine.
+For that, you will need to leave at least one thread vacant in your OpenMP thread placing map, cf. the SuperMUC-NG example below.
+
+If you do not want to use a communication thread, you may set `SEISSOL_COMMTHREAD=0`; then SeisSol polls on the progress from time to time.
+
+Load Balancing
+--------------
+
+When running with multiple ranks, SeisSol will estimate the performance of a node, to enable better load balancing for it.
+For that, it runs the so-called "Mini SeisSol" benchmark. As its name already hints at—it simulates a small test workload on each node;
```
Thomas-Ulrich, line 18: typo "at—it"

Thomas-Ulrich (approved):
LGTM, thanks!

Documentation/environment-variables.rst
davschneller: Ok, replaced the "—" with a comma

PR #988, opened 2023-11-07
fail if dynamic rupture parameters not find

Thomas-Ulrich:
Require that the friction law parameters are specified in the parameter file, and do not use default values if not found.

### DIFF
--- a/src/DynamicRupture/Parameters.h
+++ b/src/DynamicRupture/Parameters.h
@@ -74,25 +74,33 @@ inline std::unique_ptr<DRParameters> readParametersFromYaml(std::shared_ptr<YAML
     drParameters->backgroundType = getWithDefault(yamlDrParams, "backgroundtype", 0);
     drParameters->isThermalPressureOn = getWithDefault(yamlDrParams, "thermalpress", false);
     drParameters->t0 = getWithDefault(yamlDrParams, "t_0", 0.0);
-    drParameters->rsF0 = getWithDefault(yamlDrParams, "rs_f0", 0.0);
-    drParameters->rsA = getWithDefault(yamlDrParams, "rs_a", 0.0);
-    drParameters->rsB = getWithDefault(yamlDrParams, "rs_b", 0.0);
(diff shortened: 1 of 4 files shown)


src/DynamicRupture/Parameters.h
```diff
@@ -74,25 +74,33 @@ inline std::unique_ptr<DRParameters> readParametersFromYaml(std::shared_ptr<YAML
     drParameters->backgroundType = getWithDefault(yamlDrParams, "backgroundtype", 0);
     drParameters->isThermalPressureOn = getWithDefault(yamlDrParams, "thermalpress", false);
     drParameters->t0 = getWithDefault(yamlDrParams, "t_0", 0.0);
-    drParameters->rsF0 = getWithDefault(yamlDrParams, "rs_f0", 0.0);
-    drParameters->rsA = getWithDefault(yamlDrParams, "rs_a", 0.0);
```
davschneller, line 78: Is `rs_a` still needed? It doesn't appear in the new parameters

davschneller (commented):
LGTM, only a single question regarding `rs_a`

Thomas-Ulrich:
Interestingly enough, this PR allow finding that rs_muw is not specified in tpv105 precomputed:
https://github.com/SeisSol/precomputed-seissol/blob/master/tpv105/parameters.par#L14-L40

>Tue Nov 07 19:40:44, Error: The field "rs_muw" was not found, but it is required. 

Should we then use rs_muw=0 in tpv105 precomputed, or change it to 0.2 (https://github.com/SeisSol/Examples/blob/master/tpv105/parameters.par#L22) and recompute the precomputed?

src/DynamicRupture/Parameters.h
Thomas-Ulrich: I removed this line as rsA is defined as a spatially dependent parameter (yaml file).
(I now also remove the other line mentioning this parameter).

davschneller:
Does it run through other code, if `rs_muw != 0`? Then maybe, it would be worth it—if it's not too much work to re-compute the pre-compute(d).

PR #989, opened 2023-11-08
Update CMakeLists.txt

davschneller:
Two bugfixes for CMake, with regard to finding the code generators and the recent adjustment to the installation folder structure.


sebwolf-de:
I've added another fix for libxsmm_jit.

sebwolf-de (approved):

PR #995, opened 2023-11-24
 Implementation of Instantaneous time mirror for SeisSol

vikaskurapati:
Implementation of Instantaneous Time Mirror with three different modes. The following parameters would be needed to in parameters.par to toggle an ITM;

ITMToggle = 1 or 0 //1 switches the ITM on, 0 switches it off
ITMStartingTime // time in simulation when the ITM must be started
ITMTime // duration of ITM
ITMVelocityscalingfactor = scaling factor of velocity depending on the mode of reflection
ITMReflectionType = 1 or 2 or 3 or 4 // 1 reflects both waves, 2 reflects both waves with constant velocity, 3 reflects only P-waves and 4 reflects only S-waves.


src/Solver/time_stepping/TimeManager.h
```diff
@@ -186,6 +186,10 @@ class seissol::time_stepping::TimeManager {
     void printComputationTime(const std::string& outputPrefix, bool isLoopStatisticsNetcdfOutputOn);
 
     void freeDynamicResources();
+
+    inline TimeStepping* getTimeStepping() {
```
davschneller, line 190: Make reference. Or maybe also const? If there's only one parameter you want to modify, it may look more elegant to offer a method to modify it directly.

src/SeisSol.h
```diff
@@ -143,6 +144,14 @@ class SeisSol {
    * Get the flop counter
    */
   monitoring::FlopCounter& flopCounter() { return m_flopCounter; }
+  /**
+   * Reference for timeMirrorManagers to be accessed externally when required
+   */
+  std::pair<seissol::ITM::InstantaneousTimeMirrorManager,
```
davschneller, line 150: Will there only be two time mirror managers at a time? (pun _not_intended, but welcome anyways)
Or can there be even more than that? (e.g. having multiple time reversals, maybe ping-ponging each other or similar—if that does make sense?)

src/ResultWriter/AnalysisWriter.cpp
```diff
@@ -53,6 +53,11 @@ void AnalysisWriter::printAnalysis(double simulationTime) {
      initialConditionType == seissol::initializer::parameters::InitializationType::PressureInjection) {
     return;
   }
+
+  if (initialConditionType == seissol::initializer::parameters::InitializationType::AcousticTravellingwithITM) {
```
davschneller, line 57: Do we need this If statement here? It looks a bit out of place to only do it for one initial condition

src/Physics/InstantaneousTimeMirrorParameters.h
```diff
@@ -0,0 +1,32 @@
+#ifndef SEISSOL_ITMCONFIG_H
+#define SEISSOL_ITMCONFIG_H
+
+#include <memory>
+#include <yaml-cpp/yaml.h>
+
+namespace seissol::initializers::ITM {
+class ITMParameters {
```
davschneller, line 8: Probably I had asked somewhere else already, but: do we need these parameters as well as the reading in the `InputParameters.{c|h}pp` files?

src/Physics/InstantaneousTimeMirrorParameters.h
```diff
@@ -0,0 +1,32 @@
...
+                double ITMVelocityScalingFactor,
+                double ITMStartingTime,
+                bool ITMToggle,
+                int reflectionType);
+};
+
+ITMParameters readITMParametersFromYaml(std::shared_ptr<YAML::Node>& params);
+} // namespace seissol::initializers::ITM
+
+#endif // SEISSOL_ITMCONFIG_H
```
davschneller, line 32: newline at the end of the file

src/Physics/InstantaneousTimeMirrorParameters.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+double ITMParameters::getITMVelocityScalingFactor() const { return ITMVelocityScalingFactor; }
+
+double ITMParameters::getITMStartingTime() const { return ITMStartingTime; }
+
+bool ITMParameters::getITMToggle() const { return ITMToggle; }
+
+int ITMParameters::getReflectionType() const { return reflectionType; }
+
+} // namespace seissol::initializers::ITM
```
davschneller, line 52: newline at the end of the file

src/Solver/time_stepping/AbstractGhostTimeCluster.cpp
```diff
@@ -116,4 +116,13 @@ void AbstractGhostTimeCluster::printTimeoutMessage(std::chrono::seconds timeSinc
         << "correctionsSinceSync = " << neighbor.ct.stepsSinceLastSync;
   }
 }
+
+ClusterTimes& AbstractGhostTimeCluster::getClusterTimes(){
```
davschneller, line 120: Maybe it's better to make this one const, and offer a method to explicitly change the one parameter you want to change?

src/Solver/time_stepping/AbstractGhostTimeCluster.cpp
```diff
@@ -116,4 +116,13 @@ void AbstractGhostTimeCluster::printTimeoutMessage(std::chrono::seconds timeSinc
...
+  return ct;
+}
+
+std::vector<seissol::time_stepping::NeighborCluster>* seissol::time_stepping::AbstractGhostTimeCluster::getNeighborClusters(){
```
davschneller, line 124: Reference? Or do you need this vector somewhere?

src/Solver/time_stepping/AbstractGhostTimeCluster.h
```diff
@@ -43,5 +42,7 @@ class AbstractGhostTimeCluster : public AbstractTimeCluster {
   void reset() override;
   ActResult act() override;
 
+  ClusterTimes& getClusterTimes();
+  std::vector<seissol::time_stepping::NeighborCluster>* getNeighborClusters();
```
davschneller, line 46: If we want to forward them, then it's maybe better to do that in the `AbstractTimeCluster` already

src/Solver/time_stepping/ActorState.h
```diff
@@ -78,6 +78,8 @@ struct ClusterTimes {
   [[nodiscard]] long computeStepsUntilSyncTime(double oldSyncTime,
                                                double newSyncTime) const;
 
+  [[nodiscard]] double& getTimeStepSize();
```
davschneller, line 81: I'd prefer a setter to a `double&`

src/Solver/time_stepping/DirectGhostTimeCluster.h
```diff
@@ -18,6 +18,9 @@ class DirectGhostTimeCluster : public AbstractGhostTimeCluster {
                            int globalTimeClusterId,
                            int otherGlobalTimeClusterId,
                            const MeshStructure* meshStructure);
+    ClusterTimes& getClusterTimes();
+    std::vector<seissol::time_stepping::NeighborCluster>* getNeighborClusters();
```
davschneller, line 22: Obsolete if put into abstract superclass?

src/Solver/time_stepping/TimeCluster.cpp
```diff
@@ -593,7 +597,6 @@ void seissol::time_stepping::TimeCluster::computeNeighboringIntegration( seissol
   }
 
   if (usePlasticity) {
-    updateRelaxTime();
```
davschneller, line 596: This one should stay

davschneller (commented):
Hi, thanks for the PR. I've attached a bunch of comments.
One thing maybe is that I'm myself not 100%ly sure how to properly integrate the time inversion manager (or for that matter, any simulation-changing mechanism) into the code, so that we have a good structure for that.

But maybe we can also leave it for now once the comments are addressed.

src/Initializer/InitProcedure/InitModel.cpp
```diff
@@ -244,6 +244,32 @@ static void initializeCellMatrices(LtsInfo& ltsInfo) {
 
   memoryManager.recordExecutionPaths(seissolParams.model.plasticity);
 #endif
+
+  //  const bool ITMToggle = memoryManager.getITMParameters()->getITMToggle();
+  auto itmParameters = seissol::SeisSol::main.getSeisSolParameters().itmParameters;
+
+  bool ITMToggle = itmParameters.ITMToggle;
```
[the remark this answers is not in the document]
vikaskurapati: I think it is better to have it in `InitModel.cpp` rather than `InitSideConditions.cpp` as the side conditions are more related to the initial conditions IMO. ITM activation can be done theoretically on all kinds of initial conditions.

Documentation/parameters.par
```diff
@@ -10,6 +10,12 @@ Tv=0.05
 FreqCentral=0.5
 FreqRatio=100
 GravitationalAcceleration = 9.81 ! value of gravitational acceleration
+!ITM Parameters
+ITMToggle = 1 ! 1 is on, 0 is off
```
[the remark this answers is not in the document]
vikaskurapati: Renamed just in the parameters. Let me know if you want the variables to be named too

src/Initializer/InputParameters.cpp
```diff
@@ -168,6 +168,22 @@ static void readModel(ParameterReader& baseReader, SeisSolParameters& seissolPar
...
       reader.readWithDefault("usecellhomogenizedmaterial", true);
+  seissolParams.itmParameters.ITMStartingTime = reader.readWithDefault("itmstartingtime", 0.0);
+  seissolParams.itmParameters.ITMTime = reader.readWithDefault("itmtime", 0.0);
+  seissolParams.itmParameters.ITMToggle = reader.readWithDefault("itmtoggle", bool(0));
+  seissolParams.itmParameters.ITMVelocityScalingFactor =
+      reader.readWithDefault("itmvelocityscalingfactor", 1.0);
+  seissolParams.itmParameters.reflectionType = reader.readWithDefault("itmreflectiontype", 1);
+
+  if (seissolParams.itmParameters.ITMTime < 0.0) {
+    logError() << "ITM Time is less than zero. It should be positive!\n";
```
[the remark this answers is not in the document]
vikaskurapati: If `itmtoggle==0`, nothing would happen. With `itmtime == 0.0`, nothing would happen either.

src/Initializer/InputParameters.cpp
```diff
@@ -366,6 +383,7 @@ static void readInitialization(ParameterReader& baseReader, SeisSolParameters& s
   seissolParams.initialization.origin =
       seissol::initializers::convertStringToArray<double, 3>(originString);
   const auto kVecString = reader.readWithDefault("kvec", std::string("0.0 0.0 0.0"));
+  seissolParams.initialization.k = reader.readWithDefault("k", 0.0);
```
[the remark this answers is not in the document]
vikaskurapati: Yes, for our test case, we need a one dimensional wave number and not a vector.

src/Initializer/MemoryManager.h
```diff
@@ -375,10 +377,15 @@ class seissol::initializers::MemoryManager {
...
 
+    inline ITM::ITMParameters* getITMParameters(){
+      return ITMParameters.get();
+    };
+
     void setInputParams(std::shared_ptr<YAML::Node> params) {
       m_inputParams = params;
       m_dynRupParameters = dr::readParametersFromYaml(m_inputParams);
       ltsParameters = std::make_shared<time_stepping::LtsParameters>(time_stepping::readLtsParametersFromYaml(m_inputParams));
+      ITMParameters = std::make_shared<ITM::ITMParameters>(ITM::readITMParametersFromYaml(m_inputParams));
```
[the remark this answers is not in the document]
vikaskurapati: Seems like it. May be I should remove it from here? Or from `InputParameters.hpp`?

src/Initializer/MemoryManager.h
```diff
@@ -375,10 +377,15 @@ class seissol::initializers::MemoryManager {
         return ltsParameters.get();
     };
 
+    inline ITM::ITMParameters* getITMParameters(){
```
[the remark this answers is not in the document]
davschneller: Yes, the `InputParameters.hpp` is more "centralized" and the more current thing. However, the DR and the LTS parameters still have their own files up to now (I'd guess we'll only unify everything once we have a completely new parameter format). I'd put the parameter reading still preferrably into the `InputParameters` files.

src/Solver/time_stepping/TimeManager.h
vikaskurapati: This was done by design for `initializeCellLocalMatrices` and similar methods. As far as I can recall, it was not done for the ITM. We just used the `initalizeCellLocalMatrices` to do what we want to do.

src/SeisSol.h
vikaskurapati: There are two managers, one which scales up and the other which brings the simulation parameters back to normal. In theory we could have more but it makes sense to have them only in even numbers when we want to apply ITM multiple times.

src/Physics/InstantaneousTimeMirrorParameters.h
vikaskurapati: No, corrected and pushed.

src/ResultWriter/AnalysisWriter.cpp
vikaskurapati: Need? No. I wanted it to show the log info for this specific initial condition as it is quite specific as it is for only one particular case with ITM for which we did our convergence analysis. So, I think it is nice to have so that anyone in the future who wants to do any convergence analysis with ITM can see in the log when it is chosen. But it can be removed.

src/Physics/InstantaneousTimeMirrorParameters.h
vikaskurapati: File removed

src/Physics/InstantaneousTimeMirrorParameters.cpp
vikaskurapati: File removed

src/SeisSol.h
davschneller: Still, if you're not bound to a certain number of it and use them basically always the same way (and they are, as far as I can see), I'd see it as cleaner to have a vector instead of a pair here.

In the `InitModel.cpp` you can explicitly add two managers to the vector then.

src/ResultWriter/AnalysisWriter.cpp
davschneller: There should be a log entry for the initial condition at the beginning of the simulation. What you could add is to notify the user if the ITM is active (i.e. at the `ITMToggle` if in the `InitModel.cpp` file)

src/Solver/time_stepping/TimeManager.h
davschneller: Thanks, I see now—it's because you want to have a reference in the ITM manager. The `initalizeCellLocalMatrices` does not use a pointer, but a reference. It's only to store the reference for later usage that we need the pointer here.

You _could_ also pass a reference from the `TimeManager` which is then converted to a pointer in the ITM manager, but that would just move the problem. Quite honestly, it's not ideal to have it that way exposing a pointer, but I don't see much possibility to change it unless we want to refactor the basic structure of Seissol a bit (which is arguably out of scope for this PR).

Can you maybe make the pointer `const`?

davschneller:
One thing to add: this PR adds two empty files `Interoperability.{h|cpp}` which are probably relics from the FORTRAN era (and got removed in #829)—they probably slipped in here by accident again. Mind removing them?

src/Solver/time_stepping/TimeManager.h
vikaskurapati: Making the pointer `const` makes the compiler complain of invalid conversion once again. I think it is better to let it stay untouched as it is being used by other methods too

src/SeisSol.h
vikaskurapati: Well, as we're going to use just two most of the time (I don't see anyone else creating more scenarios of it in the future), I still think a pair is more appropriate than a vector. It would have made more sense to use a vector when we use more or an unknown number of the. 

The output format is not affected by the ITM. ITM just changes the material parameters for a very short period of time.

davschneller: Oh—sorry; I meant input format. I.e. if you have a list of different ITM parameters (that does not work with the current namelists, but we're, by plan, going to replace the input format anyways soon-ish and use YAML iirc instead—there we would have arrays, and could define an array of ITMs if you want to simulate that).

We can keep it as an `std::pair`, although I quite honestly still don't see any advantage here (rather, we always have allocated two dangling ITM managers lying around if the functionality is not used; it may be cleaner to not have them at all in this case). And changing it to an `std::vector` should be done with only a few lines of code. But again, it's ultimately your choice.

src/Solver/time_stepping/AbstractGhostTimeCluster.cpp
vikaskurapati: Replaced with a direct method to access the required parameter

src/Solver/time_stepping/AbstractGhostTimeCluster.cpp
vikaskurapati: The vector is needed. If I recall it, @krenzland gave me this fix when the reference was not working for what we were thinking of doing.

Documentation/parameters.par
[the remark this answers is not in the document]
davschneller: Just renaming the parameters is good imo

src/Solver/time_stepping/AbstractGhostTimeCluster.h
vikaskurapati: Done

src/Solver/time_stepping/TimeManager.h
davschneller: Did you also make it const in the `initializeTimeMirrorManagers` and the ITM manager classes themselves—and their parameters? The conversion from const ptr to const ref should work

src/Solver/time_stepping/DirectGhostTimeCluster.h
vikaskurapati: Done

src/Solver/time_stepping/ActorState.h
vikaskurapati: Done

src/Solver/time_stepping/TimeCluster.cpp
vikaskurapati: Done

vikaskurapati:
I have addressed most of the comments and modified the code as required. Please check once again.

src/Solver/time_stepping/TimeManager.h
vikaskurapati: Ach! Yes, that makes more sense. Thank you! Fixed it now.

davschneller (commented):
Hi, just some tiny (unused) code snippets and some newlines which we can remove, then we're good I'd say.

davschneller (approved):

Issue #771, opened 2023-01-02
mpi error on supermucNG


NicoSchlw:
**Describe the bug**
The simulation hangs after a certain timestep but the job is not killed. The error log shows that this is mpi related.

**Expected behavior**
The simulation should run till the end. If an error occurs the job should be canceled.

**To Reproduce**
1. Which version do you use? Provide branch and commit id.
    Error occurred with nico/2nuc `7f8c759` and fix707-2nuc `a71e59e` without enforcing the state variable to be positive
2. Which build settings do you use? Which compiler version do you use? 
    Viscoelastic2, double precision, intel mpi compiler
3. On which machine does your problem occur? If on a cluster: Which modules are loaded?
    SupermucNG:
![Screenshot from 2023-01-02 10-10-55](https://user-images.githubusercontent.com/42612980/210212198-8987d4ba-48b2-44a4-866f-94fa7fa094ee.png)
4. Provide parameter/material files.
[parameters.par.txt](https://github.com/SeisSol/SeisSol/files/10330747/parameters.par.txt)

**Screenshots/Console output**
If you suspect a problem in the numerics/physics add a screenshot of your output.

If you encounter any errors/warnings/... during execution please provide the console output.
[2528718.ridgecrest.err.txt](https://github.com/SeisSol/SeisSol/files/10330751/2528718.ridgecrest.err.txt)
[2528718.ridgecrest.out.txt](https://github.com/SeisSol/SeisSol/files/10330752/2528718.ridgecrest.out.txt)

**Additional context**
The setup did already run successfully with nico/2nuc `7f8c759` but without writing surface/volume output.
[2526877.ridgecrest.out.txt](https://github.com/SeisSol/SeisSol/files/10330798/2526877.ridgecrest.out.txt)

krenzland:
Do you have the following line in your parameter file:

`export I_MPI_SHM_HEAP_VSIZE=8192`

See #691

NicoSchlw:
Hi, I did not have this line in my submission file when the problem occurred first. But I added it afterward, changed the CFL from 0.45 to 0.35, and used mpiexec instead of srun. Since then the error did not occur again at this timestep but twice at a later timestep (I did not recognize the first time because a node failure overprinted it):

[2553177.ridgecrest.out.txt](https://github.com/SeisSol/SeisSol/files/10371958/2553177.ridgecrest.out.txt)
[...]

PR #997, opened 2023-11-30
Persistent MPI

davschneller:
As the title says, we add persistent MPI support here. It's disabled by default, but the hopes are that it can reduce transport latencies—and maybe also remedy #771 ?

It's controlled via the environment variable `SEISSOL_MPI_PERSISTENT`.


davschneller:
Comments:
* We _could_ optimize the whole procedure even further by a little bit if we use `MPI_Startall`—which would require a bit more refactoring. Let me know if I should do that. :)
* By the way, the default behavior of all existing runs should not be altered by this.
* I'm not sure (I'll have to check that), if it's beneficial to have a certain alignment for the communication regions, at least on CPUs (e.g. page boundaries).

krenzland:
Do we actually have a measurable communication latency?

Thomas-Ulrich (approved):
as far as I understand LGTM.

PR #999, opened 2023-12-18
Disable Some OpenMP Clauses for NVHPC

davschneller:
NVHPC currently incorrectly implements some of the OpenMP clauses used in SeisSol—hence this PR to disable OpenMP at these places entirely. However, we introduce a flag in this PR that will probably enable the OpenMP clauses, once the bug has been fixed in NVHPC.

Supersedes #526 .


Thomas-Ulrich (approved):
LGTM

Issue #947, opened 2023-09-07
Replacing References by Pointers in LocalData struct


kirchhart-intel:
Hello SeisSol team!

my name is Matthias Kirchhart, I am an Intel employee working together with @uphoffc on SeisSol.

I suggest replacing references with pointers in the `LocalData` struct, and would like to discuss how to implement this.

### Rationale

For better performance, in the long run, it makes sense to shift from unified shared memory allocations to explicitly managed allocations using `sycl::malloc_device()` and `sycl::malloc_host()`. It allows for more fine-grained control when and which data is transferred between the host system and GPU devices.

### Description

For starters, I want to implement explicit data transfers, i.e., memcpy, between host and GPU in the `calcReceivers()` routine in [src/Kernels/Receiver.cpp](https://github.com/SeisSol/SeisSol/blob/master/src/Kernels/Receiver.cpp). 

For this I need to copy the required DOFs from the GPU to a local buffer on the host, and pass a `LocalData` object referring to that buffer to `computeAder()`. The problem is that `LocalData` objects only contain references and not pointers. So I cannot just create a copy of the `LocaData` object and have its dofs member “point-to” the local memory.

I would like to suggest the following solution:

```
struct LocalData
{
    real *member_ptr;
    
    real& member() const noexcept { return *member_ptr; }
};
```

This looks like a “clean” solution to me. However, it requires to change all the code in SeisSol that uses LocalData. Each time the old `data.member` is accessed, the we now need to write `data.member()` with the function call brackets.

Would this be okay for you? If so, I would start preparing the necessary changes and create a pull request once ready.

An alternative could be to have two members: one a reference, the other a pointer. This however, would increase memory consumption and lead to inconsistencies if the pointer member is changed.

I am looking forward to your feedback.

### General information
- [x] I have searched the issues of this repo and believe this is not a duplicate

krenzland:
Hi Matthias, 
I see no problem with using pointers there. The manual prefetching is okay. I would make this optional as it may be beneficial on some hardware to not prefetch it explicitly (e.g. Grace Hopper?)

davschneller:
Hi, a small question: how would you plan to implement the prefetching/copying exactly?

uphoffc:
[...]

PR #1000, opened 2023-12-20
Enabled explicit Device-to-Host data copies in Receiver.cpp

kirchhart-intel:
Enabled explicit Device-to-Host data copies in Receiver.cpp

This required changing the LocalData structure to use pointers instead of references, as discussed in https://github.com/SeisSol/SeisSol/issues/947.


davschneller:
Hi, thanks for the PR! We'll need the tests to pass still; right now that is:

* formatting. In case you have `clang-format` installed on your system, you should be good by running `sh .ci/format.sh $(which clang-format) .` in the SeisSol root directory.
* A reference on a file introduced in #982 is still in the CMakeLists which has not made it to the master yet—hence the build fails here. That should be fixed pretty soon-ish; the PR is on the merge queue soon.
* Once fixing all of that, we can also run the extended tests—which only trigger for branches in the repository—so I mirrored the branch for you (cf. https://github.com/SeisSol/SeisSol/tree/mirror/matthias/issue947 ) There was only one instance of brackets missing for debug builds with assertions enabled (cf. the last commit on the branch mentioned). Otherwise, we should be good.

(also note that the SYCL copy could be replaced by a `device::DeviceInstance::getInstance().api->copyTo` operation, but I can do that myself later-on)

davschneller:
Ok; I'm gonna run it through the extended CI again; then we should be merge-ready.

davschneller (approved):

PR #1001, opened 2023-12-21
Fix GPU fault output

davschneller:
Adds the fault output for the derivative calculation. (cf. what's been started by #982) We also clean up some SYCL code in the process.


davschneller:
We'll wait until #982 is through for this here.

vikaskurapati (approved):

PR #1004, opened 2023-12-29
Sebastian/seissol lib

sebwolf-de:
Work towards SeisSol as a library. In particular, this PR contains two features:
* Remove `seissol::SeisSol::main` singleton and replace it by `seissol::SeisSol& seissolInstance`.
* Gather the parameter reader in one place. SeisSol is given a parameter struct. Currently, there is not too much change for the user, but this facilitates other parameter readers.


davschneller:
Hi, thanks for the PR! Some short questions before reviewing:

* do we really want to take `initializers` as namespace, and not `initializer` (like the folder is called)?
* Are the materialized files in `viscoelastic` and `poroelastic` wanted like that (or are the symlinks enough)? If yes—then why not also for `anisotropic`?
* What about the SeisSol MPI object? (also the MPI init should happen maybe directly after the accelerator device init, regardless of what's going on)

sebwolf-de:
* Both variants were available in the code, I've decided for initializers, but we can also go for the single noun.
* No, this shouldn't have happened, this must have slipped through somehow.
* The MPI object is still static - this is just a first step towards removing all singletons.

davschneller (commented):
Just one comment (uppercase folder names)—we're good otherwise I'd say.

I'd vouch for going to `initializer` over `initializers`, to keep consistency with the naming of the folders below. If I'm the only one with that concern however, I'd be good with `initializers` as well.

Thomas-Ulrich:
it seems it you would have opted for initializers instead of initializer, this would drop 90% of the file changes.

davschneller (approved):

PR #1007, opened 2024-01-11
Remove superfluous include

davschneller:
Lets some builds fail—hence we remove it.

Ported from #1003 .


davschneller:
(note that the fault output will currently fail—due to a tiny dependency problem. Working on it)

Thomas-Ulrich (approved):

PR #1024, opened 2024-01-29
write fault tags in fault output

Thomas-Ulrich:
- Generalize the clustering output of Ravil in an intCellDataWriter (currently supports only one array per writer).
- Use it to write locationFlags in surface output (before it was written at each time step as a real), clustering in wavefield output, and fault tag in fault output. This later is useful when you want to post-process fault output segment-wise.
- Requires an update of the XdmfWriter: https://github.com/TUM-I5/XdmfWriter/pull/11


src/ResultWriter/FreeSurfaceWriterExecutor.h
```diff
@@ -116,6 +117,10 @@ class FreeSurfaceWriterExecutor
 		m_stopwatch.pause();
 	}
 
+	void setlocationFlagData(const unsigned int *locationFlags) {
```
davschneller, line 120: uppercase `l` for `location`

davschneller (approved):
As far as my knowledge goes, LGTM.

PR #1035, opened 2024-02-05
a new abort criteria for dynamic rupture simulations

Thomas-Ulrich:
In a recent project, I'm exploring a parameter space by running a set of dynamic rupture simulations.
If e.g. the stress is not critical enough, or Dc is too large, rupture may quickly die out.
Therefore I'm interested in detecting such a situation and stopping the simulation.
Here is a draft of a functional terminator (as named in Moose), which checks the min slip rate (minSR) over the rupture area, and stops the simulation x sec after minSR < threshold.
Here threshold is hardcoded to 0.4m/s (we can't have it too low because of tpv5 like artifacts).
Currently, the code lies in the DR energy output. Any suggestion for a better location is welcome.


src/DynamicRupture/FrictionLaws/FrictionSolver.h
```diff
@@ -47,6 +47,7 @@ class FrictionSolver {
    * For reference, see: https://strike.scec.org/cvws/download/SCEC_validation_slip_law.pdf.
    */
   real deltaT[CONVERGENCE_ORDER] = {};
+  real dt;
```
sebwolf-de, line 50: So `dt = sum(deltaT)`, right?

sebwolf-de (commented):
Regarding the abort criterion, I thought you could maybe check the diff in rupture energy? Then you'd also consider the area of the respective slipping patch.

Thomas-Ulrich:
> Regarding the abort criterion, I thought you could maybe check the diff in rupture energy? Then you'd also consider the area of the respective slipping patch.

yes, probably a good idea. However the rupture area may increase by creeping, and the area of creeping may increase with the already ruptured area. it is probably the same problem with the diff in rupture energy. Therefore I prefer a local abort critierion.

PR #1039, opened 2024-02-06
Add Global ID (and larger boundary conditions)

davschneller:
Adds a global ID to each element (in preparation to checkpointing). Also, we streamline 64-bit boundary conditions.

Dependent on the PUML2 PR https://github.com/TUM-I5/PUML2/pull/14 .


src/Geometry/PUMLReader.cpp
```diff
@@ -60,17 +63,116 @@
...
+                   << bcToString(sideBC)
+                   << "boundary condition, but the neighboring element is not flagged -1";
+      return false;
+    }
+  }
+  // ignore unknown boundary conditions and warn
+  else {
+    logWarning() << "Element" << cellIdAsInFile << ", side" << side
+                 << " has a boundary condition, which I don't understand" << sideBC;
+    return true;
```
Thomas-Ulrich, line 150: I would also return error in this case.

src/Geometry/PUMLReader.cpp
```diff
@@ -60,17 +63,116 @@
...
+const int PumlFaceToSeisSol[4] = {0, 1, 3, 2};
+
+const int FaceVertexToOrientation[4][4] = {
+    {0, 2, 1, -1}, {0, 1, -1, 2}, {0, -1, 2, 1}, {-1, 0, 1, 2}};
+
+const int FirstFaceVertex[4] = {0, 0, 0, 1};
+} // namespace
+
 /**
  * @todo Cleanup this code
```
Thomas-Ulrich, line 164: still required?

src/Geometry/PUMLReader.cpp
```diff
@@ -60,17 +63,116 @@
...
+    logWarning() << "Element" << cellIdAsInFile << ", side" << side
+                 << " has a boundary condition, which I don't understand" << sideBC;
+    return true;
+  }
+  return true;
+}
+
+const int PumlFaceToSeisSol[4] = {0, 1, 3, 2};
+
+const int FaceVertexToOrientation[4][4] = {
```
Thomas-Ulrich, line 157: can you add a comment about what is orientation?

src/Geometry/PUMLReader.cpp
```diff
@@ -60,17 +63,116 @@
...
+    {0, 2, 1, -1}, {0, 1, -1, 2}, {0, -1, 2, 1}, {-1, 0, 1, 2}};
+
+const int FirstFaceVertex[4] = {0, 0, 0, 1};
```
Thomas-Ulrich, line 160: can you add a comment about what it is?

src/Geometry/PUMLReader.h
```diff
@@ -41,28 +41,48 @@
 
 #include "MeshReader.h"
 #include "Parallel/MPI.h"
-#include "PUML/PUML.h"
+#include <Initializer/Parameters/MeshParameters.h>
```
Thomas-Ulrich, line 44: shouldn't it be "Initializer/Parameters/MeshParameters.h"?

Thomas-Ulrich (changes requested):

davschneller:
Ok; I've added an `auto` option which checks the HDF5 file. (if we ever should add bare-binary Xdmf support or Hdf5 data from different files, we may need to read the XML file instead)

palgunadi1993:
I tested this PR. It works fine with my fracture network model. Thanks @davschneller !

Thomas-Ulrich (approved):

PR #1044, opened 2024-02-07
Explicit Taylor Sum Kernels

davschneller:
Currently, the some of Taylor sum kernels do very small reads and writes—which is bad. Therefore, we for now fuse them manually, until our code generation can help us here.

Note that the results will be slightly different—since we now use `real` everywhere during the computation.

(yes, I know that "some" and "sum" sound similar, non-intended pun :P)

Draft until performance verified everywhere.


src/Equations/elastic/Kernels/Time.cpp
```diff
@@ -430,6 +433,9 @@ void Time::computeBatchedIntegral(double expansionPoint,
   intKrnl.streamPtr = runtime.stream();
   intKrnl.execute();
   device.api->popStackMemory();
+#else
+  seissol::kernels::time::aux::taylorSum(true, numElements, timeIntegratedDofs, timeDerivatives, deltaTLower, deltaTUpper, runtime.stream());
+#endif
 #else
   assert(false && "no implementation provided");
```
vikaskurapati, line 440: Can we change this to `logError` instead?

vikaskurapati (approved):
LGTM, minor comments. Can be merged after those changes

PR #1053, opened 2024-02-15
Fix compilation

sebwolf-de:
Fix compilation of damaged-elastic model

### DIFF
--- a/src/DynamicRupture/FrictionLaws/BaseFrictionLaw.h
+++ b/src/DynamicRupture/FrictionLaws/BaseFrictionLaw.h
@@ -40,27 +40,26 @@ class BaseFrictionLaw : public FrictionSolver {
       SCOREP_USER_REGION_BEGIN(
           myRegionHandle, "computeDynamicRupturePrecomputeStress", SCOREP_USER_REGION_TYPE_COMMON)
       LIKWID_MARKER_START("computeDynamicRupturePrecomputeStress");
-      /// TODO: convert from strain to stress
+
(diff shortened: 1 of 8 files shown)


src/DynamicRupture/FrictionLaws/BaseFrictionLaw.h
```diff
@@ -40,28 +40,27 @@ class BaseFrictionLaw : public FrictionSolver {
...
               {{0.0}};
+#ifdef USE_DAMAGEDELASTIC
+      // TODO: convert from strain to stress
 
       using QInterpolatedShapeT =
           const real(*)[seissol::dr::misc::numQuantities][seissol::dr::misc::numPaddedPoints];
       using QStressInterpolatedShapeT =
           real(*)[seissol::dr::misc::numQuantities][seissol::dr::misc::numPaddedPoints];
+
+      using namespace seissol::dr::misc::quantity_indices;
```
vikaskurapati, line 58: Can we do it without it or is it called in too many places? Not a fan of using namespaces like this as it gets confusing when we are using other namespace calls in the same file.

vikaskurapati (approved):

src/DynamicRupture/FrictionLaws/BaseFrictionLaw.h
sebwolf-de: I agree with you in general, but I guess here it's just convenient to write `array[Q]` instead of `array[seissol::dr::misc::quantity_indices::Q]`...

PR #1059, opened 2024-02-23
Allow integration of quanitities

sebwolf-de:
We had a differentiate button for some time, now I've added the integration button.

### DIFF
--- a/postprocessing/visualization/receiver/src/Waveform.py
+++ b/postprocessing/visualization/receiver/src/Waveform.py
@@ -72,6 +72,10 @@ class Waveform:
         self.waveforms[name] = wf / self.norm[name]
 
   def differentiate(self):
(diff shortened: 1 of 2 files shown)


postprocessing/visualization/receiver/src/Waveform.py
```diff
@@ -72,6 +72,10 @@ def normalize(self):
         self.waveforms[name] = wf / self.norm[name]
 
   def differentiate(self):
+    for name, wf in self.waveforms.items():
+      self.waveforms[name] = numpy.gradient(wf, self.time)
```
davschneller, line 76: Quick question: is the `self.time` here correct? (before, it was `dt` as below—the `self.time` would make more sense to me; but I'm asking simply to make sure)

PR #1074, opened 2024-03-04
Add a Data Collector Class

davschneller:
Unifies the output for the receivers in the DR and the wave propagation part (since: we have a (usually sparse) gather operation of some device memory in both cases).

Plus, less stream/queue syncs for multiple receivers.

### DIFF
--- a/src/DynamicRupture/Output/Builders/ReceiverBasedOutputBuilder.cpp
+++ b/src/DynamicRupture/Output/Builders/ReceiverBasedOutputBuilder.cpp
@@ -108,21 +108,21 @@ void ReceiverBasedOutputBuilder::initBasisFunctions() {
   outputData->cellCount = elementIndices.size() + elementIndicesGhost.size();
 
 #ifdef ACL_DEVICE
-  std::vector<real*> preDofPtr(outputData->cellCount);
+  std::vector<real*> indexPtrs(outputData->cellCount);
 
   for (const auto& [index, arrayIndex] : elementIndices) {
-    preDofPtr[arrayIndex] = wpLut->lookup(wpDescr->derivatives, index);
-    assert(preDofPtr[arrayIndex] != nullptr);
+    indexPtrs[arrayIndex] = wpLut->lookup(wpDescr->derivatives, index);
+    assert(indexPtrs[arrayIndex] != nullptr);
   }
   for (const auto& [_, ghost] : elementIndicesGhost) {
     const auto neighbor = ghost.data;
     const auto arrayIndex = ghost.index + elementIndices.size();
-    preDofPtr[arrayIndex] = wpLut->lookup(wpDescr->faceNeighbors, neighbor.first)[neighbor.second];
-    assert(preDofPtr[arrayIndex] != nullptr);
+    indexPtrs[arrayIndex] = wpLut->lookup(wpDescr->faceNeighbors, neighbor.first)[neighbor.second];
+    assert(indexPtrs[arrayIndex] != nullptr);
   }
 
   outputData->deviceDataCollector =
-      std::make_unique<seissol::parallel::DataCollector>(preDofPtr, seissol::tensor::Q::size());
+      std::make_unique<seissol::parallel::DataCollector>(indexPtrs, seissol::tensor::Q::size());
 #endif
 
   outputData->deviceDataPlus.resize(foundPoints);
--- a/src/DynamicRupture/Output/Builders/ReceiverBasedOutputBuilder.cpp
+++ b/src/DynamicRupture/Output/Builders/ReceiverBasedOutputBuilder.cpp
@@ -1,4 +1,6 @@
 #include "DynamicRupture/Output/Builders/ReceiverBasedOutputBuilder.hpp"
+#include <Parallel/DataCollector.h>
+#include <memory>
 #include <unordered_map>
 
 namespace seissol::dr::output {
@@ -106,12 +108,7 @@ void ReceiverBasedOutputBuilder::initBasisFunctions() {
   outputData->cellCount = elementIndices.size() + elementIndicesGhost.size();
 
 #ifdef ACL_DEVICE
-  real** preDofPtr =
-      reinterpret_cast<real**>(device::DeviceInstance::getInstance().api->allocPinnedMem(
-          sizeof(real*) * outputData->cellCount));
-  outputData->deviceDataPtr =
-      reinterpret_cast<real**>(device::DeviceInstance::getInstance().api->allocGlobMem(
-          sizeof(real*) * outputData->cellCount));
+  std::vector<real*> preDofPtr(outputData->cellCount);
 
   for (const auto& [index, arrayIndex] : elementIndices) {
     preDofPtr[arrayIndex] = wpLut->lookup(wpDescr->derivatives, index);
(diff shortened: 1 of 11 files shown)


src/Parallel/DataCollector.h
```diff
@@ -0,0 +1,86 @@
...
+
+#ifdef ACL_DEVICE
+#include "device.h"
+#endif
+
+namespace seissol::parallel {
+
+class DataCollector {
+  public:
+  DataCollector(const std::vector<real*>& indexDataHost,
```
Thomas-Ulrich, line 17: don't you mean indexedDataHost?

src/Kernels/Receiver.cpp
```diff
@@ -37,18 +37,19 @@
...
 #include "Numerical_aux/BasisFunction.h"
-
+#include "Receiver.h"
 #include <Initializer/PointMapper.h>
+#include <Monitoring/FlopCounter.hpp>
 #include <Numerical_aux/Transformation.h>
+#include <Parallel/DataCollector.h>
 #include <Parallel/MPI.h>
-#include <Monitoring/FlopCounter.hpp>
+#include <SeisSol.h>
```
Thomas-Ulrich, line 47: You did not modify, but it shouldn't it be
#include "SeisSol.h"
(and others)

src/Kernels/Receiver.cpp
```diff
@@ -82,8 +83,20 @@ double seissol::kernels::ReceiverCluster::calcReceivers(  double time,
   alignas(ALIGNMENT) real timeEvaluatedAtPoint[tensor::QAtPoint::size()];
   alignas(ALIGNMENT) real timeEvaluatedDerivativesAtPoint[tensor::QDerivativeAtPoint::size()];
 #ifdef ACL_DEVICE
-  constexpr size_t dofsHostCopySize = tensor::Q::size();
-  alignas(ALIGNMENT) real dofsHostCopy[ dofsHostCopySize ];
+  if (deviceCollector.get() == nullptr) {
```
Thomas-Ulrich, line 86: what does it mean if there is no index in get?

src/DynamicRupture/Output/Builders/ReceiverBasedOutputBuilder.cpp
```diff
@@ -106,12 +108,7 @@ void ReceiverBasedOutputBuilder::initBasisFunctions() {
   outputData->cellCount = elementIndices.size() + elementIndicesGhost.size();
 
 #ifdef ACL_DEVICE
-  real** preDofPtr =
-      reinterpret_cast<real**>(device::DeviceInstance::getInstance().api->allocPinnedMem(
-          sizeof(real*) * outputData->cellCount));
-  outputData->deviceDataPtr =
-      reinterpret_cast<real**>(device::DeviceInstance::getInstance().api->allocGlobMem(
-          sizeof(real*) * outputData->cellCount));
+  std::vector<real*> preDofPtr(outputData->cellCount);
```
Thomas-Ulrich, line 111: any idea what means pre in preDofPtr?

src/Kernels/Receiver.cpp
```diff
@@ -82,8 +83,20 @@ double seissol::kernels::ReceiverCluster::calcReceivers(  double time,
...
+    deviceIndices.resize(m_receivers.size());
+    std::vector<real*> dofs;
+    std::unordered_map<real*, size_t> indexMap;
```
Thomas-Ulrich, line 89: I don't understand what is indexMap.
I mean:
`indexMap[currentDofs] = dofs.size();`
does not really suggest `indexMap[currentDofs]` points to an index but to a size?

src/Parallel/DataCollector.h
```diff
@@ -0,0 +1,86 @@
...
+
+#include "Kernels/precision.hpp"
+
+#ifdef ACL_DEVICE
+#include "device.h"
+#endif
+
+namespace seissol::parallel {
+
+class DataCollector {
```
Thomas-Ulrich, line 15: Proposed comment (ChatGPT)
// Class for collecting and managing data, facilitating data transfers between host and device

src/Parallel/DataCollector.h
```diff
@@ -0,0 +1,86 @@
...
+
+  void gather(void* stream) {
+    if (!hostAccessible && indexCount > 0) {
+#ifdef ACL_DEVICE
+      device::DeviceInstance::getInstance().algorithms.copyScatterToUniform(
+          indexDataDevice, copiedData, elemSize, elemSize, indexCount, stream);
+#endif
+    }
+  }
+
```
Thomas-Ulrich, line 57: Here ChatGPT would comment:
// Copies uniform data to a scattered region on the device
(is that correct?)

src/Parallel/DataCollector.h
```diff
@@ -0,0 +1,86 @@
...
+  ~DataCollector() {
+    if (!hostAccessible && indexCount > 0) {
+#ifdef ACL_DEVICE
+      device::DeviceInstance::getInstance().api->freeMem(indexDataDevice);
+      device::DeviceInstance::getInstance().api->freePinnedMem(copiedData);
+#endif
+    }
+  }
+
+  void gather(void* stream) {
```
Thomas-Ulrich, line 49: // Copies scattered data to a uniform region on the device 
  ?

src/Parallel/DataCollector.h
```diff
@@ -0,0 +1,86 @@
...
+
+class DataCollector {
+  public:
+  DataCollector(const std::vector<real*>& indexDataHost,
+                size_t elemSize,
+                bool hostAccessible = false)
+      : indexCount(indexDataHost.size()), indexDataHost(indexDataHost), elemSize(elemSize),
+        hostAccessible(hostAccessible) {
+#ifndef ACL_DEVICE
+    hostAccessible = true;
```
Thomas-Ulrich, line 23: should never happen, right?

Thomas-Ulrich (commented):
uninformed review, hopefully can make the code clearer.

src/Kernels/Receiver.cpp
davschneller: We add an element directly after that—i.e. `currentDofs` is then located at `indexMap[currentDofs]`.

src/Parallel/DataCollector.h
davschneller: If we were to use this class in a non-GPU environment, then yes

src/Kernels/Receiver.cpp
```diff
@@ -216,3 +195,26 @@ double seissol::kernels::ReceiverCluster::calcReceivers(  double time,
...
+  // collect all data pointers to transfer. If we have multiple receivers on the same cell, we make sure to only transfer the related data once (hence, we use the `indexMap` here)
+  deviceIndices.resize(m_receivers.size());
+  std::vector<real*> dofs;
+  std::unordered_map<real*, size_t> indexMap;
+  for (size_t i = 0; i < m_receivers.size(); ++i) {
+    real* currentDofs = m_receivers[i].data.dofs();
+    if (indexMap.find(currentDofs) == indexMap.end()) {
+      // point to the current array end
+      indexMap[currentDofs] = dofs.size();
+      dofs.push_back(currentDofs);
```
Thomas-Ulrich, line 209: I see. Wouldn't it more readable with 
```
dofs.push_back(currentDofs);
indexMap[currentDofs] = dofs.back();
```
?

davschneller: Unfortunately, `dofs.back()` is then equal to `currentDofs`—we really want `indexMap[currentDofs]` to point to the index inside the `dofs` vector

src/Parallel/DataCollector.h
davschneller: It should mean "index data, acessible on the host."

Thomas-Ulrich (approved):

PR #1075, opened 2024-03-08
Reduce double-built files

davschneller:
Right now, the SeisSol-lib and the SYCL-dependent files are sometimes the same, i.e. we compile them twice. This PR makes them only be compiled once.

Also, we add options to enable/disable building the SeisSol executable and the SeisSol proxy—and to make the main SeisSol library a shared object instead (experimentally).


CMakeLists.txt
```diff
@@ -41,9 +41,9 @@ endif()
 # 'process_users_input' returns the following:
 #
 #       switches: HDF5, NETCDF, GRAPH_PARTITIONING_LIBS, MPI, OPENMP, ASAGI, MEMKIND,
-#                 PROXY_PYBINDING, ENABLE_PIC_COMPILATION, PREMULTIPLY_FLUX
+#                 PROXY_PYBINDING, ENABLE_PIC_COMPILATION, PREMULTIPLY_FLUX, BUILD_PROXY, BUILD_EXECUTABLE, SHARED
```
sebwolf-de, line 44: `SHARED` is a good idea, but do we really need the other options. Right now, you can already choose not to build SeisSol/the proxy, by using
`make SeisSol-lib`
`make SeisSol-bin`
`make SeisSol-proxy`

davschneller: Good point—but when we do `make install` we inadvertedly install everything (except the tests which can still be enabled/disabled)—even if we e.g. only want to have `seissol-bin` in the end. If that's ever a concern.

sebwolf-de: I don't think this is a concern. If you are a power user, you'll know what to do. If you are a not-so-power user, you won't care for the proxy.

davschneller: I think you're right in the current moment—but it won't work once e.g. we'll consider that we won't need e.g. easi (or Hdf5 etc.) for the proxy at all. Then, having the additional build flags makes it possible that CMake won't look for that package then, and make the proxy a really slim build.

davschneller: I've removed the two flags for now still—making the proxy slimmer is something probably for a future PR

PR #1082, opened 2024-03-28
Abort criteria based on seismic moment rate and abort properly simulation

Thomas-Ulrich:
The Terminator based on slip rate is not always working. e.g. in case of non-vanishing high creeping slip rate around a failed nucleation.  Therefore, I propose to add a criterion based on the moment rate release.

Following David's suggestions, I also improved the way the simulation is aborted (no more truncated last output).


davschneller (approved):
LGTM; two comments independent of this PR:

* it would probably be cleaner to have extra (boolean) flags for enabling/disabling the `slipRate`/`moment` thresholds. But that's maybe partially at least also a constraint of the current output format.
* Instead of running `mpi.finalize()` immediately, one could go ahead and add an "abort" function to the `Simulator` class. That way, the application can be cleaned up properly—and maybe any final IO can be addressed as well. (if that's needed)

src/Solver/Simulator.cpp
```diff
@@ -128,6 +134,9 @@ void seissol::Simulator::simulate(seissol::SeisSol& seissolInstance) {
     if (upcomingTime < m_currentTime + l_timeTolerance) {
       logError() << "Simulator did not advance in time from" << m_currentTime << "to" << upcomingTime;
     }
+    if (m_abort) {
```
davschneller, line 137: You could also add a message here like `logInfo(seissol::MPI::mpi.rank()) << "Aborting simulation.";` or the likes.

PR #1084, opened 2024-03-28
Refactor supermuc documentation page

Thomas-Ulrich:
In this PR, I refactor the supermuc documentation page, now detailing how to install the seissol spack package.
I nevertheless keep information about how to compile seissol-env.

On NG, there are two user_spack packages. 
One with spack 0.17.1 (extremly old) and another with spack 0.22.0.dev0.

seissol-env was installed with  0.17.1.
It is tricky to install the new seissol package based on this version (too modern).
On the other hand, I have trouble getting workable modules with spack 0.22.0.dev0.
E.g. module of dependencies do not get created (easi), even if I manually try to create them (it works on heisenbug).
Anyway, this is not such a big issue (e.g. the easi python module I was needing can be using by manually adding the path to the library to PYTHONPATH).
That's why I still detail the seissol-env installation, with spack 0.17.1.

### DIFF
--- a/Documentation/supermuc.rst
+++ b/Documentation/supermuc.rst
@@ -218,8 +218,7 @@ To compile the seissol spack package on SuperMUC-NG, follow the procedure below.
 
     # install a few python modules
     python3.10 -m pip install --upgrade pip --user --proxy http://localhost:DDDDD
-    pip3.10 install --upgrade setuptools --user --proxy http://localhost:DDDDD
-    pip3.10 install --upgrade numpy --user --proxy http://localhost:DDDDD
+    pip3.10 install --upgrade setuptools numpy wheel packaging --user --proxy http://localhost:DDDDD
     pip3.10 install git+https://github.com/SeisSol/PSpaMM.git --no-build-isolation --user --proxy http://localhost:DDDDD
 
     # clone seissol-spack-aid and add the repository
--- a/Documentation/supermuc.rst
+++ b/Documentation/supermuc.rst
@@ -27,14 +27,14 @@ Load modules (including seissol-env). Add these lines to .bashrc:
   module load seissol-env/develop-intel21-impi-x2b
   export CC=mpicc 
   export CXX=mpiCC 
-  export FC=mpif90
-
 
 Install pspamm (see :ref:`pypi_behind_firewall` for the proxy):
 
 ::
 
     pip3 install git+https://github.com/SeisSol/PSpaMM.git --no-build-isolation --user --proxy http://localhost:DDDDD
+
+(`--no-build-isolation` is used to circumvent the problem described in https://github.com/SeisSol/PSpaMM/issues/13)
 
 Clone SeisSol including the submodules using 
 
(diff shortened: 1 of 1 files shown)


Documentation/supermuc.rst
```diff
@@ -36,100 +29,36 @@ Load module. Add these lines to .bashrc:
...
-
-.. code-block:: yaml
-
-    modules:
-      default:
-        roots:
-         tcl: your_custom_path_2_modules
 
-This can be useful to share packages with other user of a SuperMUC project.
+Install pspamm (see :ref:`pypi_behind_firewall` for the proxy):
```
davschneller, line 33: Is it maybe useful to also insert a comment about `--no-build-isolation` into that article?

davschneller: (we could also recommend/suggest using a virtual environment for the Python packages, if it helps—but maybe that's a bit out of scope for this topic here)

Documentation/supermuc.rst
```diff
@@ -196,3 +125,170 @@ Many post-processing scripts of SeisSol require Python dependencies.
 We describe how to use pip on SuperMUC at see :ref:`pypi_behind_firewall`.
 
 
+Using Sanitizer
```
davschneller, line 128: `Using the Sanitizer` ? (the article would seem more natural to me here)

Documentation/supermuc.rst
```diff
@@ -196,3 +125,170 @@ Many post-processing scripts of SeisSol require Python dependencies.
...
+Note that to use sanitizer (https://en.wikipedia.org/wiki/AddressSanitizer), SeisSol needs to be compiled with gcc.
+For that modules and compiler need to be switched:
+
+.. code-block:: bash
+
+    module switch seissol-env seissol-env/develop-gcc11
+
+    export CC=mpigcc
+    export CXX=mpigxx
+    export FC=mpifc
```
davschneller, line 140: `FC` is not needed anymore—also `gcc` and `gxx` should be enough actually—CMake will do the rest automatically.

Documentation/supermuc.rst
```diff
@@ -196,3 +125,170 @@ Many post-processing scripts of SeisSol require Python dependencies.
 We describe how to use pip on SuperMUC at see :ref:`pypi_behind_firewall`.
 
 
+Using Sanitizer
+---------------
+
+Note that to use sanitizer (https://en.wikipedia.org/wiki/AddressSanitizer), SeisSol needs to be compiled with gcc.
```
davschneller, line 131: It's also possible to use clang, iirc; but one of the arguments doesn't work then (the `static-libasan` iirc)

Documentation/supermuc.rst
```diff
@@ -196,3 +125,170 @@ Many post-processing scripts of SeisSol require Python dependencies.
...
+        - spec: python@3.8.11
+          buildable: False
+          modules:
+           - python/3.8.11-extended
+
+
+Compiling seissol spack package
+-------------------------------
+ 
+The seissol package is similar with the seissol-env package (it gathers all dependencies of seissol), but also compiles a specific version of seissol.
```
davschneller, line 209: `similar to`

davschneller: Also: `but also compiles a specific version of seissol itself` ... or something to highlight that SeisSol is included here

Thomas-Ulrich:
@davschneller: Thanks for the review!

PR #1085, opened 2024-03-28
add velocity-aware meshing docs

NicoSchlw:


NicoSchlw:
Thanks for approving, I added the reference

PR #1086, opened 2024-04-02
Add some constexpr values to the material declarations

davschneller:
A refactoring PR.

We just add values, like e.g. the number of quantities to the material declaration (or quantity names)—so that we can mid-term replace all the external preprocessor definitions, yielding a (hopefully) cleaner and more equation-independent code.

Also, we namespace some structs and make some tensors precision-independent.


.ci/format.sh
```diff
@@ -30,6 +30,10 @@ format() {
     
     # NOTE: once the files of a directory are (almost) fully covered, consider moving it to allowlist_dir instead
     local allowlist_file="
+        src/Equations/elastic/Model/datastructures.hpp
+        src/Equations/viscoelastic2/Model/datastructures.hpp
```
sebwolf-de, line 34: What about viscoelastic(1)?

generated_code/general.py
```diff
@@ -0,0 +1,14 @@
...
+    stiffnessTensor = Tensor('stiffnessTensor', (3, 3, 3, 3))
+    direction = Tensor('direction', (3,))
+    christoffel = Tensor('christoffel', (3,3))
+
+    computeChristoffel = christoffel['ik'] <= stiffnessTensor['ijkl'] * direction['j'] * direction['l']
+    generator.add('computeChristoffel', computeChristoffel)
+
+def includeMatrices(matricesDir):
+    matrices = parseJSONMatrixFile('{}/sampling_directions.json'.format(matricesDir))
+    return set([matrices.samplingDirections])
```
sebwolf-de, line 14: This is anisotropy related, do we need it for all equations now?!

generated_code/generate.py
```diff
@@ -176,11 +180,50 @@
   except:
     print('WARNING: ChainForge was not found. Falling back to GemmForge.')
 
+trueOutputDir = os.path.join(cmdLineArgs.outputDir, 'equation')
```
sebwolf-de, line 183: Why don't we use `cmdLineArgs.equation` here? This would pave the way for having more than one equation in the same binary?

generated_code/generate.py
```diff
@@ -176,11 +180,50 @@
...
 gemmTools = GeneratorCollection(gemm_generators)
-generator.generate(outputDir=cmdLineArgs.outputDir,
+generator.generate(outputDir=trueOutputDir,
                    namespace='seissol',
                    gemm_cfg=gemmTools,
                    cost_estimator=cost_estimators,
                    include_tensors=include_tensors)
+
+def generate_general(subfolders):
+  arch = useArchitectureIdentifiedBy('d' + cmdLineArgs.host_arch[1:])
```
sebwolf-de, line 198: What about single precision here?

src/Equations/viscoelastic2/Kernels/Local.cpp
```diff
@@ -84,7 +86,7 @@ void seissol::kernels::Local::computeIntegral(real i_timeIntegratedDegreesOfFree
   assert( ((uintptr_t)data.dofs())           % ALIGNMENT == 0 );
 #endif
 
-  real Qext[tensor::Qext::size()] __attribute__((aligned(ALIGNMENT)));
+  real Qext[tensor::Qext::size()] __attribute__((aligned(Alignment)));
```
sebwolf-de, line 89: Change to `alignas(Alignment)`?

src/Equations/viscoelastic2/Kernels/Neighbor.cpp
```diff
@@ -95,7 +97,7 @@ void seissol::kernels::Neighbor::computeNeighborsIntegral(  NeighborData&
   // alignment of the degrees of freedom
   assert( ((uintptr_t)data.dofs()) % ALIGNMENT == 0 );
 
-  real Qext[tensor::Qext::size()] __attribute__((aligned(PAGESIZE_STACK))) = {};
+  real Qext[tensor::Qext::size()] __attribute__((aligned(PagesizeStack))) = {};
```
sebwolf-de, line 100: Change to `alignas(PagesizeStack)`?

src/Equations/viscoelastic2/Kernels/Time.cpp
```diff
@@ -85,9 +87,9 @@ void seissol::kernels::Time::computeAder(double i_timeStepWidth,
    * compute ADER scheme.
    */
   // temporary result
-  real temporaryBuffer[2][tensor::dQ::size(0)] __attribute__((aligned(PAGESIZE_STACK)));
-  real temporaryBufferExt[2][tensor::dQext::size(1)] __attribute__((aligned(PAGESIZE_STACK)));
-  real temporaryBufferAne[2][tensor::dQane::size(0)] __attribute__((aligned(PAGESIZE_STACK)));
+  real temporaryBuffer[2][tensor::dQ::size(0)] __attribute__((aligned(PagesizeStack)));
+  real temporaryBufferExt[2][tensor::dQext::size(1)] __attribute__((aligned(PagesizeStack)));
+  real temporaryBufferAne[2][tensor::dQane::size(0)] __attribute__((aligned(PagesizeStack)));
```
sebwolf-de, line 92: Change to `alignas(PagesizeStack)`?

src/Initializer/InitialFieldProjection.cpp
```diff
@@ -83,7 +83,7 @@ void seissol::initializer::projectInitialField(std::vector<std::unique_ptr<physi
   #pragma omp parallel
   {
 #endif
-  real iniCondData[tensor::iniCond::size()] __attribute__((aligned(ALIGNMENT))) = {};
+  real iniCondData[tensor::iniCond::size()] __attribute__((aligned(Alignment))) = {};
```
sebwolf-de, line 86: Change to `alignas(Alignment)`?

src/Initializer/typedefs.hpp
```diff
@@ -460,16 +462,14 @@ struct BoundaryFaceInformation {
  * See, an example of a policy-based design in GlobalData.cpp
  * */
 struct MemoryProperties {
-  size_t alignment{ALIGNMENT};
-  size_t pagesizeHeap{PAGESIZE_HEAP};
-  size_t pagesizeStack{PAGESIZE_STACK};
+  size_t alignment{Alignment};
+  size_t pagesizeHeap{PagesizeHeap};
+  size_t pagesizeStack{PagesizeStack};
 };
```
sebwolf-de, line 468: Why do we need this struct? If all of these quantities are `constexpr`, we can always refer to them directly, insted of using this struct, which could be changed.

src/Kernels/Plasticity.cpp
```diff
@@ -59,22 +59,22 @@ namespace seissol::kernels {
...
-    real dudt_pstrain[tensor::QStress::size()] __attribute__((aligned(ALIGNMENT)));
+    real QStressNodal[tensor::QStressNodal::size()] __attribute__((aligned(Alignment)));
+    real QEtaNodal[tensor::QEtaNodal::size()] __attribute__((aligned(Alignment)));
+    real QEtaModal[tensor::QEtaModal::size()] __attribute__((aligned(Alignment)));
+    real meanStress[tensor::meanStress::size()] __attribute__((aligned(Alignment)));
+    real secondInvariant[tensor::secondInvariant::size()] __attribute__((aligned(Alignment)));
+    real tau[tensor::secondInvariant::size()] __attribute__((aligned(Alignment)));
+    real taulim[tensor::meanStress::size()] __attribute__((aligned(Alignment)));
+    real yieldFactor[tensor::yieldFactor::size()] __attribute__((aligned(Alignment)));
+    real dudt_pstrain[tensor::QStress::size()] __attribute__((aligned(Alignment)));
```
sebwolf-de, line 77: Change to `alignas(Alignment)`?

src/Model/plasticity.hpp
```diff
@@ -0,0 +1,40 @@
+#pragma once
```
sebwolf-de, line 1: How's our opinion on the non-standard `#pragma once`?

src/Solver/FreeSurfaceIntegrator.cpp
```diff
@@ -117,7 +117,7 @@ void seissol::solver::FreeSurfaceIntegrator::calculateOutput()
     #pragma omp parallel for schedule(static) default(none) shared(offset, surfaceLayer, dofs, displacementDofs, side)
 #endif // _OPENMP
     for (unsigned face = 0; face < surfaceLayer->getNumberOfCells(); ++face) {
-      real subTriangleDofs[tensor::subTriangleDofs::size(FREESURFACE_MAX_REFINEMENT)] __attribute__((aligned(ALIGNMENT)));
+      real subTriangleDofs[tensor::subTriangleDofs::size(FREESURFACE_MAX_REFINEMENT)] __attribute__((aligned(Alignment)));
```
sebwolf-de, line 120: Change to `alignas(Alignment)`?

src/Solver/time_stepping/TimeCluster.cpp
```diff
@@ -335,7 +335,7 @@ void seissol::time_stepping::TimeCluster::computeLocalIntegration(seissol::initi
   m_loopStatistics->begin(m_regionComputeLocalIntegration);
 
   // local integration buffer
-  real l_integrationBuffer[tensor::I::size()] __attribute__((aligned(ALIGNMENT)));
+  real l_integrationBuffer[tensor::I::size()] __attribute__((aligned(Alignment)));
```
sebwolf-de, line 338: Change to `alignas(Alignment)`?

sebwolf-de (commented):
LGTM, most comments are about `__attribute__((aligned(X)))`

.ci/format.sh
davschneller: Do we still support `viscoelastic` (1)? Otherwise, yes; I can change it.

generated_code/generate.py
davschneller: We could definitely do that. Originally, the equations were also templated, by https://github.com/SeisSol/yateto/pull/62 (cf. what was done in #932; that's where most of this PR is from). It just doesn't completely work with the subroutine.cpp file yet.

generated_code/generate.py
davschneller: That part was more meant to be precision-agnostic—i.e. the stiffness tensors etc.

(originally, it came from #932, to try to support multi-precision as well... Then forcing `double` was a good way to avoid templating in these classes as well)

generated_code/general.py
davschneller: It probably won't hurt _too_ much to have it in... But I get it—I'll disable it if we don't have anisotropy enabled.

src/Model/plasticity.hpp
davschneller: I'll remove it. (I mainly use it for convenience first—and forgot to replace it here... ;)  ... Maybe we'll need some, say, "create-me-an-empty-header-file" script sometime?)

.ci/format.sh
davschneller: Wait—we don't have a `Model/datastructures.hpp` for viscoelastic (1). :)

sebwolf-de: Yes, viscoelastic(1) is a prototype for a non-stiff pde with source term.

generated_code/generate.py
davschneller: It's changed now to `equation-{equation}-{order}-{precision}`. The `order` part may be slightly overkill (as the matrices are usually pretty similar). But the `equation` partially and the `precision` surely would be necessary, I'd guess

src/Initializer/typedefs.hpp
davschneller: They are used in the `GlobalData` allocation only—the values are set according to us being on host or device. We may be good with either keeping everything as is—or we could be moving the `struct` just there to `src/Initializer/GlobalData.h`.

.ci/format.sh
sebwolf-de: We don't have a datastructure.hpp for viscoelastic(1), we just use the same as for viscoelastic2.

generated_code/generate.py
sebwolf-de: Makes sense. Can you add a comment that you use `double` here, since this kernel is not used at runtime?

src/Initializer/typedefs.hpp
sebwolf-de: Then, I'd move the struct.

generated_code/generate.py
davschneller: Done. In the same vein, couldn't we probably keep the anisotropy-related kernels; especially when going towards multi-material anyways?

src/Initializer/typedefs.hpp
davschneller: Done.

src/Kernels/DynamicRupture.cpp
```diff
@@ -86,13 +86,13 @@ void seissol::kernels::DynamicRupture::setTimeStepWidth(double timestep)
 {
 #ifdef USE_DR_CELLAVERAGE
   static_assert(false, "Cell average currently not supported");
-  /*double subIntervalWidth = timestep / CONVERGENCE_ORDER;
```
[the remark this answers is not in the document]
davschneller: I'd guess it's more or less an unfinished idea. If noone wants to continue it, I can remove it.

Issue #546, opened 2022-05-10
tpv5 slip artifacts 


NicoSchlw:
**Describe the bug**
Tpv5 produces elements with unreasonable high absolute slip.

**Expected behavior**
A reasonable and smooth absolute slip distribution. 

**To Reproduce**
1. Which version do you use? d5656cd; but had the same problem also with 9895309
2. Which build settings do you use? Which compiler version do you use? Double precision, order 4
3. On which machine does your problem occur? If on a cluster: Which modules are loaded? I'm running it on Supermuc and compiled SeisSol following the documentation: https://seissol.readthedocs.io/en/latest/supermuc.html
![Screenshot from 2022-05-10 12-13-57](https://user-images.githubusercontent.com/42612980/167606151-0f0b1805-4081-4d9f-8c6d-4c35d41a5195.png)

4. Provide parameter/material files.

[parameters.txt](https://github.com/SeisSol/SeisSol/files/8660675/parameters.txt)

[2089910.tpv5.out.txt](https://github.com/SeisSol/SeisSol/files/8660677/2089910.tpv5.out.txt)

**Screenshots/Console output**
![Screenshot from 2022-05-09 20-05-08](https://user-images.githubusercontent.com/42612980/167604777-8671026b-fefd-4137-a613-671781ab162c.png)

If you encounter any errors/warnings/... during execution please provide the console output.
I got no errors.

**Additional context**
I used this setup: https://github.com/SeisSol/Examples/tree/master/tpv5
and increased the domain size and resolution (150m on-fault element size; order 4).

Thomas-Ulrich:
Hi,
I can confirm that it is not fixed by the new fault output. 
And it is not fixed by the new quadrature rule (see a comparison of P_n, top Dunavant, bottom strout.
![image](https://user-images.githubusercontent.com/13536910/171603329-6a00e074-e09e-42ac-b167-6020a14b2294.png)

The only know way to mitigate the problem are:
- use a symmetrical mesh (no more problem)
- modify an impedance, as in https://github.com/SeisSol/SeisSol/commit/181fc85d5c405a8c44fe21869fe736ab1f0206d5
(but this comes with a loss of accuracy).

sebwolf-de:
[...]

PR #1087, opened 2024-04-02
Add the "eta hack" as a parameter

davschneller:
Adds the patch from #546 as a parameter. Also, we emit a warning if it's used.


Thomas-Ulrich (approved):
Thank you!

Issue #1083, opened 2024-03-28
Seissol infers the wrong boundary format


Thomas-Ulrich:
**Describe the bug**
Seissol infers the wrong boundary format for the following mesh with 
[mesh.zip](https://github.com/SeisSol/SeisSol/files/14787638/mesh.zip)
(pumlboundaryformat = 'auto')
```
Thu Mar 28 11:07:31, Info:  Inferring boundary format. 
Thu Mar 28 11:07:31, Info:  Using boundary format: i64 (4xi16) 
```
(should be i32 and it works fine if I set i32)

**To Reproduce**
1. Which version do you use? Provide branch and commit id.
master 5c163ed1f6931a3ac1e5e2ff5ca63dd97bce6c0f
2. Which build settings do you use? Which compiler version do you use? 
dhw4 elastic
3. On which machine does your problem occur? If on a cluster: Which modules are loaded?
cachemiss
```
ulrich@cachemiss:/export/dump/ulrich/tpv5$ module list
Currently Loaded Modulefiles:
 1) paraview/5.11.1                       12) zstd/1.5.5-gcc-12.2.0-d3iesv2            23) pkgconf/1.9.5-gcc-12.2.0-pdjoat6       34) python/3.9.2-gcc-12.2.0-ikw34ao                         
 2) libpciaccess/0.17-gcc-12.2.0-foegz6x  13) tar/1.34-gcc-12.2.0-344crw2              24) hdf5/1.12.2-gcc-12.2.0-xwxzdqa         35) yaml-cpp/0.6.2-gcc-12.2.0-adem7ot                       
 3) libiconv/1.17-gcc-12.2.0-pztn24l      14) gettext/0.21.1-gcc-12.2.0-htnecfu        25) netcdf-c/4.7.4-gcc-12.2.0-5wzc2ic      36) easi/1.3.0-gcc-12.2.0-3egmupu                           
 4) xz/5.4.1-gcc-12.2.0-beahnap           15) openssl/3.1.1-gcc-12.2.0-xydqrfk         26) asagi/1.0.1-gcc-12.2.0-xvz2l4l         37) eigen/3.4.0-gcc-12.2.0-kaoej2y                          
[...]

PR #1088, opened 2024-04-02
Change boundary format inference rules

davschneller:
Fix/address #1083.

We add an option to check for an attribute in a PUML file which gives us the boundary type. Otherwise, we avoid `i64`, unless specified explicitly.

### DIFF
--- a/Documentation/fault-tagging.rst
+++ b/Documentation/fault-tagging.rst
@@ -70,10 +70,10 @@ However, to prevent mistakes with reading the format, we nevertheless recommend 
 
 To do that, it suffices to specify ``pumlboundaryformat = $option`` in the ``&meshnml`` section of your SeisSol parameter file, where ``$option`` is one of the following:
 
-- ``'auto'``: SeisSol will try to infer the boundary format automatically. This is the default option.
+- ``'auto'``: SeisSol will try to infer the boundary format automatically. This is the default option. That is, if an attribute ``boundary-format`` is present, its value will be used (``i32x4 == 0``, ``i32 == 1``, ``i64 == 2``). Otherwise, the type will be inferred from the rank of the boundary storage array: if it is two-dimensional, then we choose ``i32x4``; if it is rank 1, then we choose ``i32`` (note that ``i64`` will not be chosen automatically in this case).
 - ``'i32'``: 8 bits per boundary face. That is, 189 dynamic rupture tags are possible (255 boundary condition types). It is (usually) stored as a one-dimensional 32-bit integer array (one entry per cell) in the Hdf5/binary file.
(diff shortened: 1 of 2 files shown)


src/Initializer/InitProcedure/InitMesh.cpp
```diff
@@ -107,25 +107,44 @@ static void readMeshPUML(const seissol::initializer::parameters::SeisSolParamete
...
-    hid_t boundaryDataset = _eh(H5Dopen2(dataFile, "boundary", H5P_DEFAULT));
-    hid_t boundaryType = _eh(H5Dget_type(boundaryDataset));
-    hid_t boundarySpace = _eh(H5Dget_space(boundaryDataset));
-
-    auto boundaryTypeSize = _eh(H5Tget_size(boundaryType));
-    auto boundaryTypeRank = _eh(H5Sget_simple_extent_ndims(boundarySpace));
+    hid_t existenceTest = _eh(H5Lexists(dataFile, "boundary-format", H5P_DEFAULT));
+    if (existenceTest > 0) {
+      logInfo(rank) << "Boundary format given in PUML file.";
+      hid_t boundaryAttribute = _eh(H5Aopen(dataFile, "boundary-format", H5P_DEFAULT));
```
Thomas-Ulrich, line 113: wouldn't be more readable to store boundary-format as a string with I32, I64 and I32x4? 
Else
Wouldn't it make more sense to have format 0= the smallest format and 2 the largest?

Documentation/fault-tagging.rst
```diff
@@ -70,10 +70,10 @@ However, to prevent mistakes with reading the format, we nevertheless recommend
...
 - ``'i32'``: 8 bits per boundary face. That is, 189 dynamic rupture tags are possible (255 boundary condition types). It is (usually) stored as a one-dimensional 32-bit integer array (one entry per cell) in the Hdf5/binary file.
 - ``'i64'``: 16 bits per boundary face. That is, 65469 dynamic rupture tags (65535 boundary condition types). It is stored as a one-dimensional 64-bit integer array (one entry per cell) in the Hdf5/binary file.
 - ``'i32x4'``: 32 bits per boundary face. In short, you will have :math:`2^{32} - 65` different dynamic rupture tags available. The data is stored as a two-dimensional array (four entries per cell, one for each face) of 32-bit integers.
 
-To see which boundary format you have built your mesh for, you can use ``h5dump -H <yourmeshfile>.puml.h5``,
-and look at the datatype and the shape of the ``boundary`` dataset.
+To approximately see which boundary format you have built your mesh for, you can use ``h5dump -H <yourmeshfile>.puml.h5``,
```
Thomas-Ulrich, line 78: You may try inferring ...

src/Initializer/InitProcedure/InitMesh.cpp
davschneller: Ok, we use strings now. (however, pumgen is not up-to-date for the time being; but the attribute can also be added using h5py pretty quickly)

Thomas-Ulrich (approved):

PR #1096, opened 2024-04-12
TP proxy slip-weakening friction

Thomas-Ulrich:
A slip weakening law approximating Tp.


Documentation/parameters.par
```diff
@@ -54,6 +54,7 @@ etahack = 1                 ! use any value smaller than one to mitigate quasi-d
 OutputPointType = 5         ! Type (0: no output, 3: ascii file, 4: paraview file, 5: 3+4)
 SlipRateOutputType=0        ! 0: (smoother) slip rate output evaluated from the difference between the velocity on both side of the fault
                             ! 1: slip rate output evaluated from the fault tractions and the failure criterion (less smooth but usually more accurate where the rupture front is well developped)
+!TpProxyExponent = -0.33333333 ! exponent (1+d_c) in the TP proxy slip weakening law (default is -1/3)
```
davschneller, line 57: A `!` at front? I.e. a comment here?

Documentation/dynamic-rupture.rst
```diff
@@ -198,6 +198,21 @@ Linear slip weakening can be seen as a special case of rate-and-state friction w
...
+The fault strength is determined by
+
+.. math::
+ 
+  \tau = -C - \min\left(0, \sigma_n\right) \left( \mu_d + \frac{{(\mu_s - \mu_d)}}{{\left(1 + \frac{S}{d_c}\right)^{\frac{1}{3}}}} \right),
+
+
+All variables are the same as defined in previous section for :code:`FL=16`.
+The friction law also support forced rupture time.
+The exponent of (1 + d_c) in the friction law, by default set to -1/3, can be changed by setting :code:`TP_proxy_exponent` to a different value in the main parameter file (namelist :code:`DynamicRupture`).
```
davschneller, line 214: Maybe give a symbol, and also change in the formula above?

davschneller (approved):
LGTM up to some tiny comments and my degree of understanding

Documentation/parameters.par
Thomas-Ulrich: just because the default value is 1/3 (which is not exactly -0.33333333). But probably you are right, I could remove the first !

AliceGabriel: The exponent should be (-1/3) @Thomas-Ulrich

davschneller: In the code, that is the case; however, the parameter file doesn't support displaying fractions.
I.e. as long as `tpproxyexponent` is not given as parameter, we have `1/3`, in the best possible floating point approximation

Thomas-Ulrich: There was a typo in my comment. Yes the exponent is -1/3

PR #1097, opened 2024-04-17
Momentum to volume energy outputs

vikaskurapati:
Added a simple momentum calculation to the output of volume energies. `\int \rho*v` is the formula used

### DIFF
--- a/src/ResultWriter/EnergyOutput.cpp
+++ b/src/ResultWriter/EnergyOutput.cpp
@@ -6,6 +6,7 @@
 #include "Numerical_aux/Quadrature.h"
 #include "Parallel/MPI.h"
 #include "SeisSol.h"
+#include <array>
 
 namespace seissol::writer {
 
@@ -28,6 +29,10 @@ double& EnergiesStorage::plasticMoment() { return energies[7]; }
 double& EnergiesStorage::seismicMoment() { return energies[8]; }
 
 double& EnergiesStorage::potency() { return energies[9]; }
+
+double& EnergiesStorage::totalMomentumX() { return energies[10]; }
+double& EnergiesStorage::totalMomentumY() { return energies[11]; }
+double& EnergiesStorage::totalMomentumZ() { return energies[12]; }
 
 void EnergyOutput::init(
     GlobalData* newGlobal,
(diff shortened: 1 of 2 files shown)


vikaskurapati:
I guess the tests for GPU fail as we don't have momentum values in the precomputed tests. Do we calculate them for the tests? Or do we not print it in the csv?

davschneller:
Right now, the energies are not compared to the reference data (after looking at the artifacts etc.; e.g. https://gitlab.lrz.de/ci_seissol/SeisSol/-/jobs/7963872/artifacts/file/output-tpv%7B5-nuc%7D-double/tpv-energy.csv ... at least visible within the next few hours from posting): the problem seems to be that all three momentum values currently get posted into one line—thus we get a CSV line length of 5 (timestep, key, value0, value1, value2) instead of 3 (timestep, key, value). I.e. it should work if you store each momentum component individually.

vikaskurapati:
> Right now, the energies are not compared to the reference data (after looking at the artifacts etc.; e.g. https://gitlab.lrz.de/ci_seissol/SeisSol/-/jobs/7963872/artifacts/file/output-tpv%7B5-nuc%7D-double/tpv-energy.csv ... at least visible within the next few hours from posting): the problem seems to be that all three momentum values currently get posted into one line—thus we get a CSV line length of 5 (timestep, key, value0, value1, value2) instead of 3 (timestep, key, value). I.e. it should work if you store each momentum component individually.

I just modified that in the latest commit `da7fa9e8603ac4b424a6bafd143dbb098532a05e`  an hour ago. So, I guess that should work now with the format.

src/ResultWriter/EnergyOutput.cpp
```diff
@@ -367,11 +381,16 @@ void EnergyOutput::computeVolumeEnergies() {
       } else {
         // Elastic
         totalElasticKineticEnergyLocal += curWeight * curKineticEnergy;
+
```
sebwolf-de, line 384: remove

src/ResultWriter/EnergyOutput.cpp
```diff
@@ -549,6 +571,10 @@ void EnergyOutput::printEnergies() {
                       << 2.0 / 3.0 * std::log10(energiesStorage.plasticMoment()) - 6.07 << " ,"
                       << ratioPlasticMoment;
       }
+      if (totalMomentumX || totalMomentumY || totalMomentumZ) {
```
sebwolf-de, line 574: Should be an `&&` to make sure that all of these three quantities exist.

sebwolf-de (commented):
Only some minor comments.

src/ResultWriter/EnergyOutput.cpp
vikaskurapati: There might be cases where only one of them or two of them are zero and the remainging quantities are non-zero. For example, if we take a travelling wave scenario in x-direction. We get a zero in y and z directions and a non-zero momentum in x direction. I guess it still makes sense to have all the three values printed in any case.

vikaskurapati:
I removed the if condition around printing momentum values as it is possible to have total zero momentum with moving waves like sinusoidal waves. I guess we should just print it when the energy output is turned on.

PR #1098, opened 2024-04-22
alpham fixed in Receiver output

vikaskurapati:
This works now but the P_n0 on the fault receivers look a little more different from the reference. But the reference has the wrong implementation for dynamic rupture receiver output too. So, I think it makes sense to update the precomputed repository with this updated calculation after the PR is merged.

image


Aangniu:
I would agree with you on this. It is also expected that the resulted normal traction (P_n0) on the fault will be different with this change. I also agree with updating the pre-computed repository. Thanks!

PR #1099, opened 2024-05-02
harmonize documentation and code regarding tpProxyExponent

Thomas-Ulrich:
(it was either I change the documentation and move the denominator to the numerator in the equation: 
image
or changing the sign of alpha in the code. I propose changing the code, as the formula looks clearer like that.)

### DIFF
--- a/Documentation/parameters.par
+++ b/Documentation/parameters.par
@@ -54,7 +54,7 @@ etahack = 1                 ! use any value smaller than one to mitigate quasi-d
 OutputPointType = 5         ! Type (0: no output, 3: ascii file, 4: paraview file, 5: 3+4)
 SlipRateOutputType=0        ! 0: (smoother) slip rate output evaluated from the difference between the velocity on both side of the fault
                             ! 1: slip rate output evaluated from the fault tractions and the failure criterion (less smooth but usually more accurate where the rupture front is well developped)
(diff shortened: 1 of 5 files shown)


Documentation/parameters.par
```diff
@@ -54,7 +54,7 @@ etahack = 1                 ! use any value smaller than one to mitigate quasi-d
 OutputPointType = 5         ! Type (0: no output, 3: ascii file, 4: paraview file, 5: 3+4)
 SlipRateOutputType=0        ! 0: (smoother) slip rate output evaluated from the difference between the velocity on both side of the fault
                             ! 1: slip rate output evaluated from the fault tractions and the failure criterion (less smooth but usually more accurate where the rupture front is well developped)
-TpProxyExponent = -0.33333333 ! exponent alpha in the TP proxy slip weakening law (default is -1/3)
+TpProxyExponent = 0.33333333 ! exponent alpha in the TP proxy slip weakening law (default is 1/3)
```
davschneller, line 57: You could add a few `33333`'s here for good measure (say, so that we have e.g. 20 of the 3s there—that _should_ then hopefully equal 1/3 up to floating point standards; e.g. 1/3 in Python shows `0.3333333333333333`)

davschneller (approved):
LGTM; I think it also looks better like that.

Remark: in the documentation, there's currently also a `(-1/3)` which may need adjustment

PR #1102, opened 2024-05-07
Add TensorForge [Tracking PR]

davschneller:
We add TensorForge as a replacement and amalgamate of ~Yateto,~ (not anymore) gemmforge and chainforge. As a result, we can then handle tensors (and thus fused simulations etc.)—and we won't need any many Python packages anymore for GPU support.

A secondary goal (maybe part of a future PR) is to remove the explicit GPU code and move everything into the code generation (including also plasticity etc.)—resulting in (hopefully) simpler code since we can merge the CPU and GPU paths, faster code by merging kernels and reducing the latencies etc., but also simplify the build complexity on Nvidia and AMD systems, except if we want to use SYCL kernels or SYCL as runtime in the background.

A tertiary goal (maybe part of a future PR) is to get rid of the scratchpads (by kernel merging), so that we can play around with scheduling much more.

We also begin to add some basic support for `viscoelastic2` and `poroelastic` on GPUs—note that only viscoelasticity works at the moment.

Still WIP, but already with some progress (currently, the codegen is a bit broken still):

* [x] arbitrary-rank tensor support; minus full reduction
* [ ] (WIP) full reduction support
* [x] full yateto support. In particular, copy-add scales can be fused, thus possibly superseding #1044 .
* [x] OMP Target backend (dry-run)
* [ ] (preliminarily done) scalar nonlinear support (dry-run only)
* [ ] implement IR
* [ ] control structure support (ifs etc.)
* [ ] (partially done, WIP) translate DR to codegen
* [x] compiling `viscoelastic2` and `poroelastic` kernels
* [x] working `viscoelastic2` implementation
* [ ] working `poroelastic` implementation

(parts of this PR may be spun out)


davschneller:
`viscoelastic2` works now.

davschneller:
Mostly covered by #1284 and #1288 ; except for moving any nonlinear kernels into code generation.

PR #1111, opened 2024-05-14
Add Documentation Build To GHA

davschneller:
Adds the documentation build to Github Actions; previously it was still on Azure.
With that, the Azure pipeline has been removed entirely.

Also, we update fty and remove it from the explicit headers list.


vikaskurapati (approved):
Approved to my knowledge

PR #1120, opened 2024-05-23
Make Anisotropic Elasticity Work On GPUs

davschneller:
A few bugfixes to make anisotropy work on GPUs.

CI tests are included and pass.

### DIFF
--- a/src/Equations/anisotropic/Model/datastructures.hpp
+++ b/src/Equations/anisotropic/Model/datastructures.hpp
@@ -72,6 +72,14 @@ namespace seissol {
       double c56;
       double c66;
 
+      double getLambda() const override {
+        return (c11 + c22 + c33) / 3.0 - 2.0*getMu();
+      }
+
+      double getMu() const override {
+        return (c44 + c55 + c66) / 3.0;;
+      }
+
       AnisotropicMaterial() {}
 
       explicit AnisotropicMaterial(ElasticMaterial m) {
--- a/src/Equations/anisotropic/Model/datastructures.hpp
+++ b/src/Equations/anisotropic/Model/datastructures.hpp
@@ -72,11 +72,11 @@ namespace seissol {
       double c56;
       double c66;
 
-      double getLambda() const override {
-        return (c11 + c22 + c33) / 3.0 - 2.0*getMu();
+      double getLambdaBar() const override {
+        return (c11 + c22 + c33) / 3.0 - 2.0*getMuBar();
       }
 
-      double getMu() const override {
+      double getMuBar() const override {
         return (c44 + c55 + c66) / 3.0;;
       }
 
--- a/gitlab-gpu-pipeline.yml
+++ b/gitlab-gpu-pipeline.yml
@@ -85,8 +85,7 @@ gpu_build:
         - set -euo pipefail
         - if [ "$EQUATION" = viscoelastic2 ]; then MECHANISMS=3; else MECHANISMS=0; fi;
         - for PRECISION in double single; do
-            mkdir -p ./install_${BACKEND}_${PRECISION} ;
-            mkdir -p ./build_${BACKEND}_${PRECISION} && cd ./build_${BACKEND}_${PRECISION} ;
+            mkdir -p ./build_${BACKEND}_${PRECISION}_${EQUATION} && cd ./build_${BACKEND}_${PRECISION}_${EQUATION} ;
             cmake ..
             -DCMAKE_BUILD_TYPE=Release
             -DDEVICE_BACKEND=${BACKEND}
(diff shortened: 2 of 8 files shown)


davschneller:
Two open questions remain:

* I added two methods `getMu`, `getLambda` to all materials. However, is that always a good way to name them like that?
* The Dirichlet boundary conditions will currently behave a little differently due to the `ifdef`s regarding isotropic elasticity changing some parameters (impedance and so on)

vikaskurapati:
> Two open questions remain:
> 
> * I added two methods `getMu`, `getLambda` to all materials. However, is that always a good way to name them like that?
Well, I would say it is not a good way to name them this way for anisotropic as these parameters don't really exist for the anisotropic setup. Is there no way to surpass these two functions for anisotropic simulations? or just pass them without defining specifically?
> * The Dirichlet boundary conditions will currently behave a little differently due to the `ifdef`s regarding isotropic elasticity changing some parameters (impedance and so on)
We will need to fix this in a different PR with Dirichlet boundary conditions for anisotropic...

src/Equations/anisotropic/Model/datastructures.hpp
```diff
@@ -72,6 +72,14 @@ namespace seissol {
       double c56;
       double c66;
 
+      double getLambda() const override {
+        return (c11 + c22 + c33) / 3.0 - 2.0*getMu();
+      }
```
vikaskurapati, line 77: These are used only for the Free Surface Boundary condition in the end, right? If that is the case, wouldn't it be better if we just write that that surface boundary condition is not yet implemented for anisotropic material? Mainly because I am not really sure if the free surface boundary condition just behaves this way with anisotropic material or if there is more math involved for the same. 

We are also yet to solve the acoustic-anisotropic coupling problem and this could then be taken care of there...

Thomas-Ulrich: it seems getLambda is required to compute P wave speed, which is needed to get the max wave speed and therefore the min dt.

davschneller:
Thanks for the replies.

The `getLambda` etc. _can_ be used to formulate the P/S wave speed calculation—but that's not mandatory.

One way to maybe rewrite the `getLambda` / `getMu` functions would maybe to call them `getLambdaBar` or similar (average?)—that's at least what was done in the FORTRAN attenuation computation (which currently doesn't support anisotropy anyways anymore).

So yes, we can either disable the calculation for now (and just remove the lambda recording—I could do that, at the cost of some more ifdefs), then the `getMu`/`getLambda` could be removed entirely.
But if that boundary condition is not going to used anyways at the moment—we could also leave it be for now (and avoid the ifdefs); unless that's too misleading. In that case, it may be even useful to then _temporarily_ adjust the CPU version to also use the (isotropic) elastic approximation in the anisotropic case.

Whatever you'd prefer to have for now—until we get something proper. :)

Thomas-Ulrich:
getLambdaBar is fine IMO.
which boundary condition are you talking about? Free-surface? I suppose this is a must, right?

vikaskurapati:
> getLambdaBar is fine IMO. which boundary condition are you talking about? Free-surface? I suppose this is a must, right?

This is the Gravitational Free-surface boundary condition that Lukas implemented for the fully coupled Earthquake tsunami scenario if I recall it correctly...

davschneller:
Yup, it's the free-surface conditions.

I will have to say; I don't know what kind of adjustments we'll need for the anisotropic case there (probably rotate the parameters to the face?); I'd leave that to @vikaskurapati if he likes. 

But we could employ `lambdaBar` and `muBar` as a first approximation—until more exact strategies are implemented. (besides, the acoustic support for the anisotropic material is probably currently also unfinished)

davschneller:
To add some code, see here for the CPU case right now (there's only a different code path for isotropic elastic):

https://github.com/SeisSol/SeisSol/blob/3cb4872176203b8f93eeee113c8483eca6a79197/src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h#L116-L120
https://github.com/SeisSol/SeisSol/blob/3cb4872176203b8f93eeee113c8483eca6a79197/src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h#L142-L148

and the GPU case

https://github.com/SeisSol/SeisSol/blob/3cb4872176203b8f93eeee113c8483eca6a79197/src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h#L252-L258
https://github.com/SeisSol/SeisSol/blob/3cb4872176203b8f93eeee113c8483eca6a79197/src/Equations/elastic/Kernels/GravitationalFreeSurfaceBC.h#L291-L302

vikaskurapati:
> Yup, it's the free-surface conditions.
> 
> I will have to say; I don't know what kind of adjustments we'll need for the anisotropic case there (probably rotate the parameters to the face?); I'd leave that to @vikaskurapati if he likes and is ok with that.
I am fine with that. If there are any changes that would be needed later on, I can modify as per the current implementation. 
> 
> But we could employ `lambdaBar` and `muBar` as a first approximation—until more exact strategies are implemented. (besides, the acoustic support for the anisotropic material is probably currently also unfinished)

The `lambdaBar` and `muBar` sounds good to start with. I am yet to figure out the acoustic support for the anisotropic material and the anisotropic-acoustic coupling Riemann problem.

Thomas-Ulrich:
ok, free surface is a must, but I think we can leave with approximations with free surface with gravity and anisotropy.
(maybe add a warning if you think it is necessary?)

davschneller:
Name has been fixed; and I've enabled the isotropic elastic material changes in the free-surface BC for all (current) materials now.

(would probably be better to double-check if you like; to be absolutely safe)

PR #1125, opened 2024-06-12
Update rise_time_from_xdmf.py

NicoSchlw:
speed-up script with jit compilation
add median filter
fix output filename

### DIFF
--- a/postprocessing/science/rise_time_from_xdmf.py
+++ b/postprocessing/science/rise_time_from_xdmf.py
@@ -1,19 +1,28 @@
 # Script to determine the rise time (time the absolute slip rate is above a certain threshold) from seissol fault.xdmf output
-# Output of this script is a new xdmf file
 
 import numpy as np
 import seissolxdmf as sx
 import seissolxdmfwriter as sxw
 import argparse
 import os
+import timeit
+from numba import jit
+from multiprocessing import Pool, cpu_count
 
 parser = argparse.ArgumentParser(description='calculate rise times from fault xdmf file')
 parser.add_argument('filename', help='path+filename-fault.xdmf')
-parser.add_argument('--threshold', nargs=1, metavar=('threshold'), default=([0.01]), help='specify slip rate threshold' ,type=float)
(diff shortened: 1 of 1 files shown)


postprocessing/science/rise_time_from_xdmf.py
```diff
@@ -22,44 +31,107 @@ def counting_loop(row, array, threshold=0.01):
...
+                variableSmooth[i] = np.median(variable[tmp_ind3])
+        else:
+            variableSmooth[i] = np.median(variable[distances <= radius])
+    return variableSmooth
+
+
+def MedianSmoothingParallel(variable, xyz, radius=250., nprocs=4):
+    
+    chunks = [(variable, xyz, i, radius) for i in np.array_split(np.arange(variable.shape[0]), nprocs)]        
+    p = Pool(nprocs)
```
Thomas-Ulrich, line 88: consider using Context Manager for the Pool:
```
    with Pool(nprocs) as pool:
        result = pool.map(MedianSmoothing, chunks)
```

postprocessing/science/rise_time_from_xdmf.py
```diff
@@ -22,44 +31,107 @@ def counting_loop(row, array, threshold=0.01):
...
+            tmp_ind3 = tmp_ind1 & tmp_ind2
+            non_zero_portion = np.sum(tmp_ind3)/np.sum(tmp_ind1)
+            if non_zero_portion < 0.75 and variable[j] == 0:       # improve filter for branching faults
+                variableSmooth[i] = 0
+            elif non_zero_portion < 0.35:  # remove isolated slipping elements
+                variableSmooth[i] = 0
+            else:
+                variableSmooth[i] = np.median(variable[tmp_ind3])
+        else:
+            variableSmooth[i] = np.median(variable[distances <= radius])
```
Thomas-Ulrich, line 81: seems like variable[distances <= radius && variable != 0] would also make sense here?

Thomas-Ulrich (approved):
what about linting the script (with black https://github.com/psf/black?)

postprocessing/science/rise_time_from_xdmf.py
NicoSchlw: This would lead to nans for regions with only zeros

PR #1126, opened 2024-06-14
Add Off-Fault Receiver Strain Output

davschneller:
We add strain output for the receivers. Also, we refactor the implementation a bit. Besides, we add parallel receiver evaluation (if there are enough of them).


src/Kernels/Receiver.cpp
```diff
@@ -116,97 +156,111 @@ double seissol::kernels::ReceiverCluster::calcReceivers(  double time,
   }
 #endif
 
-  double receiverTime = time;
   if (time >= expansionPoint && time < expansionPoint + timeStepWidth) {
-    for (size_t i = 0; i < m_receivers.size(); ++i) {
+    // heuristic; to avoid the overhead from the parallel region
+    auto threshold = omp_get_num_threads() * 100;
```
Thomas-Ulrich, line 161: what about: std::max(1000, omp_get_num_threads() * 100) 
This would ensure that even if the number of threads is low, parallelization doesn't kick in

Thomas-Ulrich (approved):
LGTM

Thomas-Ulrich (approved):

PR #1128, opened 2024-06-18
Add LUA mirror

davschneller:
Add a mirror for LUA, if it cannot be found.


davschneller:
(note that it's probably better to pre-build the images we use anyways—i.e. make the `build-seissol/dependencies` be mostly skipped)

Thomas-Ulrich (approved):

PR #1134, opened 2024-07-12
Make the Proxy compile without OpenMP

davschneller:
The title says it all. There were some changes over the years which made the proxy not compile without OpenMP—so we fix them.


vikaskurapati (approved):
Approving as it seems like it does not break anything. The CI tests seem to have failed due to time outs.

PR #1140, opened 2024-08-12
Print Convergence Order in the welcome message

vikaskurapati:


davschneller (approved):
LGTM; but I'd place it a bit more down in the log (after the hashes and parallelization technicalities have been printed).

PR #1141, opened 2024-08-14
Add CMake Warning for Serial Hdf5, Add Partial ACPP Arguments/Package Search

davschneller:
As the title says:

* add a CMake warning for serial Hdf5 (which will currently cause the code to not compile)
* remove the deprecation warning for hipSYCL by introducing the respective ACPP arguments


davschneller:
(partial overlap to #832 ; but we do not yet fix the use of AdaptiveCpp for codegen—for now at least; maybe #1102 may be beneficial to merge here before as well)

vikaskurapati (approved):
LGTM

PR #1145, opened 2024-08-19
Update OutputParameters.cpp

vikaskurapati:
Modification of reading checkpoint parameters

### DIFF
--- a/src/Initializer/Parameters/OutputParameters.cpp
+++ b/src/Initializer/Parameters/OutputParameters.cpp
@@ -35,13 +35,13 @@ CheckpointParameters readCheckpointParameters(ParameterReader* baseReader) {
            {"mpio", CheckpointingBackend::MPIO},
            {"mpio_async", CheckpointingBackend::MPIO_ASYNC},
            {"sionlib", CheckpointingBackend::SIONLIB}});
+        const auto interval = reader->readWithDefault("checkpointinterval", 0.0);
     } else {
-      reader->markUnused({"CheckpointingBackend"});
+      reader->markUnused({"checkpointbackend", "checkpointinterval"});
     }
     return backend;
   };
   const auto backend = readBackend(enabled);
-  const auto interval = reader->readWithDefault("checkpointinterval", 0.0);
 
   warnIntervalAndDisable(enabled, interval, "checkpoint", "checkpointinterval");
 
@@ -50,7 +50,7 @@ CheckpointParameters readCheckpointParameters(ParameterReader* baseReader) {
     if (enabled) {
       fileName = reader->readOrFail<std::string>("checkpointfile", "No checkpoint fileName given.");
     } else {
-      reader->markUnused({"chekpointfileName"});
+      reader->markUnused({"chekpointfile"});
     }
     return fileName;
   };
--- a/src/Initializer/Parameters/OutputParameters.cpp
+++ b/src/Initializer/Parameters/OutputParameters.cpp
@@ -35,15 +35,17 @@ CheckpointParameters readCheckpointParameters(ParameterReader* baseReader) {
            {"mpio", CheckpointingBackend::MPIO},
            {"mpio_async", CheckpointingBackend::MPIO_ASYNC},
            {"sionlib", CheckpointingBackend::SIONLIB}});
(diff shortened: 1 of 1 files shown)


davschneller (commented):
Hi, thanks for doing that; general comment:

* I'd use two `if`s. The following structure:
  * Read `checkpoint` (unconditionally).
    * If false, then `markUnused` for `checkpointinterval`.
    * If true, read `checkpointinterval`, and check if it's enabled.
  * With the result (in `enabled`):
    * If true, read `checkpointbackend` and `checkpointfile`.
    * Otherwise, mark them both as unused.

The lambdas can be removed then.

Furthermore: small commits can be combined by rebasing (make sure to not include merge commits there) or soft resetting; all combined with forced pushing. That's especially useful for avoiding too many formatting commits.

src/Initializer/Parameters/OutputParameters.cpp
```diff
@@ -23,41 +23,28 @@ CheckpointParameters readCheckpointParameters(ParameterReader* baseReader) {
...
-  const auto backend = readBackend(enabled);
-  const auto interval = reader->readWithDefault("checkpointinterval", 0.0);
-
+  CheckpointingBackend backend = CheckpointingBackend::DISABLED;
+  double interval = 0.0;
+  std::string fileName = "";
   if (enabled) {
+    interval = reader->readWithDefault("checkpointinterval", 0.0);
     warnIntervalAndDisable(enabled, interval, "checkpoint", "checkpointinterval");
   }
```
davschneller, line 32: I'd check for `markUnused` for `checkpointinterval` already here—if `enabled==false`

vikaskurapati:
> Hi, thanks for doing that; general comment:
> 
> * I'd use two `if`s. The following structure:
>   
>   * Read `checkpoint` (unconditionally).
>     
>     * If false, then `markUnused` for `checkpointinterval`.
>     * If true, read `checkpointinterval`, and check if it's enabled.
>   * With the result (in `enabled`):
>     
>     * If true, read `checkpointbackend` and `checkpointfile`.
>     * Otherwise, mark them both as unused.
> 
> The lambdas can be removed then.
Used a slightly different approach with the ifs so that the amount of code can be reduced and we have all the `markUnused` in one place. 
> 
> Furthermore: small commits can be combined by rebasing (make sure to not include merge commits there) or soft resetting; all combined with forced pushing. That's especially useful for avoiding too many formatting commits.

src/Initializer/Parameters/OutputParameters.cpp
vikaskurapati: I wrote it this way so that all the unused variables are in one place. Would you rather have it in two different if, else statements? I would change that if it is cleaner.

davschneller: Would be better: what could happen is

* `checkpoint=1`
* `checkpointinterval=-1`

Then we mark only two fields with unused. Compare that with `checkpoint=0`. Then, also `checkpointinterval` is unused.

vikaskurapati: > Would be better: what could happen is
> 
> * `checkpoint=1`
> * `checkpointinterval=-1`
> 
> Then we mark only two fields with unused. Compare that with `checkpoint=0`. Then, also `checkpointinterval` is unused—in addition to the two other fields.

I am sorry, I think I am confused:

lets say checkpoint=1, checkpointinterval=-1, here `warnIntervalAndDisable`  turns `enabled=false` and then all the three fields are marked unused in the next if, else statement. Or did you want to not mark the `checkpointinterval` in this case?

If `checkpoint=0`, both the ifs and ignored anyway and the else statement gets activated.

Or did you mean something else entirely?

davschneller: Yes; you're right about that—it won't really matter here, as the reader right now allows a variable to be both read (i.e. used) and begin marked as unused.

To me, it would come over cleaner if the `checkpointinterval` would be marked as unused in the first `if`, since it's being queried in the other part of the `if` there. Just in case the logic changes at some point in the future.

Otherwise, it would be best to add a comment why `checkpointinterval` is marked as unused only there.

PR #1152, opened 2024-08-28
Use Persistent MPI by default

vikaskurapati:
@davschneller please feel free to reject it if we are not ready to use Persistent MPI by default everywhere


davschneller (approved):
Thanks!

I think we are—it might actually also fix other MPI problems, if they are there right now. Besides, it may enable some optimizations within some other MPI implementations as well (and you never know what they're really doing internally, somehow...).

(note that the docs will also need to be updated: cf. the last sentence in https://github.com/SeisSol/SeisSol/blob/master/Documentation/environment-variables.rst#persistent-mpi-operations )

PR #1153, opened 2024-08-28
Update environment-variables.rst

vikaskurapati:
Modified the Persistent MPI section and slight rewording of other sections

### DIFF
--- a/Documentation/environment-variables.rst
+++ b/Documentation/environment-variables.rst
@@ -9,7 +9,7 @@ Communication Thread
 By default, any SeisSol run with more than one MPI rank will use a communication thread to advance the MPI progress engine.
 For that, it is required to leave at least one thread vacant in your OpenMP thread placing map, cf. the SuperMUC-NG example below.
 
-If you do not want to use a communication thread, you may set `SEISSOL_COMMTHREAD=0` to make SeisSol poll the progress from time to time.
+If you do not want to use a communication thread, you may set ``SEISSOL_COMMTHREAD=0`` to make SeisSol poll the progress from time to time.
 
 Load Balancing
 --------------
@@ -19,7 +19,7 @@ For that, it runs the so-called "Mini SeisSol" benchmark. As its name already hi
 thus estimating the performance of all nodes relative to each other. The number of elements per node assigned during the partitioning will be resized according to these values.
 
 As a result, the partitioning of runs may become non-deterministic, and the initialization procedure may take a little longer, especially when running only on a single node with multiple ranks.
-To disable it, set `SEISSOL_MINISEISSOL=0`.
+To disable it, set ``SEISSOL_MINISEISSOL=0``.
 
 Persistent MPI Operations
 -------------------------
@@ -27,7 +27,7 @@ Persistent MPI Operations
 Since SeisSol has a static communication pattern which means, we issue the same MPI transfer requests. 
 Thus, we use persistent MPI communication which may reduce communication latency.
 
-To disable persistent communication, set `SEISSOL_MPI_PERSISTENT=0`. Then, SeisSol will use `MPI_Isend` and `MPI_Irecv` instead. To explicitly enable the persistent communication, set `SEISSOL_MPI_PERSISTENT=1`. Right now, it is enabled by default.
+To disable persistent communication, set ``SEISSOL_MPI_PERSISTENT=0``. Then, SeisSol will use ``MPI_Isend`` and ``MPI_Irecv`` instead. To explicitly enable the persistent communication, set ``SEISSOL_MPI_PERSISTENT=1``. Right now, it is enabled by default.
 
 Output
 ------
@@ -79,5 +79,5 @@ Here are some proposed environment variables to be added before invoking SeisSol
 
 A complete batch script for SuperMUC-NG can be found in the chapter about ref:`SuperMUC-NG <running_seissol_on_supermuc>`.
 
-In previous versions of SeisSol, you had to explicitly compile the software with `-DCOMMTHREAD=ON`. That is not necessary anymore, as
+In previous versions of SeisSol, you had to explicitly compile the software with ``-DCOMMTHREAD=ON``. That is not necessary anymore, as
 any configuration with more than one MPI rank uses the communication thread by default.
--- a/Documentation/environment-variables.rst
+++ b/Documentation/environment-variables.rst
@@ -7,14 +7,14 @@ Communication Thread
 --------------------
 
 By default, any SeisSol run with more than one MPI rank will use a communication thread to advance the MPI progress engine.
-For that, you must leave at least one thread vacant in your OpenMP thread placing map, cf. the SuperMUC-NG example below.
+For that, it is required to leave at least one thread vacant in your OpenMP thread placing map, cf. the SuperMUC-NG example below.
 
-If you do not want to use a communication thread, you may set `SEISSOL_COMMTHREAD=0`; then, SeisSol polls the progress from time to time.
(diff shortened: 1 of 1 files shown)


Documentation/environment-variables.rst
```diff
@@ -7,28 +7,27 @@ Communication Thread
 --------------------
 
 By default, any SeisSol run with more than one MPI rank will use a communication thread to advance the MPI progress engine.
-For that, you will need to leave at least one thread vacant in your OpenMP thread placing map, cf. the SuperMUC-NG example below.
+For that, you must leave at least one thread vacant in your OpenMP thread placing map, cf. the SuperMUC-NG example below.
 
-If you do not want to use a communication thread, you may set `SEISSOL_COMMTHREAD=0`; then SeisSol polls on the progress from time to time.
+If you do not want to use a communication thread, you may set `SEISSOL_COMMTHREAD=0`; then, SeisSol polls the progress from time to time.
```
davschneller, line 12: You can use ` `` ` here for the `SEISSOL_COMMTHREAD=0` . Also, how about "you may set `SEISSOL_COMMTHREAD=0` to make SeisSol poll the progress [...]" ? (just to remove the semicolon after the code snippet)

Documentation/environment-variables.rst
```diff
@@ -7,28 +7,27 @@ Communication Thread
 --------------------
 
 By default, any SeisSol run with more than one MPI rank will use a communication thread to advance the MPI progress engine.
-For that, you will need to leave at least one thread vacant in your OpenMP thread placing map, cf. the SuperMUC-NG example below.
+For that, you must leave at least one thread vacant in your OpenMP thread placing map, cf. the SuperMUC-NG example below.
```
davschneller, line 10: Maybe a bit finnicky to remark, but "must" feels a bit out of place here (since it's not an "inner" thing).

... how about "For that, it is required that you leave [...]" ?

(I agree that the `will need to` formulation etc. are a bit more informal)

Documentation/environment-variables.rst
```diff
@@ -7,28 +7,27 @@ Communication Thread
...
+As a result, the partitioning of runs may become non-deterministic, and the initialization procedure may take a little longer, especially when running only on a single node with multiple ranks.
 To disable it, set `SEISSOL_MINISEISSOL=0`.
 
 Persistent MPI Operations
 -------------------------
 
-Since SeisSol has a static communication pattern (in the sense of: per iteration, we issue the same MPI transfer requests),
-we may use persistent MPI communication—it may reduce the communication latency.
+Since SeisSol has a static communication pattern (in the sense that per iteration, we issue the same MPI transfer requests). 
+We may use persistent MPI communication, which may reduce communication latency.
```
davschneller, line 28: No comma before "which" needed.

Documentation/environment-variables.rst
```diff
@@ -7,28 +7,27 @@ Communication Thread
...
-As a result, the partitioning of runs may become non-deterministic, and the initialization procedure may take a little longer; especially when running only on a single node with multiple ranks.
+As a result, the partitioning of runs may become non-deterministic, and the initialization procedure may take a little longer, especially when running only on a single node with multiple ranks.
 To disable it, set `SEISSOL_MINISEISSOL=0`.
 
 Persistent MPI Operations
 -------------------------
 
-Since SeisSol has a static communication pattern (in the sense of: per iteration, we issue the same MPI transfer requests),
-we may use persistent MPI communication—it may reduce the communication latency.
+Since SeisSol has a static communication pattern (in the sense that per iteration, we issue the same MPI transfer requests).
```
davschneller, line 27: The sentence stop here will cause the sentence to sound a bit incomplete. We could e.g. remove the bracket, going for something like "SeisSol has a static communication pattern. That means, we issue the same MPI transfer requests in each iteration. Thus, we use persistent MPI [...]".

Documentation/environment-variables.rst
```diff
@@ -7,28 +7,27 @@ Communication Thread
...
 -------------------------
 
-Since SeisSol has a static communication pattern (in the sense of: per iteration, we issue the same MPI transfer requests),
-we may use persistent MPI communication—it may reduce the communication latency.
+Since SeisSol has a static communication pattern (in the sense that per iteration, we issue the same MPI transfer requests). 
+We may use persistent MPI communication, which may reduce communication latency.
 
-You may enable persistent communication by setting `SEISSOL_MPI_PERSISTENT=1`,
-and explicitly disable it with `SEISSOL_MPI_PERSISTENT=0`. Right now, it is disabled by default.
+You may enable persistent communication by setting `SEISSOL_MPI_PERSISTENT=1`, explicitly disabling it with `SEISSOL_MPI_PERSISTENT=0`. Right now, it is enabled by default.
```
davschneller, line 30: Again ` `` ` may be used here.

davschneller, line 30: Maybe the formulation here should be inverted as well: i.e. "To disable persistent communication, set ``SEISSOL_MPI_PERSISTENT=0``. Then, SeisSol will use  `MPI_Isend` and `MPI_Irecv` instead. To explicitly enable the persistent communication, set ``SEISSOL_MPI_PERSISTENT=1``. Right now, it is enabled by default."

Documentation/environment-variables.rst
```diff
@@ -51,16 +50,16 @@ variables listed in the `wiki <https://github.com/TUM-I5/ASYNC/wiki>`__.
...
 .. _optimal_environment_variables_on_supermuc_ng:
 
 Optimal environment variables on SuperMUC-NG
 --------------------------------------------
 
 On SuperMUC-NG, we recommend using SeisSol with async output in thread mode.
-Also, we recommend using hyperthreading capabilities (that is using 96 CPUs instead of 48. 2 threads out of 96 are used as communication threads).
-Here are some proposed environment variables, to be added prior to invoking SeisSol in your batch file:
+Also, we recommend using hyperthreading capabilities (that is, using 96 CPUs instead of 48. 2 threads out of 96 are used as communication threads).
+Here are some proposed environment variables to be added before invoking SeisSol in your batch file:
```
davschneller, line 62: Sure about removing that comma?

davschneller (commented):
Thanks for doing that!
I've left some comments about the phrasing here.

Documentation/environment-variables.rst
vikaskurapati: "It is required to leave"...

Documentation/environment-variables.rst
vikaskurapati: What do you mean? `` is already used? Or did you mean to use ``SEISSOL_COMMTHREAD=0``?

Documentation/environment-variables.rst
vikaskurapati: Modified but within the same sentence

Documentation/environment-variables.rst
vikaskurapati: Modified but within the same sentence

Documentation/environment-variables.rst
vikaskurapati: Yes

Documentation/environment-variables.rst
vikaskurapati: Do you mean `` Variable ``? Then, we need to change it everywhere in the document. We have been using ` Variable ` until now iirc.

Documentation/environment-variables.rst
davschneller: The `rst` syntax will interpret a single
```
`
```
as italic text; only a double ` `` ` as code

vikaskurapati: Understood. Thanks! Modified

PR #1159, opened 2024-09-02
Disabling the option of fused simulations as of now in master branch

vikaskurapati:


davschneller (approved):
LGTM (the test failure given happens from time to time—but the reason for it is still unclear)

PR #1160, opened 2024-09-03
More Extensive CPU Tests

davschneller:
We refine the CPU tests a bit; now we test against both `gcc-13` and `clang-19`. Also, we run the tests with `-Wall` now.

Also, we now run the proxy routinely in the tests and try to start the application.

Besides, there's one tiny bugfix for the proxy with poroelasticity.


vikaskurapati (approved):
LGTM as far as I understand it.

PR #1162, opened 2024-09-05
Option to enable a floating point check in the simulation. 

vikaskurapati:


src/Main.cpp
```diff
@@ -99,6 +103,9 @@ int main(int argc, char* argv[]) {
 
   seissol::MPI::mpi.init(argc, argv);
   const int rank = seissol::MPI::mpi.rank();
+#ifdef FP_CHECK
```
davschneller, line 106: Would it cause any problem to make it an environment variable instead?
(that's usually more convenient to handle than a CMake flag)

vikaskurapati: Done.

src/Main.cpp
```diff
@@ -100,6 +102,11 @@ int main(int argc, char* argv[]) {
   seissol::MPI::mpi.init(argc, argv);
   const int rank = seissol::MPI::mpi.rank();
 
+  if (utils::Env::get<bool>("FLOATING_POINT_EXCEPTION", false)) {
+    logInfo() << "REACHED HERE";
```
davschneller, line 106: `logInfo(rank) << "Enabling floating point exception handlers.";` or the likes?

vikaskurapati: Oh, my bad, it seems like I forgot to remove the test statement I used. Putting the right one now, thanks!

vikaskurapati:
``clang-tidy`` seems to not find the included ``cfenv``. I included the ``utils``.

davschneller (approved):

davschneller:
> `clang-tidy` seems to not find the included `cfenv`. I included the `utils`.

Maybe put `std::` in front of it?
(also, the `logInfo()` should be `logInfo(rank)`; otherwise the message is gonna appear for every rank)

vikaskurapati:
> > `clang-tidy` seems to not find the included `cfenv`. I included the `utils`.
> 
> Maybe put `std::` in front of it? (also, the `logInfo()` should be `logInfo(rank)`; otherwise the message is gonna appear for every rank)

Modified the ``logInfo``. `std::` does not work as ``fenableexcept`` is not a part of the std environment.

davschneller:
Ok, hadn't seen that it's only a GNU extension (cf. the comment on https://en.cppreference.com/w/cpp/numeric/fenv )—maybe some preprocessor guards will be necessary after all

vikaskurapati:
> Ok, hadn't seen that it's only a GNU extension (cf. the comment on https://en.cppreference.com/w/cpp/numeric/fenv )—maybe some preprocessor guards will be necessary after all

Added the preprocessor guards now.

src/Main.cpp
```diff
@@ -41,13 +41,16 @@
 #include "Initializer/Parameters/ParameterReader.h"
 #include "Initializer/PreProcessorMacros.h"
 #include "Modules/Modules.h"
+#include "fenv.h"
```
davschneller, line 44: `<fenv.h>`

I'd also include-guard that one.

src/Main.cpp
```diff
@@ -66,6 +68,16 @@
 #include "device.h"
 #endif
 
+// Check if on a GNU system (Linux) or other platform
+#if defined(__GNUC__) || defined(__linux__)
+#define ENABLE_FLOAT_EXCEPTIONS(rank)                                                              \
```
davschneller, line 73: It's best to move the `if` to line 116 IMO, to avoid the `ENABLE_FLOAT_EXCEPTIONS` as far as possible—macros can cause a lot of trouble and are hence best reduced to minimum.

src/Main.cpp
```diff
@@ -56,7 +58,7 @@
 #include "Initializer/InitProcedure/Init.h"
 #include "Initializer/Parameters/SeisSolParameters.h"
 #include "SeisSol.h"
-#include "utils/args.h"
+#include <cfenv>
```
davschneller, line 61: Can be removed now

davschneller, line 61: Can be removed now

src/Main.cpp
```diff
@@ -100,6 +106,20 @@ int main(int argc, char* argv[]) {
   seissol::MPI::mpi.init(argc, argv);
   const int rank = seissol::MPI::mpi.rank();
 
+  if (utils::Env::get<bool>("FLOATING_POINT_EXCEPTION", false)) {
+    // Check if on a GNU system (Linux) or other platform
+#if defined(__GNUC__) || defined(__linux__)
+#define ENABLE_FLOAT_EXCEPTIONS(rank)                                                              \
```
davschneller, line 112: Remove (i.e. just have
```
#if defined(__GNUC__) || defined(__linux__)
feenableexcept();
logInfo << "yes";
#else
logInfo << "no";
#endif
```
)
(the `#define` is not necessary here)

davschneller (commented):
Almost LGTM. The extra `#define` should be removed for readability—then we're good to go.

Also, it's best if you squash your commits a bit; to get rid especially of the formatting commits—e.g. by `git rebase -i be28e5b; git push --force`. It's ok (and good) to split changes into multiple commits, especially if they're logically different; but IMO formatting commits should be squashed to the possible minimum.
(make sure to not do that over merge commits)

davschneller (approved):

PR #1163, opened 2024-09-05
Remove Some Compiler Warnings

davschneller:
We just remove most of the warnings that appear with `-Wall`. Excluding, most notably, the vectorization failures for the DR code, as well as the errors in the libxsmm-generated code.

(note that some of them occur because the energy computation isn't yet implemented for anisotropic elastic and poroelastic material)

### DIFF
--- a/src/Initializer/TimeStepping/LtsLayout.cpp
+++ b/src/Initializer/TimeStepping/LtsLayout.cpp
@@ -599,9 +599,6 @@ void seissol::initializer::time_stepping::LtsLayout::normalizeClustering() {
   unsigned int l_maximumDifference      = 0;
   unsigned int l_dynamicRupture         = 0;
   unsigned int l_singleBuffer           = 0;
-  unsigned int l_totalMaximumDifference = 0;
-  unsigned int l_totalDynamicRupture    = 0;
-  unsigned int l_totalSingleBuffer      = 0;
 
   int l_globalContinue = 1;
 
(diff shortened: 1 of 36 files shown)


src/Initializer/TimeStepping/LtsLayout.cpp
```diff
@@ -599,9 +599,8 @@ void seissol::initializer::time_stepping::LtsLayout::normalizeClustering() {
   unsigned int l_maximumDifference      = 0;
   unsigned int l_dynamicRupture         = 0;
   unsigned int l_singleBuffer           = 0;
-  unsigned int l_totalMaximumDifference = 0;
-  unsigned int l_totalDynamicRupture    = 0;
-  unsigned int l_totalSingleBuffer      = 0;
+  // unsigned int l_totalMaximumDifference = 0;
+  // unsigned int l_totalDynamicRupture    = 0;
```
vikaskurapati, line 603: Unwanted comments? Or was this intended to work on later?

src/Initializer/TimeStepping/LtsLayout.cpp
```diff
@@ -624,9 +623,8 @@ void seissol::initializer::time_stepping::LtsLayout::normalizeClustering() {
     // TODO: missing implementation (works only for max. difference 0 or 1)
     l_singleBuffer = enforceSingleBuffer();
 
-    l_totalMaximumDifference += l_maximumDifference;
-    l_totalDynamicRupture    += l_dynamicRupture;
-    l_totalSingleBuffer      += l_singleBuffer;
+    // l_totalMaximumDifference += l_maximumDifference;
+    // l_totalDynamicRupture    += l_dynamicRupture;
```
vikaskurapati, line 627: Same as above

vikaskurapati (approved):
LGTM. Can be merged once those comments are deleted.

src/Initializer/TimeStepping/LtsLayout.cpp
davschneller: There was a commented message a few lines below which these variables were still for.

But—I've removed them now still: the whole file is best completely rewritten in the end, so those comments can go now as well.

PR #1166, opened 2024-09-10
Acceleration values in Receivers

vikaskurapati:
This PR is to output acceleration values in receivers.


vikaskurapati:
> Something that I just noticed: the `kdivMT` matrices get negated within SeisSol—maybe that's related to why the formula doesn't work as intended—even though it looks fully correct and logical to me.

Shouldn't the `kernel::derivativeTaylorExpansion` automatically take care of it like it does in `computeTaylorExpansion`? Or am I missing something else here?

PR #1173, opened 2024-09-23
Update Submodules

davschneller:
We

* update the Device submodule to fix a bug with ROCm 6.0 on MI250X onwards (potentially affecting other GPUs, or being related to the OS configuration as well). The update also includes the fix for the oneAPI graph extension.
* add a few licensing details; a bit of progress towards adapting the REUSE spec
* update the fty module. Giving us a bit more backwards compatibility (i.e. now `0d0` is properly converted to `0e0`; before that only worked for `0d+0` or `0d-0`)


CODE_OF_CONDUCT.md
```diff
@@ -1,3 +1,9 @@
+<!--
+    SPDX-FileCopyrightText: 2014 Coraline Ada Ehmke
```
Thomas-Ulrich, line 2: is it normal that it is not 2022-2024 SeisSol Group?

Thomas-Ulrich (approved):
see comment, else LGTM

CODE_OF_CONDUCT.md
davschneller: Effectively, we've just copied the file from there; didn't we? Honestly, I'm a bit uncertain myself; but I don't see any contribution from our side.

We can change it to `SeisSol Group` etc.; but I'm sure if we have a "copyright" there. The original CoC we definitely released under CC-BY 4.0 though.

(if we want to discuss these issues for thoroughly; I can also remove the statement for now from this file)

davschneller: Oh wait, I stand corrected. We _did_ modify the file. I'll add us. Or should we adjust the file even more?

davschneller: Ok; I've adjusted the header as you've suggested and only left the SeisSol Group—the attribution at the bottom should be good. I guess otherwise it might also give off the "wrong" impression.

PR #1176, opened 2024-10-02
Add Rusanov Flux

davschneller:
An ad-hoc implementation based on @sebwolf-de 's centered flux implementation. I.e. you can now choose between using Godunov and Rusanov; both just using only near the fault (faces with adjacent cells having at least one DR face), or everywhere.


Documentation/parameters.par
```diff
@@ -16,6 +16,9 @@ ITMStartingtime = 2.0 ! Time when ITM is turned on
 ITMTime = 0.01 !Time duration for which ITM is on
 ITMVelocityscalingfactor = 2 ! Scaling factor
 ITMReflectionType = 1 ! 1 reflects both waves, 2 reflects both waves with constant velocity, 3 reflects only P-waves, 4 reflects only S-waves
+
+numflux = 'godunov' ! The numerical flux near the fault. Values: 'godunov', 'rusanov'
```
Thomas-Ulrich, line 20: remove "near the fault"
(and the line below precise
The flux for cells adjacent to the fault, excluding the fault boundary flux.
or something like that?)

generated_code/aderdg.py
```diff
@@ -197,11 +197,15 @@ def starMatrix(self, dim):
     pass
 
   def addInit(self, generator):
+    godunov_spp = self.godunov_spp()
+    self.QcorrLocal = Tensor('QcorrLocal', godunov_spp.shape)
+    self.QcorrNeighbor = Tensor('QcorrNeighbor', godunov_spp.shape)
+
     fluxScale = Scalar('fluxScale')
-    computeFluxSolverLocal = self.AplusT['ij'] <= fluxScale * self.Tinv['ki'] * self.QgodLocal['kq'] * self.starMatrix(0)['ql'] * self.T['jl']
+    computeFluxSolverLocal = self.AplusT['ij'] <= fluxScale * self.Tinv['ki'] * (self.QgodLocal['kq'] * self.starMatrix(0)['ql'] + self.QcorrLocal['kl']) * self.T['jl']
```
Thomas-Ulrich, line 205: how do adding these zeros in the case of Godunov ( extra flops) will affect performance?

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -196,25 +203,73 @@ void seissol::initializer::initializeCellLocalMatrices( seissol::geometry::MeshR
...
+        const auto wavespeedNeighbor = material[cell].neighbor[side].getMaxWaveSpeed();
+        const auto wavespeed = std::max(wavespeedLocal, wavespeedNeighbor);
+
+        real centralFluxData[tensor::QgodLocal::size()]{};
+        real rusanovPlusData[tensor::QcorrLocal::size()]{};
+        real rusanovMinusData[tensor::QcorrNeighbor::size()]{};
+        auto centralFluxView = init::QgodLocal::view::create(centralFluxData);
+        auto rusanovPlusView = init::QcorrLocal::view::create(rusanovPlusData);
+        auto rusanovMinusView = init::QcorrNeighbor::view::create(rusanovMinusData);
+        for (size_t i = 0; i < 9; i++) {
```
Thomas-Ulrich, line 233: what about attenuation or poroelasticity?
(should you raise an error maybe?)

Thomas-Ulrich (approved):
LGTM, but would be nice with Zihua could have a look

Thomas-Ulrich (commented):
actually changed "Approve" to "Comment" because Rusanov does not give results that look similar as Godunov when running tpv5 (e.g. missing free surface reflection), while it works with Zihua's branch. so there still might be something fishy, right?
Ha wait, do this flux overprints the free-surface by chance? Probably not, but maybe those cells with at least one free-surface boundary conditions should not be modified?

generated_code/aderdg.py
davschneller: That's only done once during the initialization procedure. The actual computation is not modified; we effectively "just" exchange some matrices.

src/Initializer/CellLocalMatrices.cpp
davschneller: That's why it's more or less still a draft right now. :)

(didn't get around to modifying for those systems yet)

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -196,25 +203,78 @@ void seissol::initializer::initializeCellLocalMatrices( seissol::geometry::MeshR
...
+          rusanovPlusView(i, i) = wavespeed * 0.5;
+          rusanovMinusView(i, i) = -wavespeed * 0.5;
+        }
+
+        const auto fluxDefault = isSpecialBC(side) ?
+          modelParameters.fluxNearFault : modelParameters.flux;
+        
+        const auto flux = 
+          cellInformation[cell].faceTypes[side] == FaceType::FreeSurface ||
+          cellInformation[cell].faceTypes[side] == FaceType::FreeSurfaceGravity
```
Thomas-Ulrich, line 244: probably also absorbing?

davschneller: Done; added

Aangniu:
https://github.com/SeisSol/SeisSol/pull/1176/files#diff-554fea388baa28e948a2a1ddb8875411390684c325e972eeceb6586861590357R226

Aangniu (approved):

davschneller:
> https://github.com/SeisSol/SeisSol/pull/1176/files#diff-554fea388baa28e948a2a1ddb8875411390684c325e972eeceb6586861590357R226
> 
> This coefficient for Rusanov flux, whether to use max between cells. I suggest to use average.
> 
> I have a reference here: https://www.sam.math.ethz.ch/sam_reports/reports_final/reports2016/2016-57.pdf

By the formula (given for staggered grids there; but I'd guess it could translate back? My knowledge with those is a bit lacking still), we'd need to take the average of the material parameters then, and subsequently compute the maximum wavespeed of that? Or maybe the eigenvalue average as approximation?

Can both be implemented as well (but it might take a bit longer than today alas :) ).

PR #1177, opened 2024-10-03
Fix #973

### LINKED ISSUE #973: Sporadic CI failures at tpv5-poro
**Describe the bug**
The tpv5-poro test fails from time to time in the CI due to a segfault. If it happens, it always does so about 30 seconds into the run. Given that the successful runtime is about 2 min 30 s, and the run taking pretty exactly 2 mins, I'd suspect that it happens in one of the first iterations.

**Expected behavior**
The test should pass consistently.

**To Reproduce**
Steps to reproduce the behavior: submit a CI job several times (e.g. create any new branch, or commit to an existing one), eventually one of them will fail at tpv5-poro.

1. Which version do you use? Happens with all current versions sporadically, cf. e.g. https://gitlab.lrz.de/ci_seissol/SeisSol/-/jobs/6781361
2. Which build settings do you use? Cf. https://github.com/SeisSol/SeisSol/blob/4f89e56f2d6450e728fdc9cc4daf6563759ca07b/gitlab-cpu-pipeline.yml#L36-L78 with `poroelastic` material and `Release` configuration, precision does not matter. See here, https://github.com/SeisSol/SeisSol/blob/4f89e56f2d6450e728fdc9cc4daf6563759ca07b/gitlab-cpu-pipeline.yml#L115-L153 , choosing `5-poro`.
3. On which machine does your problem occur? SeisSol CPU CI (i.e. SCCS CI) node.
4. Provide parameter/material files. See https://github.com/SeisSol/precomputed-seissol/tree/master/tpv5-poro

Cf. e.g. https://gitlab.lrz.de/ci_seissol/SeisSol/-/jobs/6781361

Especially line 241 there (https://gitlab.lrz.de/ci_seissol/SeisSol/-/jobs/6781361#L241 )

**Additional context**
None.

davschneller:
An non-alignment bug sometimes caused segfaults due to LIBXSMM expecting aligned pointers. Hereby fixed.

### DIFF
--- a/src/Initializer/DynamicRupture.h
+++ b/src/Initializer/DynamicRupture.h
@@ -127,7 +127,7 @@ struct DynamicRupture {
     tree.addVar(waveSpeedsMinus, mask, 1, allocationModeDR(), true);
     tree.addVar(drEnergyOutput, mask, Alignment, allocationModeDR());
     tree.addVar(impAndEta, mask, 1, allocationModeDR(), true);
-    tree.addVar(impedanceMatrices, mask, 1, allocationModeDR(), true);
+    tree.addVar(impedanceMatrices, mask, Alignment, allocationModeDR(), true);
     tree.addVar(initialStressInFaultCS, mask, 1, allocationModeDR());
     tree.addVar(nucleationStressInFaultCS, mask, 1, allocationModeDR(), true);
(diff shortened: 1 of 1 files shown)


Thomas-Ulrich (approved):
(I trust you on this one)

PR #1179, opened 2024-10-04
Port more fault receiver metadata, from before dr/cpp

davschneller:


davschneller:
(note: may still need a tiny bit of refactoring)

Thomas-Ulrich (approved):
LGTM (and works)

PR #1182, opened 2024-10-16
Delta pulse friction law

NicoSchlw:
This PR adds FL 35, which imposes a delta pulse. It should be the basis for a future implementation of numerical Green's functions.


davschneller (commented):
Thanks; I've added a few small comments, mainly about the (inverse) surface area. From the technical point of view, it's a LGTM from me otherwise.

davschneller (approved):
LGTM now.

I'm wondering if the division by `timeIncrement` can be somehow done in advance; but that's something that affects more than this friction law anyways.

NicoSchlw:
Ok, thank you for the review!

PR #1183, opened 2024-10-17
Throw a warning in Cell Local Matrices if one of the fault mesh faces has a negative index and we blow up the solution on purpose

vikaskurapati:


src/Initializer/CellLocalMatrices.cpp
```diff
@@ -743,13 +746,19 @@ void seissol::initializer::initializeDynamicRuptureMatrices( seissol::geometry::
       } else {
         /// Blow up solution on purpose if used by mistake
         plusSurfaceArea = 1.e99; plusVolume = 1.0;
+  #ifndef NDEBUG
```
davschneller, line 749: Use `logDebug(rank)` instead

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -743,13 +746,19 @@ void seissol::initializer::initializeDynamicRuptureMatrices( seissol::geometry::
...
+        logWarning(rank) << "fault[meshFace].element is negative at meshFace: " << meshFace << ", on rank: " << rank <<", blowing up solution on purpose";
+  #endif    
       }
       if (fault[meshFace].neighborElement >= 0) {
         surfaceAreaAndVolume( i_meshReader, fault[meshFace].neighborElement, fault[meshFace].neighborSide, &minusSurfaceArea, &minusVolume );
         surfaceArea = minusSurfaceArea;
       } else {
         /// Blow up solution on purpose if used by mistake
         minusSurfaceArea = 1.e99; minusVolume = 1.0;
+#ifndef NDEBUG
```
davschneller, line 759: Use `logDebug(rank)` instead.

davschneller (commented):
That's _most likely_ done on purpose for all DR copy cells. I.e. it will probably only show the size of the DR copy layer in the end, if we count the messages per rank.

I appreciate the idea; but I'm not 100%ly sure if it'll help us much alas.

vikaskurapati:
> That's _most likely_ done on purpose for all DR copy cells. I.e. it will probably only show the size of the DR copy layer in the end, if we count the messages per rank.
> 
> I appreciate the idea; but I'm not 100%ly sure if it'll help us much alas.

Okay, then we don't need it and is fine, I guess. Let us just close the PR if it is not really useful.

PR #1184, opened 2024-10-17
Add the No-Fault Friction Law for the GPU

davschneller:
Ported from #1122 . Effectively, we only need to copy the traction around (in principle, we could also do that as a memcpy operation, on second thought).

(NOTE: it's different from a "true" DR-disablement—which can also be included in some subsequent PR)


vikaskurapati (approved):
LGTM

Issue #1185, opened 2024-10-17
cmake option MPI=OFF does not compile


Thomas-Ulrich:
I suggest removing the cmake MPI option.
[spack_error.txt](https://github.com/user-attachments/files/17416603/spack_error.txt)

davschneller:
Yes, it would probably make sense to hide them for now.
(in the CMake code, we could just ad hoc comment out the `option(${VAR} "desc" ${DEFAULT})` and replace it by a `set(${VAR} ${DEFAULT})`.

There are actually some more that are currently inactive, most notably `HDF5` or `LOG_LEVEL`. Maybe also `OPENMP` though that one seems to compile in some cases.

davschneller:
Cf. #1186

PR #1186, opened 2024-10-17
Hide some for-now-always-on options

davschneller:
We hide:

* `MPI` (assumed to be always on)
* `OPENMP` (assumed to be always on)
* `HDF5` (assumed to be always on)
* `LOG_LEVEL` (assumed to be `"warning"` for compatibility reasons)

Currently, as in #1185, we don't compile with `MPI=OFF`. The setting `OPENMP=OFF` _could_ maybe work; but is completely untested. Also `HDF5=OFF` would not allow any meshes to be read in at the current point in time (_maybe_ the cube generator would work as standalone).

Also, `LOG_LEVEL` is ignored as of now; only `LOG_LEVEL_MASTER` matters.

### DIFF
--- a/cmake/process_users_input.cmake
+++ b/cmake/process_users_input.cmake
@@ -1,12 +1,21 @@
-option(HDF5 "Use HDF5 library for data output" ON)
+# FIXME: forced to be on
+# option(HDF5 "Use HDF5 library for data output" ON)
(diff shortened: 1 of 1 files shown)


Thomas-Ulrich (approved):
LGTM

PR #1193, opened 2024-10-25
Some very tiny bugfixes

davschneller:
We change:

* remove a remnant of the circular streams/stream ring buffer (those are now allocated dynamically); and may be removed soon-ish from the Device module
* remove a remnant from a merge conflict which prevented a build without experimental kernels
* fix one host-device buffer usage bug


vikaskurapati (approved):
LGTM

PR #1194, opened 2024-10-29
Some more small bugfixes

davschneller:
* more host-device bugfixes
* fix `MPI_Datatype` and `MPI_Comm` not to use `nullptr` as default value (conflicting e.g. with Intel MPI).


vikaskurapati (approved):
LGTM

PR #1197, opened 2024-11-03
The acoustic model with 4 DOFs is implemented for the proxy application. 

JinwenPan:

PR #1198, opened 2024-11-06
The 4-DOF acoustic model is implemented for the proxy

JinwenPan:
- The issue of hardcoding is resolved through conditional statements and function overloading.
- The modifications now are based on the latest version of SeisSol. 
- `src/Equations/acoustic/Model` is added into `tidy.sh` and `format.sh`. 
- Unnecessary comments are removed. 
- A symlink is created for `src/Equations/acoustic/Kernels`.


.ci/format.sh
```diff
@@ -32,6 +32,8 @@ format() {
     local allowlist_file="
         src/Equations/elastic/Model/Datastructures.h
         src/Equations/elastic/Model/IntegrationData.h
+        src/Equations/acoustic/Model/Datastructures.h
+        src/Equations/acoustic/Model/IntegrationData.h
```
davschneller, line 36: It's best to add the whole folder here (see the first `allowlist`); the `AcousticSetup.h` may also join there.

If there's an issue with the prefixed variable names like `i_material` or the likes—the prefix and the underscore may go in all cases.

(or was there something blocking in that file?)

.ci/tidy.sh
```diff
@@ -33,6 +33,8 @@ format() {
     local allowlist_file="
         src/Equations/elastic/Model/Datastructures.h
         src/Equations/elastic/Model/IntegrationData.h
+        src/Equations/acoustic/Model/Datastructures.h
+        src/Equations/acoustic/Model/IntegrationData.h
```
davschneller, line 37: Cf. the comment above

generated_code/aderdg.py
```diff
@@ -125,13 +125,11 @@ def __init__(self, order, multipleSimulations, matricesDir):
     self.db.update(project2nFaceTo3m)
 
     selectVelocitySpp = np.zeros((self.numberOfQuantities(), 3))
```
davschneller, line 127: You could e.g. take `extractVelocities` here instead and transpose the resulting matrix—i.e. the `mapToVelocities` method.

generated_code/aderdg.py
```diff
@@ -180,6 +178,14 @@ def extractTractions(self):
   def mapToTractions(self):
     return self.extractTractions().T
 
+  def setBottomToIdentity(self, selectVelocitySpp):
```
davschneller, line 181: Would be better `setVelocitiesToIdentity`; but cf. comment above: you can take `extractVelocities` and transpose instead.

generated_code/aderdg.py
```diff
@@ -125,13 +125,11 @@ def __init__(self, order, multipleSimulations, matricesDir):
...
     selectVelocitySpp = np.zeros((self.numberOfQuantities(), 3))
-    selectVelocitySpp[6:9,0:3] = np.eye(3)
+    self.setBottomToIdentity(selectVelocitySpp)
     self.selectVelocity = Tensor('selectVelocity', selectVelocitySpp.shape, selectVelocitySpp, CSCMemoryLayout)
 
     self.selectTractionSpp = np.zeros((self.numberOfQuantities(), 3), dtype=bool)
-    self.selectTractionSpp[0,0] = True
-    self.selectTractionSpp[3,1] = True
-    self.selectTractionSpp[5,2] = True
+    self.setSelectTractionSppToTrue(self.selectTractionSpp)
```
davschneller, line 132: Likewise, you could take the `mapToTractions` method here for `self.selectTractionSpp`.

Most likely, you won't even need to convert to booleans afterwards.

generated_code/NodalBoundaryConditions.py
```diff
@@ -53,7 +54,10 @@ def addKernels(generator, aderdg, include_tensors, matricesDir, dynamicRuptureMe
     # To be used as Tinv in flux solver - this way we can save two rotations for
     # Dirichlet boundary, as ghost cell dofs are already rotated
     identity_rotation = np.double(aderdg.transformation_spp())
-    identity_rotation[0:9, 0:9] = np.eye(9)
+    if isinstance(aderdg, acoustic.AcousticADERDG):
```
davschneller, line 57: That works well.

However, you can even go a bit further and set `identity_rotation[0:quantities, 0:quantities] = np.eye(quantities)`, where `quantities = aderdg.numberOfQuantities()`.

(we do have some material with 13 quantities in—that way, we'd cover it here as well already)

src/Equations/acoustic/Kernels
```diff
@@ -0,0 +1 @@
+/dss/dsshome1/03/ge84lej2/fork-seissol/SeisSol/src/Equations/elastic/Kernels/
```
davschneller, line 1: Great to have it like that.

However, it'll need to be a relative symlink (i.e. something like `../../elastic/Kernels/`).

src/Equations/acoustic/Model/AcousticSetup.h
```diff
@@ -0,0 +1,134 @@
...
+         R(0,0) = std::sqrt(local.lambda * local.rho);
+         R(1,0) = -local.lambda;
+         R(0,1) = std::sqrt(neighbor.lambda * neighbor.rho);
+         R(1,1) = neighbor.lambda;
+         R(2,2) = local.lambda;
+         R(3,3) = local.lambda;
+
+         if (faceType == FaceType::FreeSurface) {
+          //  MaterialType materialtype = testIfAcoustic(local.mu) ? MaterialType::Acoustic : MaterialType::Elastic;
+          //  getTransposedFreeSurfaceGodunovState(materialtype, QgodLocal, QgodNeighbor, R);
```
davschneller, line 105: Does that method still need to be called here?

Feel free to extend the `getTransposedFreeSurfaceGodunovState` then (should be in the `Model` folder)—iirc there was already some code path for the pseudo-acoustic material there.

src/Equations/acoustic/Model/Datastructures.h
```diff
@@ -0,0 +1,109 @@
...
+namespace seissol::model {
+struct AcousticMaterial : Material {
+  static constexpr std::size_t NumQuantities = 4;
+  static constexpr std::size_t NumberPerMechanism = 0;
+  static constexpr std::size_t Mechanisms = 0;
+  static constexpr MaterialType Type = MaterialType::Acoustic;
+  static constexpr LocalSolver Solver = LocalSolver::CauchyKovalevski;
+  static inline const std::string Text = "acoustic";
+  static inline const std::array<std::string, NumQuantities> Quantities = {
+      "xx", "v1", "v2", "v3"};
```
davschneller, line 63: Doesn't the first component denote the pressure instead? Then maybe it's better to go for `p` or something like that? @vikaskurapati what would you suggest here?

src/Equations/acoustic/Model/Datastructures.h
```diff
@@ -0,0 +1,109 @@
...
+
+  double lambda;
+  double mu;
```
davschneller, line 66: That variable may still go—any explicit call to `mu` should be handleable by `getMuBar()`.

src/Initializer/ParameterDB.cpp
```diff
@@ -289,6 +290,14 @@ void MaterialParameterDB<ElasticMaterial>::addBindingPoints(
   adapter.addBindingPoint("lambda", &ElasticMaterial::lambda);
 }
 
+template <>
+void MaterialParameterDB<AcousticMaterial>::addBindingPoints(
+    easi::ArrayOfStructsAdapter<AcousticMaterial>& adapter) {
+  adapter.addBindingPoint("rho", &AcousticMaterial::rho);
+  adapter.addBindingPoint("mu", &AcousticMaterial::mu);
```
davschneller, line 297: Again, remove; as above.

src/SourceTerm/PointSource.cpp
```diff
@@ -98,8 +98,10 @@ void seissol::sourceterm::transformMomentTensor(
     }
   }
 
+#ifndef USE_ACOUSTIC
```
davschneller, line 101: Ok, may stay like that. I guess adjusting the point sources properly by material is something for the near future.

davschneller (commented):
(for reference, this is a follow-up PR to #1197 )

Thanks alot for all the changes. It's really coming together now—I think we may be merge-ready rather soon-ish; but @vikaskurapati should have the final say in that regard, probably.

I've added some additional comments. Mainly:

* you may format the whole `acoustic/Model` folder, if it's not too much work for you.
* The velocity/traction methods may be unified with the existing ones.

... and some other, tiny comments.

Also, I'm gonna enable the CI now. One more thing: you can add the acoustic equation to https://github.com/SeisSol/SeisSol/blob/91377508af4412914d707b04481f8b678b1c4044/.github/workflows/build-seissol.yml#L24 . Then, it's gonna be built in the CI with acoustic material only—and the proxy is being let run.

Thanks again for your contribution!

davschneller:
I've let the basic tests run now—it's looking good.

The poroelastic tests did not compile due to the `generated_code/SurfaceDisplacement.py`—since the velocity extraction is used there hard-coded to 3 velocity components (instead of 4; or maybe 6?).

And I'm not fully sure if it's desirable to generalize that one. (@sebwolf-de any opinions? I.e. would there be anything surface-displacement-y required for poroelastic materials; or is that a purely acoustic thing?)

Hence, I think it's fine for now to take `mapToVelocities()[:,:3]` for the `selectVelocities` for now (even though it's not the cleanest solution). Anything else can be left for someone to do in the future.

sebwolf-de:
Of course the surface displacement is crucial for every material model. How would you create seismograms/the free surface output without them?
The correct solution would be to write a function `mapToVelocities` for each material model.

davschneller:
> Of course the surface displacement is crucial for every material model. How would you create seismograms/the free surface output without them? The correct solution would be to write a function `mapToVelocities` for each material model.

That's already the case. :)

The question is more or less: is it good like it is now for the poroelastic material? (i.e. 4 velocity components; as taken for the DR)

sebwolf-de:
These lines https://github.com/SeisSol/SeisSol/blame/91377508af4412914d707b04481f8b678b1c4044/generated_code/poroelastic.py#L43-L58 were written for the DR kernels, were in the poroelastic case, also the normal component of the fluid velocity (quantity 10) and the pressure (quantity 10) are relevant.
If you want to extract the surface displacement, it is sufficient to have a look at the solid particle velocities (quantities 6,7,8). So in essence, as you proposed.

davschneller:
Ok, thanks alot! That's what was needed here.

@JinwenPan you're good with going by `mapToVelocities()[:,:3]` for now.

(if you'd want to, you could add a method `mapToSolidVelocities()` or the likes which does that automatically; but one of us can also do that in a later PR)

src/Equations/acoustic/Model/Datastructures.h
vikaskurapati: That sounds good. But we need to be careful that the "-ve" should be somehow taken care of. As in acoustic scenario, pressure p = -sigma_xx. So it may in fact be okay to just stick to xx and then use pressure whenever we post-process. Or we could call the pressure, negative pressure by default and work around that.

generated_code/acoustic.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
+##
+# @file
+# This file is part of SeisSol.
+#
+# @author Carsten Uphoff (c.uphoff AT tum.de, http://www5.in.tum.de/wiki/index.php/Carsten_Uphoff,_M.Sc.)
```
vikaskurapati, line 6: @JinwenPan, you are the author of this file. Add your name along with Carsten's in case you used his code and made modifications.

generated_code/acoustic.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# @section LICENSE
+# Copyright (c) 2016-2018, SeisSol Group
```
vikaskurapati, line 9: Change year to 2024 in the end. Or better adapt license from one of the recent license changes David made.

generated_code/acoustic.py
```diff
@@ -0,0 +1,78 @@
...
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+# @section DESCRIPTION
```
vikaskurapati, line 38: Why is the Description section empty?

generated_code/acoustic.py
```diff
@@ -0,0 +1,78 @@
...
+    }
+    self.db.update(
+      parseXMLMatrixFile('{}/star_acoustic.xml'.format(matricesDir), clones)
+    )
+
+    memoryLayoutFromFile(memLayout, self.db, clones)
+    self.kwargs = kwargs
+
+  def numberOfQuantities(self):
+    return 4
```
vikaskurapati, line 60: Can you please write a comment with the four quantities here?

src/Equations/acoustic/Model/AcousticSetup.h
```diff
@@ -0,0 +1,132 @@
+/**
+ * @file
+ * This file is part of SeisSol.
+ *
+ * @author Carsten Uphoff (c.uphoff AT tum.de, http://www5.in.tum.de/wiki/index.php/Carsten_Uphoff,_M.Sc.)
+ * @author Sebastian Wolf (wolf.sebastian AT in.tum.de, https://www5.in.tum.de/wiki/index.php/Sebastian_Wolf,_M.Sc.)
```
vikaskurapati, line 6: Please add your name too

src/Equations/acoustic/Model/AcousticSetup.h
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * @section LICENSE
+ * Copyright (c) 2015 - 2020, SeisSol Group
```
vikaskurapati, line 9: Year

src/Equations/acoustic/Model/AcousticSetup.h
```diff
@@ -0,0 +1,132 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * @section DESCRIPTION
+ **/
```
vikaskurapati, line 39: Description

src/Equations/acoustic/Model/Datastructures.h
```diff
@@ -0,0 +1,107 @@
+/**
+ * @file
+ * This file is part of SeisSol.
+ *
+ * @author Carsten Uphoff (c.uphoff AT tum.de,
+ *http://www5.in.tum.de/wiki/index.php/Carsten_Uphoff,_M.Sc.)
+ * @author Sebastian Wolf (wolf.sebastian AT in.tum.de,
+ *https://www5.in.tum.de/wiki/index.php/Sebastian_Wolf,_M.Sc.)
```
vikaskurapati, line 8: Add yourself here.

src/Equations/acoustic/Model/Datastructures.h
```diff
@@ -0,0 +1,107 @@
...
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * @section DESCRIPTION
```
vikaskurapati, line 40: Describe what this code is doing briefly

src/Equations/acoustic/Model/Datastructures.h
```diff
@@ -0,0 +1,107 @@
...
+
+namespace seissol::model {
+struct AcousticMaterial : Material {
+  static constexpr std::size_t NumQuantities = 4;
+  static constexpr std::size_t NumberPerMechanism = 0;
+  static constexpr std::size_t Mechanisms = 0;
+  static constexpr MaterialType Type = MaterialType::Acoustic;
+  static constexpr LocalSolver Solver = LocalSolver::CauchyKovalevski;
+  static inline const std::string Text = "acoustic";
+  static inline const std::array<std::string, NumQuantities> Quantities = {
```
vikaskurapati, line 62: If we are sticking to "p" here, are you using the original pressure or negative of pressure in line with the stress-velocity formulation? This needs to be clearly mentioned as otherwise it would be really confusing when we later couple the elastic-acoustic formulation.

src/Equations/acoustic/Model/Datastructures.h
```diff
@@ -0,0 +1,107 @@
...
+    stiffnessTensorView(0, 0, 0, 0) = lambda;
+    stiffnessTensorView(0, 0, 1, 1) = lambda;
+    stiffnessTensorView(0, 0, 2, 2) = lambda;
+    stiffnessTensorView(1, 1, 0, 0) = lambda;
+    stiffnessTensorView(1, 1, 1, 1) = lambda;
+    stiffnessTensorView(1, 1, 2, 2) = lambda;
+    stiffnessTensorView(2, 2, 0, 0) = lambda;
+    stiffnessTensorView(2, 2, 1, 1) = lambda;
+    stiffnessTensorView(2, 2, 2, 2) = lambda;
+  }
```
vikaskurapati, line 95: Please write in comments the stiffness Tensor you used. If it is still in line with the stress-velocity formulation used for elastic materials or if you changed it accordingly for the pressure-velocity formulations(the key difference would be a change of sign in some places). We have to make sure it is consistent and easy to understand for your successor.

src/Physics/InitialField.cpp
```diff
@@ -537,7 +537,7 @@ void seissol::physics::Ocean::evaluate(double time,
     if (std::abs(g - 9.81e-3) > 10e-15) {
       logError() << "Ocean scenario only supports g=9.81e-3 currently!";
     }
-    if (materialData.local.mu != 0.0) {
+    if (materialData.local.getMuBar() != 0.0) {
```
vikaskurapati, line 540: Here, we used a greater than 1e-15 comparison above everywhere to compare to zero. I think we can do the same to stay consistent.

src/Physics/InstantaneousTimeMirrorManager.cpp
```diff
@@ -70,8 +70,8 @@ void InstantaneousTimeMirrorManager::syncPoint(double currentTime) {
 }
 
 void InstantaneousTimeMirrorManager::updateVelocities() {
-#ifdef USE_ANISOTROPIC
-  logError() << "This feature has not been implemented for anisotropic yet";
+#if defined(USE_ANISOTROPIC) || defined(USE_ACOUSTIC)
```
vikaskurapati, line 73: ITM works directly for acoustic materials too. You can remove this part here.

vikaskurapati (changes requested):
Looks good to me mostly. We just need to confirm the stiffness matrices' you had in your mind while you were building the proxy are the same/consistent as the ones in elastic stress-velocity formulation making your dofs '-p, v1, v2, v3' or the acoustic pressure-velocity formulation making your dofs 'p, v1, v2, v3'. The calculations in proxy won't change due to this difference as far as I understand but this needs to be consistent if someone wants to continue this work. Once you confirm this and add a few comments to clarify this. I believe you are good to go.

generated_code/aderdg.py
```diff
@@ -124,15 +124,10 @@ def __init__(self, order, multipleSimulations, matricesDir):
...
-    selectVelocitySpp[6:9,0:3] = np.eye(3)
+    selectVelocitySpp = self.mapToVelocities()[:,:3]
     self.selectVelocity = Tensor('selectVelocity', selectVelocitySpp.shape, selectVelocitySpp, CSCMemoryLayout)
 
-    self.selectTractionSpp = np.zeros((self.numberOfQuantities(), 3), dtype=bool)
-    self.selectTractionSpp[0,0] = True
-    self.selectTractionSpp[3,1] = True
-    self.selectTractionSpp[5,2] = True
-
+    self.selectTractionSpp = self.mapToTractions()
```
davschneller, line 130: Apologies; I forgot that one here (and didn't expect it to fail).

Once you also set `self.mapToTractions()[:,:3]`, we should be good to go.

davschneller (approved):
Ok, the tests pass now; all good from my point of view.

Thanks again for your contribution!

vikaskurapati (approved):

PR #1199, opened 2024-11-08
fix workflow getting triggered on forks

Thomas-Ulrich:
I've updated master on my fork, and get every day a report from git  like:
>[Thomas-Ulrich/SeisSol] Run failed: github-repo-stats - master (72084d8)

This should only be run on the main repository.


davschneller (approved):
LGTM; in fact https://github.com/SeisSol/SeisSol/blob/master/.github/workflows/github-to-gitlab-push-sync.yml may also need it

PR #1203, opened 2024-11-11
Set a positive pickdt default

davschneller:
Set `pickdt=0.005` if not given (as in the doc parameter file); previously, it had been `0`, causing length error bugs when reserving receiver space.

Furthermore, we now block non-negative `pickdt` values—and interpret them as disabling the receiver recording.


Thomas-Ulrich (approved):
LGTM. Thank you!

Issue #1181, opened 2024-10-16
Inconsistent results between 16 and 64 MPI ranks


Thomas-Ulrich:
**Describe the bug**
on the master branch (5e75cf7) as well as older branches (v1.1.3), a specific setup with 16mio cells diverges when using 64 ranks, and run smoothly (final Mw~1.3) on 16 ranks.
on v1.1.2, no bug.

**Expected behavior**
same behaviour independent of number of ranks.

**To Reproduce**

seissol installed with:
```
spack install -j 40 seissol@1.1.3 convergence_order=4 dr_quad_rule=dunavant equations=elastic precision=single %intel  ^easi +python
```
on supermuc NG

setup available at /hppfs/work/pn49ha/di73yeq4/setup_francesco

(see also bashScript_NG_16ranks.sh  bashScript_NG_64ranks.sh for launch scripts).

EDIT: tested commit merging https://github.com/SeisSol/SeisSol/pull/982: error is already there

Thomas-Ulrich:
This is the fault slip on 16 ranks
![image](https://github.com/user-attachments/assets/ab0ee7b6-85ff-4329-8ba4-b89808258357)
and on 64 ranks 
![image](https://github.com/user-attachments/assets/0e4773dd-84b4-4558-bfad-4731e92c1ef9)
with v1.0.1 on superNG

Thomas-Ulrich:
The boundary conditions looks good
![image](https://github.com/user-attachments/assets/69ffc0ab-cbb0-493d-8a27-536d744273a1)

davschneller:
That is very strange. By chance—do you have a clustering file both of a 16-rank and a 64-rank run at hand?

davschneller:
Two things would come to mind as well:

* does it happen with GTS?
* if it's not too much work: does it happen before the actor model was introduced? (i.e. with tag sc21 or the likes)

davschneller:
Not 100%ly sure if it helps, but would maybe https://github.com/SeisSol/SeisSol/tree/davschneller/cluster-mpi-problem-fix-attempt help?

I'm maybe getting a bit paranoid right now; but the case of having ghost regions with different clusters may not be handled properly; e.g. they may not present in the data regions (and especially the tree storage) at all.

Maybe there's also a relation to #839 .

Thomas-Ulrich:
it seems to fix the problem!

davschneller:
That... might be just an (at least) 8-year-old bug then.

[...]

Issue #839, opened 2023-04-13
Detected Inf/NaN in free surface output or not depending on the number of nodes


Thomas-Ulrich:
**Describe the bug**
during the Texascale, we ran a scenario of the 7.8 Turkey earthquake.
When using all nodes of Frontera (8192) with 2 ranks per nodes, we get Nan when writing the first surface output:

```
Wed Apr 12 20:28:45, Info:  Writing faultoutput at time 0.
Wed Apr 12 20:28:46, Info:  Writing faultoutput at time 0. Done.
Wed Apr 12 20:28:47, Info:  Waiting for last wave field.
Wed Apr 12 20:28:47, Info:  Writing wave field at time 0.
Wed Apr 12 20:28:47, Info:  Writing wave field at time 0. Done.
Wed Apr 12 20:28:49, Info:  Writing free surface at time 0.
Wed Apr 12 20:28:49, Info:  Writing free surface at time 0. Done.
Wed Apr 12 20:28:49, Info:  Writing energy output at time 0
Wed Apr 12 20:28:51, Info:  Writing energy output at time 0 Done.
Wed Apr 12 20:28:56, Info:  Writing free surface at time 0.02.
Wed Apr 12 20:28:56, Info:  Writing free surface at time 0.02. Done.
Wed Apr 12 20:28:56, Error: Detected Inf/NaN in free surface output. Aborting. Wed Apr 12 20:28:56, Error: Detected Inf/NaN in free surface output. Aborting. Wed Apr 12 20:28:56, Error: Detected Inf/NaN in free surface output. Aborting. Wed Apr 12 20:28:56, Error: Detected Inf/NaN in free surface output. Aborting.
```
with 8000 nodes it runs without problem.

**Expected behavior**
no node dependence.

**To Reproduce**
1. Which version do you use? Provide branch and commit id.
master, 8d6e4551ea1a070b5b7d8261582f2b532e63dccd
2. Which build settings do you use? Which compiler version do you use? 
intel 
```
ADDRESS_SANITIZER_DEBUG          OFF
 ASAGI                            ON
 CMAKE_BUILD_TYPE                 Release
 CMAKE_INSTALL_PREFIX             /usr/local
 COMMTHREAD                       ON
 COVERAGE                         OFF
 DEVICE_ARCH                      none
 DEVICE_BACKEND                   none
 DR_QUAD_RULE                     dunavant
 EQUATIONS                        viscoelastic2
[...]

Issue #818, opened 2023-03-14
inf/Nan in receiver output (GPU and CPU)


francescomosconi:
**Describe the bug**
The simulation hangs with error inf/NaN in receiver output(10;10;10). 

**Expected behavior**
The simulation should run till the end

**To Reproduce**
1. Which version do you use? Provide branch and commit id.
Seissol version is 202103_Sumatra-1090-g7ae55cfa (last compilation: 2 weeks ago), master branch
2. Which build settings do you use? Which compiler version do you use? 
GPU setting with CUDA and MPI compilers. Single precision. We tested polynomial order 4, 6 and 7. 
3. On which machine does your problem occur? If on a cluster: Which modules are loaded?
Marconi100:
<img width="837" alt="Schermata 2023-03-14 alle 12 52 27" src="https://user-images.githubusercontent.com/84501864/224992777-a0bc6a8c-c2f6-4996-af8c-c01ca6fe535c.png">

4. Provide parameter/material files.
[parameters.txt](https://github.com/SeisSol/SeisSol/files/10967937/parameters.txt)

**Screenshots/Console output**
The seissol.err is provide in:
[err_rec_output.txt](https://github.com/SeisSol/SeisSol/files/10967958/err_rec_output.txt)
By visualizing the results in paraview this appere just before the error occur:
<img width="819" alt="Schermata 2023-03-11 alle 12 12 18" src="https://user-images.githubusercontent.com/84501864/224994299-a00cc6da-9066-4a12-b635-6669363f253d.png">

**Additional context**
We test different setup on the cluster from 24 to 64 nodes to avoid memory problem, but the error still occurs in a different simulation time. Even by vary the Dc the error occur in different time (we use FL=16).
The mesh was checked and we also try to decrease CFL from 0.5 to CFL=0.25.

Additionally we reduce the timesteps before the inf/Nan but we obtain this error in closing of hipSYCL (see it in seissol_out.txt and seissol_err.txt files, maybe it could be related)
from /m100/home/userexternal/rdorozhi/Downloads/gcc11.2.0/hipSYCL/src/runtime/cuda/cuda_allocator.cpp:143 @ query_pointer(): cuda_allocator: query_pointer(): pointer is unknown by backend (error code = CUDA:0)
[seissol_err.txt](https://github.com/SeisSol/SeisSol/files/10968266/seissol_err.txt)
[seissol_out.txt](https://github.com/SeisSol/SeisSol/files/10968270/seissol_out.txt)

[...]

PR #1204, opened 2024-11-13
Extend Local Cluster Array

davschneller:
Add

* all clusters that are only present in ghost halos
* all cluster gaps, to make everything contiguous afterwards (to make the actor model happy)

to the local cluster id array. So far, it seemingly only contained all copy/interior cell cluster ids (i.e. those which are really computed on the local rank).

Potential relation to #1181 and #839 .


Thomas-Ulrich (approved):
I confirm this fixes https://github.com/SeisSol/SeisSol/issues/1181

Thomas-Ulrich:
I need to give a try on https://github.com/SeisSol/SeisSol/issues/818 too

PR #1208, opened 2024-11-14
Unify variable names for receivers and wavefield

davschneller:
A suggestion. It's to source the variable names entirely from the material class. Right now, that would mean that both receivers and wavefield have (displayed here for poroelastic, since it includes elastic):

```
s_xx, s_yy, s_zz, s_xy, s_yz, s_xz, v1, v2, v3, p, v1_f, v2_f, v3_f, # (poro)elastic
ep_xx, ep_yy, ep_zz, ep_xy, ep_yz, ep_xz, eta # plasticity
```

The only significant change would be that `sigma_xx`, `xx` -> `s_xx` and so on.

### DIFF
--- a/postprocessing/validation/compare-receivers.py
+++ b/postprocessing/validation/compare-receivers.py
@@ -20,12 +20,12 @@ def velocity_norm(receiver):
 
 def stress_norm(receiver):
     return np.sqrt(
-        receiver["xx"] ** 2
-        + receiver["yy"] ** 2
-        + receiver["zz"] ** 2
-        + receiver["xy"] ** 2
-        + receiver["yz"] ** 2
-        + receiver["xz"] ** 2
+        receiver["s_xx"] ** 2
+        + receiver["s_yy"] ** 2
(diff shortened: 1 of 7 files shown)


Thomas-Ulrich:
sigma_xx or s_xx you be a bit more readable, but xx is also quite clear.

Thomas-Ulrich (approved):

davschneller:
Two thoughts:

* some prefix like `sigma_` or `s_` might help to reduce any disambiguities, if we introduce another stress-like variable (symmetric 3×3 tensor) in the future.
* On the other hand, having no prefix or `sigma_` would reduce the amount of scripts that need to be adjusted (as we'd only change either the wavefield or receiver variable names).

davschneller:
Ok, changed them to `s_xx` etc.

If it's good to you like that—then feel free to enable the merge process (otherwise, I'll change it again).

Thomas-Ulrich (approved):

Issue #1209, opened 2024-11-15
cannot compile seissol %oneapi


Thomas-Ulrich:
Trying to install seissol@1.3 %oneapi on supermuc NG:
```
spack install -j 40 seissol convergence_order=4 dr_quad_rule=dunavant equations=elastic precision=single %oneapi ^easi +python
```
Here is the beginning of the error log:
```
di73yeq4@login02:~> spack install -j 40 seissol convergence_order=4 dr_quad_rule=dunavant equations=elastic precision=single %oneapi ^easi +python
[+] /dss/lrzsys/sys/spack/release/23.1.0/opt/skylake_avx512/intel-oneapi-mpi/2021.9.0-oneapi-2e53iqg
[+] /dss/lrzsys/sys/spack/release/23.1.0/opt/skylake_avx512/libaec/1.0.6-oneapi-vppbhpy
[+] /dss/lrzsys/sys/spack/release/23.1.0/opt/skylake_avx512/pkgconf/1.8.0-oneapi-aleopfo
[+] /dss/lrzsys/sys/spack/release/23.1.0/opt/skylake_avx512/zlib/1.2.13-oneapi-iysylpx
[+] /hppfs/work/pn49ha/di73yeq4/user_spack23.1/linux-sles15-skylake_avx512/bzip2/1.0.8-oneapi-2023.1.0-twsu5wr
[+] /hppfs/work/pn49ha/di73yeq4/user_spack23.1/linux-sles15-skylake_avx512/lz4/1.9.4-oneapi-2023.1.0-2kikiiq
[+] /hppfs/work/pn49ha/di73yeq4/user_spack23.1/linux-sles15-skylake_avx512/snappy/1.1.10-oneapi-2023.1.0-lfaupjz
[+] /hppfs/work/pn49ha/di73yeq4/user_spack23.1/linux-sles15-skylake_avx512/zlib-ng/2.1.4-oneapi-2023.1.0-tbqz43r
[+] /hppfs/work/pn49ha/di73yeq4/user_spack23.1/linux-sles15-skylake_avx512/zstd/1.5.5-oneapi-2023.1.0-w3dqt74
[+] /hppfs/work/pn49ha/di73yeq4/user_spack23.1/linux-sles15-skylake_avx512/numactl/2.0.14-oneapi-2023.1.0-iicgwax
[...]

PR #1210, opened 2024-11-15
Specify iterator type explicitly

davschneller:
Related to #1209 .

The question would be: would we want to merge it—or request everyone to just use a newer compiler?

Hence draft for now; may be closed.

### DIFF
--- a/src/Common/Iterator.h
+++ b/src/Common/Iterator.h
@@ -18,6 +18,10 @@ namespace seissol::common {
 
 // TODO: remove once C++23 lands in SeisSol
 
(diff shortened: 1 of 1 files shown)


Thomas-Ulrich:
thank you. At least with spack, we can make a patch

davschneller:
If it's still in 2023.2, we can also merge that one here; the code change is pretty tiny after all.

Thomas-Ulrich:
ok, I give it a try, and if it works, I will merge.

Thomas-Ulrich (approved):
works!

PR #1214, opened 2024-11-20
Overhaul the CI (lite)

davschneller:
We overhaul the CI a bit. That is:

* add ICX and NVHPC to the CPU testing
* test building for CUDA (sm_60), ROCm (gfx906) and oneAPI (skl)

Further end-to-end tests like in the Gitlab CI are included, but our current Kubernetes setup is still a bit broken. Thus, they're still disabled here. (besides, we lack a testing environment for ROCm, and potentially also oneAPI/Intel)

Needs https://github.com/SeisSol/Device/pull/40 .


.ci/gha-gpu-nv/Dockerfile
```diff
@@ -0,0 +1,140 @@
...
+
+# ROCm/HIP
+# RUN apt-get install -y libnuma-dev gnupg2 && \
+#     wget https://repo.radeon.com/amdgpu-install/6.2/ubuntu/noble/amdgpu-install_6.2.60200-1_all.deb && \
+#     apt install -y ./amdgpu-install_6.2.60200-1_all.deb && \
+#     amdgpu-install -y --usecase=hiplibsdk --no-dkms && \
+#     rm ./amdgpu-install_6.2.60200-1_all.deb && \
+#     echo 'export PATH=$PATH:/opt/rocm/bin:/opt/rocm/rocprofiler/bin:/opt/rocm/opencl/bin' | tee -a /etc/profile.d/rocm.sh && \
+#     echo 'export HIPCC_VERBOSE=7' | tee -a /etc/profile.d/rocm.sh && \
+#     echo 'export HIP_PLATFORM=nvidia' | tee -a /etc/profile.d/rocm.sh
```
vikaskurapati, line 130: Do we still need these comments?

.github/workflows/build-seissol-gpu.yml
```diff
@@ -0,0 +1,365 @@
...
+          - name: tpv104
+            equation: elastic
+            energies: true
+            receivers: true
+            fault: true
+          #- name: tpv105
+          #  equation: elastic
+          #  energies: true
+          #  receivers: true
+          #  fault: true
```
vikaskurapati, line 212: Same with these comments

src/tests/DynamicRupture/FrictionLaws/FrictionSolverCommon.t.h
```diff
@@ -4,6 +4,9 @@
 #include "DynamicRupture/Misc.h"
 #include "tests/TestHelper.h"
 
+// TODO: re-enable?
```
vikaskurapati, line 7: What does this comment mean? We can't have dynamic rupture with Acoustic materials. So, I would say that the ifdef underneath makes sense. (Unless we move to multi-material models later, and then we need to discuss the new structure, but for now, it should be fin,e IMO)

src/tests/DynamicRupture/FrictionLaws/FrictionSolverCommon.t.h
```diff
@@ -76,7 +79,7 @@ TEST_CASE("Friction Solver Common") {
 
     // Assure that qInterpolatedPlus and qInterpolatedMinus are const.
     for (size_t o = 0; o < ConvergenceOrder; o++) {
-      for (size_t q = 0; q < 9; q++) {
+      for (size_t q = 0; q < misc::NumQuantities; q++) {
```
vikaskurapati, line 82: Is it right here to put misc::NumQuantitites even for poroelastic or visco-elastic materials?

vikaskurapati (approved):
LGTM, very minor comments about comments which could be ignored.

src/tests/DynamicRupture/FrictionLaws/FrictionSolverCommon.t.h
davschneller: Yes, thanks; that's what I needed to confirm (only had it fuzzily in my mind).

I had tried putting a `skip` flag below instead (like with the other test further below; cf. https://github.com/SeisSol/SeisSol/blob/49d2cae9175c48105c68508ed427074b728ade8c/src/tests/Model/GodunovState.t.h#L38 ); but that didn't quite work out yet; hence the define here.

src/tests/DynamicRupture/FrictionLaws/FrictionSolverCommon.t.h
```diff
@@ -61,7 +64,7 @@ TEST_CASE("Friction Solver Common") {
 
   for (size_t o = 0; o < ConvergenceOrder; o++) {
     for (size_t p = 0; p < misc::NumPaddedPoints; p++) {
-      for (size_t q = 0; q < 9; q++) {
+      for (size_t q = 0; q < misc::NumQuantities; q++) {
```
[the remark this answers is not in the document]
davschneller: Should be, yes. The underlying tensor has `NumQuantities` as shape in dimension 1—so we're essentially checking if the whole thing stays constant (which the currently implementation does for all materials).

(but, more generally speaking, I agree—sometimes it's truly not 100%ly clear whether a wild `9` is a legacy constant or specifically wanted in some case)

.github/workflows/build-seissol-gpu.yml
```diff
@@ -0,0 +1,365 @@
...
+            container: seissol/gha-gpu-nv:davschneller-gpu-image
+            runner: ubuntu-24.04
+          #- arch: sm_60
+          #  backend: cuda
+          #  syclcc: acpp
+          #  cc: nvc
+          #  cxx: nvc++
+          #  fc: nvfortran
+          #  container: seissol/gha-gpu-nvhpc:davschneller-gpu-image
+          #  runner: sccs-ci-nv-sm60
```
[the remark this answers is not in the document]
davschneller: I had to comment it. Due to AdaptiveCpp with NVHPC, it wants an NVIDIA card present iirc (or needs some more obscure config option)—hence the `runner` option is set differently here. Alas, the self-hosted runners are still unstable. :/
I'll add a comment why it's left out right now.

.ci/gha-gpu-nv/Dockerfile
davschneller: In princple... Not really, no. But maybe it's good to just keep it in for a tiny bit longer—in case someone wants to build an extended image (?).
It was in the original GPU image—but installing HIP for CUDA in addition to everything else consumes more memory than the online image builders currently offer. Of course, you could build offline—but I'd opt to go for avoiding the CUDA-via-HIP strategy (CUDA-via-CUDA, and maybe CUDA-via-OpenMP or CUDA-via-SYCL, is already enough). In the end, it doesn't really give us too much testing capabilities for the AMDGPU ROCm stack.

.github/workflows/build-seissol-gpu.yml
```diff
@@ -0,0 +1,365 @@
...
+          #- name: tpv5-visco
+          #  equation: viscoelastic2
+          #  energies: true
+          #  receivers: true
+          #  fault: true
+          #- name: tpv5-poro
+          #  equation: poroelastic
+          #  energies: true
+          #  receivers: true
+          #  fault: true
```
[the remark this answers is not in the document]
davschneller: In principle, these constitute tests that don't work on the GPU yet with `master`. I.e. we could leave them out for now—I thought it might be useful to keep them in here, so that it's more convenient to (re-)enable them once TensorForge etc. are merged.

Issue #1212, opened 2024-11-20
Problem running Northridge example on SuperMUC


cabril85:
Hello,

I am running the Northridge example, but SeisSol crashes when reading the NRF source file

The following is the failure report on the *.out file

Tue Nov 19 16:44:00, Info:  <---------------------------------------------------------> 
Tue Nov 19 16:44:00, Info:  <                      Point sources                      > 
Tue Nov 19 16:44:00, Info:  <---------------------------------------------------------> 
Tue Nov 19 16:44:00, Info:  Reading ./northridge.nrf 
Tue Nov 19 16:44:00, Info:  Finding meshIds for point sources... 
Tue Nov 19 16:44:00, Info:  Cleaning possible double occurring point sources for MPI... 

**===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   RANK 0 PID 112832 RUNNING AT i01r01c01s12
=   KILLED BY SIGNAL: 6 (Aborted)
===================================================================================**

I attach the original *.srf source file I am using
[nr6.70-s0000-h0000.txt](https://github.com/user-attachments/files/17827152/nr6.70-s0000-h0000.txt)

Thomas-Ulrich:
Could you fill in the issue template?
Without this information, it is not reproducible 

**Describe the bug**
A clear and concise description of what the bug is. Tell us what the code actually does.

**Expected behavior**
A clear and concise description of what you expected to happen. Tell us what you want the code to do.

**To Reproduce**
Steps to reproduce the behavior:
1. Which version do you use? Provide branch and commit id.
2. Which build settings do you use? Which compiler version do you use? 
3. On which machine does your problem occur? If on a cluster: Which modules are loaded?
4. Provide parameter/material files.

**Screenshots/Console output**
If you suspect a problem in the numerics/physics add a screenshot of your output.

If you encounter any errors/warnings/... during execution please provide the console output.

**Additional context**
Add any other context about the problem here.

Thomas-Ulrich:
Another idea is to run it with attached debugger:
```#mpiexec -n $SLURM_NTASKS gdb -batch -ex run -ex bt --args SeisSol_Release_dskx_4_elastic parameters.par```
and see if you get more information (because now there is no information apart from that it crashes).

**Comment by Thomas-Ulrich** [comment]
In any case, I m able to reproduce the error:
/hppfs/work/pn49ha/ra58tuy2/North_Iceland/Northridge

```
[...]

PR #1215, opened 2024-11-20
Demote the error when point sources are not found to a warning

davschneller:
Cf. the discussion in #1212 .

(though maybe it would be better to also specify which point sources were not found then)


Thomas-Ulrich:
we can close, right?

davschneller:
We can, yes. On the contrary, maybe it would be good to have an error for not-found on-fault receivers as well?

Thomas-Ulrich:
Yes it would probably make sense.

davschneller:
(however, note: it may cause spontaneous errors with receivers on MPI boundaries in some cases)

PR #1216, opened 2024-11-20
Update the Utils Submodule (and Others)

davschneller:
We update the utils module—and by extension also Device, PUML and xdmfwriter. Thus, we now don't need any more rank specifiers for `logInfo` and `logWarning` messages.

Draft until all other subpackages are merged.


Thomas-Ulrich (approved):
LGTM for the seissol part

PR #1231, opened 2024-12-31
Lint/Format the Relevant Python Files

davschneller:
We apply flake8 to the `generated-code` folder, as well as the `proxy-runners` folder in the `auto-tuning/src` directory. Executable Python files get their code moved into a `main` method. Also, we add a flake8 build check.

Note that some flake8 rules are excluded, as the made the code unnecessarily complicated. Furthermore, feel free to protest with regards to the currently-set line length. And _please_ do, if you think the current formatting is lacking.

(maybe it's better to increase the line length limit, so that the current Yateto kernel descriptions come off cleaner?)

Currently, the pre/postprocessing folders are excluded from the formatting as well—depending on if we want to move them out of the SeisSol repository at one point.


generated-code/kernels/equations/viscoelastic2.py
```diff
@@ -2,191 +2,413 @@
...
+
+        powers = [Scalar(f"power({i})") for i in range(self.order)]
+
+        derivativeTaylorExpansionEla = Add()
+        # derivativeTaylorExpansionAne = Add()
+        for d in range(0, self.order):
+            derivativeTaylorExpansionEla += powers[d] * dQ[d]["kp"]
+            # derivativeTaylorExpansionAne += powers[d] * dQane[d]['kpm']
+        derivativeTaylorExpansionElaExpr = self.I["kp"] <= derivativeTaylorExpansionEla
+        # derivativeTaylorExpansionAneExpr = self.Iane['kpm'] <= derivativeTaylorExpansionAne
```
vikaskurapati, line 360: Unnecessary comments?

vikaskurapati (approved):

PR #1236, opened 2025-01-07
Fix the Coverage Computation

davschneller:
* Fix generated kernel test generation
* Make the coverage computation work with test drivers (i.e. testing commands, like needed for NVHPC)
* Fix the coverage computation
* Remove the generated kernels and easi from the coverage calculation (feel free to protest here, if it feels wrong to someone)
* Add the coverage computation to the CI (including a downloadable overview)

(note: the coverage is currently pretty low, since we only consider the unit tests here. The full CI should test through much more parts of the code, via the end-to-end tests—it might be useful to somehow integrate those into the unit tests or the coverage computation somehow.)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,66 @@
+# SPDX-FileCopyrightText: 2024-2025 SeisSol Group
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+name: coverage
+on:
+  - push
+
+jobs:
+  coverage:
+    name: coverage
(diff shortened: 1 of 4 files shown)


vikaskurapati (commented):
Do the CI tests fail for the reason in #1235 ? I am guessing that should fix these CI tests, am I right?

davschneller:
Yes, they're fixed by #1235

vikaskurapati (approved):
LGTM as the CI tests are fixed by #1235

PR #1237, opened 2025-01-07
Normalize CMake Target Names

davschneller:
...to: kebab case (all lowercase). B/c it's easier to type. (also, some parts like `seissol-io` are already in that style)

[note: the exported target names don't change—those are set explicitly; however, we can of course do that as well]


vikaskurapati (commented):
#1235 should fix the failing CI tests, shouldn't they? I just want to confirm once before merging.

vikaskurapati (approved):
LGTM as CI tests are fixed with #1235

PR #1244, opened 2025-01-14
Fix the new IO Reader

davschneller:
* Corrected a bug with setting the chunksizes
* Add some code for ignoring padding (disabled for now)

### DIFF
--- a/src/IO/Instance/Checkpoint/CheckpointManager.cpp
+++ b/src/IO/Instance/Checkpoint/CheckpointManager.cpp
@@ -72,8 +72,8 @@ std::function<writer::Writer(const std::string&, std::size_t, double)>
 }
 
 double CheckpointManager::loadCheckpoint(const std::string& file) {
-  std::size_t storesize = 1;
-  void* datastore = std::malloc(1);
+  std::size_t storesize = 0;
+  void* datastore = nullptr;
 
   logInfo(seissol::MPI::mpi.rank()) << "Loading checkpoint...";
   logInfo(seissol::MPI::mpi.rank()) << "Checkpoint file:" << file;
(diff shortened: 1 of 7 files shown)


src/IO/Instance/Checkpoint/CheckpointManager.cpp
```diff
@@ -101,12 +122,23 @@ double CheckpointManager::loadCheckpoint(const std::string& file) {
       const std::size_t count = reader.dataCount(variable.name);
       const std::size_t currsize = count * variable.datatype->size();
       if (currsize > storesize) {
-        datastore = std::realloc(datastore, currsize);
+        std::free(datastore);
+        datastore = std::malloc(currsize);
+        if (datastore == nullptr) {
+          logError() << "Realloc failed; maybe you are reading too much (checkpoint) data?";
```
vikaskurapati, line 128: Can we differentiate whether it is checkpoint data or if some other data is causing an issue? If yes, it would be great to clarify this in the error message so the user knows which one is causing the problem and correct it.

src/IO/Instance/Checkpoint/CheckpointManager.h
```diff
@@ -51,8 +54,60 @@ class CheckpointManager {
...
+  }
+
+  template <typename S, typename T>
+  void registerTransformedData(const std::string& name,
+                               initializer::LTSTree* tree,
+                               initializer::Variable<T> var,
+                               const std::function<void(void*, const void*)>& pack,
+                               const std::function<void(void*, const void*)>& unpack) {
+    if (var.mask != initializer::LayerMask(Ghost)) {
+      logError() << "Invalid layer mask for a checkpointing variable (i.e.: NYI).";
```
vikaskurapati, line 68: Just for my curiosity, what is NYI?

vikaskurapati (approved):

src/IO/Instance/Checkpoint/CheckpointManager.h
davschneller: NYI = "Not yet implemented"

PR #1245, opened 2025-01-15
Towards Adopting the REUSE Spec

davschneller:
A.k.a. the big license header shortening. For readability.

In principle, we do:

* replace the whole BSD-3 license text by a copyright notice, the license text, and a reference to where the license files can be found. Move the copyrights also into an SPDX clause.
* move author lists to the SPDX format, and point to the AUTHORS and CITATION files. (I'd be still in favor to make that the only author reference; but I guess that won't happen—i.e. we may even want to start extending the author list with this generation's names again (?))
* normalize the header include guard names. We unify it, the header include guards now follow the format `SEISSOL_${FILEPATH-CAPITALIZED}_` . E.g. for `src/Parallel/Pin.h` , we have `SEISSOL_SRC_PARALLEL_PIN_H_`.

Feel free to want to adjust the generic reference texts.

Note that we're close, but not yet quite at the adoption of the REUSE spec (hence no CI test yet). We still need to deal with the preprocessing and postprocessing folders (i.e. either move them, or add license identifiers everywhere); as well as some of the CMake files. (those pose a particular problem; since two of them seem to be taken from GPL-licensed projects; even though they barely contain any licensing info on their own)


vikaskurapati (approved):
Looks good to me!

PR #1248, opened 2025-01-21
Always link in GKlib explicitly (if found separately)

davschneller:
Fix yet another METIS/ParMETIS+GKlib linker bug. Especially, if GKlib is linked in dynamically.


davschneller:
Solves the problem when GKlib is being linked dynamically

vikaskurapati (approved):

PR #1253, opened 2025-01-23
Reduce some numerical errors in the FVW calculation

davschneller:
Replace the `exp(x) - exp(-x)` by `sinh` again—since the former is inaccurate for `x` close to 0—and might yield negative values there. Thus breaking the subsequent `log` application.

Possibly related to #361 and #342.

### DIFF
--- a/src/DynamicRupture/FrictionLaws/FastVelocityWeakeningLaw.h
+++ b/src/DynamicRupture/FrictionLaws/FastVelocityWeakeningLaw.h
@@ -62,18 +62,18 @@ class FastVelocityWeakeningLaw
     const real steadyStateFrictionCoefficient =
         muW + (lowVelocityFriction - muW) /
                   std::pow(1.0 + misc::power<8, double>(localSlipRate / localSrW), 1.0 / 8.0);
-    // For compiling reasons we write SINH(X)=(EXP(X)-EXP(-X))/2
-    // Need double precision here, otherwise this will evaluate to infinity
+    // TODO: check again, if double precision is necessary here (earlier, there were cancellation
+    // issues)
     const real steadyStateStateVariable =
-        localA * std::log(this->drParameters->rsSr0 / localSlipRate *
-                          (std::exp(steadyStateFrictionCoefficient / localA) -
-                           std::exp(-steadyStateFrictionCoefficient / localA)));
+        localA * std::log(this->drParameters->rsSr0 / localSlipRate * 2 *
+                          std::sinh(steadyStateFrictionCoefficient / localA));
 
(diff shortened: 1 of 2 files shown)


davschneller:
(whether we can remove the enforced `double` for the GPU implementation remains open for now)

vikaskurapati:
> (whether we can remove the enforced `double` for the GPU implementation remains open for now—but maybe we indeed can)

The PR seems ready to be merged but have we tested it on one of the test cases? IMO, it makes sense to decide on the double too before merging if we can run them on a proper mesh.

davschneller:
> > (whether we can remove the enforced `double` for the GPU implementation remains open for now—but maybe we indeed can)
> 
> The PR seems ready to be merged but have we tested it on one of the test cases? IMO, it makes sense to decide on the double too before merging if we can run them on a proper mesh.

In principle yes; however I'm not super certain if there are problems by the subtractions in the previous two steps (which required `double` as a pseudo-remedy before). In principle, I'm currently lacking exactly those test cases. :)

I've mainly compared if the results on `tpv104` stay the same so far—and they seem to match by the energies.

In doubt, we can have another PR after that one here.

src/DynamicRupture/FrictionLaws/FastVelocityWeakeningLaw.h
```diff
@@ -62,18 +62,18 @@ class FastVelocityWeakeningLaw
...
                   std::pow(1.0 + misc::power<8, double>(localSlipRate / localSrW), 1.0 / 8.0);
-    // For compiling reasons we write SINH(X)=(EXP(X)-EXP(-X))/2
-    // Need double precision here, otherwise this will evaluate to infinity
+    // TODO: check again, if double precision is necessary here (earlier, there were cancellation
+    // issues)
     const real steadyStateStateVariable =
-        localA * std::log(this->drParameters->rsSr0 / localSlipRate *
-                          (std::exp(steadyStateFrictionCoefficient / localA) -
-                           std::exp(-steadyStateFrictionCoefficient / localA)));
+        localA * std::log(this->drParameters->rsSr0 / localSlipRate * 2 *
```
Thomas-Ulrich, line 68: 2.0

davschneller: That would actually enforce double precision when using single precision. Probably, a `static_cast<real>(2)` might make more sense here—my guess so far.
Though the `2` should be converted correctly already.

davschneller:
Any more objections here? I guess #342 might still need a different fix by now (that is, to somehow limit the low velocity friction to non-negative values).

Thomas-Ulrich (approved):
LGTM.

PR #1256, opened 2025-01-27
Add automatic CPU/GPU Switchpoint Detection

davschneller:
We change:

* add a small scaling test for CPU-GPU latency-throughput switching (given that we work on a unified memory system like the GH Superchip or the MI300A). For now, it only considers the wavepropagation part (the rest will need #1122), and the CPU layout may still need some optimization.
* unify the code base of the proxy and Mini SeisSol (i.e. "Mini SeisSol" becomes a mere proxy call)


vikaskurapati (approved):

vikaskurapati:
Looks good to me but the conflicts are to be resolved yet.

PR #1258, opened 2025-01-31
Remove some legacy code

davschneller:
Added in #923, it was added to mitigate some bugs on LUMI with old ROCm (early/mid 5) back then. However, those should be fixed by now. Hence, we remove it.

### DIFF
--- a/Documentation/gpus.rst
+++ b/Documentation/gpus.rst
@@ -125,7 +125,7 @@ the following will force SeisSol to allocate 1.5GB of stack GPU memory for tempo
     export DEVICE_STACK_MEM_SIZE=1.5
     mpirun -n <M x N> ./SeisSol_dsm70_cuda_* ./parameters.par
 
-The following device-specific environment variable are supported right now:
+The following device-specific environment variable is supported right now:
 
 * ``SEISSOL_PREFERRED_MPI_DATA_TRANSFER_MODE``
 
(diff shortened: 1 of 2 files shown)


Documentation/gpus.rst
```diff
@@ -125,10 +125,9 @@ the following will force SeisSol to allocate 1.5GB of stack GPU memory for tempo
     export DEVICE_STACK_MEM_SIZE=1.5
     mpirun -n <M x N> ./SeisSol_dsm70_cuda_* ./parameters.par
 
-The following device-specific environment variables are supported right now:
+The following device-specific environment variable are supported right now:
```
vikaskurapati, line 128: The following device-specific environment variable "is" supported right now:

vikaskurapati (changes requested):
One very minor grammatical change.

vikaskurapati:
It seems that the CI fails because of tinytc. I would recommend making it optional for now as we have not thoroughly tested it yet.

davschneller:
> It seems that the CI fails because of tinytc. I would recommend making it optional for now as we have not thoroughly tested it yet.

That's been done by #1262 :)

Documentation/gpus.rst
davschneller: Addressed.

vikaskurapati:
> > It seems that the CI fails because of tinytc. I would recommend making it optional for now as we have not thoroughly tested it yet.
> 
> That's been done by #1262 :)
> 
> (this branch here just isn't updated yet; but it shouldn't need to be to be merged)

Okay, approved in that case

PR #1266, opened 2025-02-04
Make SIMD-aligned block sparsity the default for non-fused simulations on the CPU with PSpaMM

davschneller:
To put forward a small discussion: to enable by default a bit more sparsity on the CPU (if PSpaMM is a codegenerator)

* NZ-FLOP/s increase. I.e. we reduce the runtime of non-fused CPU simulations by a tiny bit (given that you've installed the latest PSpaMM).
* HW-FLOP/s get reduced.


davschneller:
(I'll move the data out into a comment here; previously it's been part of the PR description)

Some preliminary example values; 0.5 s tpv5, order 6 double, four MI300A CPUs (1 rank in total):
* 3 min 27.8853 s  |  3.6176 HW-TFLOP/s (PSpaMM, all dense)
* 2 min 53.4152 s  |  2.7593 HW-TFLOP/s (block sparsity) 
From a Grace-Grace CPU (1 rank per socket):
* 3 min 48.9366 s | 3.0203 HW-TFLOP/s (dense, LIBXSMM_JIT, no prefetching)
* 2 min 17.2289 s | 2.4582 HW-TFLOP/s (block-sparse, PSpaMM+LIBXSMM_JIT, PSpaMM with prefetching)
* (PSpaMM only with sparsity performs pretty similar)

generated-code/kernels/memlayout.py
```diff
@@ -26,7 +26,8 @@ class Candidate(object):
         "equations": 2,
         "order": 3,
         "pe": 4,
-        "multipleSimulations": 5,
+        "gemmgen": 5,
+        "multipleSimulations": 6,
```
vikaskurapati, line 30: Would this change in anyway break the multiplesimulations implementations?

vikaskurapati (approved):
Looks good to me in a broad sense. But the title makes me wonder, isn't aligned sparsity already the default for fused-simulations on CPUs? This basically means that we are making it a default for every case now, isn't it? 

And why is it only possible to be done with PSpaMM? Can we also make this default with LibXSMM and use the sparsity there too?

generated-code/kernels/memlayout.py
davschneller: It wont; if there's a fitting config starting with `ms`, it'll be preferred over the `pspamm.xml` config.
(in fact, the `ms` configs could even be unified by now as well)

davschneller:
> Looks good to me in a broad sense. But the title makes me wonder, isn't aligned sparsity already the default for fused-simulations on CPUs? This basically means that we are making it a default for every case now, isn't it?

Not quite; for the fused simulations, we have elementwise sparsity. Here, we try a different approach (which wouldn't make sense)

To explain the implementation here: we basically treat sparsity as removing all operations which multiply by known zero elements; thus effectively not contributing towards the result (but sometimes still being beneficial, thanks to the SIMD registers).

Thus, since we load the B matrix (C = A@B) elementwise (we broadcast it into a SIMD register or directly inside the operation). I.e. crossing out all zero operations means that we simply avoid all zero elements. That's e.g. the case for the fused simulations, by transposing the matrix operations (normally, it'd be `kdivMT @ Q @ star`; roughly speaking—after transposing we sort of have `kdivMT` on the right side).
In contrast to that, the A matrix gets loaded contiguously into SIMD registers. Thus, we'd only have a multiply-by-zero if the whole vector register is zero (i.e. a chunk of usually 2-16 elements)—or by choosing another matrix multilplication method. At the same time, elementwise sparsity would mean that we have to unpack some values into the full SIMD registers which can be both complicated and not save too much memory accesses (probably) in the end, since A usually is globally constant in our case. Thus, we can relax the sparsity constraint from elementwise to block-sparsity, by filling out the SIMD registers which have at least one non-zero element. E.g. if our SIMD register takes two elements, and we have the row [1,2,0,0,0,3]; then we'd store it as [[1, 2], [0, 3]]. I.e. the middle two values get removed; but the fifth value is used for padding the register we'll load as a whole.

In both cases, we can preserve the code structure of the dense-by-dense matrix multiplication, but just remove all multiply-by-zero operations.

(to pun around a bit: when you take "SIMD-aligned block sparsity"; take the acronym (S.a.b.s.), voice the last "s" (sabz) and translate it, you end up with "green"—hence there's a `green.xml` config file for that very reason)

> And why is it only possible to be done with PSpaMM? Can we also make this default with LibXSMM and use the sparsity there too?

I tried doing that—but the generic sparsity implementation in LIBXSMM (unless I'm missing some implementation somewhere) is unfortunately pretty outdated (it's on AVX-level; without e.g. FMA—save for AVX512); otherwise something like PSpaMM wouldn't probably even exist. Thus, the performance was roughly equivalent, if not a bit slower.
There are sparsity methods in the LIBXSMM_JIT interface; but I haven't found a fitting one for the block sparsity model considered here. If there are, then I'd be happy to look for an implementation. (note: there _do_ exist dedicated methods for the cases we used in the fused simulations—maybe they could actually give a benefit there)

vikaskurapati:
Thanks for the detailed explanation but a few things are still not very clear to me. 
> > Looks good to me in a broad sense. But the title makes me wonder, isn't aligned sparsity already the default for fused-simulations on CPUs? This basically means that we are making it a default for every case now, isn't it?
> 
> Not quite; for the fused simulations, we have elementwise sparsity. Here, we try a different approach (which wouldn't make sense)
> 

Understood, but how about the other way around. Say could we do something like this even for fused-simulations? Or would that not make sense at all for fused-simulations?

> To explain the implementation here: we basically treat sparsity as removing all operations which multiply by known zero elements; thus effectively not contributing towards the result (but sometimes still being beneficial, thanks to the SIMD registers).
> 
> Thus, since we load the B matrix (C = A@B) elementwise (we broadcast it into a SIMD register or directly inside the operation). I.e. crossing out all zero operations means that we simply avoid all zero elements. That's e.g. the case for the fused simulations, by transposing the matrix operations (normally, it'd be `kdivMT @ Q @ star`; roughly speaking—after transposing we sort of have `kdivMT` on the right side). In contrast to that, the A matrix gets loaded contiguously into SIMD registers. Thus, we'd only have a multiply-by-zero if the whole vector register is zero (i.e. a chunk of usually 2-16 elements)—or by choosing another matrix multilplication method. At the same time, elementwise sparsity would mean that we have to unpack some values into the full SIMD registers which can be both complicated and not save too much memory accesses (probably) in the end, since A usually is globally constant in our case. Thus, we can relax the sparsity constraint from elementwise to block-sparsity, by filling out the SIMD registers which have at least one non-zero element. E.g. if our SIMD register takes two elements, and we have the row [1,2,0,0,0,3]; then we'd store it as [[1, 2], [0, 3]]. I.e. the middle two values get removed; but the fifth value is used for padding the register we'll load as a whole.
> 

This is understandable, using the block sparsity structure. Is there a working way of making this default for fused-simulations too? Or some method where we let the code generator choose which is better for that configuration?

> In both cases, we can preserve the code structure of the dense-by-dense matrix multiplication, but just remove all multiply-by-zero operations.
> 
> (to pun around a bit: when you take "SIMD-aligned block sparsity"; take the acronym (S.a.b.s.), voice the last "s" (sabz) and translate it, you end up with "green"—hence there's a `green.xml` config file for that very reason)
> 
> > And why is it only possible to be done with PSpaMM? Can we also make this default with LibXSMM and use the sparsity there too?
> 
> I tried doing that—but the generic sparsity implementation in LIBXSMM (unless I'm missing some implementation somewhere) is unfortunately pretty outdated (it's on AVX-level; without e.g. FMA—save for AVX512); otherwise something like PSpaMM wouldn't probably even exist. Thus, the performance was roughly equivalent, if not a bit slower. There are sparsity methods in the LIBXSMM_JIT interface; but I haven't found a fitting one for the block sparsity model considered here. If there are, then I'd be happy to look for an implementation. (note: there _do_ exist dedicated methods for the cases we used in the fused simulations—maybe they could actually give a benefit there)

PR #1274, opened 2025-02-07
Refactor model info printing a tiny bit

davschneller:
That is:

* print the precision built with
* print the flux used for the run
* remove the old yes/no model printing inherited from the FORTRAN code, and replace it by checking `MaterialT::Text` instead

### DIFF
--- a/src/Initializer/Parameters/ModelParameters.cpp
+++ b/src/Initializer/Parameters/ModelParameters.cpp
@@ -6,7 +6,9 @@
 // SPDX-FileContributor: Author lists in /AUTHORS and /CITATION.cff
 
 #include "ModelParameters.h"
+#include <Equations/Datastructures.h>
 #include <Initializer/Parameters/ParameterReader.h>
+#include <Model/CommonDatastructures.h>
 #include <utils/logger.h>
 
 namespace seissol::initializer::parameters {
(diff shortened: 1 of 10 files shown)


src/Initializer/Parameters/ModelParameters.cpp
```diff
@@ -59,12 +61,13 @@ ModelParameters readModelParameters(ParameterReader* baseReader) {
...
-  if constexpr (isModelViscoelastic()) {
+  constexpr auto IsViscoelastic = model::MaterialT::Type == model::MaterialType::Viscoelastic;
+
+  const auto freqCentral = reader->readIfRequired<double>("freqcentral", IsViscoelastic);
+  const auto freqRatio = reader->readIfRequired<double>("freqratio", IsViscoelastic);
+  if constexpr (IsViscoelastic) {
     if (freqRatio <= 0) {
-      logError()
-          << "The freqratio parameter must be positive---but that is currently not the case.";
+      logError() << "The freqratio parameter must be positive; but that is currently not the case.";
```
vikaskurapati, line 70: Isn't the "but that is currently not the case" redundant? We are already throwing an error and mentioning that it must be positive.

vikaskurapati (approved):

PR #1286, opened 2025-02-18
Add the new logo to the README

davschneller:
Replace the existing header by the new logo.

Looks like this: https://github.com/SeisSol/SeisSol/blob/davschneller/new-logo/README.md

(only potential downside: might feel a bit barren; without full header)


vikaskurapati:
image

It looks like this in dark mode. Do we want to continue that way or change it?

davschneller:
I see. Four thoughts:

* we could brighten up the font a bit (but I don't know by how much)
* add a non-transparent background (maybe not that preferred)
* add a small write border around the text
* combine CSS with the SVG, and give it a different, brighter color on dark backgrounds (cf. https://stackoverflow.com/questions/67187091/creating-svg-that-appears-black-in-light-mode-and-light-in-dark-mode ). I.e. it'll adjust its color dynamically.

vikaskurapati:
> I see. Four thoughts:
> 
> * we could brighten up the font a bit (but I don't know by how much)
> * add a non-transparent background (maybe not that preferred)
> * add a small write border around the text
If you meant a small white border around the text, I like this idea more.
> * combine CSS with the SVG, and give it a different, brighter color on dark backgrounds (cf. https://stackoverflow.com/questions/67187091/creating-svg-that-appears-black-in-light-mode-and-light-in-dark-mode ). I.e. it'll adjust its color dynamically.
well, if it is not too much work, this may work too.

davschneller:
Done; a border's been added. The only small change right now (which can be still adjusted) is that there's now a small, white "artifact" within the fault. Can either be kept or removed (with a tiny bit of work); whatever seems more fit.

vikaskurapati (approved):

PR #1293, opened 2025-03-01
Enforce more Godunov boundaries (when Rusanov is chosen)

davschneller:
Enforce Godunov boundaries at:

* elastic-acoustic boundaries
* outflow boundaries

(not 100%ly sure if the latter is really necessary)


fabian-kutschera:
LGTM (necessary for simulations with water layer).

fabian-kutschera (approved):

PR #1295, opened 2025-03-07
WIP: HDF5 receiver output

jrekoske:
This PR supports writing a single HDF5 output file for point-wise receiver output, instead of writing individual ASCII files. I know a few items are needed to be finished before being able to merge, so please reach out if anyone is interested to help. 

Remaining TODO includes:
- We probably want to still provide support for the ASCII output for backwards compatibility, so we need two distinct classes for ASCII and HDF5 output.
- The current code doesn't handle errors properly using SeisSol functionality.

### DIFF
--- a/src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
+++ b/src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
@@ -7,10 +7,10 @@ ParallelHdf5ReceiverWriter::ParallelHdf5ReceiverWriter(MPI_Comm comm,
                                                        const std::string& filename,
                                                        hsize_t totalReceivers,
                                                        hsize_t numVariables)
-    : comm_(comm) // <-- fixes the bug
+    : comm_(comm)
 {
 
-  dims_[0] = 1000000; // max time steps. TODO: calculate this
+  dims_[0] = H5S_UNLIMITED;
   dims_[1] = totalReceivers;
   dims_[2] = numVariables;
 
@@ -23,7 +23,16 @@ ParallelHdf5ReceiverWriter::ParallelHdf5ReceiverWriter(MPI_Comm comm,
   H5Pclose(plistId);
 
   // Create dataspace
-  filespaceId_ = H5Screate_simple(RANK, dims_, nullptr);
+  hsize_t initialDims[RANK] = {0, totalReceivers, numVariables};
+  hsize_t maxDims[RANK] = {H5S_UNLIMITED, totalReceivers, numVariables};
+  filespaceId_ = H5Screate_simple(RANK, initialDims, maxDims);
+
+  // Create a dataset creation property list for "ReceiverData".
+  hid_t dcplId = H5Pcreate(H5P_DATASET_CREATE);
+  hsize_t chunkDims[RANK] = {5000, totalReceivers, numVariables};
+
+  // 4. Enable chunking
+  H5Pset_chunk(dcplId, 3, chunkDims);
 
   // Create dataset "ReceiverData"
   dsetId_ = H5Dcreate(fileId_,
@@ -31,36 +40,198 @@ ParallelHdf5ReceiverWriter::ParallelHdf5ReceiverWriter(MPI_Comm comm,
                       H5T_NATIVE_DOUBLE,
                       filespaceId_,
                       H5P_DEFAULT,
-                      H5P_DEFAULT,
+                      dcplId,
                       H5P_DEFAULT);
+
+  H5Pclose(dcplId);
+
+  // Create dataspace for pointIds
+  hsize_t pointIdDims_[1] = {totalReceivers}; // 1D array for pointIds
+  pointIdSpaceId_ = H5Screate_simple(1, pointIdDims_, nullptr);
+  pointIdDsetId_ = H5Dcreate(fileId_,
+                             "PointIds",
+                             H5T_NATIVE_ULLONG, // Use unsigned long long for pointIds
+                             pointIdSpaceId_,
+                             H5P_DEFAULT,
+                             H5P_DEFAULT,
+                             H5P_DEFAULT);
 }
 
 void ParallelHdf5ReceiverWriter::writeChunk(hsize_t timeOffset,
-                                            hsize_t receiverOffset,
                                             hsize_t timeCount,
+                                            hsize_t receiverOffset,
                                             hsize_t localReceiverCount,
                                             const std::vector<double>& data) {
+  // Get my MPI rank for debugging
+  int myRank = -1;
+  MPI_Comm_rank(comm_, &myRank); // comm_ is presumably stored from the constructor
 
-  // Check data size
-  if (data.size() != static_cast<size_t>(timeCount * localReceiverCount * dims_[2])) {
-    throw std::runtime_error("writeChunk() data size mismatch!");
+  std::cout << "[Rank " << myRank << "] Entering writeChunk()"
+            << " with timeOffset=" << timeOffset << " timeCount=" << timeCount
+            << " receiverOffset=" << receiverOffset << " localReceiverCount=" << localReceiverCount
+            << " dataSize=" << data.size() << std::endl;
+  std::fflush(stdout); // Force immediate flush to see this output in real time
+
+  // ----------------------------------------------------
+  // 1) Extend the dataset along the unlimited dimension
+  // ----------------------------------------------------
+  std::cout << "[Rank " << myRank << "] Getting current dataset dimensions..." << std::endl;
+  std::fflush(stdout);
+
+  // Get the current size of the dataset
+  hsize_t currentDims[RANK];
+  {
+    hid_t currentSpaceId = H5Dget_space(dsetId_);
+    checkStatus(currentSpaceId, "H5Dget_space in writeChunk");
+    H5Sget_simple_extent_dims(currentSpaceId, currentDims, nullptr);
+    H5Sclose(currentSpaceId);
   }
+  std::cout << "[Rank " << myRank << "] Current dataset dims before extend: (" << currentDims[0]
+            << ", " << currentDims[1] << ", " << currentDims[2] << ")" << std::endl;
+  std::fflush(stdout);
 
-  // Define the hyperslab
+  hsize_t localExtent = timeOffset + timeCount;
+  hsize_t globalExtent = 0;
+
+  unsigned long long _localExtent = (unsigned long long)localExtent;
+  unsigned long long _globalExtent = (unsigned long long)globalExtent;
+
+  // Use MPI_UNSIGNED_LONG_LONG if 'hsize_t' is effectively 64-bit
+  // (comm_ is your MPI communicator stored in the class)
+  MPI_Allreduce(&_localExtent,  // sendbuf
+                &_globalExtent, // recvbuf
+                1,             // count
+                MPI_UNSIGNED_LONG_LONG,
+                MPI_MAX,
+                comm_);
+
+  localExtent = (hsize_t)_localExtent;
+  globalExtent = (hsize_t)_globalExtent;
+
+  // Compute new dimensions
+  hsize_t newDims[RANK];
+  newDims[0] = globalExtent;
+  newDims[1] = dims_[1]; // totalReceivers (fixed)
+  newDims[2] = dims_[2]; // numVariables   (fixed)
+
+  std::cout << "[Rank " << myRank << "] Extending dataset to dims: (" << newDims[0] << ", "
+            << newDims[1] << ", " << newDims[2] << ")" << std::endl;
+  std::fflush(stdout);
+
+  herr_t status = H5Dset_extent(dsetId_, newDims);
+  checkStatus(status, "H5Dset_extent in writeChunk");
+
+  std::cout << "[Rank " << myRank << "] Finished H5Dset_extent()" << std::endl;
+  std::fflush(stdout);
+
+  // ----------------------------------------------------
+  // 2) Re-acquire the dataset's new filespace
+  // ----------------------------------------------------
+  hid_t filespaceId = H5Dget_space(dsetId_);
+  checkStatus(filespaceId, "H5Dget_space (after set_extent) in writeChunk");
+
+  std::cout << "[Rank " << myRank << "] Re-acquired dataset filespace." << std::endl;
+  std::fflush(stdout);
+
+  // ----------------------------------------------------
+  // 3) Define the hyperslab in the file where we'll write
+  // ----------------------------------------------------
   hsize_t start[RANK] = {timeOffset, receiverOffset, 0};
   hsize_t count[RANK] = {timeCount, localReceiverCount, dims_[2]};
 
-  H5Sselect_hyperslab(filespaceId_, H5S_SELECT_SET, start, nullptr, count, nullptr);
+  std::cout << "[Rank " << myRank << "] Selecting hyperslab with start=(" << start[0] << ", "
+            << start[1] << ", " << start[2] << ") "
+            << " and count=(" << count[0] << ", " << count[1] << ", " << count[2] << ")"
+            << std::endl;
+  std::fflush(stdout);
 
-  // Create a matching memory dataspace
+  status = H5Sselect_hyperslab(filespaceId, H5S_SELECT_SET, start, nullptr, count, nullptr);
+  checkStatus(status, "H5Sselect_hyperslab in writeChunk");
+
+  std::cout << "[Rank " << myRank << "] Hyperslab selection successful." << std::endl;
+  std::fflush(stdout);
+
+  // ----------------------------------------------------
+  // 4) Create a matching memory dataspace for our data
+  // ----------------------------------------------------
   hid_t memspaceId = H5Screate_simple(RANK, count, nullptr);
+  checkStatus(memspaceId, "H5Screate_simple in writeChunk");
 
-  // Collective write
+  std::cout << "[Rank " << myRank << "] Created memory dataspace." << std::endl;
+  std::fflush(stdout);
+
+  // ----------------------------------------------------
+  // 5) Perform the write
+  // ----------------------------------------------------
+  hid_t xferPlist = H5Pcreate(H5P_DATASET_XFER);
+  checkStatus(xferPlist, "H5Pcreate(H5P_DATASET_XFER) in writeChunk");
+
+  status = H5Pset_dxpl_mpio(xferPlist, H5FD_MPIO_COLLECTIVE);
+  checkStatus(status, "H5Pset_dxpl_mpio in writeChunk");
+
+  std::cout << "[Rank " << myRank << "] About to call H5Dwrite (independent mode)..." << std::endl;
+  std::fflush(stdout);
+
+  status = H5Dwrite(dsetId_, H5T_NATIVE_DOUBLE, memspaceId, filespaceId, xferPlist, data.data());
+  checkStatus(status, "H5Dwrite in writeChunk");
+
+  std::cout << "[Rank " << myRank << "] H5Dwrite call completed successfully." << std::endl;
+  std::fflush(stdout);
+
+  // ----------------------------------------------------
+  // 6) (Optional) Flush to ensure data is fully written
+  // ----------------------------------------------------
+  std::cout << "[Rank " << myRank << "] Flushing file..." << std::endl;
+  std::fflush(stdout);
+
+  status = H5Fflush(fileId_, H5F_SCOPE_GLOBAL);
+  checkStatus(status, "H5Fflush in writeChunk");
+
+  std::cout << "[Rank " << myRank << "] Flush completed." << std::endl;
+  std::fflush(stdout);
+
+  // ----------------------------------------------------
+  // 7) Cleanup
+  // ----------------------------------------------------
+  status = H5Pclose(xferPlist);
+  checkStatus(status, "H5Pclose(xferPlist) in writeChunk");
+
+  status = H5Sclose(memspaceId);
+  checkStatus(status, "H5Sclose(memspaceId) in writeChunk");
+
+  status = H5Sclose(filespaceId);
+  checkStatus(status, "H5Sclose(filespaceId) in writeChunk");
+
+  std::cout << "[Rank " << myRank << "] Exiting writeChunk()." << std::endl;
+  std::fflush(stdout);
+}
+
+void ParallelHdf5ReceiverWriter::writePointIds(hsize_t receiverOffset,
+                                               hsize_t localReceiverCount,
+                                               const std::vector<unsigned long long>& pointIds) {
+
+  std::cout << "localReceiverCount: " << localReceiverCount
+            << ", pointIds.size(): " << pointIds.size() << std::endl;
+
+  if (pointIds.size() != localReceiverCount) {
+    throw std::runtime_error("writePointIds() size mismatch!");
+  }
+
+  // Define the hyperslab
+  hsize_t start[1] = {receiverOffset};
+  hsize_t count[1] = {localReceiverCount};
+
+  H5Sselect_hyperslab(pointIdSpaceId_, H5S_SELECT_SET, start, nullptr, count, nullptr);
+
+  // Create memory dataspace
+  hid_t memspaceId = H5Screate_simple(1, count, nullptr);
+
+  // Collective write for pointIds
   hid_t xferPlist = H5Pcreate(H5P_DATASET_XFER);
   H5Pset_dxpl_mpio(xferPlist, H5FD_MPIO_COLLECTIVE);
 
-  herr_t status =
-      H5Dwrite(dsetId_, H5T_NATIVE_DOUBLE, memspaceId, filespaceId_, xferPlist, data.data());
+  herr_t status = H5Dwrite(
+      pointIdDsetId_, H5T_NATIVE_ULLONG, memspaceId, pointIdSpaceId_, xferPlist, pointIds.data());
   checkStatus(status, "H5Dwrite");
 
   // Flush
@@ -72,9 +243,49 @@ void ParallelHdf5ReceiverWriter::writeChunk(hsize_t timeOffset,
   H5Sclose(memspaceId);
 }
 
-ParallelHdf5ReceiverWriter::~ParallelHdf5ReceiverWriter() {
-  // For example, close open HDF5 objects:
-  H5Sclose(filespaceId_);
-  H5Dclose(dsetId_);
-  H5Fclose(fileId_);
+void ParallelHdf5ReceiverWriter::writeCoordinates(std::vector<Eigen::Vector3d> points) {
+  // Convert Eigen vectors into a contiguous buffer of doubles
+  std::vector<double> coords(points.size() * 3);
+  for (size_t i = 0; i < points.size(); ++i) {
+    coords[i * 3 + 0] = points[i](0);
+    coords[i * 3 + 1] = points[i](1);
+    coords[i * 3 + 2] = points[i](2);
+  }
+
+  // Create a 2D dataspace: (nPoints x 3)
+  hsize_t dims[2];
+  dims[0] = static_cast<hsize_t>(points.size()); // number of points
+  dims[1] = 3;                                   // x,y,z
+  hid_t dataspaceId = H5Screate_simple(2, dims, nullptr);
+
+  // Create the dataset.
+  hid_t datasetId = H5Dcreate(fileId_,           // existing HDF5 file handle (from the class)
+                              "Coordinates",     // name of the new dataset
+                              H5T_NATIVE_DOUBLE, // store as double precision
+                              dataspaceId,
+                              H5P_DEFAULT, // link creation property list
+                              H5P_DEFAULT, // dataset creation property list
+                              H5P_DEFAULT  // dataset access property list
+  );
+
+  // Write the entire coords array in one shot
+  herr_t status = H5Dwrite(datasetId,
+                           H5T_NATIVE_DOUBLE,
+                           H5S_ALL, // memory dataspace = entire buffer
+                           H5S_ALL, // file dataspace = entire dataset
+                           H5P_DEFAULT,
+                           coords.data());
+  checkStatus(status, "H5Dwrite in writeCoordinates");
+
+  // Close HDF5 handles
+  H5Dclose(datasetId);
+  H5Sclose(dataspaceId);
 }
+
+ParallelHdf5ReceiverWriter::~ParallelHdf5ReceiverWriter() {
+  H5Dclose(dsetId_);         // Close the main dataset
+  H5Dclose(pointIdDsetId_);  // Close the point ID dataset
+  H5Sclose(filespaceId_);    // Close the main dataspace
+  H5Sclose(pointIdSpaceId_); // Close the point ID dataspace
+  H5Fclose(fileId_);         // Finally, close the file
+  }
--- a/src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
+++ b/src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
@@ -1,0 +1,80 @@
+// ParallelHdf5ReceiverWriter.cpp
+
+#include "ParallelHdf5ReceiverWriter.h"
+#include <iostream>
+
+ParallelHdf5ReceiverWriter::ParallelHdf5ReceiverWriter(MPI_Comm comm,
+                                                       const std::string& filename,
+                                                       hsize_t totalReceivers,
+                                                       hsize_t numVariables)
+    : comm_(comm) // <-- fixes the bug
+{
+
+  dims_[0] = 1000000; // max time steps. TODO: calculate this
+  dims_[1] = totalReceivers;
+  dims_[2] = numVariables;
+
+  // Create property list for parallel I/O
+  hid_t plistId = H5Pcreate(H5P_FILE_ACCESS);
+  H5Pset_fapl_mpio(plistId, comm_, MPI_INFO_NULL);
+
+  // Create file
+  fileId_ = H5Fcreate(filename.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, plistId);
+  H5Pclose(plistId);
+
+  // Create dataspace
+  filespaceId_ = H5Screate_simple(RANK, dims_, nullptr);
+
+  // Create dataset "ReceiverData"
+  dsetId_ = H5Dcreate(fileId_,
+                      "ReceiverData",
+                      H5T_NATIVE_DOUBLE,
+                      filespaceId_,
+                      H5P_DEFAULT,
+                      H5P_DEFAULT,
+                      H5P_DEFAULT);
+}
+
+void ParallelHdf5ReceiverWriter::writeChunk(hsize_t timeOffset,
+                                            hsize_t receiverOffset,
+                                            hsize_t timeCount,
+                                            hsize_t localReceiverCount,
+                                            const std::vector<double>& data) {
+
+  // Check data size
+  if (data.size() != static_cast<size_t>(timeCount * localReceiverCount * dims_[2])) {
+    throw std::runtime_error("writeChunk() data size mismatch!");
+  }
+
+  // Define the hyperslab
+  hsize_t start[RANK] = {timeOffset, receiverOffset, 0};
+  hsize_t count[RANK] = {timeCount, localReceiverCount, dims_[2]};
+
+  H5Sselect_hyperslab(filespaceId_, H5S_SELECT_SET, start, nullptr, count, nullptr);
+
+  // Create a matching memory dataspace
+  hid_t memspaceId = H5Screate_simple(RANK, count, nullptr);
+
+  // Collective write
+  hid_t xferPlist = H5Pcreate(H5P_DATASET_XFER);
+  H5Pset_dxpl_mpio(xferPlist, H5FD_MPIO_COLLECTIVE);
+
+  herr_t status =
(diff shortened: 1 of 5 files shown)


davschneller:
Hi, thanks for the PR. Some ad hoc comments regarding the code style (which I guess you'd have planned to address anyways):

* `std::cout` should be replaced by `logInfo()` or `logDebug()` (with an include to `<utils/logger.h>`); also the `std::endl` can then be removed as well as the `fflush`. (`std::endl` already does that for you as well)
* the `static`s outside the class declaration (in `ReceiverWriter.cpp`) should be moved. It's better to have them be tracked inside the class itself.
* try to reduce the use of C-style arrays, unless explicitly needed—`std::array` and/or `std::vector` can provide you mostly with the same functionality. The only "inconvenience" is that you'd need to use the `data()` method when converting it back to a C-style array, for e.g. H5 function args.
* make sure to apply `clang-format` and `clang-tidy` to all new files (it suffices to run those tools—but I can also enable the CI for you here).

We also provide some IO infrastructure within the `src/IO` folder; I did also add some code there to write out a Pytables file (which is Hdf5). Maybe it can be of help here as well; though that's not a must for the PR here—I can do it some time later.
(writing all receivers in parallel and integrating them into the larger IO infrastructure is a longer TODO of mine as well; but I got it pretty much shelved so far—so thanks for pushing ahead on that issue)

AliceGabriel:
Hi @davschneller, regarding your last comment -
 do you mean this: https://github.com/SeisSol/SeisSol/blob/master/src/IO/Writer/File/Hdf5Writer.cpp ?

davschneller:
> Hi @davschneller, regarding your last comment - do you mean this: https://github.com/SeisSol/SeisSol/blob/master/src/IO/Writer/File/Hdf5Writer.cpp ?

Yes, that goes into the right direction and can be used if you manage to handle the `DataSource` class (maybe via a `WriteBuffer`).
Beyond that, there's also file here: https://github.com/SeisSol/SeisSol/blob/master/src/IO/Instance/Point/TableWriter.h . If you manage to set up a periodic writer with that (that's where I did get stuck last time iirc; plus the interface needs polishing), then you can use the existing IO infrastructure, including the "asynchronous" Hdf5 writing.
But again, I wouldn't consider it required or needed for this PR—because code-wise it's probably rather involved, and I wouldn't want to burden anyone but myself with that.

(note that the whole `src/IO` folder turned out to be rather complex; that's so that it can "offload" the IO to other processes or aggregate them without "knowing" about the data—hence it has a whole infrastructure around e.g. datatyping and where to write what and how, i.e. involving also a serialization and deserialization mechanism, etc.)

AliceGabriel:
Hello, I am resurfacing this PR by @jrekoske. We need to update this pull request and merge it so Isabel Reinicke can start using it.

AliceGabriel:
@davschneller thanks for the comments on alternative implementations and synergies. good suggestions! For now we would like to merge this PR to make this specific functionality available in main, but subsequent PRs can look into other tools/optimization :)

davschneller:
For that to happen, we will at minimum need it to be updated to the latest SeisSol master, pass our linters (i.e. `clang-format` and `clang-tidy`; and pass our `pre-commit` checks), and also add the switch for the CSV output for compatibility reasons. And, more optionally, also adjust the code to look more like the rest of SeisSol. (besides extending the functionality to the on-fault receivers, probably)

I can offer to clone this PR to the main SeisSol repo and polish it up for the merge—but we'll most likely need to close here for that (I need to open a new PR from the main repository). Timeline being either this weekend or the end of next week probably.

jrekoske:
Hi @davschneller , sorry that this got delayed. I will push some changes to this PR soon with some updates and to address your comments.

davschneller:
Thanks for the updates. Some more quick notes w.r.t. the CI (if you're not working on them already):

* There were some files deleted by accident—those should stay in place. Also check the other files for missed updates; e.g. the shorter headers etc.
* it's easiest (besides clang-tidy) to apply the `pre-commit` check for our linters; that one will run a few other updates besides `clang-format` that are also mandatory for PRs right now. It can be e.g. installed via `pip`:
```
pip install pre-commit
pre-commit install
pre-commit run --all-files
```
* make sure that the headers that point to internal SeisSol files use `#include "..."` and not `#include <...>`

src/ResultWriter/ReceiverWriter.h
```diff
@@ -17,6 +17,8 @@
 #include "Monitoring/Stopwatch.h"
 
 #include <Eigen/Dense>
+#include <hdf5.h>
```
davschneller: Try to avoid the forward here if it's only about `hsize_t`. Maybe `size_t` works?

src/ResultWriter/ReceiverWriter.h
```diff
@@ -25,48 +27,93 @@ struct LocalIntegrationData;
 struct GlobalData;
 class SeisSol;
 namespace initializer::parameters {
-struct ReceiverOutputParameters;
+struct ReceiverOutputParameters; // forward declaration
 } // namespace initializer::parameters
 } // namespace seissol
 
+class ParallelHdf5ReceiverWriter;
```
davschneller, line 35: Put into a namespace; like e.g. `seissol::writer`

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -77,88 +78,20 @@ std::vector<Eigen::Vector3d> parseReceiverFile(const std::string& receiverFileNa
   return points;
 }
 
-std::string ReceiverWriter::fileName(unsigned pointId) const {
```
davschneller, line 80: Generally, SeisSol should still be able to write CSV file receivers; most likely too many workflows still depend on that (and we'd need to rewrite a lot of tooling etc.).
Therefore, it's best to keep all these methods in; and include a parameter file switch for it. (in `src/Initializer/Parameters/OutputParameters.{h|cpp}`)

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -177,6 +110,7 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
                                const LTS::Backmap& backmap,
                                const CompoundGlobalData& global) {
   std::vector<Eigen::Vector3d> points;
+  const auto rank = seissol::Mpi::mpi.rank();
```
davschneller: Do we still need the rank here?

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +178,155 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
     }
   }
+
+  // -------------------------------------------------------
+  // Now, sum up total # of receivers across ranks
+  hsize_t localCountH = static_cast<hsize_t>(localReceiverCount);
+  MPI_Allreduce(&localCountH,
+                &m_totalReceivers,
+                1,
+                MPI_UNSIGNED_LONG_LONG,
```
davschneller: Use `Mpi::castToMpiType<hsize_t>()`

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +178,155 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
+                MPI_SUM,
+                seissol::Mpi::mpi.comm());
+
+  logInfo(rank) << "Total number of receivers: " << m_totalReceivers;
+
+  // We also need the offset in the "receivers" dimension for this rank
+  MPI_Scan(&localCountH,
+           &m_localReceiverOffset,
+           1,
+           MPI_UNSIGNED_LONG_LONG,
```
davschneller: Same as above.

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +178,155 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
+                &m_totalReceivers,
+                1,
+                MPI_UNSIGNED_LONG_LONG,
+                MPI_SUM,
+                seissol::Mpi::mpi.comm());
+
+  logInfo(rank) << "Total number of receivers: " << m_totalReceivers;
+
+  // We also need the offset in the "receivers" dimension for this rank
+  MPI_Scan(&localCountH,
```
davschneller: You can also, of you initialize `m_localReceiverOffset` to 0, use `MPI_Exscan` here and avoid the offset by `m_localReceiverOffset -= localCountH;`.

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +178,155 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
+  m_localReceiverOffset -= localCountH; // so the rank's chunk starts at "offset"
+
+  // We can now open the single HDF5 file if not done yet:
+  if (m_hdf5Writer == nullptr) {
+
+    // Find the local maximum ncols
+    unsigned localNcols = 0;
+    for (auto& cluster : m_receiverClusters) {
+      if (cluster->ncols() > 0) {
+        unsigned c = cluster->ncols();
```
davschneller: `localNcols = std::max(localNcols, cluster->ncols());`

(and remove the if below)

Also, you don't need the `if (cluster->ncols() > 0)`; that's automatically included in the max.

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +178,155 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
+           MPI_UNSIGNED_LONG_LONG,
+           MPI_SUM,
+           seissol::Mpi::mpi.comm());
+  m_localReceiverOffset -= localCountH; // so the rank's chunk starts at "offset"
+
+  // We can now open the single HDF5 file if not done yet:
+  if (m_hdf5Writer == nullptr) {
+
+    // Find the local maximum ncols
+    unsigned localNcols = 0;
```
davschneller: Prefer, if possible, e.g. `uint32_t` or `size_t`.

jrekoske:
@davschneller Thanks, I believe it's now meeting the format requirements. Would you be able to take over from here? I did not add the switch for the CSV output.

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +178,155 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
+        }
+      }
+    }
+
+    // Gather the global maximum ncols
+    unsigned globalNcols = 0;
+    MPI_Allreduce(&localNcols, &globalNcols, 1, MPI_UNSIGNED, MPI_MAX, seissol::Mpi::mpi.comm());
+
+    // Now use globalNcols
+    logInfo(rank) << "Global number of columns: " << globalNcols;
```
davschneller: 1. it's `logInfo()` now. (no more `rank` needed; that was some submodule update sometime in 2025)
2. Make more specific or remove—`Global number of columns` may not be obvious to what it would mean.

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +178,155 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
+
+void ReceiverWriter::syncPoint(double /*currentTime*/) {
+
+  size_t totalNewSamples = 0;
+  size_t localReceiverCount = 0;
+
+  for (auto& cluster : m_receiverClusters) {
+    for (auto& receiver : *cluster) {
+      size_t thisReceiverSamples = receiver.output.size() / cluster->ncols();
+      if (thisReceiverSamples > totalNewSamples) {
```
davschneller: Use `max` here again

src/ResultWriter/ParallelHdf5ReceiverWriter.h
```diff
@@ -0,0 +1,57 @@
...
+  static constexpr int RANK = 3;
+
+  MPI_Comm comm_;
+  hid_t fileId_;
+  hid_t dsetId_;
+  hid_t filespaceId_;
+  hid_t pointIdSpaceId_;
+  hid_t pointIdDsetId_;
+
+  hsize_t dims_[RANK];
```
davschneller: Use `std::array`

src/ResultWriter/ParallelHdf5ReceiverWriter.h
```diff
@@ -0,0 +1,57 @@
...
+  void writePointIds(hsize_t receiverOffset,
+                     hsize_t localReceiverCount,
+                     const std::vector<hsize_t>& pointIds);
+
+  void writeCoordinates(std::vector<Eigen::Vector3d> points);
+
+  ~ParallelHdf5ReceiverWriter();
+
+  private:
+  static constexpr int RANK = 3;
```
davschneller: Write as `Rank`

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,175 @@
...
+                                                       hsize_t totalReceivers,
+                                                       hsize_t numVariables,
+                                                       hsize_t totalTimeSteps)
+    : comm_(comm), dims_{totalTimeSteps, totalReceivers, numVariables} {
+  // Create property list for parallel I/O
+  hid_t plistId = H5Pcreate(H5P_FILE_ACCESS);
+  H5Pset_fapl_mpio(plistId, comm_, MPI_INFO_NULL);
+
+  // Create file
+  fileId_ = H5Fcreate(filename.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, plistId);
```
davschneller: All of these might also throw errors maybe; so you might need a `checkStatus` pretty much everywhere.

How about an error handler like e.g. https://github.com/SeisSol/SeisSol/blob/5e03868e0f8e367dd35492ca11d1aa988f7b1b6c/src/IO/Writer/File/Hdf5Writer.cpp#L31-L39 and wrap each function? The advantage is, you don't even need to write extra messages as in `checkStatus`, as you'll be given the source code point of failure (which can be extended to even give you the function name by the macro).

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,175 @@
...
+                      "ReceiverData",
+                      H5T_NATIVE_DOUBLE,
+                      filespaceId_,
+                      H5P_DEFAULT,
+                      H5P_DEFAULT,
+                      H5P_DEFAULT);
+
+  // Create dataspace for pointIds
+  std::array<hsize_t, 1> pointIdDims = {totalReceivers};
+  pointIdSpaceId_ = H5Screate_simple(1, pointIdDims.data(), nullptr);
```
davschneller: Works. Though for a 1-D array, you can also use e.g. `&totalReceivers` .

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,175 @@
...
+  checkStatus(xferPlist, "H5Pcreate(H5P_DATASET_XFER) in writeChunk");
+
+  status = H5Pset_dxpl_mpio(xferPlist, H5FD_MPIO_COLLECTIVE);
+  checkStatus(status, "H5Pset_dxpl_mpio in writeChunk");
+
+  status = H5Dwrite(dsetId_, H5T_NATIVE_DOUBLE, memspaceId, filespaceId, xferPlist, data.data());
+  checkStatus(status, "H5Dwrite in writeChunk");
+
+  // Flush to ensure data is fully written
+  status = H5Fflush(fileId_, H5F_SCOPE_GLOBAL);
```
davschneller: Do we need the flush here in all cases? Or can we let it write over the next iteration as well?

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+  // Create memory dataspace
+  hid_t memspaceId = H5Screate_simple(1, count.data(), nullptr);
+
+  // Collective write for pointIds
+  hid_t xferPlist = H5Pcreate(H5P_DATASET_XFER);
+  H5Pset_dxpl_mpio(xferPlist, H5FD_MPIO_COLLECTIVE);
+
+  herr_t status = H5Dwrite(
+      pointIdDsetId_, H5T_NATIVE_ULLONG, memspaceId, pointIdSpaceId_, xferPlist, pointIds.data());
```
davschneller: Would be good to have `H5T_NATIVE_ULLONG` match the type of the array (`hsize_t`) by name as well.

There are also functions somewhere in SeisSol IIRC to give you the correct type here; or you choose e.g. `uint64_t` as point IDs instead.

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,175 @@
...
+  checkStatus(status, "H5Dwrite in writePointIds");
+
+  // Flush
+  status = H5Fflush(fileId_, H5F_SCOPE_GLOBAL);
```
davschneller: Same question as above—do we really need that? It feels like flushing a lot will rather cause a lower performance (empirically; though I haven't really benchmarked).

src/ResultWriter/ParallelHdf5ReceiverWriter.h
```diff
@@ -0,0 +1,57 @@
...
+  MPI_Comm comm_;
+  hid_t fileId_;
+  hid_t dsetId_;
+  hid_t filespaceId_;
+  hid_t pointIdSpaceId_;
+  hid_t pointIdDsetId_;
+
+  hsize_t dims_[RANK];
+
+  void checkStatus(herr_t status, const std::string& msg) const {
```
davschneller: Can be made either `static` and/or moved into the cpp file.

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +178,155 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
+    // Gather the global maximum ncols
+    unsigned globalNcols = 0;
+    MPI_Allreduce(&localNcols, &globalNcols, 1, MPI_UNSIGNED, MPI_MAX, seissol::Mpi::mpi.comm());
+
+    // Now use globalNcols
+    logInfo(rank) << "Global number of columns: " << globalNcols;
+
+    // Create the HDF5 writer
+    // Compute total number of time steps from simulation end time and sampling interval
+    hsize_t totalTimeSteps = static_cast<hsize_t>(std::ceil(m_endTime / m_samplingInterval));
```
davschneller: Just to be on the safe side: have you tested with a lot of other output functionality enabled; especially at different output times?
There was a bug of going out of capacity with the on-fault receivers in a similar scenario; though they operate a little different.

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +178,155 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
+}
+
+// --------------------------------------------------------------------------
+
+void ReceiverWriter::syncPoint(double /*currentTime*/) {
+
+  size_t totalNewSamples = 0;
+  size_t localReceiverCount = 0;
+
+  for (auto& cluster : m_receiverClusters) {
```
davschneller: `const`?

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +178,155 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
+    for (auto& receiver : *cluster) {
```
davschneller: `const` ?

src/ResultWriter/ReceiverWriter.h
```diff
@@ -25,48 +27,93 @@ struct LocalIntegrationData;
...
+  /**
+   * \brief Registers receivers by reading their positions and mapping them to
+   *        mesh cells, then (optionally) creates the single parallel HDF5 file.
+   */
   void addPoints(const seissol::geometry::MeshReader& mesh,
                  const LTS::Backmap& backmap,
                  const CompoundGlobalData& global);
 
+  /**
+   * \brief Returns the ReceiverCluster for a given cluster ID and layer type.
```
davschneller: Layer type not anymore; that was changed by IIRC #1411 .

src/ResultWriter/ReceiverWriter.h
```diff
@@ -25,48 +27,93 @@ struct LocalIntegrationData;
...
+class ParallelHdf5ReceiverWriter;
+
 namespace seissol::writer {
+// Forward declarations for helper functions
 Eigen::Vector3d parseReceiverLine(const std::string& line);
 std::vector<Eigen::Vector3d> parseReceiverFile(const std::string& receiverFileName);
 
+/**
+ * \brief Writes out receiver data in a single parallel HDF5 file
+ * instead of multiple ASCII files.
```
davschneller: Maybe (though the CSV files should come back), there doesn't need to be many references to the "old" format if you plan to remove it—or have your own classes for that.

src/ResultWriter/ReceiverWriter.h
```diff
@@ -25,48 +27,93 @@ struct LocalIntegrationData;
...
+ */
 class ReceiverWriter : public seissol::Module {
   private:
   seissol::SeisSol& seissolInstance;
 
   public:
   explicit ReceiverWriter(seissol::SeisSol& seissolInstance) : seissolInstance(seissolInstance) {}
 
+  /**
+   * \brief Initializes internal data.
```
davschneller: It does a bit more; as without this function, the receivers won't be recorded at all—and it's called from somewhere else.

src/ResultWriter/ReceiverWriter.h
```diff
@@ -25,48 +27,93 @@ struct LocalIntegrationData;
...
+   *
+   * @param fileNamePrefix Prefix for the output file name.
```
davschneller: Including also the folder it's stored in.

src/ResultWriter/ReceiverWriter.h
```diff
@@ -25,48 +27,93 @@ struct LocalIntegrationData;
...
+  /**
+   * \brief Returns the ReceiverCluster for a given cluster ID and layer type.
+   */
   kernels::ReceiverCluster* receiverCluster(std::size_t id);
+
   //
   // Hooks
   //
-  void syncPoint(double /*currentTime*/) override;
+  /// Called at each synchronization point (i.e., sampling) to flush data.
```
davschneller: Not necessarily sampling; that happens much more frequently usually (at `pickdt` vs `receiveroutputinterval`).

src/ResultWriter/ReceiverWriter.h
```diff
@@ -25,48 +27,93 @@ struct LocalIntegrationData;
...
+  /// Current time offset for HDF5 writes
+  hsize_t m_nextTimeOffset{0};
+
+  /// Total number of receivers across all ranks
+  hsize_t m_totalReceivers{0};
+
+  /// This rank's offset in the receiver dimension
+  hsize_t m_localReceiverOffset{0};
+
+  /// Stopwatch for timing I/O
```
davschneller: Only for the receiver IO; we also have other IO components which may (or may not) have a timer on their own

src/ResultWriter/ReceiverWriter.h
```diff
@@ -25,48 +27,93 @@ struct LocalIntegrationData;
...
 namespace initializer::parameters {
-struct ReceiverOutputParameters;
+struct ReceiverOutputParameters; // forward declaration
 } // namespace initializer::parameters
 } // namespace seissol
 
+class ParallelHdf5ReceiverWriter;
+
 namespace seissol::writer {
+// Forward declarations for helper functions
```
davschneller: Those aren't really forward declarations; IIRC they're only out here because they are needed somewhere else in some completely different file (if not, we could make them static and remove the declarations here)

src/ResultWriter/ReceiverWriter.h
```diff
@@ -25,48 +27,93 @@ struct LocalIntegrationData;
 struct GlobalData;
 class SeisSol;
 namespace initializer::parameters {
-struct ReceiverOutputParameters;
+struct ReceiverOutputParameters; // forward declaration
```
davschneller: Maybe the comment can be removed; that should normally be clear (or at least I don't know of any other way how to interpret such declarations

davschneller (changes requested):
Thanks for the PR. I've attached a first round of issues/questions regarding the code.

Main thing is that the "old" CSV receiver writer should stay—as that would probably topple many current workflows which we wouldn't want to have. I.e. either have the `ReceiverWriter` class do both, or spin out the Hdf5 Receiver writer into its own class and reference it in the `src/SeisSol.h` and `src/Initializer/InitProcedure/InitIO.cpp` .

I was thinking (and I'm not sure) if it might be useful to squash some commits; especially around removing the `.github/workflows` files; so it seems like it never happened (however, rebasing around merges might be tricky...).

davschneller:
> @davschneller Thanks, I believe it's now meeting the format requirements. Would you be able to take over from here? I did not add the switch for the CSV output.

Hi @jrekoske ; sorry—I've only seen this reply after submitting my review.
I can offer you to take care of the adaption to SeisSol (i.e. keep the CSV/ASCII writer in; and add a switch to the parameters); if your advisor is ok with that. I'll probably open a pull request at your fork for that—or I'll need to "port over" your branch to the SeisSol main repository and continue working there (which will entail a new PR opened probably by me; or we add you to the main SeisSol repository).
For the other issues I've remarked; I think they should be manageable—let me know if you have any questions about or issues with them.

src/ResultWriter/ReceiverWriter.h
jrekoske: `hsize_t` is only used for counters in this class so would it make sense to use `size_t` here and then cast the counters later to `hsize_t`?

src/ResultWriter/ReceiverWriter.cpp
jrekoske: It's not used here so I will remove

src/Initializer/Parameters/OutputParameters.cpp
```diff
@@ -144,6 +144,14 @@ ReceiverOutputParameters readReceiverParameters(ParameterReader* baseReader) {
   auto enabled = reader->readWithDefault("receiveroutput", true);
   warnIntervalAndDisable(enabled, interval, "receiveroutput", "receiveroutputinterval");
 
+  const auto format = reader->readWithDefaultStringEnum<ReceiverOutputFormat>(
+      "receiverformat",
+      "hdf5",
```
davschneller: Are we sure we want to make the `hdf5` reader the default already? That might break many current workflows; and probably most are "ok" with the CSV files (as they are usually deemed more accessible than Hdf5).

Maybe something to put up for discussion e.g. in the SeisSol meeting.

docs/off-fault-receivers.rst
```diff
@@ -14,17 +14,22 @@ Off fault receivers
...
 .. code-block:: Fortran
 
   &Output
+  receiverFormat = 'hdf5' ! Can be 'hdf5' or 'ascii'
   pickdt = 0.01 ! Pickpoint Sampling
   pickDtType = 1 ! Pickpoint Type
   RFileName = 'receivers.dat' ! Record Points in extra file
   /
 
+If `receiverFormat = 'ascii'`, each receiver trace is written to a separate `.dat` file. If `receiverFormat = 'hdf5'` (the default),
```
davschneller: Maybe `csv` instead of `ascii` ; as `ascii` could mean various things; including e.g. JSON, YAML, XML, or something else.

docs/off-fault-receivers.rst
```diff
@@ -14,17 +14,22 @@ Off fault receivers
...
+all receiver traces are combined into a single HDF5 output file (`_receivers.h5`). The HDF5 file includes attributes
```
davschneller: May be just myself, but maybe go more for `-receivers.h5` ? :) (the `-` is easier to type than the `_`)

docs/off-fault-receivers.rst
```diff
@@ -14,17 +14,22 @@ Off fault receivers
...
+(`DimNames` and `VariableNames`) that define the array dimensions (Time, Receivers, Variables) and output quantities.
```
davschneller, line 31: One more generic question: does the Hdf5 output follow any given pre-defined format; like e.g. PyTables? (or pandas) That might be useful for later interoperability with e.g. other tools (as the format this PR will introduce will probably stay).

src/Initializer/Parameters/OutputParameters.h
```diff
@@ -66,8 +66,11 @@ struct PickpointParameters {
   bool collectiveio{false};
 };
 
+enum class ReceiverOutputFormat { Ascii, Hdf5 };
```
davschneller: Cf. the comment for `Csv` vs `Ascii` above

jrekoske:
I didn't realize that Copilot was going to start a review so sorry about that -- but it caught a few things that I messed up. I've done my best to address your comments, and it's ready for another review from my perspective @davschneller . 

I have added back the ASCII writer and I added an option `ReceiverFormat` in the parameter file (feel free to rename this). I also added some documentation about the HDF5 file that is produced with this PR.

src/ResultWriter/ReceiverWriter.h
```diff
@@ -10,63 +10,111 @@
...
+  /// Additional derived quantities (e.g., rotation, strain)
   std::vector<std::shared_ptr<kernels::DerivedReceiverQuantity>> derivedQuantities;
+
   std::vector<std::shared_ptr<kernels::ReceiverCluster>> m_receiverClusters;
+
+  /// Parallel HDF5 writer for receiver data
+  std::unique_ptr<ParallelHdf5ReceiverWriter> m_hdf5Writer;
+
+  /// Current time offset for HDF5 writes
+  std::size_t m_nextTimeOffset{0};
```
davschneller: FYI; there's a PR (#1481) to switch all of these member variables to the notation you're already using in `ParallelHdf5ReceiverWriter.h` ; i.e. e.g. `nextTimeOffset_` . You can already use them here, if you want (`clang-tidy` won't complain about that). If not, then I'll do it when updating #1481.

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -238,24 +229,213 @@ void ReceiverWriter::addPoints(const seissol::geometry::MeshReader& mesh,
...
+                    MPI_MAX,
+                    seissol::Mpi::mpi.comm());
+
+      m_hdf5Writer =
+          std::make_unique<ParallelHdf5ReceiverWriter>(seissol::Mpi::mpi.comm(),
+                                                       hdf5FileName(m_fileNamePrefix),
+                                                       static_cast<hsize_t>(m_totalReceivers),
+                                                       static_cast<hsize_t>(globalNcols));
+
+      // Construct variable names and write to HDF5 attribute
```
davschneller: Careful: the `writeHeaders` code above uses an outdated version of the receiver name finding. Maybe separate out the receiver header determination (i.e. getting `fullNames`) into a method on its own? And then just call it in `writeHeaders` and here.

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,249 @@
...
+  // or write collectively. HDF5 metadata operations generally need to be collective or
+  // safely handled. Since `fileId_` is opened with MPI-IO, H5Acreate must be called collectively.
+  hid_t attr = _eh(H5Acreate(dsetId_, "VariableNames", atype, space, H5P_DEFAULT, H5P_DEFAULT));
+
+  _eh(H5Awrite(attr, atype, combinedExt.c_str()));
+
+  _eh(H5Aclose(attr));
+
+  // Add DimNames indicating standard shape dimensions: (time_steps, receiver_points, variables)
+  std::string dimNames = "time_steps,receiver_points,variables";
```
davschneller: Prefer array of strings of length 3

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+void ParallelHdf5ReceiverWriter::writeVariableNames(const std::vector<std::string>& names) {
+  int rank;
+  MPI_Comm_rank(comm_, &rank);
+
+  // Combine names into a single comma-separated string
+  std::string combinedExt;
+  for (size_t i = 0; i < names.size(); ++i) {
+    if (i > 0)
+      combinedExt += ",";
```
davschneller: Make array of strings. We might now know if for some really obscure reason, there's a comma in some name some day. And then this procedure might fail.

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+    hid_t xferPlist = _eh(H5Pcreate(H5P_DATASET_XFER));
+    _eh(H5Pset_dxpl_mpio(xferPlist, H5FD_MPIO_COLLECTIVE));
+
+    _eh(H5Dwrite(dsetId_, H5T_NATIVE_DOUBLE, memspaceId, filespaceId, xferPlist, data.data()));
+
+    _eh(H5Pclose(xferPlist));
+    _eh(H5Sclose(memspaceId));
+    _eh(H5Sclose(filespaceId));
+    return;
```
davschneller: This may be a bit more by my personal opinion, but I'm not really a fan of the premature returns for the sake of code clarity; see here on why: https://www.fluentcpp.com/2018/08/24/how-to-design-early-returns-in-c-based-on-procedural-programming/
(tldr: if you look at some long function near the end, you can't be sure anymore if that place is going to get executed or not due to the control flow exiting at some much earlier place; plus, e.g. in this case, you'll need to close the necessary data fields in two places instead of closing a longer `if`)
I.e. the `else` block may be clearer (contrary to what e.g. LLVM normally recommends).

IMO the only real place where premature returns are justified is plain error handling (e.g. even Rust has respective short-hand notations for that; and that's usually a more single-return-ish language)—but not really alternative code paths; that makes the total code unfortunately look a bit to goto-y.

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  // Define the hyperslabs in the file where we'll write scattered across receivers
+  _eh(H5Sselect_none(filespaceId));
```
davschneller: Good point. I'll leave a short remark here (just in case it'll become relevant at some point; no change needed here): the `H5Sselect_none` did, to my knowledge, not really work in some cases; there you'd need to also use a null space object. Though I'm not certain whether that is relevant here.
(it's e.g. used in the `src/IO/Writer/File/Hdf5Writer.cpp` class; but that might need to go for multiple rounds)

src/ResultWriter/ReceiverWriter.cpp
```diff
@@ -43,11 +45,16 @@
 
 namespace seissol::writer {
 
-Eigen::Vector3d parseReceiverLine(const std::string& line) {
+ReceiverWriter::ReceiverWriter(seissol::SeisSol& seissolInstance)
+    : seissolInstance(seissolInstance) {}
+
+ReceiverWriter::~ReceiverWriter() = default;
+
+static Eigen::Vector3d parseReceiverLine(const std::string& line) {
```
davschneller: Our `clang-tidy` might complain here (for the `static`); maybe move to an empty `namespace` instead?

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,261 @@
...
+#include <array>
+#include <utils/logger.h>
+
+namespace {
+#define _eh(x) _ehh((x), __FILE__, __LINE__)
+
+hid_t _ehh(hid_t data, const char* file, int line) {
+  if (data < 0) {
+    H5Eprint(H5Eget_current_stack(), stdout);
+    throw std::runtime_error("HDF5 error at " + std::string(file) + ":" + std::to_string(line));
```
davschneller: Make sure to either additionally or instead call `logError()` in SeisSol; that makes the abort look a bit better (and it's the de-facto default error mechanism in SeisSol right now).

src/ResultWriter/ParallelHdf5ReceiverWriter.cpp
```diff
@@ -0,0 +1,261 @@
...
+    _eh(H5Dwrite(dsetId_, H5T_NATIVE_DOUBLE, memspaceId, filespaceId, xferPlist, data.data()));
+
+    _eh(H5Pclose(xferPlist));
+    _eh(H5Sclose(memspaceId));
+  } else {
+    // Define the hyperslabs in the file where we'll write scattered across receivers
+    _eh(H5Sselect_none(filespaceId));
+    for (size_t r = 0; r < localReceiverCount; ++r) {
+      if (pointIds[r] >= dims_[1]) {
+        throw std::runtime_error(
```
davschneller: Same as the comment to the other `throw`

jrekoske:
Hi @davschneller, it's all good to go from my side now. Thanks for the additional comments, I have addressed these and should be good to trigger CI now.

davschneller:
Thanks for the changes. I've run it through the CI; so all that's left is to make the tests pass:

* you'll need to re-forward the `parseReceiverLine` etc. (sorry, I've also not seen that IIRC); that one is required in the tests
* update the branch; in particular, the `cleanDoubles` method became obsolete (and doesn't exist anymore)
* fix the `clang-tidy` issues; which can be preferrably done by running it again after all the previous changes are done

And that should be it.

PR #1302, opened 2025-03-13
Update the DR Update Time Formulation to Match the FORTRAN Code

davschneller:
A potential small porting bug: the simulation-in-iteration time was given by `baseUpdateTime + deltaT[currentIndex]`; however it might better be `baseUpdateTime + sum(deltaT[:currentIndex+1])` by the FORTRAN code.

Compare the code snippet here: https://github.com/SeisSol/SeisSol/blob/eadae61c2432b89e528579944c3d3c60a7a67ba9/src/Physics/Evaluate_friction_law.f90#L420-L433 . Note that `tn = tn + deltat(itimegp)`; i.e. `tn` accumulates the time value.

Affects everything a bit with nucleation enabled, and LSW doubly. (it does unfortunately not fix the current tpv16 test failures)


Thomas-Ulrich:
as far as I remember, nTimeGP was set to 1 in the fortran code because it was assumed that the time loop was useless and impacting performance. but Sebastian and Ravil decided to put it back as the overhead was not important.

Thomas-Ulrich (approved):

PR #1304, opened 2025-03-15
Fix the energy output

davschneller:
The energy output and CI verification broke by #385. Thus, we fix it here.

Besides, we merge some MPI calls, and enable the energy output for all equations. (and print a warning if the energy is probably computed inaccurately)

### DIFF
--- a/src/ResultWriter/EnergyOutput.cpp
+++ b/src/ResultWriter/EnergyOutput.cpp
@@ -462,10 +462,10 @@ void EnergyOutput::computeVolumeEnergies() {
         const double curMomentumY = rho * v;
         const double curMomentumZ = rho * w;
 
-        if (std::abs(material.local.mu) < 10e-14) {
+        if (std::abs(material.local.getMuBar()) < 10e-14) {
           // Acoustic
           constexpr int PIdx = 0;
-          const auto k = material.local.lambda;
+          const auto k = material.local.getLambdaBar();
           const auto p = numSub(qp, PIdx);
           const double curAcousticEnergy = (p * p) / (2 * k);
           totalAcousticEnergyLocal += curWeight * curAcousticEnergy;
@@ -484,8 +484,8 @@ void EnergyOutput::computeVolumeEnergies() {
 
           auto getStress = [&](int i, int j) { return numSub(qp, getStressIndex(i, j)); };
 
-          const auto lambda = material.local.lambda;
-          const auto mu = material.local.mu;
+          const auto lambda = material.local.getLambdaBar();
+          const auto mu = material.local.getMuBar();
           const auto sumUniaxialStresses = getStress(0, 0) + getStress(1, 1) + getStress(2, 2);
           auto computeStrain = [&](int i, int j) {
             double strain = 0.0;
@@ -726,7 +726,9 @@ void EnergyOutput::checkAbortCriterion(
   }
 }
 
-void EnergyOutput::writeHeader() { out << "time,variable,simulation_index,measurement" << std::endl; }
+void EnergyOutput::writeHeader() {
+  out << "time,variable,simulation_index,measurement" << std::endl;
+}
 
 void EnergyOutput::writeEnergies(double time) {
   for (size_t sim = 0; sim < multisim::NumSimulations; sim++) {
@@ -734,8 +736,8 @@ void EnergyOutput::writeEnergies(double time) {
     if (shouldComputeVolumeEnergies()) {
       out << time << ",gravitational_energy," << fusedSuffix << ","
           << energiesStorage.gravitationalEnergy(sim) << "\n"
-          << time << ",acoustic_energy," << fusedSuffix << "," << energiesStorage.acousticEnergy(sim)
-          << "\n"
+          << time << ",acoustic_energy," << fusedSuffix << ","
+          << energiesStorage.acousticEnergy(sim) << "\n"
           << time << ",acoustic_kinetic_energy," << fusedSuffix << ","
           << energiesStorage.acousticKineticEnergy(sim) << "\n"
           << time << ",elastic_energy," << fusedSuffix << "," << energiesStorage.elasticEnergy(sim)
--- a/src/ResultWriter/EnergyOutput.cpp
+++ b/src/ResultWriter/EnergyOutput.cpp
@@ -579,19 +579,12 @@ void EnergyOutput::reduceEnergies() {
 #ifdef USE_MPI
   const auto rank = MPI::mpi.rank();
   const auto& comm = MPI::mpi.comm();
-
-  const auto count = static_cast<int>(energiesStorage.energies.size());
-  if (rank == 0) {
-    MPI_Reduce(MPI_IN_PLACE, energiesStorage.energies.data(), count, MPI_DOUBLE, MPI_SUM, 0, comm);
-  } else {
-    MPI_Reduce(energiesStorage.energies.data(),
-               energiesStorage.energies.data(),
-               count,
-               MPI_DOUBLE,
-               MPI_SUM,
-               0,
-               comm);
-  }
+  MPI_Allreduce(MPI_IN_PLACE,
+                energiesStorage.energies.data(),
+                static_cast<int>(energiesStorage.energies.size()),
+                MPI_DOUBLE,
+                MPI_SUM,
+                comm);
 #endif
 }
 
@@ -599,20 +592,12 @@ void EnergyOutput::reduceMinTimeSinceSlipRateBelowThreshold() {
 #ifdef USE_MPI
   const auto rank = MPI::mpi.rank();
   const auto& comm = MPI::mpi.comm();
-  for (size_t sim = 0; sim < multisim::NumSimulations; sim++) {
-    if (rank == 0) {
-      MPI_Reduce(
-          MPI_IN_PLACE, &minTimeSinceSlipRateBelowThreshold[sim], 1, MPI_C_REAL, MPI_MIN, 0, comm);
-    } else {
-      MPI_Reduce(&minTimeSinceSlipRateBelowThreshold[sim],
-                 &minTimeSinceSlipRateBelowThreshold[sim],
-                 1,
-                 MPI_C_REAL,
-                 MPI_MIN,
-                 0,
-                 comm);
-    }
-  }
+  MPI_Allreduce(MPI_IN_PLACE,
+                minTimeSinceSlipRateBelowThreshold.data(),
+                static_cast<int>(minTimeSinceSlipRateBelowThreshold.size()),
+                MPI_C_REAL,
+                MPI_MIN,
+                comm);
 #endif
 }
 
@@ -706,8 +691,9 @@ void EnergyOutput::printEnergies() {
   }
 }
 
-void EnergyOutput::checkAbortCriterion(const real (&timeSinceThreshold)[multisim::NumSimulations],
-                                       const std::string& prefixMessage) {
+void EnergyOutput::checkAbortCriterion(
+    const std::array<real, multisim::NumSimulations>& timeSinceThreshold,
+    const std::string& prefixMessage) {
   const auto rank = MPI::mpi.rank();
   bool abort = true;
   for (size_t sim = 0; sim < multisim::NumSimulations; sim++) {
--- a/src/ResultWriter/EnergyOutput.cpp
+++ b/src/ResultWriter/EnergyOutput.cpp
@@ -12,6 +12,7 @@
 #include "Parallel/MPI.h"
 #include "SeisSol.h"
 #include <Common/Constants.h>
+#include <Equations/Datastructures.h>
 #include <Geometry/MeshDefinition.h>
 #include <Geometry/MeshTools.h>
 #include <Initializer/BasicTypedefs.h>
@@ -415,7 +416,6 @@ void EnergyOutput::computeVolumeEnergies() {
     for (std::size_t elementId = 0; elementId < elements.size(); ++elementId) {
       const real volume = MeshTools::volume(elements[elementId], vertices);
       const CellMaterialData& material = ltsLut->lookup(lts->material, elementId);
-#if defined(USE_ELASTIC) || defined(USE_VISCOELASTIC2)
       auto& cellInformation = ltsLut->lookup(lts->cellInformation, elementId);
       auto& faceDisplacements = ltsLut->lookup(lts->faceDisplacements, elementId);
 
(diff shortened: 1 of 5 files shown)


src/ResultWriter/EnergyOutput.cpp
```diff
@@ -622,6 +608,11 @@ void EnergyOutput::printEnergies() {
       seissolInstance.getSeisSolParameters().output.energyParameters.terminalPrecision;
 
   const auto shouldPrint = [](double thresholdValue) { return std::abs(thresholdValue) > 1.e-20; };
+  if (model::MaterialT::Type != model::MaterialType::Elastic &&
+      model::MaterialT::Type != model::MaterialType::Viscoelastic &&
+      model::MaterialT::Type != model::MaterialType::Acoustic && shouldComputeVolumeEnergies()) {
+    logInfo() << "WARNING: the volume energies printed here may be inaccurate.";
```
vikaskurapati, line 614: How about using LogWarning() directly instead?

vikaskurapati (changes requested):

src/ResultWriter/EnergyOutput.cpp
davschneller: Yup; thought about that. However, the de facto main difference between a warning and an info message is the output stream they go to:

* warnings (and errors) go to stderr
* infos (and debug messages, if enabled) go to stdout

In that sense, if messages should stay together, they should be best printed on the same stream. Hence, I didn't deem the message "warning-y" enough to declare it a warning (that'll appear quite a lot as it's printed right now—for proximity to the energy output messages).
In that case, e.g. if someone were to pipe stdout only, they'd get the energy values, but no warning that they might be needed to be treated with care; thus potentially causing slightly-off conclusions.

However, feel free to offer a compelling counter argument.

vikaskurapati: > For post-processing, there's the CSV file (which was repaired by this PR)—so it wouldn't be influenced by that (or only very marginally). Regarding once at the beginning vs. every time: that's also a good question
> 

The CSV file definitely helps for post-processing. But piping out the output in that case shouldn't need the warning printed so often as the piping out for multiple simulations is used generally used in a case where a user wants to set up a bunch of different setups in an automated way. If someone is just testing, they would not just pipe it out in my opinion. I would really avoid printing this multiple times and just print once in the beginning. As an stderr or an stdout, that is dependent on the nature, but this still feels like something which should rather be flused to stderr given the use case that we say that the volume energies are not reliable as the material for which they are using is not compatible with this volume energies function.

> I _do_ understand what you mean—however stderr output isn't necessarily stderr output. :) E.g. take the `AMD_LOG_LEVEL` variable which is used for tracing ROCm code—for example SeisSol; to see where it breaks. All the log output gets send not to stdout, but stderr; i.e. if you'd look at the stdout log, you'd just see all simulation info, without any other issues. Without the warnings; which may, in the case here, be a tiny problem, since the message here is relevant to the numerical values, but not simulation-threatening. Don't get me wrong; I get that it's getting a bit counter-intuitive; but maybe the divide is somewhere else, maybe the definition of "warning == stderr message" (and "error == stderr message + abort") is the wrong notion to start with.
> 

Do you mean in the terminal output? Or just piping the output? Or as a slurm job? Because the terminal output should give everything on the terminal. Never used the piping as I am not really a fan of it and the slurm job's outputs are configurable to output into different or the same file. I just put everything into the same file for convenience. I do understand that stderr does not mean it is going to abort but it in this case, it is more of an unexpected behaviour or something which is not implemented. So, may be it is still good to have in the warning section but I don't have big objections for having it flushed to output. It just reads slightly inconsistent is all, IMO. 

> If that would mean that we reduce the warnings in general down to those that are _absolutely_ simulation-threatening/technical warnings—yes, that would be good IMO. ... Maybe it's something to even discuss a bit more broadly somewhere else.

I think anything that is technical and relates to unexpected or deviational behaviour, should stay in warnings and that anything that is threatening or would break something should be an error and only those which are needed to explain the state/settings/values of the simulation should be in the stdout. We could discuss this broadly somewhere else. :)

davschneller: > > For post-processing, there's the CSV file (which was repaired by this PR)—so it wouldn't be influenced by that (or only very marginally). Regarding once at the beginning vs. every time: that's also a good question
> 
> The CSV file definitely helps for post-processing. But piping out the output in that case shouldn't need the warning printed so often as the piping out for multiple simulations is used generally used in a case where a user wants to set up a bunch of different setups in an automated way. If someone is just testing, they would not just pipe it out in my opinion. I would really avoid printing this multiple times and just print once in the beginning. As an stderr or an stdout, that is dependent on the nature, but this still feels like something which should rather be flused to stderr given the use case that we say that the volume energies are not reliable as the material for which they are using is not compatible with this volume energies function.

I do pipe out for testing; cf. the given use case below. Just to see if SeisSol gets to the part that interests me to then just abort the whole program. Then, open the file and see what's printed there last, and try to draw a conclusion. Just did all of that a few days ago.

Furthermore, I do get that it can be visually a bit ; however at least IMO, wrong things should be marked somehow as it's easy to forget down the road. But, I'm open for more non-intrusive things.
E.g. an asterisk in front of the values; like: `Energy value (J, %1, %2): *123, *43, *54`. IIRC the linguists like to do that with unnatural sentences; but I'm not sure if anyone in the natural science community would use it or think of it the same way.
(plus maybe a message in the beginning that can then go to `logWarning`—then I'd agree that it's better there. Otherwise, if it's a `logWarning` in the simulation, it would have just spammed the stderr)

> > I _do_ understand what you mean—however stderr output isn't necessarily stderr output. :) E.g. take the `AMD_LOG_LEVEL` variable which is used for tracing ROCm code—for example SeisSol; to see where it breaks. All the log output gets send not to stdout, but stderr; i.e. if you'd look at the stdout log, you'd just see all simulation info, without any other issues. Without the warnings; which may, in the case here, be a tiny problem, since the message here is relevant to the numerical values, but not simulation-threatening. Don't get me wrong; I get that it's getting a bit counter-intuitive; but maybe the divide is somewhere else, maybe the definition of "warning == stderr message" (and "error == stderr message + abort") is the wrong notion to start with.
> 
> Do you mean in the terminal output? Or just piping the output? Or as a slurm job? Because the terminal output should give everything on the terminal. Never used the piping as I am not really a fan of it and the slurm job's outputs are configurable to output into different or the same file. I just put everything into the same file for convenience. I do understand that stderr does not mean it is going to abort but it in this case, it is more of an unexpected behaviour or something which is not implemented. So, may be it is still good to have in the warning section but I don't have big objections for having it flushed to output. It just reads slightly inconsistent is all, IMO.
> 

E.g. in slurm, but Python does something similar with some settings. But yes; piping is pretty similar—you'd pipe out stdout, and the warnings still get printed onto the command line. Great way on filtering e.g. some low-noise tool messages.
Vice-versa is also possible; i.e. having the SeisSol logs intact; while you'd obtain the tool log in a file. As some examples with possible use cases.
But yes; for the rest of the answer, see above.

> > If that would mean that we reduce the warnings in general down to those that are _absolutely_ simulation-threatening/technical warnings—yes, that would be good IMO. ... Maybe it's something to even discuss a bit more broadly somewhere else.
> 
> I think anything that is technical and relates to unexpected or deviational behaviour, should stay in warnings and that anything that is threatening or would break something should be an error and only those which are needed to explain the state/settings/values of the simulation should be in the stdout. We could discuss this broadly somewhere else. :)

vikaskurapati: > > > For post-processing, there's the CSV file (which was repaired by this PR)—so it wouldn't be influenced by that (or only very marginally). Regarding once at the beginning vs. every time: that's also a good question
> > 
> > 
> > The CSV file definitely helps for post-processing. But piping out the output in that case shouldn't need the warning printed so often as the piping out for multiple simulations is used generally used in a case where a user wants to set up a bunch of different setups in an automated way. If someone is just testing, they would not just pipe it out in my opinion. I would really avoid printing this multiple times and just print once in the beginning. As an stderr or an stdout, that is dependent on the nature, but this still feels like something which should rather be flused to stderr given the use case that we say that the volume energies are not reliable as the material for which they are using is not compatible with this volume energies function.
> 
> I do pipe out for testing; cf. the given use case below. Just to see if SeisSol gets to the part that interests me to then just abort the whole program. Then, open the file and see what's printed there last, and try to draw a conclusion. Just did all of that a few days ago.
> 

I don't understand the reasoning behind this, if you are piping to test, you should not pipe just the stdout but also the stderr. The warnings in that case should not be ignored. But in any case, this should not be the sole reason for flushing out warnings to stdout or vice versa.

> Furthermore, I do get that it can be visually a bit ; however at least IMO, wrong things should definitely be marked somehow as it's easy to forget down the road—then, unintentionally, "falsehood has become truth", so to say.
> 
We would be marking it, doing it once is enough to mark it instead of printing it everytime. And IMO, throwing it in a warning is marking enough.

> But, though I'm not a super fan of it, maybe we can settle for something more non-intrusive? E.g. an asterisk in front of the values; like: `Energy value (J, %1, %2): *123, *43, *54`. IIRC the linguists like to do that with unnatural sentences; but I'm not sure if anyone in the natural science community would use it or think of it the same way. (plus maybe a message in the beginning that can then go to `logWarning`—then I'd agree that it's better there. Otherwise, if it's a `logWarning` in the simulation, it would have just spammed the stderr)
> 

I am not really a fan of this idea with numbers but if others like it, I am not going to oppose it.

> > > I _do_ understand what you mean—however stderr output isn't necessarily stderr output. :) E.g. take the `AMD_LOG_LEVEL` variable which is used for tracing ROCm code—for example SeisSol; to see where it breaks. All the log output gets send not to stdout, but stderr; i.e. if you'd look at the stdout log, you'd just see all simulation info, without any other issues. Without the warnings; which may, in the case here, be a tiny problem, since the message here is relevant to the numerical values, but not simulation-threatening. Don't get me wrong; I get that it's getting a bit counter-intuitive; but maybe the divide is somewhere else, maybe the definition of "warning == stderr message" (and "error == stderr message + abort") is the wrong notion to start with.
> > 
> > 
> > Do you mean in the terminal output? Or just piping the output? Or as a slurm job? Because the terminal output should give everything on the terminal. Never used the piping as I am not really a fan of it and the slurm job's outputs are configurable to output into different or the same file. I just put everything into the same file for convenience. I do understand that stderr does not mean it is going to abort but it in this case, it is more of an unexpected behaviour or something which is not implemented. So, may be it is still good to have in the warning section but I don't have big objections for having it flushed to output. It just reads slightly inconsistent is all, IMO.
> 
> E.g. in slurm, but Python does something similar with some settings. But yes; piping is pretty similar—you'd pipe out stdout, and the warnings still get printed onto the command line. Great way on filtering e.g. some low-noise tool messages. Vice-versa is also possible; i.e. having the SeisSol logs intact; while you'd obtain the tool log in a file. As some examples with possible use cases. But yes; for the rest of the answer, see above.
>
In Slurm, you have an option of flushing both into the same file and in different files. If I remember correctly, this can be done in bash/terminal, too, with pipes. I just don't remember it on the top of my head. It depends on what the user wants/needs to do. But in any case, they should be at least looking at both.  

> > > If that would mean that we reduce the warnings in general down to those that are _absolutely_ simulation-threatening/technical warnings—yes, that would be good IMO. ... Maybe it's something to even discuss a bit more broadly somewhere else.
> > 
> > 
> > I think anything that is technical and relates to unexpected or deviational behaviour, should stay in warnings and that anything that is threatening or would break something should be an error and only those which are needed to explain the state/settings/values of the simulation should be in the stdout. We could discuss this broadly somewhere else. :)

davschneller: > > > > For post-processing, there's the CSV file (which was repaired by this PR)—so it wouldn't be influenced by that (or only very marginally). Regarding once at the beginning vs. every time: that's also a good question
> > > 
> > > 
> > > The CSV file definitely helps for post-processing. But piping out the output in that case shouldn't need the warning printed so often as the piping out for multiple simulations is used generally used in a case where a user wants to set up a bunch of different setups in an automated way. If someone is just testing, they would not just pipe it out in my opinion. I would really avoid printing this multiple times and just print once in the beginning. As an stderr or an stdout, that is dependent on the nature, but this still feels like something which should rather be flused to stderr given the use case that we say that the volume energies are not reliable as the material for which they are using is not compatible with this volume energies function.
> > 
> > 
> > I do pipe out for testing; cf. the given use case below. Just to see if SeisSol gets to the part that interests me to then just abort the whole program. Then, open the file and see what's printed there last, and try to draw a conclusion. Just did all of that a few days ago.
> 
> I don't understand the reasoning behind this, if you are piping to test, you should not pipe just the stdout but also the stderr. The warnings in that case should not be ignored. But in any case, this should not be the sole reason for flushing out warnings to stdout or vice versa.
> 

Ah, I forgot; that was with `AMD_LOG_LEVEL=3`. Thus, there was a ton of messages in stderr (which I piped away), but just the normal SeisSol execution flow in stdout (which I kept on the command line); which just needed to wait until we got into the simulation loop—that can be best read from the command line. Also, it's a very small-scale run.

...I guess that may be where the difference in our opinions may lie (feel free to disagree here): interactive vs. batch job simulations.
If we just consider an sbatch log file, I totally get your point of view. If we consider someone running a small simulation on `salloc`, it gets much more difficult with just a single warning message without further annotations. Imagine e.g. someone who just looks at the latest log message "out of convenience" while running a simulation interactively—they may not remember the initial warning message.

(also, while everyone "should" be looking at both; that can easily get lost. Because it's not enforced. "should" may not be enough)

To anyone else: any opinion on this matter? :)

Thomas-Ulrich: no strong opinion. I'm always piping both stdout and stderr to the same log file.

vikaskurapati: > Ah, I forgot; that was with `AMD_LOG_LEVEL=3`. Thus, there was a ton of messages in stderr (which I piped away), but just the normal SeisSol execution flow in stdout (which I kept on the command line); which just needed to wait until we got into the simulation loop—that can be best read from the command line. Also, it's a very small-scale run.
> 
> ...I guess that may be where the difference in our opinions may lie (feel free to disagree here): interactive vs. batch job simulations. If we just consider an sbatch log file, I totally get your point of view. If we consider someone running a small simulation on `salloc`, it gets much more difficult with just a single warning message without further annotations. Imagine e.g. someone who just looks at the latest log message "out of convenience" while running a simulation interactively—they may not remember the initial warning message.
> 

Yeah, while I do understand your point, I don't believe that is a good enough reason to keep spamming the output with this. We can't change something because of an "out of convenience" aspect, IMO. If it is a warning but you want it to be seen for sure in stdout because the stderr is piped out when ignored, doing it once should suffice. We can't take care of every "the user might not read this" thing. 

> (also, while everyone "should" be looking at both; that can easily get lost. Because it's not enforced. "should" may not be enough)
> 
> To anyone else: any opinion on this matter? :)

davschneller: Ok, made it a bit less intrusive (which yes, I see—that could be really annoying when analyzing a log file; though it isn't when analyzing a live terminal output), and added a message.

Also a kind reminder to fix the issue. :) I tried already some time ago to make it much more general https://github.com/SeisSol/SeisSol/tree/davschneller/generalized-energy ; but the results were still pretty off.

I often run such "out of convenience" tests for rapid development; so it's an active use case at least for me as a developer.
Also, given that there may be plenty of users who just run SeisSol on their local machine or interactively, so I fear it's not that far off (what we definitely shouldn't do IMO is to just assume that our way of running the software is the "only and right one" (also true for me of course); because who knows what other ways are out there).

vikaskurapati (approved):
Looks good to me now, thank you!

PR #1307, opened 2025-03-20
Fault output/on-fault receiver messages, on-fault point finding overhaul

davschneller:
We

* add some messages for on-fault receivers being treated now
* refine the receiver containment mechanism to hopefully work better at the halo boundaries (copy/ghost layers). In essence, it's just porting over the hyperplane formulation from tetrahedron to (extruded) triangle.


fabian-kutschera:
LGTM.

fabian-kutschera (approved):

PR #1308, opened 2025-03-21
more explicit memory requirements

Thomas-Ulrich:
I needed to write a paragraph about that for an HPC proposal.
Please check my calculations as the result slightly differs from the one I wrote a few years ago.

### DIFF
--- a/Documentation/memory-requirements.rst
+++ b/Documentation/memory-requirements.rst
@@ -13,7 +13,7 @@ General
 -------
 
 Memory requirements per node are difficult to predict because the elements are not distributed equally when Local Time-Stepping (LTS) is enabled.
-However, an adequate rule of thumb to estimate memory requirements is to multiply the number of elements with the number of degrees of freedom per element and then multiply this number with a factor of 10.
+However, an adequate rule of thumb to estimate memory requirements is to multiply the number of elements with an estimate of the number of variable per elements and then multiply this number with a factor of 10.
 Given the number of basis function per element function of the order of accuracy n:
 
 .. math::
@@ -21,10 +21,11 @@ Given the number of basis function per element function of the order of accuracy
    N_b = \frac{n(n+1)(n+2)}{6}
 
 And the number of variables (with :math:`Q_p` number of relaxation mechanisms used for viscoelastic attenuation, typically 3):
+(3 velocity, 6 stresses, 9 buffers)
 
 .. math::
 
-   N_v = 9 + 6 Q_p
+   N_v = 9 + 6 Q_p + 9
 
 
 A run using the viscoelastic wave equation with 100 million elements of order 5 counts
@@ -32,7 +33,7 @@ A run using the viscoelastic wave equation with 100 million elements of order 5 
 
 .. math::
 
-   N_{\text{dof}} = N_v N_b N_{\text{cells}} = 35 \times 27 \times 100,000,000 = 9.45 \times 10^{10} \text{ DOF}
+   N_{\text{dof}} = N_v N_b N_{\text{cells}} = 35 \times 36 \times 100,000,000 = 9.45 \times 10^{10} \text{ DOF}
 
 
 and therfore would require about 0.9 terabytes (:math:`=9.45 \times 10^{10}/1024^4`) of memory.


davschneller:
Hmm; do you remember how you calculated the number back then?

Two things for the current calculation:

* floats consume 4 B, doubles consume 8 B. Thus, it's actually `0.9 * 8 TiB = 7.2 TiB`. (?) Feels a bit over-estimating the original memory requirement number.
* the integration buffers (`buffers`) or the space-time evolution (`derivatives`) take up memory. Probably, the majority of the cells will just have buffers which'll again consume `9 * N_b` of space. Only at LTS boundaries, DR faces, and some boundary conditions, you'd have the full space-time derivatives which I think in the visco case would be just of size `9 * N_b * n`. (but, to be honest, I'd need to check again) Thus, assuming that we use GTS, we'd already need `35*36*10**8 = 1.26 * 10**11 DOFs`.
(for anything more visco-elasticity-specific, I'd probably need to check a bit more)

Maybe also, the section is tilted a bit towards just viscoelasticity; maybe it would be good to add a note there about that? (e.g. poro will have `N_v=13`)

davschneller:
Calculating the "rule of thumb" with order 6 gives 1.37 TiB which could be rounded to 1.4 TiB—could it be an order typo maybe?

Thomas-Ulrich:
I think it was an empirical estimate of Ravil.

I've added to my job script on NG to track memory:
```
while true; do
    free -m | awk 'NR==2{print strftime("%Y-%m-%d %H:%M:%S"), "Memory Used:", $3, "MB"}' >> mem_usage_$SLURM_JOB_ID.log
    sleep 10
done &
```
and got with LTS:
LTS2, 8 nodes, single, o4, 223,088 cells, 11309 MB
LTS2, 8 nodes, double, o4, 223,088 cells, 11507 MB
(probably showing that there is some baseline whatever the mesh size)
LTS2, 8 nodes, single, o4, 12,413,140 cells, 27420 MB
LTS2, 8 nodes, double, o4, 12,413,140 cells, 35796 MB
GTS, 8 nodes, double, o4, 12,413,140 cells, 34237 MB
LTS2, 16 nodes, single, o4, 12,413,140 cells, 19746 MB

(this is mem per node, would need to be multiply by 8 or 16).

Let see let say for:
LTS2, 8 nodes, single, o4, 12413140 cells, 27420 MB

>>> 4*5*9*12413140*10/(16*1024**2)
1331 MB/nodes

(far from the estimate).
Maybe i should test that one 1 node first.

davschneller:
Per cell, you'll also have the PDE matrices as well; for elastic, you'd get about `36 * 3 + 81 * 8`, for the star matrices and flux matrices.

Plus, the DR which consumes _much_ more memory.

davschneller:
In sum, I'd get around 2000 values (floats/doubles) for o4 with visco2 / 3 relaxation mechanisms. Which would yield about 24 GiB per node, for 12.5 mln cells, 8 nodes; ignoring anything else like infrastructure data, ghost/copy, and the dynamic rupture.



In any way—it would probably be possible to show the theoretical per-cell/per-layer memory consumption for the DR and waveprop rather easily when starting a simulation; if that's wanted or also helpful here.

Thomas-Ulrich:
I just realized there is about 10600 Mb of memory used by default on NG on each node (independent of using seissol).
than is 11309MB used =  about 700Mb used by seissol only.

davschneller:
In addition to that, the single-double comparison could roughly reveal how much memory is consumed for the DOFs+DR etc. By having `memory(double) - memory(single)` is the single memory consumption, given that you'd ignore any partitioning issues.

Thus, we'd have 200 MiB for the DOFs and DR and PDE data, leaving roughly 500 MiB for metadata (e.g. `CellLocalInformation`), mesh data, IO data, and more.

Thomas-Ulrich:
Ok, Mhm I don't know with all this new information if I get a way to compute a better estimate...
If I read you well I get:

`4 or 8 * N_cell * [ n(n+1)(n+2)/6 *(18+3Qp) + 2000*min(Qp,1)]`


For the record here are some new estimates on 1 node, removing the initial node memory.
```
GTS, 1 node, double, 0.223e6 cells, o4, 3148 MB
GTS, 1 node, single, 0.223e6 cells, o4, 2159 MB 
(diff 989MB).
GTS, 1 node, double, 2.009e6 cells, o4, 22874 MB 
GTS, 1 node, single, 2.009e6 cells, o4, 14967 MB 
(diff 7907MB)
```

davschneller:
Ah, sorry; I meant 2000 for everything; including DOFs, for o4.

Without that, it's I think about 1341 values, i.e. `4 or 8 * N_cell * [ n(n+1)(n+2)/6 *(18+3Qp) + 1341]`. However, that way, I get roughly double the actual memory consumption for both cases (comparing with the differences). Which is somehow pretty strange.

Anyways; I've tried looking through the data fields—and it shows: the estimate above seems to be going into the right direction (we got the fixed variables mostly covered; except for the metadata). Somehow only the buckets (the buffers/derivatives) are much larger; but that might be due to LTS+DR for the small test case that I have.

The memory consumption per DR face is double the consumption per mesh cell (but smaller in total).

davschneller:
If it's still relevant: did #1338 help with the analysis here? (or does it show anything useful?)

PR #1309, opened 2025-03-21
`fusedOriginalIndex` -> `simulationIndex`

davschneller:
Instead of storing the original point source index, store the simulation it is used in—that's all we need.

Reason: we can avoid one modulo.

(still, IMO the fused way of point source distribution is pretty inconsistent with running a single simulation tbh)

### DIFF
--- a/src/SourceTerm/Manager.cpp
+++ b/src/SourceTerm/Manager.cpp
@@ -343,12 +343,14 @@ auto loadSourcesFromFSRM(const char* fileName,
       sources.numberOfSources = numberOfSources;
       sources.mInvJInvPhisAtSources.resize(numberOfSources);
       sources.tensor.resize(numberOfSources);
-      sources.simulationIndex.resize(numberOfSources);
       sources.onsetTime.resize(numberOfSources);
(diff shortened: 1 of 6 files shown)


src/SourceTerm/Manager.cpp
```diff
@@ -468,7 +468,7 @@ auto loadSourcesFromNRF(const char* fileName,
       sources.tensor.resize(numberOfSources);
       sources.A.resize(numberOfSources);
       sources.stiffnessTensor.resize(numberOfSources);
-      sources.fusedOriginalIndex.resize(numberOfSources); // only used in case of fused simulations
+      sources.simulationIndex.resize(numberOfSources);
```
vikaskurapati, line 471: can you please let the comment stay? Just for the ease of understanding. The variable naming is okay, but I would really prefer if the comment stayed.

vikaskurapati (approved):

src/SourceTerm/Manager.cpp
davschneller: Sure thing; it's re-added. Thanks for the review!

PR #1317, opened 2025-03-30
Add/mainline multi-nucleation support

davschneller:
Effectively mainlining of the 2nuc branches. Right now, it's fixed to maximum 16 nucleations; but that limit can be increased freely (that limit existing in the first place is to avoid handling the GPU memory in the parameters struct in a slightly more complicated way)

It will require some more parameters; mostly, it's just copying the 2nuc notation of things and generalizing it.
* `nucleationcount` saying how many nucleations there are
* `t{i}_0` and `s{i}_0`; `t_0` being the respective rise duration; and `s_0` the respective onset time. It can also be specified for nucleation 0 now.
* any fault parameters in easi are also supplied the way the 2nuc branches do, i.e. with suffixed numbers for nucleation 2 and upwards.

Feel free to suggest a better way of handling it.


Aangniu (approved):
Thanks, David. It looks good to me :)

davschneller:
Thanks alot for your review!

PR #1320, opened 2025-04-08
Fused simulations with Dynamic Rupture and Plasticity on CPUs

vikaskurapati:
This PR implements fused simulations/multiple simulations with Dynamic Rupture and Plasticity on CPUs for elastic materials. We observed a speedup on grace-grace CPUs of Vista and Skylake CPUs of Frontera when we used lower orders like 3,4 but a slowdown on higher orders like 6. 

We run the WP in a fused way and Dynamic Rupture and Plasticity kernels in a serial way due to non-linearities. A further PR would be required for porting of this to GPUs and then probably more optimizations to avoid running Dynamic Rupture and Plasticity serially

Still draft because a lot of CI tests still fail


vikaskurapati:
Closing this as #1364 fixes the receiver output, and this method is not intended to be merged into the mainline SeisSol currently. The energy outputs and plasticity corrections will come in later PRs. 

Note: Please don't delete the branch, as it was used for benchmarking and UQ workflows. I will need the branch in the future.

PR #1332, opened 2025-04-24
Update the ASAGI cmake finder

davschneller:
Cf. https://github.com/TUM-I5/ASAGI/pull/17


davschneller:
(the ASAGI init needed to be moved; so that Netcdf and Hdf5 are linked afterwards—so that the Netcdf symbols can be found)

vikaskurapati (approved):

PR #1333, opened 2025-04-25
Transfer some more defines to generated constexprs

davschneller:
Replace some more defines. And move others into config files that CMake will fill with values once it runs.

As a result, we're able to clean up some more code; and remove some more defines in some places. In particular, there are currently three files that are filled with build-time data: `Version.h.in` contains some build-time info from CMake. `Alignment.h.in` contains the alignment which is supposed to stay constant, even if we had multiple different kernels or configurations in SeisSol. And finally, `Config.h.in` which contains parameters that are defined by the code generation.
(maybe `Alignment.h.in` and `Version.h.in` can be unified)
In the future, an idea would be to automatically generate the `Config.h` from the code generation even.

In the past, those CMake defines were usually a problem e.g. when building different files in different targets. Then, the defines need to be enabled everywhere which can be forgotten quickly, and thus cause confusion (the same header files are used in the same supposed way, but the result is different) or hard-to-find bugs. Hence the change—also, it makes the project even cleaner.
(the usage of the defines might be a relic from the C++-FORTRAN interop era still; partially at least)

Besides, we replace a few `real` occurrences (which are non-performance-critical) with `double`s; to improve the quality of some constant matrices.


davschneller:
(seems to have a dependency on https://github.com/SeisSol/Device/pull/46 to land first)

vikaskurapati (approved):

PR #1334, opened 2025-04-25
Update SeisSol after updating the Device Submodules

davschneller:
Update the device interop after the latest Device module changes from

* https://github.com/SeisSol/Device/pull/45
* https://github.com/SeisSol/Device/pull/46

Besides, it fixes the plasticity FLOP/s counting (in fact: so far, no FLOP/s for plasticity were accounted for on the CPU).

### DIFF
--- a/src/Kernels/Plasticity.cpp
+++ b/src/Kernels/Plasticity.cpp
@@ -265,8 +265,12 @@ void Plasticity::computePlasticityBatched(
     auto prevDofs = runtime.memoryHandle<real>(prevDofsSize);
 
     real** dofsPtrs = (entry.get(inner_keys::Wp::Id::Dofs))->getDeviceDataPtr();
-    device.algorithms.copyScatterToUniform(
-        dofsPtrs, prevDofs.get(), DofsSize, DofsSize, numElements, defaultStream);
+    device.algorithms.copyScatterToUniform(const_cast<const real**>(dofsPtrs),
+                                           prevDofs.get(),
+                                           DofsSize,
+                                           DofsSize,
+                                           numElements,
+                                           defaultStream);
 
     // Convert modal to nodal
     real** modalStressTensors = (entry.get(inner_keys::Wp::Id::Dofs))->getDeviceDataPtr();
--- a/src/Kernels/Plasticity.cpp
+++ b/src/Kernels/Plasticity.cpp
@@ -236,13 +236,14 @@ unsigned Plasticity::computePlasticity(double oneMinusIntegratingFactor,
   return 0;
 }
 
-unsigned Plasticity::computePlasticityBatched(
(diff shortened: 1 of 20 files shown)


Documentation/build-seissol.rst
```diff
@@ -73,9 +73,9 @@ Building for GPUs
 
 When building for GPUs, you will need to install SYCL and gemmforge, chainforge as well.
 Also, you will need to supply the device backend and the device architecture.
-The backend can be ``cuda``, ``hip``, or one of the SYCL implementations, ``hipsycl`` or ``oneapi``.
+The backend can be ``cuda``, ``hip``, or one of the SYCL implementations, ``acpp`` or ``oneapi``.
 
-Generally, we recommend compiling for simgle precision on GPUs.
+Generally, we recommend compiling for single precision on GPUs.
```
vikaskurapati, line 78: Is this for better performance? Could we modify the note here, mentioning that we observe that the performance is better for single precision on GPUS, and that is why we recommend this? There are quite a few scenarios, e.g., RidgeCrest, which fail with single precision, and it would be nice for anyone who looks at it to know why we recommend this.

src/Kernels/Plasticity.cpp
```diff
@@ -297,20 +301,22 @@ unsigned Plasticity::computePlasticityBatched(
...
                                                      defaultStream);
 
     // count how many elements needs to be adjusted
-    unsigned numAdjustedElements = device.algorithms.reduceVector(
-        isAdjustableVector, numElements, ::device::ReductionType::Add, defaultStream);
+    device.algorithms.reduceVector(yieldCounter,
+                                   isAdjustableVector.get(),
+                                   true,
+                                   numElements,
+                                   ::device::ReductionType::Add,
```
vikaskurapati, line 318: Which namespace is this derived from? Could we avoid such empty namespace calls? Imo, it makes the code quite confusing to read.

vikaskurapati (approved):
Minor comments

Documentation/build-seissol.rst
davschneller: Removed it.

It's good to use single-precision on consumer-grade GPUs (b/c their FP64 path is slow), and generally, if your mesh can take it. Still, probably not enough to put up a recommendation here.

src/Kernels/Plasticity.cpp
davschneller: The empty namespace call means we're at the root namespace (or whatever it's called in the correct terminology), and by that says that it's explicitly not nested

E.g. if we'd have a `seissol::device` namespace, it would take that one and cause a compiler error.
Not sure why it's done here (haven't tried compiling without), but in some cases, it's needed for disambiguation.

PR #1336, opened 2025-04-28
More clang-tidy

davschneller:
There's more `clang-tidy` functionality to be enabled (and that also seemed to be partially applied in the past; for those using CLion as IDE). Thus, this preparation PR which should fix many of the issues in advance.


src/IO/Datatype/Datatype.h
```diff
@@ -32,7 +32,7 @@ class Datatype : public std::enable_shared_from_this<Datatype> {
   virtual ~Datatype();
   virtual std::size_t size() const = 0;
   virtual YAML::Node serialize() const = 0;
-  virtual Array unwrap(std::size_t maxDimensions = 256);
+  virtual Array unwrap(std::size_t maxDimensions);
```
vikaskurapati, line 35: Why remove the default value here?

davschneller: Because of the `virtual` here; some clang-tidy rule complained about that one—so I removed it (it isn't used anywhere right I think anywhere)

src/Initializer/CellLocalMatrices.cpp
```diff
@@ -682,9 +682,8 @@ void initializeDynamicRuptureMatrices(const seissol::geometry::MeshReader& meshR
           1.0 / (1.0 / impAndEta[ltsFace].zs + 1.0 / impAndEta[ltsFace].zsNeig);
 
       switch (plusMaterial->getMaterialType()) {
-      case seissol::model::MaterialType::Elastic: {
-        break;
-      }
+      case seissol::model::MaterialType::Elastic:
+        [[fallthrough]];
       case seissol::model::MaterialType::Viscoelastic: {
         break;
```
vikaskurapati, line 688: Can we do the visco to [[fallthrough]] too, in that case?

davschneller: The `[[fallthrough]]` here is there because the `switch`/`case` constructs can be pretty confusing, and it's easy to forget a `break;`. So instead, here it indicates explicitly to _not_ `break`. So yes, for `visco`, we need that. In essence, the two cases got combined.

vikaskurapati (approved):

PR #1345, opened 2025-05-13
Add a floating communication thread option

davschneller:
In essence, a commthread that doesn't need to be pinned, but gets yielded after each round of `MPI_Test`. Thus, one less core to be occupied.


vikaskurapati (approved):
Looks good to me. I have a couple of questions, though:

1. Does this mean we have one communication thread per rank floating, and gets freed after every MPI Test call? And does it also do computations when freed?
2. How does this change for persistent/non persistent MPI? Does anything even change?
3. Do we not need to have the pinning environment variable anymore? If that is the case, modifying the documentation may be useful along with the PR.

PR #1347, opened 2025-05-14
Rename ViscoMode type

### LINKED ISSUE #1346: error: declaration of 'constexpr const seissol::ViscoMode seissol::Config::ViscoMode' changes meaning of 'ViscoMode'
### Describe the bug
The latest master does not compile?

```spack install -j 40 seissol@master convergence_order=4 dr_quad_rule=dunavant equations=elastic precision=double ^easi +python```
with gcc@12 on our local server exception

Output log in:

[spack-build-out.txt](https://github.com/user-attachments/files/20207322/spack-build-out.txt)

davschneller:
Fix #1346 .

### DIFF
--- a/src/Common/Typedefs.h
+++ b/src/Common/Typedefs.h
@@ -11,7 +11,7 @@ namespace seissol {
 
 enum class DRQuadRuleType { Stroud, Dunavant, WitherdenVincent };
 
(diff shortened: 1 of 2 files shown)


Thomas-Ulrich (approved):
Thank you, it fixes the problem indeed!

PR #1348, opened 2025-05-16
A an `etahack` stopping mechanic

davschneller:
A mechanic to disable the `etahack`, as soon as we've exceeded a certain duration in the simulation.

Still needs a slightly better implementation, but it seems to already do its job right now.


fabian-kutschera:
Confirm working. Use this functionality by adapting the parameter file, e.g., in the following way:

> etahack = 0.97 !1.00 is default      ! use any value smaller than one to mitigate quasi-divergent solutions in friction laws
etastop = 10 ! seconds

fabian-kutschera (approved):

PR #1351, opened 2025-05-20
Allow Ninja Multi-Config Builds (Debug+Release)

davschneller:
Refers to https://cmake.org/cmake/help/latest/generator/Ninja%20Multi-Config.html . All new builds are put into `$BUILD_ROOT/Release`, `$BUILD_ROOT/Debug` and so on—if you configure with `cmake -G "Ninja Multi-Build`.

It needed one operation to be moved.

Moreover, I've removed the `Release` and `Debug` etc. from the Seissol exe name because of that—you can now build both without regenerating your CMake files each time.
Feel free to disagree with that change.

(and yes, this PR's _only_ about the build type, nothing SeisSol specific)

### DIFF
--- a/.github/workflows/build-seissol-cpu.yml
+++ b/.github/workflows/build-seissol-cpu.yml
@@ -21,8 +21,7 @@ jobs:
     env:
       EXENAME: |
         ${{
-          format('{0}_{1}{2}_{3}_{4}{5}',
-            matrix.build_type,
+          format('{0}{1}_{2}_{3}{4}',
             matrix.precision == 'single' && 's' || 'd',
             matrix.arch,
             matrix.order,
@@ -178,8 +177,7 @@ jobs:
     env:
       EXENAME: |
         ${{
-          format('{0}_{1}{2}_{3}_{4}{5}',
-            matrix.build_type,
+          format('{0}{1}_{2}_{3}{4}',
             matrix.precision == 'single' && 's' || 'd',
             matrix.arch,
             matrix.order,
--- a/.github/workflows/build-seissol-gpu.yml
+++ b/.github/workflows/build-seissol-gpu.yml
@@ -21,8 +21,7 @@ jobs:
     env:
       EXENAME: |
         ${{
-          format('{0}_{1}{2}_{3}_{4}_{5}{6}',
-            matrix.build_type,
+          format('{0}{1}_{2}_{3}_{4}{5}',
             matrix.precision == 'single' && 's' || 'd',
             matrix.setup.arch,
             matrix.setup.backend,
@@ -197,8 +196,7 @@ jobs:
     env:
       EXENAME: |
         ${{
-          format('{0}_{1}{2}_{3}_{4}_{5}{6}',
-            matrix.build_type,
+          format('{0}{1}_{2}_{3}_{4}{5}',
             matrix.precision == 'single' && 's' || 'd',
             matrix.setup.arch,
             matrix.setup.backend,
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,12 +77,12 @@ endif()
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
-set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
-        "Debug" "Release" "RelWithDebInfo") # MinSizeRel is useless for us
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Release CACHE STRING "" FORCE)
   message(STATUS "Set build type to Release as none was supplied.")
 endif()
+set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS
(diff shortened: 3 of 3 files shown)


vikaskurapati:
> Moreover, I've removed the `Release` and `Debug` etc. from the Seissol exe name because of that—you can now build both without regenerating your CMake files each time. Feel free to disagree with that change.
> 

Yes, I prefer having the Release and Debug in the binary explicitly mentioned. If it is all the same for you(or others), I request you to revert this change. 

> (and yes, this PR's _only_ about the build type, nothing SeisSol specific)

vikaskurapati (changes requested):

davschneller:
To give a counter perspective (on why the removal makes sense IMO), here you go. :)

If we have `Release`/`Debug` in the name, it's always a hassle to switch between them, from a programmer's PoV. E.g. you program, then you switch to a debug build and want to test that one. So you need to re-run CMake or copy your CMake folder.
Then, you're done debugging, and switch back to release mode. But now, you need to re-run CMake yet again.

The new builds allow to skip that.

With the new multiconfig builds that this PR introduces, you can still have `Debug` and `Release` builds at the same time—even with the same CMake config. It's just placed into different folders. I.e. instead of having the files (and possibly installed files) named differently, you can build it differently on demand.

`$BUILD/SeisSol_Release_XYZ`, `$BUILD/SeisSol_Debug_XYZ`, you'll get `$BUILD/Release/SeisSol_XYZ`, and `$BUILD/Debug/SeisSol_XYZ`; given that you use the Ninja multi-config from the link.
I expect that any "usual" user won't bother about having a `Release` or `Debug` build.

Hence, in total, I'm very much in favor of removing it, actually. And yes; before this PR, only partially; but it was already rather annoying there to work with.

vikaskurapati:
> To give a counter perspective (on why the removal makes sense IMO), here you go. :)
> 
> If we have `Release`/`Debug` in the name, it's always a hassle to switch between them, from a programmer's PoV. E.g. you program, then you switch to a debug build and want to test that one. So you need to re-run CMake or copy your CMake folder. Then, you're done debugging, and switch back to release mode. But now, you need to re-run CMake yet again.
> 
> The new builds allow to skip that.
> 
> With the new multiconfig builds that this PR introduces, you can still have `Debug` and `Release` builds at the same time—even with the same CMake config. It's just placed into different folders. I.e. instead of having the files (and possibly installed files) named differently, you can build it differently on demand.
> 
> `$BUILD/SeisSol_Release_XYZ`, `$BUILD/SeisSol_Debug_XYZ`, you'll get `$BUILD/Release/SeisSol_XYZ`, and `$BUILD/Debug/SeisSol_XYZ`; given that you use the Ninja multi-config from the link. I expect that any "usual" user won't bother about having a `Release` or `Debug` build.
> 

That is, if we use the Ninja multi-config all the time. I don't generally use ninja, and just prefer the normal CMake for my workflows and testing, which is why I prefer having the normal Debug and Release in the binary names. With what you are saying, you could still have `$BUILD/Release/SeisSol_Release_XYZ` and `$BUILD/Release/SeisSol_Release_XYZ`.

Another idea is to have the names in binaries if we don't use the multi-config, and remove them if you are using them. That way, everyone is happy. 

> Hence, in total, I'm very much in favor of removing it, actually. And yes; before this PR, only partially; but it was already rather annoying there to work with.

davschneller:
High recommendation for `ninja`. Good build tool (clang uses CMake and only supports ninja there). And easy to install. :)

...but anyways; I forgot one point (added it in the edit only); what had been bothering me mostly about the `Release`/`Debug` in the name is if you want to debug something while programming; and why I'm in favor of the removal here. Suppose you build in `Release`; then you switch to debug, but you need to adjust all your scripts (that won't even change with the multibuild); since you call a different binary. And then again go back to `Release` and _again_ change the binary name again. And forgetting to change your paths while debugging can be annoying or cause you to look at dated binaries. Or it even makes you chase ghost errors.

I gotta say that I don't anymore see a many reasons (other than continuity) on why we'd want to keep a `Debug` and `Release` build like that anymore. It's IMO almost the same for the different archs; besides CPU vs GPU (and precision).
In the end, I guess most users/systems will (currently) probably build and install an array of SeisSols with different equations and orders and the two precisions for either CPU or GPU; but with a single arch, and in a single build type (`Release`, probably).

In total, that's why I wanted to remove it for all build types. But still, it is possible to hide dropping `Debug`/`Release` behind a CMake option; maybe until we've removed even more parts of the long SeisSol names.

davschneller:
Maybe: how about putting the name changes behind a CMake flag for the time being?
(i.e. for ninja/multi-conf; it'll still be `SeisSol_Release_XYZ`; unless you switch to the new naming)
It's a pretty close suggestion to removing the build type for the multi-confs; but we won't run into the risk of differing names on the build system used.

If we do that, then I'd add some more typing-friendliness changes to the executable/library names there which would probably be pretty confusing if just merged upstream (e.g. name it `seissol`, not `SeisSol`; for typing friendliness). One day, if everyone is happy with the change or we complete breaking changes (e.g. 2.0.0, or multimateriality in some future time), we can make the switch to the new exe naming permanent or abandon it again.

If you're good with that suggestion, I'll roll back the last commit in this PR.

davschneller:
Ok, sounds good to me; that'll all be done in another PR.

PR #1353, opened 2025-05-26
Always assume MPI

davschneller:
The MPI-less build of SeisSol is not that well supported; and MPI is present pretty much on almost every system. Plus, distinguishing between an MPI-less and MPI-using version creates a lot of extra code paths which are not really worth the maintenance given the setups that we have.

Hence, we remove most of the `USE_MPI` checks. If we really want an MPI-less build, it would probably be better to create a one-rank fake MPI implementation (i.e. which implements all MPI calls, but runs them only locally). We'll still need the `USE_MPI` flag nonetheless; for the submodules. The only two MPI-related checks that are left in are for ASAGI which is built separately and may need MPI on its own.

Petsc does the same thing; cf. https://gitlab.com/petsc/petsc/-/tree/main/include/petsc/mpiuni .

### DIFF
--- a/src/DynamicRupture/Output/OutputManager.cpp
+++ b/src/DynamicRupture/Output/OutputManager.cpp
@@ -93,12 +93,8 @@ std::string buildIndexedMPIFileName(const std::string& namePrefix,
                                     const std::string& nameSuffix,
                                     const std::string& fileExtension = std::string()) {
   std::stringstream suffix;
-#ifdef PARALLEL
   suffix << nameSuffix << '-' << makeFormatted<int, WideFormat>(index) << '-'
          << makeFormatted<int, WideFormat>(seissol::MPI::mpi.rank());
-#else
-  suffix << nameSuffix << '-' << makeFormatted<int, WideFormat>(index);
-#endif
   return buildFileName(namePrefix, suffix.str(), fileExtension);
 }
 
--- a/src/Parallel/MPI.h
+++ b/src/Parallel/MPI.h
@@ -29,7 +29,7 @@
 namespace seissol {
 
 #ifndef USE_MPI
-typedef MPIDummy MPI;
+using MPI = MPIDummy;
 #else // USE_MPI
 
 /**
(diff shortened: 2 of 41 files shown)


src/DynamicRupture/Output/OutputManager.cpp
```diff
@@ -93,12 +93,8 @@ std::string buildIndexedMPIFileName(const std::string& namePrefix,
                                     const std::string& nameSuffix,
                                     const std::string& fileExtension = std::string()) {
   std::stringstream suffix;
-#ifdef PARALLEL
   suffix << nameSuffix << '-' << makeFormatted<int, WideFormat>(index) << '-'
          << makeFormatted<int, WideFormat>(seissol::MPI::mpi.rank());
-#else
-  suffix << nameSuffix << '-' << makeFormatted<int, WideFormat>(index);
-#endif
```
vikaskurapati, line 101: This may not be realted to this PR but One question I always had is why output the MPI rank in the receiver files. Do we really need it? Can we get away with it and just have the receiver number?

davschneller: I think by now, we might be able to.

src/Parallel/MPI.h
```diff
@@ -29,7 +29,7 @@
 namespace seissol {
 
 #ifndef USE_MPI
-typedef MPIDummy MPI;
+using MPI = MPIDummy;
 #else // USE_MPI
```
vikaskurapati, line 33: Do we still need these ifdefs? Are they used here for the submodules?

vikaskurapati (approved):

src/Parallel/MPI.h
davschneller: Ah, good point. I think I've forgotten those.

src/DynamicRupture/Output/OutputManager.cpp
davschneller: Then they shall be gone as well.

(just for _absolutest_ clarity; did you ever notice a receiver ID appearing twice?)

I've also adjusted the validation scripts now; will test them later today.

vikaskurapati: As far as I remember, no. But the receiver file names having the MPI ranks in the end were kinda annoying. If they were meant to be there for a historic reason, I don't know. @Thomas-Ulrich, do you have any idea?

PR #1354, opened 2025-05-28
Clean up the Memory Implementation a bit

davschneller:
Part 1 of streamlining the memory subsystem a bit; in a preparation to make it more flexible (e.g. allow multiple GPUs or other devices per rank, or something similar).

In particular, this PR does:

* unify `Variable<T>`, `Bucket`, `Scratchpad` in their access and definition. (i.e. `ltsTree.add` and `ltsTree.var` for everything)
* make the `LTSTree` carry all info about the data; the `Variable<T>` objects, are now merely handles (i.e. a heap integer). Thus, referring to a variable that has not been added to the storage before (which is a pretty mean bug) should not be possible anymore.
* rename some functions to make them shorter. Also, remove some few ifdefs.
* first steps towards the adoption of storage variables with different types per layer.
* reformulation of some loops from cell-elementwise form to a layer-wise form. Will make the adoption to differently-typed layers easier, and makes the computations slightly more local in terms of memory.
* introduce an alternative to the `Interface`s and `Loader` structs (cf. `InterfaceHelper.h`). Not yet propagated.
* introduce a way to utilize type traits for the storage definition [1], similarly to what mneme does (cf. e.g. #344 ). However, it's not yet propagated either (since it'll need adjustments in the whole codebase). The difference between mneme and the current implementation here is that all typing is always still partly dynamic (i.e. there's a map that maps type IDs to storage indices; no parameter pack-based static mapping). Thus, all variables will still need to be added explicitly right now (but maybe that's already good enough). ... However, the whole process may still be a first step towards the long-term adoption of mneme (probably still off-limits by now), or some other form of unification in the storage system.

A future PR will remove the `InterfaceHelper.h` as well (partially done; but the combined PR got pretty large). Same for propagating the type traits to remove the `lts` object reference everywhere.

Besides, we also fix 2-3 non-critical bugs which occur only with assertions enabled.

[1] That means: instead of having `Variable<real[123]> dofs;` and `ltsTree.var(lts.dofs)`, you'll get `struct LTS { struct Dofs : public VariableDescriptor { using Type = real[123]; }; };` and `ltsTree.var<LTS::Dofs>()`.
Big advantage: no more `LTS` objects to carry around; thus one parameter less basically everywhere. (side note: with the new setup, we could also go static with the first setup to a similar effect; but that wouldn't be that pretty either)


src/Memory/Tree/LTSInternalNode.h
```diff
@@ -128,10 +128,10 @@ class LTSInternalNode : public Node {
 
   [[nodiscard]] ConstLeafIterator endLeaf() const { return ConstLeafIterator(end()); }
 
-  [[nodiscard]] unsigned getNumberOfCells(LayerMask layerMask = LayerMask()) const {
-    unsigned numCells = 0;
+  [[nodiscard]] std::size_t size(LayerMask layerMask = LayerMask()) const {
+    std::size_t numCells = 0;
```
vikaskurapati, line 132: Why the change of the function name from getNumberOfCells to size? The earlier was a better function name, IMO.

davschneller: Two reasons were there:

* it's used a lot, so a shorter name makes the code a bit more readable—`getNumberOfCells` is just not really worth the characters for querying the size/count of a datastructure IMO.
* for why `size` to be precise: it's more in line with e.g. `std::vector` or `std::array` which also use `size` to indicate the number of elements/cells they carry.

PR #1297, opened 2025-03-09
A first High-Order Material Implementation

davschneller:
We add the possibility to run simulations with a high-order (i.e. non-constant) material. Warning: I'm still lacking proper test cases for the new features. Still, tpv5 works as before, and tpv33 shows slightly different energy values (as one might expect).

The approach is based on https://academic.oup.com/gji/article/182/1/250/561912?login=false (suggestion came from Thomas; in December 2023—sorry for taking so long to get to it now).
That is, we allow the flux matrices to take on a polynomial form; essentially amounting to a still linear, but slightly more involved PDE, and do the same for the flux matrices.
The original paper only showed 2D simulations; we extend them canonically to 3D (which required no change).
Also, the implementation is extended to the DR—again, rather canonically by just updating the flux and projection matrices. But: we now also allow the impedances to differ per point. (currently, the flux matrix eigenvalues are interpolated; which may not be super ideal—I may need to revamp that more in the poroelastic style and compute the eigenvalues for each boundary point manually; the compatibility still needs to be checked)

The current implementation is pretty neatly integrated thanks to what #1275 allows us to do: by specifying a material on top of another material (sort of a "meta material"). It might also be thinkable to pretty easily implement e.g. fused materials that way.
As a side effect, it shouldn't take much work to make this feature here work with all given equations in SeisSol (in principle, it's already possible to run all equations like that—but it's completely untested (only for poro it's known to still take one or two commits))

Besides, we compress the matrix files.

Current caveats and TODOs:

* only interpolation possible for the initial conditions (that should be fixable easily)
* the point sources will still need a tiny bit of adjustment (interpolating the stiffness tensor—currently, we take the average only)
* adjust the volume energy calculation
* needs `-DMEMORY_LAYOUT=dense.xml` (working on it; it'll require a more involved sparse matrix implementation in the Yateto CPU backend)
* it's slow. Some Yateto optimizations failed; cf. also the previous point. However, it's of course also a ton of additional work to be done.

The GPU implementation might already work out of the box with #1284 ; that's still out to test.

If you want to test the implementation; you're very welcome to do so. Just be aware that you'll need more computing power than usual and currently expected.

PR #1361, opened 2025-06-12
Add some more material infos

davschneller:
Some more material infos; ported from #1297 .

* store the number of stress quantities available
* add a "post-initialization material initialization" method (for e.g. attenuation)


src/Equations/acoustic/Model/Datastructures.h
```diff
@@ -29,12 +29,17 @@ struct AcousticMaterial : public Material {
...
   static constexpr std::size_t NumElasticQuantities = 4;
   static constexpr std::size_t NumberPerMechanism = 0;
+  static constexpr std::size_t TractionQuantities = 1;
   static constexpr std::size_t Mechanisms = 0;
   static constexpr MaterialType Type = MaterialType::Acoustic;
   static inline const std::string Text = "acoustic";
   // The stress-velocity formulation of the elastic model is reused.
   // By definition, the normal stress and pressure are negatives of each other.
   static inline const std::array<std::string, NumQuantities> Quantities = {"-p", "v1", "v2", "v3"};
+  static constexpr std::size_t Parameters = 1 + Material::Parameters;
```
vikaskurapati, line 39: This is slightly confusing. I think it is better to have parameters set to the values, i.e., 2 here, 22 in anisotropic, etc. I understand that rho is common, but from the first look it is not clear.

davschneller: Hmm; I get what you mean.

The reason I wrote it like this is that, suppose someone (for whatever strange reason) decides to either remove `rho` from the parent class (and doesn't add it here), or adds an additional parameter. Then they'd need to explicitly update this place as well—which is very easy to forget and an easy error source as a result.

... would it help if we rename the base `Material` class to something else instead—or still unclear?

PR #1364, opened 2025-06-16
Fix the receiver output of fused simulations with dynamic rupture. 

vikaskurapati:
The energy output, and plasticity fixes will come in future PRs. 

Draft because the compare-receivers python script still needs modifications


src/DynamicRupture/Output/ReceiverBasedOutput.h
```diff
@@ -54,8 +54,11 @@ class ReceiverOutput {
     size_t ltsId{};
     int nearestGpIndex{};
     int nearestInternalGpIndex{};
+    int gpIndexFused{};
```
davschneller: It would be good to rename that one to show that it's also relevant for non-fused simulations (and in fact, the main index now).
Maybe just `index`/`dataIndex`?

src/DynamicRupture/Output/ReceiverBasedOutput.h
```diff
@@ -54,8 +54,11 @@ class ReceiverOutput {
...
+    int internalGpIndexFused{};
```
davschneller, line 58: Do we need that one even? (we can still keep it IMO—if you need it in the future)

src/ResultWriter/EnergyOutput.cpp
```diff
@@ -326,6 +326,7 @@ void EnergyOutput::computeDynamicRuptureEnergies() {
         return timeDerivativeMinusHost + QSize * i;
       };
 #else
+      // TODO: do this once and reuse
```
davschneller: Clarify that this comment speaks to fused sims

src/DynamicRupture/Output/OutputManager.cpp
```diff
@@ -326,6 +314,45 @@ void OutputManager::initPickpointOutput() {
...
+
+    if (allReceiversInOneFilePerRank) {
+      suffix += std::to_string(pointIndex);
+    }
+
+    if constexpr (seissol::multisim::MultisimEnabled) {
+      suffix += std::to_string(simIndex);
+    }
+
+    return suffix;
```
davschneller, line 330: It might be good, even though that'll change the format again by a bit, to separate between digits; e.g. by a `-`. Suppose you have a lot of receivers (200 per rank) and fuse 16 simulations. Then, what would `quantity1010` refer to? Either simulation `10`, aggregated receiver `10`; or simulation `0`, in receiver `101`.

src/DynamicRupture/Output/OutputManager.cpp
```diff
@@ -348,28 +375,27 @@ void OutputManager::initPickpointOutput() {
...
           const auto& point = const_cast<ExtVrtxCoords&>(receiver.global);
 
           // output coordinates
-          file << "# " << globalIndex << "," << simIndex << " x1\t" << makeFormatted(point[0])
-               << '\n';
-          file << "# " << globalIndex << "," << simIndex << " x2\t" << makeFormatted(point[1])
-               << '\n';
-          file << "# " << globalIndex << "," << simIndex << " x3\t" << makeFormatted(point[2])
-               << '\n';
+          if (simIndex == 0) {
```
davschneller, line 390: Just adding it as a side comment: in case you have someone who _really_ uses a jagged output, this one here might omit info.
E.g. if someone would set the `indices` array so that only simulation with index 1 is outputted. In some future iteration of SeisSol.

Probably, it can still be left as it is for now—but it's IMO important to consider that as well.

src/DynamicRupture/Output/ReceiverBasedOutput.cpp
```diff
@@ -88,6 +88,7 @@ void ReceiverOutput::calcFaultOutput(
 #pragma omp parallel for
 #endif
   for (size_t i = 0; i < outputData->receiverPoints.size(); ++i) {
+    // TODO: query the dofs, only once per simulation
```
davschneller: That's not only a TODO for once per simulation, but once per face.
(note: the GPU aggregators already do exactly that; but on the CPU, we copy for each point again)

src/DynamicRupture/Output/ReceiverBasedOutput.cpp
```diff
@@ -433,25 +449,26 @@ real ReceiverOutput::computeRuptureVelocity(Eigen::Matrix<real, 2, 2>& jacobiT2d
 
     auto* rt = getCellData(local, drDescr->ruptureTime);
     for (size_t jBndGP = 0; jBndGP < misc::NumBoundaryGaussPoints; ++jBndGP) {
-      const real chi = chiTau2dPoints(jBndGP, 0);
-      const real tau = chiTau2dPoints(jBndGP, 1);
+      const real chi = seissol::multisim::multisimTranspose(chiTau2dPoints, 0, jBndGP);
```
davschneller: `seissol::multisim::multisimTranspose(chiTau2dPoints, jBndGP, 0);`; same in the line below. (unless I missed something here and it's transposed in the non-fused case)

src/DynamicRupture/Output/ReceiverBasedOutput.cpp
```diff
@@ -433,25 +449,26 @@ real ReceiverOutput::computeRuptureVelocity(Eigen::Matrix<real, 2, 2>& jacobiT2d
...
     auto m2inv =
         seissol::init::M2inv::view::create(const_cast<real*>(seissol::init::M2inv::Values));
     for (size_t d = 0; d < NumDegFr2d; ++d) {
       projectedRT[d] *= m2inv(d, d);
     }
 
-    const real chi = chiTau2dPoints(local.nearestInternalGpIndex, 0);
-    const real tau = chiTau2dPoints(local.nearestInternalGpIndex, 1);
-
+    const real chi =
```
davschneller, line 468: Cf. the comment above; the `multisimTranspose` transposes for fused, and does _not_ transpose for serial.

davschneller (commented):
Looks good at first glance IMO; I've added some minor comments as a pre-review. :)

src/DynamicRupture/Output/ReceiverBasedOutput.h
vikaskurapati: How about `gpIndex`? I think the `gp` part should not be ignored. We could name it `gpIndex` as `index` is already taken

davschneller: Ok; sounds good to me.

src/DynamicRupture/Output/OutputManager.cpp
vikaskurapati: Modified the format. That means we need to modify the precomputed seissol once again. Will do that once the compare-receivers code is fixed

src/DynamicRupture/Output/OutputManager.cpp
vikaskurapati: Sure, if there is someone who would want to do that in the future, we could modify it.

vikaskurapati:
Modified the compare-receivers script, but alas that means the precomputed-seissol needs to be updated yet again. I will do that once everything seems fine, and then add this case in the CI too.

postprocessing/validation/compare-receivers.py
```diff
@@ -282,22 +289,26 @@ def faultreceiver_diff(args, i, quantities):
...
+                )
     return errors
 
 
 def find_all_receivers(directory, prefix, faultreceiver=False):
     if faultreceiver:
         file_candidates = glob.glob(f"{directory}/{prefix}-faultreceiver-*.dat")
     else:
-        file_candidates = glob.glob(f"{directory}/{prefix}-receiver-*.dat")
+        file_candidates = glob.glob(f"{directory}/{prefix}-receiver-*-*.dat")
```
davschneller: To make sure—the off-fault receivers contain the MPI index again?

davschneller (approved):
LGTM; one minor comment

postprocessing/validation/compare-receivers.py
vikaskurapati: Oh, thanks for pointing it out. It seems like I merged with master too late, which created the precomputed values with the older naming. But now, it is modified. But this script should take care of both older precomputed and newer ones, such that we don't need to update the entire repository. Please take a look and let me know if we can merge now.

src/DynamicRupture/Output/ReceiverBasedOutput.cpp
```diff
@@ -433,25 +449,29 @@ real ReceiverOutput::computeRuptureVelocity(Eigen::Matrix<real, 2, 2>& jacobiT2d
 
     auto* rt = getCellData(local, drDescr->ruptureTime);
     for (size_t jBndGP = 0; jBndGP < misc::NumBoundaryGaussPoints; ++jBndGP) {
-      const real chi = chiTau2dPoints(jBndGP, 0);
-      const real tau = chiTau2dPoints(jBndGP, 1);
+      // const real chi = seissol::multisim::multisimTranspose(chiTau2dPoints, 0, jBndGP);
+      // const real tau = seissol::multisim::multisimTranspose(chiTau2dPoints, 1, jBndGP);
```
davschneller: One more tiny comment: those two commented lines are a duplicate of those below

vikaskurapati: Done. Thanks!

PR #1366, opened 2025-06-19
Format the `src/Solver` folder and refactor it a bit

davschneller:
Since #1122 is taking a bit longer than (originally) anticipated, we hereby format the `src/Solver` ahead of time and adjust the folder name as well.
Also, we introduce some datastructures and changes that are already done in #1122, and just needed to be ported.
Besides, we format and tidy `Initializer/typedefs.h` .

Also, beginning preparations for multi-and-different-rate clustering.

(the remaining `src/Initializer` files will be done in an upcoming PR; the branch is `davschneller/init-refactor` )


src/Solver/TimeStepping/TimeCluster.cpp
vikaskurapati, line 1: It seems that the diff lost that these are renaming, and not new files created. Could you fix this before merging? The same with the other time-stepping files underneath.

src/Solver/TimeStepping/AbstractTimeCluster.cpp
vikaskurapati, line 1: It seems that the diff lost that these are renaming, and not new files created. Could you fix this before merging? The same with the other time-stepping files underneath.

src/Solver/TimeStepping/CommunicationManager.cpp
vikaskurapati, line 1: It seems that the diff lost that these are renaming, and not new files created. Could you fix this before merging? The same with the other time-stepping files underneath.

davschneller:
Ok; sure. Probably, that'll require an additional PR though (#1369 )—so that all changes are tracked a bit better.

davschneller:
Ok; #1369 has seemingly resolved the file replacements. Merging, once the tests have passed.

PR #1368, opened 2025-06-23
Rework the GPU Temporary Memory Allocations

davschneller:
(Dependent on #1354)
Fixes a memory leak when using e.g. plasticity (most likely due to the stream allocators, somehow—thus we switch to scratchpads for now)—and also removes three small, but redundant plasticity kernels.

Draft until #1354 is done, and until the special (elastic-acoustic) boundary conditions are also handled.


src/Initializer/MemoryManager.cpp
```diff
@@ -585,7 +590,7 @@ void seissol::initializer::MemoryManager::deriveRequiredScratchpadMemoryForWp(LT
       ++integratedDofsCounter;
 
       // include data provided by ghost layers
-      for (unsigned face = 0; face < 4; ++face) {
+      for (int face = 0; face < 4; ++face) {
```
vikaskurapati, line 593: why the change from unsigned to int? Is it better not to have a size_t here too?

davschneller: In total, it shouldn't matter too much in this very case since we're only counting to 4—i.e. it's mostly stylistic (`int` is shorter :) And seems a tiny bit more "common"). In fact, if the loop is short enough, there's a chance that the compiler will just unroll it; or at least _know_ that it's small enough to definitely fit into both an integer etc.

I just checked; in the `-O0` case, there's 1-2 extra instructions; but in an optimized build (`-O1` or higher), they shouldn't matter. Same for loops that are a tiny bit longer, like `1024`.

vikaskurapati: Sure, but I remember seeing diffs where int was changed to unsigned, for similar loops. What I am trying to say is, we need to pick one and stay consistent, such that we do not create unnecessary diffs in such trivial cases.

vikaskurapati (approved):

src/Initializer/MemoryManager.cpp
davschneller: Do you remember which commits those were? At least for me personally, I tend to not use `unsigned`/`std::size_t` in such small loops, _unless_ the upper bound is `unsigned`/`std::size_t`.

W.r.t. consistency: yes, fully aboard. Potentially, there might be something in one of the C++ styleguides with regard to these things; but I'm not 100%ly certain.

vikaskurapati: Unfortunately, no. But in any case, we could just fix the style to size_t, and keep it the same in all future commits. Anyway, if you disagree, feel free to merge, as this is not a big issue in this scenario. :)

PR #1374, opened 2025-06-25
Abstract the Time Bases used by the Solvers

davschneller:
Refactor the time integral/evaluation computation; but moving the relevant/differing logic into the coefficient computation. In essence, that brings the `STP` time basis to equal footing with the monomial basis and lets us remove some ifdefs.

Besides, it might also slightly correct the behavior at LTS cluster boundaries for non-monomial bases (in the STP case, a monomial basis seemed to be still used for computing time slices from bigger time clusters for the fluxes). ~EDIT: seems to also have affected the GTS case. (however, the actual difference in e.g. energy seems to be only marginal)~

EDIT 2: again only concerning STP: there's a different difference for the DR part concerning that `GaussLegendre` creates points in `[-1, 1]` ; but `sampleTimeBasisFunctions` expects them in `[0, 1]`. That's also changed in this PR; i.e. the points from `[-1, 1]` are now input to a function with domain `[-1, 1]` .


davschneller:
~The codecov still fails (but it works on my local machine; however I use `gcov` in version 15); but other than that—this one should be reviewable.~ Fixed by mitigation for now.

davschneller:
For poroelasticity/STP/"Legendre-in-time":

If we ignore the DR deviation (i.e. the map `[-1, 1]` to `[0, 1]` described in the "EDIT2" above), then the following shows where things differ:

Assume GTS; but do not compute the time-integrated buffers during the prediction operation; but compute it via the full space-time derivatives during correction instead. In practice, that means that one will need to set the flags `0b1000001111` in the LTS setup of each cell in the code manually; the code will do the rest for us. (NOTE: for LTS, the same setup will require more bits to be set; as we don't handle the larger-to-smaller cluster case correctly otherwise—setting something around `0b11110000` for all respective cells might help)

The results should still match afterwards; however, that's not the case for the current master:

* master ([d52dcea](https://github.com/SeisSol/SeisSol/commit/d52dcead99f66356721b9f167f4b72d9418e3bcc)) with this change deviates slightly (already in the approximated energies; which are basically only the "elastic subset" energy or similar—but it still provides a good control value).
* this PR gives about the same approximate "energy" values on the terminal; if we're ignoring the differences caused by the DR deviation.

Both compared to master with GTS (without `0b1000001111`).

src/Kernels/LinearCK/TimeBase.h
```diff
@@ -30,13 +30,15 @@ namespace seissol::kernels::solver::linearck {
 class Spacetime : public SpacetimeKernel {
   public:
   void setGlobalData(const CompoundGlobalData& global) override;
-  void computeAder(double timeStepWidth,
+  void computeAder(const real* coeffs,
```
vikaskurapati, line 33: As the function definition has now changed, can you add what the change of the variables means, and which variable denotes what here?

src/Kernels/LinearCK/TimeBase.h
```diff
@@ -61,44 +63,15 @@ class Spacetime : public SpacetimeKernel {
...
-                                     size_t numElements,
-                                     seissol::parallel::runtime::StreamRuntime& runtime) override;
-
-  void flopsTaylorExpansion(std::uint64_t& nonZeroFlops, std::uint64_t& hardwareFlops) override;
-
-  protected:
-#ifdef ACL_DEVICE
-  device::DeviceInstance& device = device::DeviceInstance::getInstance();
-#endif
+  void evaluate(const real* coeffs,
```
vikaskurapati, line 66: Adding what coeffs refer to, and what the function evaluate is evaluating would be a good thing to add to this function definition.

src/Kernels/TimeCommon.h
```diff
@@ -71,15 +47,15 @@ struct TimeCommon {
   static void computeIntegrals(Time& time,
                                unsigned short ltsSetup,
                                const FaceType faceTypes[4],
-                               double timeStepStart,
-                               double timeStepWidth,
+                               const real* timeCoeffs,
+                               const real* subtimeCoeffs,
```
vikaskurapati, line 51: The function definition changed, but not the documentation and explanation above. Please change that too, along with it.

src/Kernels/TimeCommon.cpp
```diff
@@ -68,51 +68,19 @@ void TimeCommon::computeIntegrals(Time& time,
       }
       // integrate the DOFs in time via the derivatives and set pointer to local buffer
       else {
-        time.computeIntegral(currentTime[dofneighbor + 1],
-                             currentTime[0],
-                             currentTime[0] + timeStepWidth,
-                             timeDofs[dofneighbor],
-                             integrationBuffer[dofneighbor]);
+        const auto* coeffs =
+            ((ltsSetup >> (dofneighbor + 4)) % 2) != 0 ? timeCoeffs : subtimeCoeffs;
```
vikaskurapati, line 79: Can you add a brief comment on what this line is doing?

src/Numerical/Functions.cpp
```diff
@@ -167,6 +168,38 @@ std::array<double, 3> gradTetraDubinerP(const std::array<unsigned, 3>& i,
           ddalpha(1.0, -1.0, 1.0, -1.0, 2.0)};
 }
 
+double shiftedLegendre(int n, double x, int d) {
+  // cf. https://en.wikipedia.org/wiki/Legendre_polynomials#Shifted_Legendre_polynomials
```
vikaskurapati, line 172: May be better to add this in the header file briefly also describing what this function is supposed to do

src/Numerical/Functions.h
```diff
@@ -122,6 +122,8 @@ double TetraDubinerP(const std::array<unsigned, 3>& i, const std::array<double,
 std::array<double, 3> gradTetraDubinerP(const std::array<unsigned, 3>& i,
                                         const std::array<double, 3>& xi);
 
+double shiftedLegendre(int n, double x, int d);
```
vikaskurapati, line 138: Documentation for this function like the remaining functions, please. Briefly describing what it is doing and then the link you put in the function can be helpful here.

src/Numerical/Functions.h
```diff
@@ -140,6 +142,10 @@ template <std::size_t D>
 std::array<double, D> gradDubinerP(const std::array<unsigned, D>& i,
                                    const std::array<double, D>& xi);
 
+template <std::size_t D>
+std::array<double, D> antigradDubinerP(const std::array<unsigned, D>& i,
```
vikaskurapati, line 167: Add brief explanation here for this function, please

src/Numerical/Quadrature.h
```diff
@@ -62,6 +62,22 @@ inline void GaussLegendre(double* points, double* weights, unsigned n) {
   }
 }
 
+inline std::pair<std::vector<double>, std::vector<double>>
```
vikaskurapati, line 71: The details of the function, how it is different from the above, and what each variable contains is a nice thing to have for this function.

src/Numerical/StableSum.h
vikaskurapati, line 1: This new structure needs documentation, too. An example usage of how one might use this in the codebase is something that will help.

src/Numerical/TimeBasis.h
vikaskurapati, line 1: This header file also has new functions, class definitions, etc. Similar to other files, documenting the intention behind each of them is required for someone continuing on this later.

src/Proxy/Allocator.cpp
```diff
@@ -146,8 +150,6 @@ void ProxyData::initGlobalData() {
   localKernel.setGlobalData(globalData);
   neighborKernel.setGlobalData(globalData);
   dynRupKernel.setGlobalData(globalData);
-
-  dynRupKernel.setTimeStepWidth(Timestep);
```
vikaskurapati, line 150: I am confused. Why don't we need to set the time step width anymore?

src/Solver/TimeStepping/TimeCluster.h
```diff
@@ -52,6 +52,7 @@ class TimeCluster : public AbstractTimeCluster {
   private:
   // Last correction time of the neighboring cluster with higher dt
   double lastSubTime;
+  double neighborTimestep;
```
vikaskurapati, line 58: Is this the timestep of the neighbor cluster? Or, something else?

vikaskurapati (changes requested):
Minor comments about documentation.

Once precomputed-seissol is updated, the change in results documented, and CI hash is updated, we are good to go in my opinion.

src/Proxy/Allocator.cpp
davschneller: That's now done with the calls to `ShiftedGaussLegendre` and `collocatePoints` at the respective places directly before the kernel OMP loops (same for `TimeCluster`); the respective data is passed as parameters.

In principle, this PR makes the `dynRupKernel` stateless.

davschneller:
Ok; added some documentation to (mostly) address the comments—thanks for the approve!

src/Solver/TimeStepping/TimeCluster.h
davschneller: Added a comment; it's the timestep of the neighbor with the largest cluster ID.

PR #1380, opened 2025-06-28
Fix Elastic-Acoustic Checkpointing

davschneller:
That is: move the face displacements to the surface memory storage.
As a result, we don't have a memory bucket anymore; but a separate storage structure. Also, we can cut some code in the memory manager.

Besides, fix the Free-Surface-Gravity BC on GPUs after #1368 (a scratchpad allocation was missing there).

### DIFF
--- a/src/Initializer/InitProcedure/InitIO.cpp
+++ b/src/Initializer/InitProcedure/InitIO.cpp
@@ -365,9 +365,10 @@ void setupOutput(seissol::SeisSol& seissolInstance) {
       }
     });
 
-    writer.addCellData<uint32_t>("locationFlag", {}, [=](uint32_t* target, std::size_t index) {
-      target[0] = surfaceLocationFlag[index];
-    });
+    writer.addCellData<std::uint8_t>(
+        "locationFlag", {}, [=](std::uint8_t* target, std::size_t index) {
+          target[0] = surfaceLocationFlag[index];
+        });
 
     writer.addCellData<std::size_t>("global-id", {}, [=](std::size_t* target, std::size_t index) {
       const auto meshId = surfaceMeshIds[index];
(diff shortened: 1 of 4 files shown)


davschneller:
I've tested it locally with a elastic-acoustic mesh with CPU+GPU locally rather briefly (which is to be added to the CI); though a "default" tpv5 should effectively suffice as well.

vikaskurapati:
> I've tested it locally with a elastic-acoustic mesh with CPU+GPU locally rather briefly (which is to be added to the CI); though a "default" tpv5 should effectively suffice as well.

Sounds good. It may be a good idea to put that in precomputed-seissol and in the CI tests already with this PR so that the upcoming PRs or any development is tested for this feature too.

davschneller:
Ok—sounds good to me. It's actually about due time to add that to the CI as well (2-3 years now probably; but I didn't really have one ready earlier).

However, then it might be best to get #1400 (and the ~respective Xdmfwriter PR~ handled that one myself) in first; they will add some additional testing features which should turn out useful for the elastic-acoustic verification (and others as well). Including e.g. volume and surface mesh comparison (but albeit not high-order yet); and all of that also in multi-rank scenarios.

Edit: https://github.com/SeisSol/precomputed-seissol/pull/15

davschneller:
Ok then; both #1400 and precomputed-seissol/15 are merged; I've now added TTPV1 to the CI here as well.

vikaskurapati:
> Ok then; both #1400 and precomputed-seissol/15 are merged; I've now added TTPV1 to the CI here as well.

OK. Let us wait for the tests to finish, and then we can merge

davschneller:
(the one test that's failing is due to the CI hardware backing it; not due to SeisSol—the same has happened on master today already as well)

PR #1381, opened 2025-06-28
Enforce the double-precision ref solution for CI test cases

davschneller:
For

* tpv16
* tpv105

the double precision reference solution matches better.

### DIFF
--- a/.github/workflows/build-seissol-cpu.yml
+++ b/.github/workflows/build-seissol-cpu.yml
@@ -193,42 +193,49 @@ jobs:
             energies: true
             receivers: true
             fault: true
+            alwaysdouble: false
           - name: tpv5-nuc
             equation: elastic
             multisim: 1
             energies: true
             receivers: true
             fault: true
+            alwaysdouble: false
           - name: tpv5-visco
             equation: viscoelastic2
             multisim: 1
             energies: true
             receivers: true
             fault: true
+            alwaysdouble: false
           - name: tpv5-poro
             equation: poroelastic
             multisim: 1
             energies: false # incomplete, hence disabled
             receivers: true
             fault: true
+            alwaysdouble: false
           - name: tpv6
             equation: elastic
             multisim: 1
             energies: true
             receivers: true
             fault: true
+            alwaysdouble: false
           - name: tpv13
             equation: elastic
             multisim: 1
             energies: true
             receivers: true
             fault: true
+            alwaysdouble: false
           - name: tpv16
             equation: elastic
             multisim: 1
             energies: true
-            receivers: true
+            receivers: false # possibly broken
             fault: true
+            alwaysdouble: true
           #- name: tpv33
           #  equation: elastic
           #  multisim: 1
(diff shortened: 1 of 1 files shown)


vikaskurapati:
Just to clarify, all the CI cases, even when run with single precision, will be compared to double precision solutions, is that correct? Would that not make the comparison off?

davschneller:
Yes, that's what this PR does. For tpv16 and tpv105, we'll compare against the double precision solution; even for the single precision computation.

For all others, we compare against the respective computation.

However:

* for tpv16, the behavior changed after #1333 (I honestly don't know why—I could only think of some more double precision being used); now the single-precision solutions match the double-precision solution
* for tpv105, it was changed by utilizing `expm1` in #1288 ; now there _also_ the single-precision solutions match the double-precision solution better.
  * however, the tpv105 reference solution seems to be broken anyways as well, since it was computed with the C++ DR, but without #1302 . That made the tests fail in all cases.
  * note that there's still a bug with TP and MPI in the receivers for some reason

vikaskurapati:
> Yes, that's what this PR does. For tpv16 and tpv105, we'll compare against the double precision solution; even for the single precision computation.
> 
> For all others, we compare against the respective computation.
> 
> However:
> 
> * for tpv16, the behavior changed after [Transfer some more defines to generated constexprs #1333](https://github.com/SeisSol/SeisSol/pull/1333) (I honestly don't know why—I could only think of some more double precision being used); now the single-precision solutions match the double-precision solution

In this case, updating the precomputed solution is better than this PR, IMO. Instead of adding yet another flag, and comparing two different things.

> * for tpv105, it was changed by utilizing `expm1` in [Add GPU-ISR, GPU-R+S/TP, [dry] CPU+GPU-R+S/SevWL (FL7); add a pure CUDA and HIP implementation for SeisSol #1288](https://github.com/SeisSol/SeisSol/pull/1288) ; now there _also_ the single-precision solutions match the double-precision solution better.
>   
>   * however, the tpv105 reference solution seems to be broken anyways as well, since it was computed with the C++ DR, but without [Update the DR Update Time Formulation to Match the FORTRAN Code #1302](https://github.com/SeisSol/SeisSol/pull/1302) . That made the tests fail in all cases.

If something is fixed, we can rather update the precomputed solution rather than comparing single precision with double precision. 

>   * note that there's still a bug with TP and MPI in the receivers for some reason

davschneller:
Well; it isn't really comparing two different things IMO; but rather comparing against a more accurate version—so in principle, our single-precision solutions got better.

But yup; I get what you mean. And I agree; it's better to update the reference solutions.

Besides, we'll definitely need to recompute for tpv105 anyways; then this should be gone (except for the MPI bug).
For tpv16—we could either copy double to single (that might already work) or recompute after verifying one more time.

vikaskurapati:
> Well; it isn't really comparing two different things IMO; but rather comparing against a more accurate version—so in principle, our single-precision solutions got better.
> 
> But yup; I get what you mean. And I agree; it's better to update the reference solutions.

Yes, and once they are done, may be someone from the geophysics side can confirm  that the updates make sense so that we are sure.

> 
> Besides, we'll definitely need to recompute for tpv105 anyways; then this should be gone (except for the MPI bug). For tpv16—we could either copy double to single (that might already work) or recompute after verifying one more time.

I would say recompute after verifying one more time is better. In my experience, it is better to track what exactly we are computing instead of copying the solutions.

davschneller:
Ok, then it's settled—great to have the recomputation scheduled.

I'm closing this PR here.

PR #1386, opened 2025-07-01
Unify DR output timing

davschneller:
Fix some CI bugs; like for tpv13 (fused+nonfused)—in essence set the output interval to the lowest-present DR cluster (as in the single-rank case).

May still need some very final confirmation testing (and the DR faultreceiver scheduling seems to be rather broken in general).

### DIFF
--- a/src/Solver/TimeStepping/ActorState.cpp
+++ b/src/Solver/TimeStepping/ActorState.cpp
@@ -98,8 +98,9 @@ bool DynamicRuptureScheduler::mayComputeInterior(long curCorrectionSteps) const 
 }
 
 bool DynamicRuptureScheduler::mayComputeFaultOutput(long curCorrectionSteps) const {
-  return curCorrectionSteps == lastCorrectionStepsInterior &&
-         curCorrectionSteps == lastCorrectionStepsCopy && curCorrectionSteps > lastFaultOutput;
+  return !hasDynamicRuptureFaces() ||
+         (curCorrectionSteps == lastCorrectionStepsInterior &&
+          curCorrectionSteps == lastCorrectionStepsCopy && curCorrectionSteps > lastFaultOutput);
 }
 
 void DynamicRuptureScheduler::setLastCorrectionStepsInterior(long steps) {


src/Solver/TimeStepping/ActorState.cpp
```diff
@@ -98,8 +98,9 @@ bool DynamicRuptureScheduler::mayComputeInterior(long curCorrectionSteps) const
 }
 
 bool DynamicRuptureScheduler::mayComputeFaultOutput(long curCorrectionSteps) const {
-  return curCorrectionSteps == lastCorrectionStepsInterior &&
-         curCorrectionSteps == lastCorrectionStepsCopy && curCorrectionSteps > lastFaultOutput;
+  return !hasDynamicRuptureFaces() ||
+         (curCorrectionSteps == lastCorrectionStepsInterior &&
+          curCorrectionSteps == lastCorrectionStepsCopy && curCorrectionSteps > lastFaultOutput);
```
vikaskurapati, line 103: I am sorry, I don't understand this one here. Does this mean even when the number of correction steps is not equal to the interior, copy, if it has a DR face, it computes the output now on?

vikaskurapati (approved):

src/Solver/TimeStepping/ActorState.cpp
davschneller: Actually, it only does so if it _doesn't_ have a DR face to it.

vikaskurapati:
I tested it on precomputed-seissol on my WS without MPI, and got the same results. If you have tested it with MPI, please go ahead and merge.

src/Solver/TimeStepping/ActorState.cpp
vikaskurapati: My bad. I misread it. So, now we write DR output, if it does not have a DR face in it, sounds kinda counterintuitive, but I see the point.

davschneller: It kind of is a bit counter-intuitive, yeah. It's sort of the following situation:

* we have DR faces -> we update our copy+interior layer, and output once that has happened.
* we don't have DR faces -> no updates happen anyways; so we don't need that check.

Though to be honest, the whole logic seems a bit counter-intuitive to me right now. As if it would be even better (though out of scope for this PR at least, and quasi-orthogonal to it) to just output each DR pickpoint at its individual timestep, and not at a global one (or rank-local as before)—or to really finally implement a full time-dependent pickpoint output (which might be pretty hard).

vikaskurapati:
> I tested it on precomputed-seissol on my WS without MPI, and got the same results. If you have tested it with MPI, please go ahead and merge.

Tested this with MPI on superMUC with 2 nodes for tpv13, and tpv13 fused cases from precomputed-seissol. The compare-receivers script shows relative errors less than 0.05. So we can merge this now.

PR #1387, opened 2025-07-01
Begin reworking the MPI comm

davschneller:
Introduce `MPI_Startall` where possible; and begin capsulating things away from the `MeshStructure*` pointers.

(`MPI_Test` is not replaced; just in case the MPI implementations aren't far enough already)

### DIFF
--- a/src/Solver/TimeStepping/AbstractGhostTimeCluster.h
+++ b/src/Solver/TimeStepping/AbstractGhostTimeCluster.h
@@ -14,12 +14,14 @@
 
 #include "Parallel/MPI.h"
 
+#include "Solver/TimeStepping/HaloCommunication.h"
+
 namespace seissol::time_stepping {
 class AbstractGhostTimeCluster : public AbstractTimeCluster {
   protected:
-  const int globalClusterId;
-  const int otherGlobalClusterId;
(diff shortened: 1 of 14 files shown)


src/Solver/TimeStepping/AbstractGhostTimeCluster.h
```diff
@@ -14,12 +14,14 @@
...
+#include "Solver/TimeStepping/HaloCommunication.h"
+
 namespace seissol::time_stepping {
 class AbstractGhostTimeCluster : public AbstractTimeCluster {
   protected:
-  const int globalClusterId;
-  const int otherGlobalClusterId;
-  const MeshStructure* meshStructure;
+  std::size_t globalClusterId;
+  std::size_t otherGlobalClusterId;
```
vikaskurapati, line 23: Not const anymore?

davschneller: Yes; it doesn't seem to be of much use. And also, due to the following check here: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html (currently disabled; but I think we should eventually enable it)

PR #1388, opened 2025-07-01
Cleanup math functions

davschneller:
The disambiguation is no longer required (as it seems); and the code has caused compiler errors recently.
Hence, it hereby vanishes.


NicoSchlw (approved):
Thanks, this fixes the compilation with ICC 2021.4.0

PR #1389, opened 2025-07-01
Update setup.sh

NicoSchlw:


Documentation/build-dependencies.rst
```diff
@@ -97,8 +97,11 @@ for building—adjust to your actual location.
 
 .. code-block:: bash
 
-  # For the Intel compiler
+  # Set your compiler, e.g., for the Intel compiler
   # source /opt/intel/compiler/VERSION/bin/compilervars.sh intel64
+  # or
```
davschneller: trailing whitespace (causes the sphinx linter to fail)

davschneller (approved):
LGTM; two comments:

* setting `CC=icx` and `CXX=icpx` may not always have the same effect as loading the `compilervars.sh` script. The latter may set more variables like library paths etc. (which can cause different behavior with regard to e.g. MPI or Netcdf)

* the sphinx linter needs to pass so that we can merge :)

NicoSchlw:
Hmm ok, is there a better way to make sure the right compilers are used everywhere? I had a hard time finding `compilervars.sh` on Supermuc

davschneller:
> Hmm ok, is there a better way to make sure the right compilers are used everywhere? I had a hard time finding `compilervars.sh` on Supermuc

I think `CC=icx` etc. should already be good. Or/and maybe loading the respective module.

Can't say it 100%ly for sure—however the module system should already take care of everything save for `CC,CXX` (but take everything I say here with a slight grain of salt—I'm not an LRZ employee :) ).

PR #1390, opened 2025-07-02
Remove memory copies from the DR (GPU) implementation

davschneller:
Remove the small memory copies (mainly quadrature coefficients and data pointers) initiated by the GPU/DR implementation—instead pre-allocate the necessary arguments per layer and pass the time coefficients as kernel parameters. Thus, we remove a bit of latency and prepare the friction law kernels so that they can be launched more "asynchronously".


src/Initializer/Typedefs.h
```diff
@@ -241,6 +241,12 @@ struct GlobalData {
 
   real* selectAne{nullptr};
   real* selectEla{nullptr};
+
+  real* resampleMatrix{nullptr};
```
vikaskurapati, line 246: One thing that would be useful when we create pointer based datastructures is the expected sizes of these as comments.

davschneller: That can be dangerous in case we plan to change things in the future. Comments like those are pretty easily forgotten I'm afraid (i.e. someone only changes the `GlobalData.cpp` and all other code; but not this file here).
And in the end, everyone just goes to `GlobalData.cpp` anyways to look up the definition. :)

(if those issues weren't the case—agreed; I've added the tensors they are representing as comment—that should give enough hints on where to find the size)

Maybe, we can refactor the whole `GlobalData` in the style of the `LTSTree` / `Storage` mid-term. Cf. #1408 a bit; I can try to integrate/reformat them in some subsequent PR. And also clean up the redundant code in the `GlobalData.cpp` .

PR #1392, opened 2025-07-02
Remove some more defines; reduce material data memory consumption

davschneller:
* Remove all but one `USE_ACOUSTIC` defines (and make the ITM material adjustment a bit slimmer). Not 100%ly sure yet about the getter/setter setup here—but it does its job for now.
  * alternative names (from #1297) would be `getRhoBar` instead of `getDensity` and `setMuLambda` instead of `setLameParameters` . Whatever you'd prefer.
* Remove one `USE_VISCOELASTIC2` define
* Reduce material memory consumption by not duplicating each material up to 4 times
* Add some more getters/setters for materials; in preparation of more general "meta materials" like from #1297 .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,10 +281,6 @@ target_compile_definitions(seissol-common-properties INTERFACE LOGLEVEL=${LOG_LE
 target_compile_definitions(seissol-common-properties INTERFACE LOG_LEVEL=${LOG_LEVEL_MASTER}
                                                      LOGLEVEL0=${LOG_LEVEL_MASTER})
 
-if(NUMBER_OF_FUSED_SIMULATIONS GREATER 1)
-  target_compile_definitions(seissol-common-properties INTERFACE MULTIPLE_SIMULATIONS=${NUMBER_OF_FUSED_SIMULATIONS})
-endif()
-
 # enable interproc. opts for small cores
 #if cpu in ['knc', 'knl', 'skx']:
 #    flags.extend(['-ip'])
--- a/codegen/kernels/point.py
+++ b/codegen/kernels/point.py
@@ -18,14 +18,15 @@ def addKernels(generator, aderdg):
     numberOfQuantities = aderdg.numberOfQuantities()
     # Point sources
     mStiffnessTensor = Tensor("stiffnessTensor", (3, 3, 3, 3))
-    mSlip = Tensor("mSlip", (3,))
     mNormal = Tensor("mNormal", (3,))
     mArea = Scalar("mArea")
     basisFunctionsAtPoint = Tensor("basisFunctionsAtPoint", (numberOf3DBasisFunctions,))
     basisFunctionDerivativesAtPoint = Tensor(
         "basisFunctionDerivativesAtPoint", (numberOf3DBasisFunctions, 3)
     )
-    mInvJInvPhisAtSources = Tensor("mInvJInvPhisAtSources", (numberOf3DBasisFunctions,))
+    mInvJInvPhisAtSources = Tensor(
+        "mInvJInvPhisAtSources", (numberOf3DBasisFunctions,), alignStride=True
+    )
     JInv = Scalar("JInv")
 
     generator.add(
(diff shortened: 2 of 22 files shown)


src/Equations/anisotropic/Model/Datastructures.cpp
```diff
@@ -181,4 +182,10 @@ double AnisotropicMaterial::getSWaveSpeed() const {
 }
 
 MaterialType AnisotropicMaterial::getMaterialType() const { return MaterialType::Anisotropic; }
+
+void AnisotropicMaterial::setLameParameters(double mu, double lambda) {
+  // no idea.
+  logError() << "Setting the Lamé parameters for anisotropic materials is not yet implemented.";
```
vikaskurapati, line 188: Just FYI: if the comment was about ITM simulations, we will need to scale all the material tensor coefficients. I had a bachelor's student who worked on it, and she could not get different waves to reflect separately.

src/Equations/elastic/Model/Datastructures.h
```diff
@@ -90,6 +90,8 @@ struct ElasticMaterial : Material {
   [[nodiscard]] double getSWaveSpeed() const override { return std::sqrt(mu / rho); }
 
   [[nodiscard]] MaterialType getMaterialType() const override { return Type; }
+
+  void setLameParameters(double mu, double lambda) override { this->lambda = lambda; }
```
vikaskurapati: For elastic materials, this needs to have a this->mu = mu;

src/Initializer/TimeStepping/LtsLayout.cpp
```diff
@@ -724,16 +724,17 @@ void seissol::initializer::time_stepping::LtsLayout::getCellInformation( CellLoc
...
         unsigned int l_clusterId = m_clusteredCopy[l_cluster][l_region].first[1];
 
+        const auto rank = m_clusteredCopy[l_cluster][l_region].first[0];
+
         // set mapping invalid
         o_ltsToMesh[l_ltsCell] = std::numeric_limits<std::size_t>::max();
 
         secondaryInformation[l_ltsCell].clusterId = l_clusterId;
-        secondaryInformation[l_ltsCell].meshId = l_ghostCell;
+        // secondaryInformation[l_ltsCell].meshId will be set later
```
vikaskurapati: I don't get the point of this comment. Is this a hint to yourself for later?

davschneller: It should've meant that it'll be set somewhere later in the method. Adjusted it, to hopefully be clearer in meaning.

(the whole file should (hopefully) be finally obsoletified by #1411 )

src/Equations/elastic/Model/Datastructures.h
davschneller: Oh no, what did I miss there... Thanks so much for spotting it.

src/Equations/anisotropic/Model/Datastructures.cpp
davschneller: Ah, I see. ... Can we get even more waves with anisotropy (I'm probably totally out of the loop here)?

Any suggestions on how the interface to this function might evolve? (in case we should adjust right now already—no need to though)

vikaskurapati: > Ah, I see. ... Can we get even more waves with anisotropy (I'm probably totally out of the loop here)?
> 

Yes, in isotropic elastic materials, we have two kinds of waves -- P-wave and  S-wave. But the S-waves generated have the same eigenvalues (wave speeds). So when you simulate it, in the end, we see two kinds of waves. But that is not the case anymore in the case of anisotropic elastic materials. Here we observe quasi-P and quasi-S waves, and the quasi-S waves generally have different eigenvalues(wave speeds) depending on the polarization, causing them to be observed as two separate waves. This sort of creates problems for ITM. In isotropic elastic materials, the eigenvectors(a physical term called impedance affects these almost directly) are sort of decoupled, i.e., we can change the materials with some arithmetic such that the impedance of one wave changes but not for the other, or cause both of them to reflect, etc. But for anisotropic, as we do not have a direct analytical eigen vector calculation, we could not reproduce reflecting different waves as per choice, but we could only reflect all waves successfully in the initial test cases. 

> Any suggestions on how the interface to this function might evolve? (in case we should adjust right now already—no need to though)

As of now, I would suggest not going into this already when we do not know if this is going to work. But if it works, the interface would basically need us to modify all the terms of the material tensor c_{ij}.

PR #1395, opened 2025-07-05
PUML Periodic Boundary Conditions

davschneller:
Implement periodic boundary conditions for PUML by splitting into a geometric and a topological mesh. (which makes the seemingly hard-to-implement problem pretty simple to get done)

In effect, two options for entering the topological mesh are possible right now:
* add a new connectivity array which contains identified vertices
* add a vertex identification array

Note that the periodicity is also included in the partitioning already.

Needs the respective PUML PR to be merged first (and still needs one bug fixed).

### DIFF
--- a/src/Geometry/PUMLReader.cpp
+++ b/src/Geometry/PUMLReader.cpp
@@ -10,6 +10,7 @@
 #include <Common/Iterator.h>
 #include <Geometry/MeshReader.h>
 #include <Initializer/Parameters/MeshParameters.h>
+#include <PUML/Topology.h>
 #include <PUML/TypeInference.h>
 #include <PUML/Upward.h>
 #include <algorithm>
@@ -103,7 +104,8 @@ inline std::string bcToString(int id) {
  * @param sideBC: boundary condition tag at the side to check
  * @param cellIdAsInFile: Original cell id as it is given in the h5 file
  */
-inline bool checkMeshCorrectnessLocally(const PUML::TETPUML::face_t& face,
+template <PUML::TopoType Topo>
+inline bool checkMeshCorrectnessLocally(const typename PUML::PUML<Topo>::face_t& face,
                                         const int* cellNeighbors,
                                         int side,
                                         int sideBC,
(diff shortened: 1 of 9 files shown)


davschneller:
(see https://github.com/SeisSol/Meshing/pull/49 for a prototype implementation; replacing `cube_c` for Netcdf)
(the implementation into PUMgen for Simmodeler and GMSH is still a "TODO" by now)

docs/PUML-mesh-format.rst
```diff
@@ -87,13 +105,26 @@ The following convention for defining a face ID is used:
 
    s_vert[0,:] = [0,2,1];   s_vert[1,:] = [0,1,3];    s_vert[2,:] = [1,2,3]; s_vert[3,:] = [0,3,2];
 
-Netcdf Input Format (Deprecated)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+In general,
```
Thomas-Ulrich: missing end of sentence?

docs/PUML-mesh-format.rst
```diff
@@ -87,13 +105,26 @@ The following convention for defining a face ID is used:
...
+
+Periodicity is not encoded via the above boundary conditions;
+instead the PUML file contains additional topology
+information in these cases.
+
+The default (i.e. for non-periodic meshes) is to take the geometric mesh as topological mesh (topology format ``geometric``).
+
+There are two ways to encode periodicity:
+
+- cell-wise: i.e. an extra, "topological" connectivity array, called ``topology``, in the mesh file. (topology format ``identify-cell``)
```
Thomas-Ulrich: `periodicity-cellmap and format topology-from-cells`
or even better 
`periodicity-facemap and format topology-from-faces`

and same with vertex?

docs/parameters.par
```diff
@@ -104,6 +104,7 @@ PPFileName = 'tpv33_faultreceivers.dat'
 &MeshNml
 MeshFile = 'tpv33_gmsh'         ! Name of mesh file
 pumlboundaryformat = 'auto'      ! the boundary data type for PUML files
+pumltopologyformat = 'auto'      ! the topology type for PUML files
```
Thomas-Ulrich: maybe add (auto, Geometric, IdentifyCell or IdentifyVertex, the latter two for periodic boundaries)?

src/Geometry/MeshReader.cpp
```diff
@@ -351,4 +352,37 @@ void MeshReader::computeTimestepIfNecessary(const seissol::SeisSol& seissolInsta
   }
 }
 
+void MeshReader::verifyMesh() {
```
Thomas-Ulrich: verifyMeshOrientation?

src/Initializer/InitProcedure/InitMesh.cpp
```diff
@@ -209,6 +257,21 @@ void readMeshPUML(const seissol::initializer::parameters::SeisSolParameters& sei
     logInfo() << "Using boundary format: i32x4 (4xi32)";
   }
 
+  if (topologyFormat == seissol::initializer::parameters::TopologyFormat::Auto) {
+    logInfo() << "Inferring topology format; i.e. set it to the connectivity.";
```
Thomas-Ulrich: Why would this case ever occur?
Auto should have been inferred to something more specific above.

src/Geometry/PUMLReader.cpp
```diff
@@ -239,32 +310,41 @@ void PUMLReader::partition(PUML::TETPUML& puml,
...
-void PUMLReader::getMesh(const PUML::TETPUML& puml) {
+void PUMLReader::getMesh(const PumlMesh& meshTopology, const PumlMesh& meshGeometry) {
   SCOREP_USER_REGION("PUMLReader_getmesh", SCOREP_USER_REGION_TYPE_FUNCTION);
 
   const int rank = MPI::mpi.rank();
 
-  const std::vector<PUML::TETPUML::cell_t>& cells = puml.cells();
-  const std::vector<PUML::TETPUML::face_t>& faces = puml.faces();
-  const std::vector<PUML::TETPUML::vertex_t>& vertices = puml.vertices();
+  const std::vector<PumlMesh::cell_t>& cells = meshTopology.cells();
```
Thomas-Ulrich, line 336: so what would be the difference between meshTopology.cells() and meshGeometry.cells().
I m not sure I can follow.

davschneller: E.g. if you look at a periodic cube:

* the geometric mesh will just give you the normal cube, as shown in e.g. paraview.
* the topological mesh will connect the cells over the boundaries. (the vertex coordinates have no real meaning here anymore; but they still exist for now—until further refactoring of PUML, probably)

(not 100%ly super sure about the terminology; but it's true that only one of the meshes carries the geometric info—and the other de facto the topological data that's actually being used)

src/Initializer/InitProcedure/InitMesh.cpp
davschneller: You're right, it doesn't anymore. Removed now.

src/Geometry/MeshReader.cpp
davschneller: Done (in the upcoming commit)

docs/parameters.par
davschneller: Ok, done in the upcoming commit

docs/PUML-mesh-format.rst
davschneller: `periodicity` might not be fully correct I'm afraid; since we're not constricted to "just" periodicity. (were we ever for the Netcdf format though?)
E.g. what's also doable is: have several, disjoint cubes which are connected to another by topology.
I think we could even implement a Moebius band-like topology with the technique here.

Hence, I think `identify` describes what we're doing better than `periodicity`.

davschneller:
Thanks for the review; and yes—it's that PUML2 PR.

src/Geometry/PUMLReader.cpp
Thomas-Ulrich: ok, so meshTopology contains only a face look-up table or a vertex look-up table, and meshGeometry contains the usual mesh information (vertex, connectivity, boundaries group)?

docs/PUML-mesh-format.rst
```diff
@@ -77,7 +95,7 @@ Other boundary formats (``int64``) have a 16-bit offset and use 0xffff as a mask
 | 3: dynamic rupture
 | 4: dirichlet
 | 5: absorbing
-| 6: periodic
+| 6: periodic (behaves similarly to 0)
```
Thomas-Ulrich: then maybe consider:

identified (used for topological face mapping, e.g. periodic or stitched domains)

docs/PUML-mesh-format.rst
```diff
@@ -87,13 +105,24 @@ The following convention for defining a face ID is used:
 
    s_vert[0,:] = [0,2,1];   s_vert[1,:] = [0,1,3];    s_vert[2,:] = [1,2,3]; s_vert[3,:] = [0,3,2];
 
-Netcdf Input Format (Deprecated)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Periodic Boundaries
```
Thomas-Ulrich: Topological Connections (for periodic or stitched domains)

docs/PUML-mesh-format.rst
```diff
@@ -87,13 +105,24 @@ The following convention for defining a face ID is used:
...
+-------------------
+
+Periodicity is not encoded via the above boundary conditions;
```
Thomas-Ulrich: Periodicity -> Topological Connections

docs/PUML-mesh-format.rst
```diff
@@ -87,13 +105,24 @@ The following convention for defining a face ID is used:
...
+instead the PUML file contains additional topology
+information in these cases.
+
+The default (i.e. for non-periodic meshes) is to take the geometric mesh as topological mesh (topology format ``geometric``).
```
Thomas-Ulrich: i.e. no topological Connections

docs/PUML-mesh-format.rst
Thomas-Ulrich: then rename topology to topological-connections.
and identify-cell to identify-face?

Thomas-Ulrich (approved):
Thanks here are some follow up comments

PR #1397, opened 2025-07-08
Forward the "How To Cite" to the README and the Docs

davschneller:
Forward the new "How To Cite" page to:

* the README
* the readthedocs

Also, a tiny bit of README cleanup. Furthermore, we remove the duplicate (and outdated) publication list from the readthedocs (and thus also the SeisSol repository).


davschneller:
(included into #1408 and merged via that one)

PR #1398, opened 2025-07-09
Fix SeisSol finalization for the low-order free-surface output

davschneller:


Aangniu (approved):
Tested that it solves the termination problem.

PR #1399, opened 2025-07-09
Normalize more file and folder names

davschneller:
* leading folders are to be always lower case. Sub-code folders are always uppercase. Current exceptions are: equation folders and the `src/tests` folder. (though I may some time soon just put the latter one level up)
* executables get their own directory (`app`); to unite the previous `src/Main.cpp` and the remaining `auto-tuning` folder containing only the proxy executable itself by now.
* `generated-code` becomes `codegen`; the generated files now get placed into `codegen/GeneratedCode` instead of `src/generated_code` as before. Thus, the build folder structure can now match the source folder structure in that regard. Also, a bit of header cleanup.

All just preparations for a larger CMake refinement.

### DIFF
--- a/.github/workflows/build-seissol-cpu.yml
+++ b/.github/workflows/build-seissol-cpu.yml
@@ -10,7 +10,7 @@ on:
   - push
 
 env:
-  precomputed-hash: 57577066fddab59afc5670397549b1b3e6d6e7ac
+  precomputed-hash: fb41141ee0b38b48b10cc74c8c817860d4a745c5
 
 jobs:
   seissol-build-test:
@@ -324,6 +324,14 @@ jobs:
             volume: false
             surface: false
           - name: tpv13-fused
+            equation: elastic
+            multisim: 8
+            energies: true
+            receivers: true
+            fault: false # incomplete, hence disabled
+            volume: false
+            surface: false
+          - name: tpv13-fused-plasticity
             equation: elastic
             multisim: 8
             energies: true
--- a/codegen/kernels/plasticity.py
+++ b/codegen/kernels/plasticity.py
@@ -45,21 +45,20 @@ def addKernels(generator, aderdg, matricesDir, PlasticityMethod, targets):
         alignStride=True,
     )
 
-    initialLoading = Tensor("initialLoading", (6,))
-
-    replicateIniLShape = (numberOfNodes,)
-    replicateIniLSpp = np.ones(
-        aderdg.Q.insertOptDim(replicateIniLShape, (aderdg.Q.optSize(),))
-    )
-    replicateInitialLoading = OptionalDimTensor(
-        "replicateInitialLoading",
+    initialLoading = OptionalDimTensor(
+        "initialLoading",
         aderdg.Q.optName(),
         aderdg.Q.optSize(),
         aderdg.Q.optPos(),
-        replicateIniLShape,
-        spp=replicateIniLSpp,
-        alignStride=True,
+        (6,),
     )
+
+    replicateIniLShape = (numberOfNodes,)
+    replicateIniLSpp = np.ones(
+        replicateIniLShape
+    )
+
+    replicateInitialLoading = Tensor("replicateInitialLoading", replicateIniLShape, spp=replicateIniLSpp, alignStride=True)
 
     iShape = (numberOfNodes, 6)
     QStressNodal = OptionalDimTensor(
@@ -170,17 +169,16 @@ def addKernels(generator, aderdg, matricesDir, PlasticityMethod, targets):
 
         if aderdg.multipleSimulations > 1:
             # for now, copy the tensors into here and rename them; until gemmforge/chainforge is deprecated
-            initialLoading = Tensor("initialLoadingM", (6,))
-            replicateInitialLoading = OptionalDimTensor(
-                "replicateInitialLoadingM",
+            initialLoadingM = OptionalDimTensor(
+                "initialLoadingM",
                 aderdg.Q.optName(),
                 aderdg.Q.optSize(),
                 aderdg.Q.optPos(),
-                replicateIniLShape,
-                spp=replicateIniLSpp,
-                alignStride=True,
+                (6,),
             )
-            matreplace = replicateInitialLoading["k"] * initialLoading["p"]
+            replicateInitialLoadingM = Tensor("replicateInitialLoadingM", replicateIniLShape, spp=replicateIniLSpp, alignStride=True)
+
+            matreplace = replicateInitialLoadingM["k"] * initialLoadingM["p"]
         else:
             # suffix `M` stands for `Matrix`
             replicateInitialLoadingM = Tensor(
--- a/docs/parameters.par
+++ b/docs/parameters.par
@@ -15,6 +15,8 @@ UseCellHomogenizedMaterial = 1
 !off-fault plasticity parameters (ignored if Plasticity=0)
 Plasticity=0
 Tv=0.05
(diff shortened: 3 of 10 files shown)


vikaskurapati:
> * leading folders are to be always lower case. Sub-code folders are always uppercase. Current exceptions are: equation folders and the `src/tests` folder. (though I may some time soon just put the latter one level up)
> * executables get their own directory (`app`); to unite the previous `src/Main.cpp` and the remaining `auto-tuning` folder containing only the proxy executable itself by now.
> * `generated-code` becomes `codegen`; the generated files now get placed into `codegen/GeneratedCode` instead of `src/generated_code` as before. Thus, the build folder structure can now match the source folder structure in that regard. Also, a bit of header cleanup.
> 
> All just preparations for a larger CMake refinement.

Hi,

I did a quick scroll, and most of the changes seem to be with naming. I will take a deeper look once again, but could you fix the CI compilation issues and merge conflicts, please? Thank you!

davschneller:
Ok; done. Thanks alot for the review in advance!

.github/workflows/restructuredtext-lint.yml
vikaskurapati, line 1: Just out of curiosity, why is this file deleted? Is it replaced by some other file?

docs/basic-code-structure.rst
```diff
@@ -33,15 +34,25 @@ SourceTerm    Everything related to kinematic rupture models.
...
+============= =============
+
+app/
+------------
+
+============= =============
+Folder        Description
+============= =============
+Proxy         The code directory for SeisSol Proxy. Compiled when compiling SeisSol.
+Main          The main SeisSol executable.
```
vikaskurapati: Sorry if I am nitpicking, but should this not be the "the code director for SeisSol executable" or something along those lines?

src/DynamicRupture/FrictionLaws/CpuImpl/LinearSlipWeakening.cpp
```diff
@@ -10,12 +10,13 @@
 #include "Kernels/Precision.h"
 #include "Memory/Descriptor/DynamicRupture.h"
 #include "Memory/Tree/Layer.h"
+#include <GeneratedCode/init.h>
+#include <GeneratedCode/kernel.h>
```
vikaskurapati: Something I always wondered when including headers in SeisSol, how do we decide "xxx.h" vs. <xxx.h> ?

And should we not arrange these in alphabetical order?

src/DynamicRupture/FrictionLaws/GpuImpl/FrictionSolverDetails.cpp
```diff
@@ -11,8 +11,8 @@
 #include "DynamicRupture/Misc.h"
 #include "Initializer/Parameters/DRParameters.h"
 #include "Kernels/Precision.h"
+#include <GeneratedCode/init.h>
```
vikaskurapati: Same as the above comment, how do we decide "xxx.h" vs. <xxx.h> ?

And should we not arrange these in alphabetical order?

src/DynamicRupture/Initializer/BaseDRInitializer.cpp
```diff
@@ -16,15 +17,14 @@
 #include "Memory/Tree/Layer.h"
 #include "Numerical/Transformation.h"
 #include "SeisSol.h"
-#include "generated_code/kernel.h"
 #include <Eigen/Dense>
 #include <Equations/Datastructures.h>
+#include <GeneratedCode/init.h>
```
vikaskurapati: In this file, both the extra headers have a different style. I am skipping this particular comment for all further files, to not sound redundant. May be another sweep to make them consistent would be nice.

src/sources.cmake
```diff
@@ -172,7 +172,6 @@ set(SYCL_ONLY_SRC_FILES
   ${CMAKE_CURRENT_SOURCE_DIR}/src/Kernels/PointSourceClusterOnDevice.cpp)
 
 target_compile_options(seissol-common-properties INTERFACE ${EXTRA_CXX_FLAGS})
-target_include_directories(seissol-common-properties INTERFACE ${CMAKE_CURRENT_BINARY_DIR}/src/generated_code)
```
vikaskurapati, line 175: Just for my understanding: why is this now not required anymore?

vikaskurapati (approved):
Thank you, LGTM. Most of the comments are about clarification questions.

.github/workflows/restructuredtext-lint.yml
davschneller: It was already disabled; effectively the `sphinx-lint.yml` does its job (almost).

The problem is that rstlint doesn't understand sphinx files, and thus throws a lot of errors.

(the RST linter was one of the tests in the [SQAAAS](https://sqaaas.eosc-synergy.eu/) tests; but it effectively didn't matter. But that's why it was added in the first place)

docs/basic-code-structure.rst
davschneller: Ok, let me change it. :)

src/DynamicRupture/FrictionLaws/CpuImpl/LinearSlipWeakening.cpp
davschneller: Two things happen here:

* `"???.h"` comes before `<???.h>` and uppercase before lowercase (`clang-format` sorts it that way)
* the decision is be more or less arbitrary for us—normally `<???.h>` is for externally-included headers; and `"???.h"` for local headers. In our case, someone unfortunately seems to have messed up this rule, and thus `clang-tidy` will include all headers via `<???.h>`. (maybe we should change that some day)

src/sources.cmake
davschneller: This include allowed us to do `#include <init.h>` or `#include <tensor.h>` instead of `#include <GeneratedCode/init.h>` and so on—the latter is now enforced.

src/DynamicRupture/Initializer/BaseDRInitializer.cpp
davschneller: Ok, let me change it.

(we may want to unify all of the headers some day—however it might be best to find out what causes `clang-tidy` to behave the way it does first)

src/DynamicRupture/FrictionLaws/CpuImpl/LinearSlipWeakening.cpp
vikaskurapati: So, do we consider the build time generated code as external or internal for SeisSol's internal files? We can enforce further development based on this decision.

davschneller: We treat them as internal if I get your question right. Both `"???.h"` and `<???.h>` work for them. Everything that we have the code of in the Seissol repo (or generated by it) can be included by a `"???.h"`; even submodules.

vikaskurapati: Okay. That is exactly what I wanted to know, thank you. I will try and keep that in mind. May be we can enforce `"xxx.h"` in further development/pull request reviews, in that case.

davschneller: Yes, definitely—I think it's best to fix clang-tidy/clangd first though (b/c those are the main source of new automatically-added includes). I can try to see if we can get those tools to use `"???.h"` via some simple CMake changes once this one here is merged (and maybe #1409 ).

PR #1400, opened 2025-07-11
Add global ID output

davschneller:
Adds global ID output. Still needs a little bit to complete (i.e. to add it to the high-order IO volume+surface); as well as the xdmfwriter PR. Maybe also fault script comparison adjustment.


davschneller:
Thanks for the review. Regarding the (off-fault) surface output: unfortunately, we don't have any (off-fault) surface output for any of the precomputed tests alas (there's no `tpv-surface.xdmf` anywhere but in the upcoming TTPV1). It's just been switched off when computing them. Hence, I had to disable them here. Interestingly though, the volume output was enabled—even if only for the velocity data.
The fault surface is covered by the `fault` test (looking for `tpv-fault.xdmf` which exists for all of them).

[maybe a TODO: to move the markers on what's present and what not into the precomputed repo]

PR #1403, opened 2025-07-15
Begin adding GPU sparsity support

davschneller:
Split by CPU and GPU sparsity configs.

Begin adding sparsity to GPU codegens; i.e. mainly for TensorForge now.
(so far, it only supported instruction-cache-level sparsity; i.e. encoding the values directly in the instruction sequence—effectively yielding something like a "matrix-free" variant of the kernels)


davschneller:
Would be good to also merge https://github.com/SeisSol/yateto/pull/98 .

cmake/process_users_input.cmake
```diff
@@ -31,10 +31,11 @@ option(LIKWID "Link with the likwid marker interface for proxy" OFF)
 option(INTEGRATE_QUANTITIES "Compute cell-averaged integrated velocity and stress components (currently breaks compilation)" OFF)
 option(ADDRESS_SANITIZER_DEBUG "Use address sanitzer in debug mode" OFF)
 
-option(TESTING "Compile unit tests" OFF)
-option(TESTING_GENERATED "Include kernel tests generated by yateto" OFF)
+option(TESTING "Compile unit tests and collect them for execution" OFF)
+option(TESTING_GENERATED "Include kernel unit tests as generated by yateto" OFF)
 option(COVERAGE "Generate targed for code coverage using lcob" OFF)
 set(TESTING_COMMAND "" CACHE STRING "Prefix to the test binary, so that CMake can list the tests")
+option(TESTING_BUILD_ONLY "Only compile the unit tests, but not collect them (only in effect if TESTING=ON)" OFF)
```
Thomas-Ulrich: "but do not"
maybe collect means run here?
(because I would think you need to collect first to be able to compile?)

Thomas-Ulrich (approved):
LGTM (for what I understand).

cmake/process_users_input.cmake
davschneller: It's sort of connected to that: `ctest` (the CMake test executor) needs to run the unit test system (doctest) once to get the names of all tests.
That can cause troubles if we're compiling e.g. for a different architecture—hence this option for the CI, to check if we at least compile.

PR #1404, opened 2025-07-16
Fix dynamic rupture friction energy calculations for multiple simulations

vikaskurapati:
The space weights were being copied wrong for multiple simulations which caused the outputs for energies to be wrong for multiple simulations.

### DIFF
--- a/src/DynamicRupture/FrictionLaws/FrictionSolverCommon.h
+++ b/src/DynamicRupture/FrictionLaws/FrictionSolverCommon.h
@@ -577,8 +577,7 @@ SEISSOL_HOSTDEVICE inline void computeFrictionEnergy(
 #ifndef ACL_DEVICE
 #pragma omp simd
 #endif
-    for (size_t index = Range::Start; index < Range::End;
-         index += Range::Step) {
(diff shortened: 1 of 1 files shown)


src/DynamicRupture/FrictionLaws/FrictionSolverCommon.h
```diff
@@ -573,12 +573,14 @@ SEISSOL_HOSTDEVICE inline void computeFrictionEnergy(
   using namespace dr::misc::quantity_indices;
   for (size_t o = 0; o < ConvergenceOrder; ++o) {
     const auto timeWeight = timeWeights[o];
-
 #ifndef ACL_DEVICE
 #pragma omp simd
 #endif
-    for (size_t index = Range::Start; index < Range::End; index += Range::Step) {
-      const size_t i{startIndex + index};
+    for (size_t index = Range::Start; index < Range::End;
```
davschneller: Maybe set the comment one line before the loop, so that it doesn't wrap around here.

And also, to note: the range statement is being used everywhere in this file actually in the same way—so maybe it would be good to put a comment near the definition of the range types, or you'd need to put it everywhere.

davschneller (approved):

PR #1405, opened 2025-07-16
Reduce OpenMP ifdefs; add receiver calculation onto GPU streams

davschneller:
* rework analytical boundary condition a bit; avoid re-allocating far too often
* encapsule OpenMP calls more idiomatically
* add the on-fault and off-fault receiver calculation onto some GPU streams

### DIFF
--- a/src/Kernels/LinearCK/Local.cpp
+++ b/src/Kernels/LinearCK/Local.cpp
@@ -96,6 +96,9 @@ struct ApplyAnalyticalSolution {
       nodesVec[i][2] = nodes[i * 3 + 2];
     }
 
+    // NOTE: not yet tested for multisim setups
+    // (only implemented to get the build to work)
+
     for (std::size_t s = 0; s < multisim::NumSimulations; ++s) {
       auto slicedBoundaryDofs = multisim::simtensor(boundaryDofs, s);
       initConditions->at(s % initConditions->size())
--- a/.github/workflows/build-seissol-cpu.yml
+++ b/.github/workflows/build-seissol-cpu.yml
@@ -10,7 +10,7 @@ on:
   - push
 
 env:
-  precomputed-hash: fb41141ee0b38b48b10cc74c8c817860d4a745c5
+  precomputed-hash: 4cc0993d4ba3b8f0aabc84af78e405af9d98f902
 
 jobs:
   seissol-build-test:
@@ -338,6 +338,14 @@ jobs:
             receivers: true
             fault: false # incomplete, hence disabled
             volume: false
+            surface: false
+          - name: loh1-ic
+            equation: elastic
+            multisim: 1
+            energies: true
+            receivers: true
+            fault: false
+            volume: true
             surface: false
         precision:
           - single
--- a/.github/workflows/build-seissol-gpu.yml
+++ b/.github/workflows/build-seissol-gpu.yml
@@ -10,7 +10,7 @@ on:
   - push
 
 env:
-  precomputed-hash: 57577066fddab59afc5670397549b1b3e6d6e7ac
+  precomputed-hash: 4cc0993d4ba3b8f0aabc84af78e405af9d98f902
 
 jobs:
   seissol-build-test:
@@ -350,6 +350,14 @@ jobs:
             volume: false
             surface: false
             order: 3
+          - name: loh1-ic
+            equation: elastic
+            multisim: 1
+            energies: true
+            receivers: true
+            fault: false
+            volume: true
+            surface: false
         precision:
           - single
         arch:


docs/initial-condition.rst
```diff
@@ -6,19 +6,68 @@
...
+    v2: 64
+    v3: 81
+
+Of course, one may use Lua functions or ASAGI as input as well for several variables.
+The parameter ``hastime`` indicates that the easi file has ``t`` as a time input variable besides the spatial coordinates (``x,y,z``).
+This way, the easi boundary condition can be used for comparing against after time has passed (e.g. for convergence tests).
+However, it requires easi 1.5.0 or higher to function correctly.
+
+As a caveat, the Easi initial condition currently does not support the analytical boundary condition.
+
```
vikaskurapati, line 49: Was the implementation of these initial conditions a part of this PR?

In any case, it will be nice if you add a test case for this to the precomputed-seissol repository. If there is no small test case, you could still add it and not put it in the CI tests so that anyone who suspects they broke something related to this feature could go back and test for it.

src/DynamicRupture/Output/Builders/PickPointBuilder.h
```diff
@@ -13,6 +13,7 @@
 #include "ReceiverBasedOutputBuilder.h"
 
 #include <Common/Iterator.h>
+#include <Parallel/Runtime/Stream.h>
```
vikaskurapati, line 16: Not related to this PR, but as we discussed in another PR, if we can slowly ensure that all new headers added are of the consistent style of `"xxx.h"`, it would be nice. I understand if you don't want to do it here, that is why this is left as a comment, and not as a request changes.

src/DynamicRupture/Output/DataTypes.h
```diff
@@ -15,6 +15,7 @@
 #include "Memory/Tree/Layer.h"
 #include "Parallel/DataCollector.h"
 #include <Eigen/Dense>
+#include <Parallel/Runtime/Stream.h>
```
vikaskurapati, line 18: Same as the header comment above.

src/DynamicRupture/Output/OutputManager.cpp
```diff
@@ -30,6 +30,7 @@
 #include "Memory/Tree/Lut.h"
 #include "ResultWriter/FaultWriterExecutor.h"
 #include "SeisSol.h"
+#include <Parallel/Runtime/Stream.h>
```
vikaskurapati, line 33: Same as the header comment above. I won't write more of these comments to avoid redundancy.

src/Kernels/LinearCK/Local.cpp
```diff
@@ -78,31 +79,32 @@ void Local::setGlobalData(const CompoundGlobalData& global) {
...
+      nodesVec[i][1] = nodes[i * 3 + 1];
+      nodesVec[i][2] = nodes[i * 3 + 2];
+    }
 
-      assert(initCondition != nullptr);
-      initCondition->evaluate(time, nodesVec, localData.material(), slicedBoundaryDofs);
+    for (std::size_t s = 0; s < multisim::NumSimulations; ++s) {
+      auto slicedBoundaryDofs = multisim::simtensor(boundaryDofs, s);
+      initConditions->at(s % initConditions->size())
+          ->evaluate(time, nodesVec, NodeCount, localData.material(), slicedBoundaryDofs);
```
vikaskurapati, line 105: Have you tested this with multiple simulations?

src/Solver/TimeStepping/TimeCluster.cpp
```diff
@@ -81,7 +79,7 @@ TimeCluster::TimeCluster(unsigned int clusterId,
     : AbstractTimeCluster(
           maxTimeStepSize, timeStepRate, seissolInstance.executionPlace(clusterData->size())),
       // cluster ids
-      usePlasticity(usePlasticity), seissolInstance(seissolInstance),
+      usePlasticity(usePlasticity), seissolInstance(seissolInstance), streamRuntime(4),
```
vikaskurapati, line 82: Just for my clarification, what does the 4 here refer to? Why exactly 4?

vikaskurapati (commented):
Looks good to me for OpenMP ifdefs, and receiver calculations parts. But the PR has code which is more than what is written in the PR description. Please add the tests which you used to verify the initial conditions' implementations, and easi boundary conditions so that it is complete.

docs/initial-condition.rst
davschneller: Ok, I'll do that.

src/DynamicRupture/Output/Builders/PickPointBuilder.h
davschneller: Ok thanks; yes—I'd want postpone it a bit, to be honest. Maybe until we switch `clang-tidy` etc. ; then after some time, we can switch it for all headers.

src/Solver/TimeStepping/TimeCluster.cpp
davschneller: The `4` means how many "parallel" streams we use. I.e. in some sections like the neighbor fluxes, we can use multiple streams since we have multiple kernels that don't depend on each other.

Since the usual default in hardware queues is 4 streams, we set the number to exactly that.

docs/initial-condition.rst
davschneller: Actually, it wasn't added in this PR but already last year. With the plan of someday (possibly in a 2.0.0) replacing all the current hard-coded initial conditions. Cf. https://github.com/SeisSol/precomputed-seissol/pull/19 .

It'll still need to be extended to the analytical BCs; I'm gonna do that probably next. (after #1405 and #1401 are done most likely)

src/Kernels/LinearCK/Local.cpp
davschneller: Compiled it mainly—the focus here was to update the implementation and avoid passing a vector down the line. I'm not sure if anyone has even tested just "default" initial conditions with multiple simulations recently (w/o imposing them at the boundary).

(maybe some test to add by you?)

I'm gonna refactor the analytical BC again anyway next (to add the easi boundaries for them as well); so I'm going to wait with adding a proper test case until then from my side—then I can also add a fused case for all of it if it hasn't happened until then already.

davschneller:
Tiny nitpickity: they _were_ referenced in the first bullet point. :)

If I were to split up all the PRs even more cleanly, we might have ended up with 50 open PRs at once.
I would have liked to do so; but since the whole reviewing takes time, it would've become really difficult to manage all open PRs (it is already a bit tedious now).

docs/initial-condition.rst
vikaskurapati: Sure, having an intermediate test-case to look back to is also nice. I will look at that soon.

src/DynamicRupture/Output/Builders/PickPointBuilder.h
vikaskurapati: Of course. What I meant was that we could keep it in mind for all newly added headers. But okay, I will stop raising this point again from now on.

src/Kernels/LinearCK/Local.cpp
vikaskurapati: > Compiled it mainly—the focus here was to update the implementation and avoid passing a vector down the line. I'm not sure if anyone has even tested just "default" initial conditions with multiple simulations recently (w/o imposing them at the boundary).
> 

I tried it by hacking it in a bit in the linked fused-simulations branch for some debugging. I unfortunately don't remember the full details but they seemed to have worked when all of them were the same. 

> (maybe some test to add by you?)

Sure, but I tend to go by necessity, so I will add them when someone needs them. 

> 
> I'm gonna refactor the analytical BC again anyway next (to add the easi boundaries for them as well); so I'm going to wait with adding a proper test case until then from my side—then I can also add a fused case for all of it if it hasn't happened until then already.

Sounds good to me. In that case, can I request you to add a small comment above this that this was done mainly to get the compilation to work, and that this has not been actually tested? Or do you think that comment is also redundant, given that no one else will probably work on it?

vikaskurapati:
> Tiny nitpickity: they _were_ referenced in the first bullet point. :)
> 

Sorry, seemed to have missed it.

> If I were to split up all the PRs even more cleanly, we might have ended up with 50 open PRs at once. I would have liked to do so; but since the whole reviewing takes time, it would've become really difficult to manage all open PRs (it is already a bit tedious now).

It seems that you misunderstood what I am trying to say. I don't want you to split up all the PRs. What I requested is to add whatever you used to test these new features somewhere so that anyone who modifies something related to those can test later to ensure they don't break anything. 

That being said, I am approving this now. You can consider the comments if you wish to modify something; otherwise, Thanks!

vikaskurapati (approved):

src/Kernels/LinearCK/Local.cpp
davschneller: Yes, that sounds good. I'll add it (yes, noone besides you (or me in exceptions) will work on it—but just in case we're far too busy to ever address it and leave it to our sucessors; though I don't hope that we'll get to that).

src/DynamicRupture/Output/Builders/PickPointBuilder.h
davschneller: By the way—I know now how to enforce the `"???.h"` for new headers via `clang-tidy` ; I'll need to rule out any possible side effects still.
But that means we could switch soon.

docs/initial-condition.rst
davschneller: Ok; included and updated.

davschneller:
Ok—sorry for misunderstanding; and thanks for the approval. I've addressed the comments—and plan to extend the easi boundary conditions pretty soon.

Also, I get what you mean—and I truly hope that we can accomplish a mostly-verified SeisSol version in a reasonable timeframe soon-ish. (even though we're all rather busy)

src/DynamicRupture/Output/Builders/PickPointBuilder.h
vikaskurapati: Sounds good. Then I will keep in mind to use `"xxx.h"` for any new included headers. Thanks!

PR #1406, opened 2025-07-16
Fix fault output interval, take 2

davschneller:
Unfortunately, #1386 went into the right direction; but ran the output twice—once for the interior, once for the copy layer. Thus: strategy switch; we now "emulate" the DR scheduler to run, even if it doesn't have any DR faces.

I've tested it with tpv13-fused locally, and it seems to work—whereas the current master does explicitly _not_ work.


davschneller:
... note that there seems to be a bigger problem in the background with the pickpoint fault output (hopefully not the DR itself) once MPI is involved. Might be a cluster scheduling issue actually. ... It's solvable however. Stay tuned for "take 3". Alas, I'm sorry.

PR #1407, opened 2025-07-17
Fix Pickpoint Output Scheduling

davschneller:
Fix a number of smaller issues around the pickpoint output:

* fix the iterations to always use a valid time slice, i.e. DOFs; and not just rely on the 0-th derivative. (which, in the STP case, actually stores the integral over the last time interval; not the DOFs of the previous timestep)
* switch the energy output and the fault output to use the current DOFs (not the previous ones which can be computed from the derivatives/STP—i.e. we now transfer the DOFs at each sync point where at least one of those is used)
* switch the point output to a time-based output interval, and output at a sync point only. (that does unfortunately _not_ yet entail a `pickdt` that's more variable; but it's a step towards that direction anyways)
* collect the pickpoint output per cluster, not for all clusters at the same time

For the latter point, a short explaination: there is/was a bug in the pickpoint output scheduling (probably already since at least 1.0.0); leading to slightly differing results between single-rank and multi-rank runs, as well as multi-rank runs among each other. Most likely this bug is also one part of the cause for the tpv105 tests failing.
Reason: so far, the pickpoint was always scheduled via the lowest cluster which had DR cells. However, thus, we depend on the cluster update ordering to function correctly in an MPI setting—which isn't given there. Maybe it's not even correct in the single-rank setting (but at least it's consistent there).
As a result, however, the pickpoint calculation always uses the "old" wavepropagation solution with the "new" DR output (instead of potentially mixing them). Can probably be adjusted if wanted. For GTS, both before this PR and after this PR, we get the same result in all cases.

### DIFF
--- a/src/Initializer/Parameters/OutputParameters.h
+++ b/src/Initializer/Parameters/OutputParameters.h
@@ -59,7 +59,7 @@ struct FreeSurfaceOutputParameters {
 
 struct PickpointParameters {
   int printTimeInterval{1};
-  double interval{VeryLongTime};
+  double writeInterval{VeryLongTime};
   std::array<bool, 12> outputMask{true, true, true};
   std::optional<std::string> pickpointFileName;
   bool aggregate{false};
@@ -70,7 +70,7 @@ struct ReceiverOutputParameters {
   bool enabled{false};
   bool computeRotation{false};
   bool computeStrain{false};
-  double interval{VeryLongTime};
+  double writeInterval{VeryLongTime};
   double samplingInterval{0};
   std::string fileName;
   bool collectiveio{false};
--- a/src/DynamicRupture/Output/Builders/PickPointBuilder.h
+++ b/src/DynamicRupture/Output/Builders/PickPointBuilder.h
@@ -171,7 +171,7 @@ class PickPointBuilder : public ReceiverBasedOutputBuilder {
   }
 
   void initTimeCaching() override {
-    const auto intervalOrEnd = std::min(pickpointParams.interval, endtime_);
+    const auto intervalOrEnd = std::min(pickpointParams.writeInterval, endtime_);
     const auto neededCacheLevel =
         static_cast<std::size_t>(std::ceil(intervalOrEnd / timestep_) + 1);
 
--- a/src/DynamicRupture/Output/Builders/ReceiverBasedOutputBuilder.h
+++ b/src/DynamicRupture/Output/Builders/ReceiverBasedOutputBuilder.h
@@ -28,7 +28,6 @@ namespace seissol::dr::output {
 class ReceiverBasedOutputBuilder {
   public:
   virtual ~ReceiverBasedOutputBuilder() = default;
-  void build(std::shared_ptr<ReceiverOutputData> outputData);
 
   void setMeshReader(const seissol::geometry::MeshReader* reader);
   void setLtsData(LTS::Storage& userWpStorage,
--- a/src/DynamicRupture/Output/OutputManager.cpp
+++ b/src/DynamicRupture/Output/OutputManager.cpp
@@ -299,12 +299,9 @@ void OutputManager::initPickpointOutput() {
   ppOutputBuilder->build(ppOutputData);
   const auto& seissolParameters = seissolInstance.getSeisSolParameters();
 
-  seissolInstance.pickpointWriter().enable(seissolParameters.output.pickpointParameters.interval);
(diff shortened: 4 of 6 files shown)


davschneller:
I tried it for our precomputed tpv105 locally (in double precision). The following (relative) errors to the reference occur now:

```
1 rank
  absolute slip friction coefficient peak sliprate  traction rupture time  sliprate      slip normal velocity  pressure temperature
1      1.145553             0.001797       1.19225  0.000293          0.0  0.024112  1.145559        0.007868  1.174437         0.0
2      0.049081              0.00403      0.025907  0.001374     0.024981  0.010894  0.049081        0.011172  0.041842    0.000324
3      0.886624             0.001546      1.037506  0.000284          0.0  0.039825  0.886652        0.008079  0.933297         0.0
4      0.006242             0.005323      0.004128   0.00229     0.008261  0.006479  0.006242        0.015058  0.004691    0.000657

8 ranks
  absolute slip friction coefficient peak sliprate  traction rupture time  sliprate      slip normal velocity  pressure temperature
1      1.145554             0.001797       1.19225  0.000293          0.0  0.024112  1.145559        0.007868  1.174437         0.0
2      0.049081              0.00403      0.025907  0.001374     0.024981  0.010894  0.049081        0.011172  0.041842    0.000324
3      0.886624             0.001546      1.037506  0.000284          0.0  0.039825  0.886652        0.008079  0.933297         0.0
4      0.006242             0.005323      0.004128   0.00229     0.008261  0.006479  0.006242        0.015058  0.004691    0.000657
```

They match really closely now (for a different example, compare the current end-to-end tests on master); but don't seem to be identical; see the first receiver in the absolute slip error. ... It _would_ be thinkable to somehow attribute this to a summing error; but possibly more is going on.

Thus, more debugging, probably.

(and yes, the errors are massive by now. The reason for that is twofold: firstly, the reference solution had a bug in it which made it use an incorrect timestep. Fixed in #1302 . Secondly, the difference introduced by the changes here; though they might be an oversight of mine (_really_ possible). Or maybe even not.)

src/Solver/TimeStepping/TimeManager.cpp
```diff
@@ -93,10 +93,13 @@ void TimeManager::addClusters(const initializer::ClusterLayout& clusterLayout,
                 MPI_MIN,
                 Mpi::mpi.comm());
 
+  const auto drOutputTimestep = drClusterOutput == std::numeric_limits<std::size_t>::max()
+                                    ? std::numeric_limits<double>::infinity()
```
Thomas-Ulrich, line 97: We may get the same issue as on LUMI-G:
https://github.com/SeisSol/SeisSol/issues/1451
(to keep in mind for later debugging ^^)

src/Initializer/Parameters/OutputParameters.cpp
```diff
@@ -119,7 +119,8 @@ PickpointParameters readPickpointParameters(ParameterReader* baseReader) {
   auto* reader = baseReader->readSubNode("pickpoint");
 
   const auto printTimeInterval = reader->readWithDefault("printtimeinterval", 1);
-  const auto maxPickStore = reader->readWithDefault("maxpickstore", 50);
```
Thomas-Ulrich, line 122: no idea what this parameter was doing anyway

src/Initializer/Parameters/OutputParameters.h
```diff
@@ -59,7 +59,7 @@ struct FreeSurfaceOutputParameters {
 
 struct PickpointParameters {
   int printTimeInterval{1};
-  int maxPickStore{50};
+  double interval{VeryLongTime};
```
Thomas-Ulrich: writeInterval would be clearer

src/Initializer/Parameters/OutputParameters.cpp
davschneller: That parameter was, in essence, controlling how frequently we write the onfault receiver/pickpoint data to disk. In the default case (and probably all where the default wasn't changed), it would be 50 cycles of the lowest-ID local time cluster (i.e. usually cluster 0, either copy or interior).

(IIRC that parameter wasn't even really documented in the docs)

Thus, the onfault receiver/pickpoint output was the only output that wrote output _during_ the running simulation so far, and not a synchronization points (that is, until this PR).

src/Solver/TimeStepping/TimeManager.cpp
davschneller: I guess we'll know once we face something in that area. ^^

PR #1408, opened 2025-07-22
Switch to Type-Based Variable Indexing

davschneller:
Effectively part 2 and a follow up to #1354 .

Notable changes:

* change accesses from variable to type based. I.e. `var(lts.cellInfo)` becomes `var<LTS::CellInfo>()`. Thus, we need to carry around one less argument in total.
* subsequently, also replace the `InterfaceHelper` defines by that. Also, we get closer to adapting mneme by changing to all of these constructs.
* flatten the receiver and point source to-cluster maps.
* begin flattening the LTS tree as a whole (since we don't need that much structure after all). Since constructs and orderings are still a bit temporary; I'm gonna replace those next.

A part 3 will follow.


docs/related-publications.rst
```diff
@@ -6,33 +6,21 @@
...
-
-Käser, M., M. Dumbser, J. de la Puente, and H. Igel (2007), An arbitrary high-order discontinuous Galerkin method for elastic waves on unstructured meshes - III. Viscoelastic attenuation, Geophys. J. Int., 168(1), 224-242, doi: `10.1111/j.1365-246X.2006.03193.x  <https://doi.org/10.1111/j.1365-246X.2006.03193.x>`_..
-
-Käser, M., and M. Dumbser (2006), An arbitrary high-order discontinuous Galerkin method for elastic waves on unstructured meshes - I. The two-dimensional isotropic case with external source terms, Geophys. J. Int., 166(2), 855-877, doi: `10.1111/j.1365-246X.2006.03051.x  <https://doi.org/10.1111/j.1365-246X.2006.03051.x>`_.
-
-Dumbser, M. and Käser, M. (2006), An arbitrary high‐order discontinuous Galerkin method for elastic waves on unstructured meshes – II. The three‐dimensional isotropic case. Geophysical J. Int., 167: 319-336. doi: `10.1111/j.1365-246X.2006.03120.x <https://doi.org/10.1111/j.1365-246X.2006.03120.x>`_.
+How To Cite
+-----------
```
vikaskurapati, line 16: I don't remember the exact discussion. But was it decided to remove the list and put a link on the documentation for the related publications?

src/Common/Iterator.h
```diff
@@ -242,6 +243,59 @@ constexpr auto enumerate(RangeT&& iterator) {
       std::forward<RangeT>(iterator));
 }
 
+template <typename T>
+class FilteredIterator {
```
vikaskurapati, line 322: Please document the new class definition and its purpose. A small example of how it is to be used is also helpful.

src/Common/Iterator.h
```diff
@@ -242,6 +243,59 @@ constexpr auto enumerate(RangeT&& iterator) {
...
+    while (this->base != end && !filter(*this->base)) {
+      ++this->base;
+    }
+  }
+
+  constexpr auto operator++() {
+    // advance, and skip if needed
+    ++base;
+    while (base != end && !filter(*base)) {
+      ++base;
```
vikaskurapati, line 347: I am not 100% sure about it, but shouldn't the check be done before incrementing it? What if the operator is applied when it is already at the end?

src/Common/Templating.h
```diff
@@ -0,0 +1,120 @@
...
+#define SEISSOL_SRC_COMMON_TEMPLATING_H_
+
+#include <type_traits>
+
+namespace seissol {
+// (the variadic pattern matching/type unpacking is partially inspired by
+// https://stackoverflow.com/questions/66944744/syntax-to-unpack-tuple-on-parameter-pack-and-variadic-template
+// )
+
+// Also, say hello to a bit of templating. Just a tiny bit.
```
vikaskurapati: Does this comment really help here?

src/DynamicRupture/Factory.h
```diff
@@ -22,13 +22,13 @@ class SeisSol;
...
- * ltsTree: holds all the data, like parameters (e.g. friction coefficients) or results (e.g.
- * sliprate) initializer: reads the parameter file and stores default values in ltsTree frictionLaw:
+ * storage: holds all the data, like parameters (e.g. friction coefficients) or results (e.g.
+ * sliprate) initializer: reads the parameter file and stores default values in storage frictionLaw:
  * evaluates friction during the actual simulation output: manages, when we need to write which
  * quantity to which output
  */
 struct DynamicRuptureTuple {
-  std::unique_ptr<seissol::initializer::DynamicRupture> ltsTree;
+  std::unique_ptr<seissol::DynamicRupture> storage;
```
vikaskurapati, line 31: Just for clarification: is it no longer related to lts tree?

src/DynamicRupture/FrictionLaws/FrictionSolver.h
```diff
@@ -100,8 +97,7 @@ class FrictionSolver {
   DREnergyOutput* __restrict energyData{};
   DRGodunovData* __restrict godunovData{};
   real (*__restrict initialPressure)[misc::NumPaddedPoints]{};
-  real (*__restrict nucleationPressure[initializer::parameters::MaxNucleactions])
-      [misc::NumPaddedPoints]{};
+  real (*__restrict nucleationPressure)[misc::NumPaddedPoints]{};
```
vikaskurapati, line 100: Why are we removing multiple nucleations here?

src/DynamicRupture/FrictionLaws/FrictionSolver.h
```diff
@@ -79,8 +77,7 @@ class FrictionSolver {
   real mFullUpdateTime{};
   // CS = coordinate system
   real (*__restrict initialStressInFaultCS)[6][misc::NumPaddedPoints]{};
-  real (*__restrict nucleationStressInFaultCS[seissol::initializer::parameters::MaxNucleactions])
-      [6][misc::NumPaddedPoints]{};
+  real (*__restrict nucleationStressInFaultCS)[6][misc::NumPaddedPoints]{};
```
vikaskurapati, line 80: Does this mean we no longer support multiple nucleations again?

src/DynamicRupture/FrictionLaws/CpuImpl/BaseFrictionLaw.h
```diff
@@ -100,14 +100,15 @@ class BaseFrictionLaw : public FrictionSolver {
...
-                                      nucleationStressInFaultCS[i][ltsFace],
-                                      initialPressure[ltsFace],
-                                      nucleationPressure[i][ltsFace],
-                                      updateTime,
-                                      this->drParameters->t0[i],
-                                      this->drParameters->s0[i],
-                                      this->deltaT[timeIndex]);
+          common::adjustInitialStress(
+              initialStressInFaultCS[ltsFace],
+              nucleationStressInFaultCS[ltsFace * this->drParameters->nucleationCount + i],
```
vikaskurapati, line 105: I don't think the refactor here is correct. If the multiple nucleations are removed, we can get rid of the loop of nucleation count remove the second dimension. Please take another look.

src/DynamicRupture/FrictionLaws/CpuImpl/BaseFrictionLaw.h
```diff
@@ -100,14 +100,15 @@ class BaseFrictionLaw : public FrictionSolver {
...
+              initialPressure[ltsFace],
+              nucleationPressure[ltsFace * this->drParameters->nucleationCount + i],
```
vikaskurapati, line 107: Same as above.

src/DynamicRupture/FrictionLaws/FrictionSolver.cpp
```diff
@@ -33,39 +33,34 @@ FrictionSolver::FrictionTime FrictionSolver::computeDeltaT(const std::vector<dou
...
-  qInterpolatedPlus = layerData.var(dynRup->qInterpolatedPlus, place);
-  qInterpolatedMinus = layerData.var(dynRup->qInterpolatedMinus, place);
-  initialPressure = layerData.var(dynRup->initialPressure, place);
-  for (std::size_t i = 0; i < dynRup->nucleationPressure.size(); ++i) {
-    nucleationPressure[i] = layerData.var(dynRup->nucleationPressure[i], place);
-  }
+  impAndEta = layerData.var<DynamicRupture::ImpAndEta>(place);
+  impedanceMatrices = layerData.var<DynamicRupture::ImpedanceMatrices>(place);
+  initialStressInFaultCS = layerData.var<DynamicRupture::InitialStressInFaultCS>(place);
+  nucleationStressInFaultCS = layerData.var<DynamicRupture::NucleationStressInFaultCS>(place);
```
vikaskurapati, line 41: Removing multiple nucleations?

src/DynamicRupture/FrictionLaws/FrictionSolver.cpp
```diff
@@ -33,39 +33,34 @@ FrictionSolver::FrictionTime FrictionSolver::computeDeltaT(const std::vector<dou
...
+  imposedStatePlus = layerData.var<DynamicRupture::ImposedStatePlus>(place);
+  imposedStateMinus = layerData.var<DynamicRupture::ImposedStateMinus>(place);
+  energyData = layerData.var<DynamicRupture::DREnergyOutputVar>(place);
+  godunovData = layerData.var<DynamicRupture::GodunovData>(place);
+  dynStressTime = layerData.var<DynamicRupture::DynStressTime>(place);
+  dynStressTimePending = layerData.var<DynamicRupture::DynStressTimePending>(place);
+  qInterpolatedPlus = layerData.var<DynamicRupture::QInterpolatedPlus>(place);
+  qInterpolatedMinus = layerData.var<DynamicRupture::QInterpolatedMinus>(place);
+  initialPressure = layerData.var<DynamicRupture::InitialPressure>(place);
+  nucleationPressure = layerData.var<DynamicRupture::NucleationPressure>(place);
```
vikaskurapati, line 63: Same as above

src/DynamicRupture/FrictionLaws/GpuImpl/BaseFrictionSolver.h
```diff
@@ -109,9 +110,11 @@ class BaseFrictionSolver : public FrictionSolverDetails {
       for (uint32_t i = 0; i < ctx.data->drParameters.nucleationCount; ++i) {
         common::adjustInitialStress<GpuRangeType>(
             ctx.data->initialStressInFaultCS[ctx.ltsFace],
-            ctx.data->nucleationStressInFaultCS[i][ctx.ltsFace],
+            ctx.data
+                ->nucleationStressInFaultCS[ctx.ltsFace * ctx.data->drParameters.nucleationCount +
```
vikaskurapati, line 114: The refactor does not translate directly. If we are removing multiple nucleations support, we can remove the outer loop.

src/DynamicRupture/FrictionLaws/GpuImpl/BaseFrictionSolver.h
```diff
@@ -109,9 +110,11 @@ class BaseFrictionSolver : public FrictionSolverDetails {
...
+                                            i],
             ctx.data->initialPressure[ctx.ltsFace],
-            ctx.data->nucleationPressure[i][ctx.ltsFace],
+            ctx.data->nucleationPressure[ctx.ltsFace * ctx.data->drParameters.nucleationCount + i],
```
vikaskurapati, line 117: Same as above

src/DynamicRupture/FrictionLaws/GpuImpl/FrictionSolverInterface.h
```diff
@@ -22,8 +22,7 @@ struct FrictionLawData {
   const ImpedanceMatrices* __restrict impedanceMatrices{};
   // CS = coordinate system
   real (*__restrict initialStressInFaultCS)[6][misc::NumPaddedPoints]{};
-  const real (*__restrict nucleationStressInFaultCS
-                  [seissol::initializer::parameters::MaxNucleactions])[6][misc::NumPaddedPoints]{};
+  const real (*__restrict nucleationStressInFaultCS)[6][misc::NumPaddedPoints]{};
```
vikaskurapati, line 25: No more multiple nucleations support?

src/DynamicRupture/FrictionLaws/GpuImpl/FrictionSolverInterface.h
```diff
@@ -42,8 +41,7 @@ struct FrictionLawData {
   DREnergyOutput* __restrict energyData{};
   const DRGodunovData* __restrict godunovData{};
   real (*__restrict initialPressure)[misc::NumPaddedPoints]{};
-  const real (*__restrict nucleationPressure[seissol::initializer::parameters::MaxNucleactions])
-      [misc::NumPaddedPoints]{};
+  const real (*__restrict nucleationPressure)[misc::NumPaddedPoints]{};
```
vikaskurapati, line 44: No more multiple nucleations support?

src/DynamicRupture/FrictionLaws/GpuImpl/FrictionSolverInterface.h
```diff
@@ -96,42 +94,36 @@ class FrictionSolverInterface : public seissol::dr::friction_law::FrictionSolver
...
-    data->qInterpolatedPlus = layerData.var(dynRup->qInterpolatedPlus, place);
-    data->qInterpolatedMinus = layerData.var(dynRup->qInterpolatedMinus, place);
-    data->initialPressure = layerData.var(dynRup->initialPressure, place);
-    for (std::size_t i = 0; i < dynRup->nucleationPressure.size(); ++i) {
-      data->nucleationPressure[i] = layerData.var(dynRup->nucleationPressure[i], place);
-    }
+    data->impAndEta = layerData.var<DynamicRupture::ImpAndEta>(place);
+    data->impedanceMatrices = layerData.var<DynamicRupture::ImpedanceMatrices>(place);
+    data->initialStressInFaultCS = layerData.var<DynamicRupture::InitialStressInFaultCS>(place);
+    data->nucleationStressInFaultCS =
```
vikaskurapati, line 103: No more multiple nucleations support?

src/DynamicRupture/FrictionLaws/GpuImpl/FrictionSolverInterface.h
```diff
@@ -96,42 +94,36 @@ class FrictionSolverInterface : public seissol::dr::friction_law::FrictionSolver
...
+    data->imposedStatePlus = layerData.var<DynamicRupture::ImposedStatePlus>(place);
+    data->imposedStateMinus = layerData.var<DynamicRupture::ImposedStateMinus>(place);
+    data->energyData = layerData.var<DynamicRupture::DREnergyOutputVar>(place);
+    data->godunovData = layerData.var<DynamicRupture::GodunovData>(place);
+    data->dynStressTime = layerData.var<DynamicRupture::DynStressTime>(place);
+    data->dynStressTimePending = layerData.var<DynamicRupture::DynStressTimePending>(place);
+    data->qInterpolatedPlus = layerData.var<DynamicRupture::QInterpolatedPlus>(place);
+    data->qInterpolatedMinus = layerData.var<DynamicRupture::QInterpolatedMinus>(place);
+    data->initialPressure = layerData.var<DynamicRupture::InitialPressure>(place);
+    data->nucleationPressure = layerData.var<DynamicRupture::NucleationPressure>(place);
```
vikaskurapati, line 126: No more multiple nucleations support?

src/DynamicRupture/Initializer/BaseDRInitializer.cpp
```diff
@@ -107,52 +105,56 @@ void BaseDRInitializer::initializeFault(const seissol::initializer::DynamicRuptu
...
     // rotate initial stress to fault coordinate system
     if (initialStressParameterizedByTraction) {
-      rotateTractionToCartesianStress(dynRup, layer, initialStress);
+      rotateTractionToCartesianStress(layer, initialStress);
     }
 
-    auto* initialStressInFaultCS = layer.var(dynRup->initialStressInFaultCS);
-    rotateStressToFaultCS(dynRup, layer, initialStressInFaultCS, initialStress);
+    auto* initialStressInFaultCS = layer.var<DynamicRupture::InitialStressInFaultCS>();
+    rotateStressToFaultCS(layer, initialStressInFaultCS, 0, 1, initialStress);
```
vikaskurapati, line 126: I am unsure about the purpose of the index 0, count 1 here. It seems to have been introduced due to the multiple nucleations, but I think the base data structure size of the nucleation stresses seems to be off. Please take another look and modify it. 

Although it may be consistent in this file due to how the array of vectors is defined.

src/DynamicRupture/Initializer/BaseDRInitializer.cpp
```diff
@@ -107,52 +105,56 @@ void BaseDRInitializer::initializeFault(const seissol::initializer::DynamicRuptu
...
+    auto* initialPressure = layer.var<DynamicRupture::InitialPressure>();
     for (std::size_t ltsFace = 0; ltsFace < layer.size(); ++ltsFace) {
       for (std::uint32_t pointIndex = 0; pointIndex < misc::NumPaddedPoints; ++pointIndex) {
         initialPressure[ltsFace][pointIndex] = initialStress.p[ltsFace][pointIndex];
         for (unsigned i = 0; i < drParameters->nucleationCount; ++i) {
-          auto* nucleationPressure = layer.var(dynRup->nucleationPressure[i]);
-          nucleationPressure[ltsFace][pointIndex] = nucleationStresses[i].p[ltsFace][pointIndex];
+          auto* nucleationPressure = layer.var<DynamicRupture::NucleationPressure>();
+          nucleationPressure[ltsFace * drParameters->nucleationCount + i][pointIndex] =
+              nucleationStresses[i].p[ltsFace][pointIndex];
```
vikaskurapati, line 147: This is also off considering multiple nucleations.

src/DynamicRupture/Initializer/BaseDRInitializer.h
```diff
@@ -83,36 +83,29 @@ class BaseDRInitializer {
...
 
   protected:
   /**
    * Add additional parameters to be read from the easi file
    * This will be specialized in the derived friction law initializers
    * @param parameterToStorageMap reference to a std::unordered_map<std::string, double*>, which
    * maps the parameter name, to the address in memory, where the parameter shall be stored
-   * @param dynRup pointer to the respective dynamic rupture datastructure
-   * @param it reference to an LTSTree leaf_iterator
+   * @param layer reference to an Storage layer
```
vikaskurapati: a storage. ;)

src/DynamicRupture/Initializer/BaseDRInitializer.h
```diff
@@ -83,36 +83,29 @@ class BaseDRInitializer {
...
-                              const seissol::initializer::DynamicRupture* dynRup,
-                              seissol::initializer::Layer& layer);
+                              DynamicRupture::Layer& layer);
 
   /**
-   * Finds all faceIDs in one iterator. This is the mapping idInLTSTree -> idInMesh
-   * @param dynRup pointer to the respective dynamic rupture datastructure
-   * @param it reference to an LTSTree leaf_iterator
+   * Finds all faceIDs in one iterator. This is the mapping idInStorage -> idInMesh
+   * @param layer reference to an Storage layer
```
vikaskurapati: a storage. ;)

src/DynamicRupture/Initializer/BaseDRInitializer.h
```diff
@@ -126,11 +119,9 @@ class BaseDRInitializer {
    * slipRateMagnitude
    * traction1
    * traction2
-   * @param dynRup pointer to the respective dynamic rupture datastructure
-   * @param it reference to an LTSTree leaf_iterator
+   * @param layer reference to an Storage layer
```
vikaskurapati: a storage. ;)

src/DynamicRupture/Initializer/BaseDRInitializer.h
```diff
@@ -151,27 +142,24 @@ class BaseDRInitializer {
   private:
   /**
    * Rotates the fault-aligned traction to cartesian stress coordinates
-   * @param dynRup pointer to the respective dynamic rupture datastructure
-   * @param it reference to an LTSTree leaf_iterator
+   * @param layer reference to an Storage layer
```
vikaskurapati: a storage. ;)

src/DynamicRupture/Initializer/BaseDRInitializer.h
```diff
@@ -151,27 +142,24 @@ class BaseDRInitializer {
...
-  void rotateTractionToCartesianStress(const seissol::initializer::DynamicRupture* dynRup,
-                                       seissol::initializer::Layer& layer,
-                                       StressTensor& stress);
+  void rotateTractionToCartesianStress(DynamicRupture::Layer& layer, StressTensor& stress);
 
   /**
    * Rotates the stress tensor to a fault aligned coordinate system and stores it in stressInFaultCS
-   * @param dynRup pointer to the respective dynamic rupture datastructure
-   * @param it reference to an LTSTree leaf_iterator
+   * @param layer reference to an Storage layer
```
vikaskurapati: a storage. ;) 

Please also add what index and count are referring to.

src/DynamicRupture/Initializer/ImposedSlipRatesInitializer.cpp
```diff
@@ -81,35 +75,35 @@ void ImposedSlipRatesInitializer::initializeFault(
...
+      auto* nucleationPressure = layer.var<DynamicRupture::NucleationPressure>();
       for (std::size_t ltsFace = 0; ltsFace < layer.size(); ++ltsFace) {
         for (std::uint32_t pointIndex = 0; pointIndex < misc::NumPaddedPoints; ++pointIndex) {
           for (unsigned int dim = 0; dim < 6; ++dim) {
-            nucleationStressInFaultCS[ltsFace][dim][pointIndex] = 0;
+            nucleationStressInFaultCS[ltsFace * drParameters->nucleationCount + i][dim]
+                                     [pointIndex] = 0;
           }
-          nucleationPressure[ltsFace][pointIndex] = 0;
+          nucleationPressure[ltsFace * drParameters->nucleationCount + i][pointIndex] = 0;
```
vikaskurapati, line 86: Same as above for multiple nucleations.

src/DynamicRupture/Initializer/ImposedSlipRatesInitializer.h
```diff
@@ -20,45 +20,37 @@ class ImposedSlipRatesInitializer : public BaseDRInitializer {
...
+  void initializeFault(DynamicRupture::Storage& drStorage) override;
 
   /**
    * Add additional parameters to be read from the easi file
    * This will be specialized in the derived friction law initializers
    * @param parameterToStorageMap reference to a std::unordered_map<std::string, double*>, which
    * maps the parameter name, to the address in memory, where the parameter shall be stored
-   * @param dynRup pointer to the respective dynamic rupture datastructure
-   * @param it reference to an LTSTree leaf_iterator
+   * @param layer reference to an Storage layer
```
vikaskurapati: a storage ;)

src/DynamicRupture/Output/OutputManager.h
```diff
@@ -63,19 +62,17 @@ class OutputManager {
...
-  seissol::initializer::LTSTree* wpTree{nullptr};
-  seissol::initializer::Lut* wpLut{nullptr};
-  seissol::initializer::LTSTree* drTree{nullptr};
-  seissol::initializer::DynamicRupture* drDescr{nullptr};
+  LTS::Storage* wpStorage{nullptr};
+  LTS::Backmap* wpBackmap{nullptr};
+  DynamicRupture::Storage* drStorage{nullptr};
 
-  FaceToLtsMapType faceToLtsMap{};
+  FaceToLtsMapType faceToLtsMap;
```
vikaskurapati, line 69: Why remove the default initializer?

src/DynamicRupture/Output/OutputManager.h
```diff
@@ -63,19 +62,17 @@ class OutputManager {
...
   std::vector<std::size_t> globalFaceToLtsMap;
   seissol::geometry::MeshReader* meshReader{nullptr};
 
   size_t iterationStep{0};
   static constexpr double timeMargin{1.005};
-  std::string backupTimeStamp{};
+  std::string backupTimeStamp;
```
vikaskurapati, line 75: Why remove the default initializer?

src/DynamicRupture/Output/ReceiverBasedOutput.h
```diff
@@ -109,6 +105,17 @@ class ReceiverOutput {
     }
   }
 
+  template <typename StorageT>
+  std::remove_extent_t<typename StorageT::Type>* getCellData(const LocalInfo& local) {
```
vikaskurapati, line 102: Documentation would be nice. :)

src/Memory/Descriptor/DynamicRupture.h
```diff
@@ -35,315 +37,330 @@ inline auto allocationModeDR() {
...
+  struct ImpAndEta : public initializer::Variable<seissol::dr::ImpedancesAndEta> {};
+  struct ImpedanceMatrices : public initializer::Variable<seissol::dr::ImpedanceMatrices> {};
   // size padded for vectorization
   // CS = coordinate system
-  Variable<real[6][dr::misc::NumPaddedPoints]> initialStressInFaultCS;
-  std::vector<Variable<real[6][dr::misc::NumPaddedPoints]>> nucleationStressInFaultCS;
+  struct InitialStressInFaultCS : public initializer::Variable<real[6][dr::misc::NumPaddedPoints]> {
+  };
+  struct NucleationStressInFaultCS
+      : public initializer::Variable<real[6][dr::misc::NumPaddedPoints]> {};
```
vikaskurapati, line 66: How are multiple nucleations being handled here?

src/Memory/Descriptor/DynamicRupture.h
```diff
@@ -35,315 +37,330 @@ inline auto allocationModeDR() {
...
-  Variable<real[dr::misc::NumPaddedPoints]> traction2;
-  Variable<real[ConvergenceOrder][tensor::QInterpolated::size()]> qInterpolatedPlus;
-  Variable<real[ConvergenceOrder][tensor::QInterpolated::size()]> qInterpolatedMinus;
-
-  Scratchpad<real> idofsPlusOnDevice;
-  Scratchpad<real> idofsMinusOnDevice;
-
-  virtual void addTo(LTSTree& tree) {
+  struct InitialPressure : public initializer::Variable<real[dr::misc::NumPaddedPoints]> {};
+  struct NucleationPressure : public initializer::Variable<real[dr::misc::NumPaddedPoints]> {};
```
vikaskurapati, line 69: How are multiple nucleations being handled here?

src/Memory/Descriptor/DynamicRupture.h
```diff
@@ -35,315 +37,330 @@ inline auto allocationModeDR() {
...
+    storage.add<DREnergyOutputVar>(mask, Alignment, allocationModeDR());
+    storage.add<ImpAndEta>(mask, Alignment, allocationModeDR(), true);
+    storage.add<ImpedanceMatrices>(mask, Alignment, allocationModeDR(), true);
+    storage.add<InitialStressInFaultCS>(mask, Alignment, allocationModeDR());
+    storage.add<InitialPressure>(mask, Alignment, allocationModeDR());
+    storage.add<RuptureTime>(mask, Alignment, allocationModeDR());
+
+    storage.add<NucleationStressInFaultCS>(
+        mask, Alignment, allocationModeDR(), true, nucleationCount);
+    storage.add<NucleationPressure>(mask, Alignment, allocationModeDR(), true, nucleationCount);
```
vikaskurapati, line 129: I am still unclear how multiple nucleations are handled here. :sweat_smile:

src/Memory/Tree/Backmap.h
```diff
@@ -0,0 +1,113 @@
...
+#include <cstddef>
+#include <limits>
+#include <optional>
+#include <utility>
+#include <utils/logger.h>
+#include <vector>
+
+namespace seissol::initializer {
+
+struct StoragePosition {
```
vikaskurapati, line 25: Documentation for a new structure would be nice to add

src/Memory/Tree/LTSTree.h
```diff
@@ -317,20 +382,96 @@ class LTSTree : public LTSInternalNode {
...
     size_t maxClusterSize{0};
-    for (const auto& leaf : leaves(mask)) {
+    for (const auto& leaf : this->leaves(mask)) {
       const auto currClusterSize = static_cast<size_t>(leaf.size());
       maxClusterSize = std::max(currClusterSize, maxClusterSize);
     }
     return maxClusterSize;
   }
+
+  class IteratorWrapper {
```
vikaskurapati, line 412: Two new classes, some documentation on their purpose, and how they're supposed to be used would be nice.

src/Memory/Tree/Layer.h
vikaskurapati, line 1: This file has a few new structures. Is it possible to add their documentation and/or unit tests for them?

src/ResultWriter/ReceiverWriter.h
```diff
@@ -66,8 +65,7 @@ class ReceiverWriter : public seissol::Module {
   double m_samplingInterval;
   std::vector<std::shared_ptr<kernels::DerivedReceiverQuantity>> derivedQuantities;
   // Map needed because LayerType enum casts weirdly to int.
```
vikaskurapati: Comment no longer relevant?

src/SourceTerm/Typedefs.h
```diff
@@ -92,8 +92,8 @@ struct PointSources {
 };
 
 struct CellToPointSourcesMapping {
-  //! Pointer to DOFs
-  real (*dofs)[tensor::Q::size()]{};
+  //! Pointer to DOFs (tensor::Q)
+  real* dofs;
```
vikaskurapati, line 96: Any reason we move from pointing to the array to pointing to the first element here?

vikaskurapati (commented):
Most of the comments are about multiple nucleations scenarios. I am unsure if your new `Variable` implementation already takes care of it. If it does, I apologise for the multiple comments regarding it.

In any case, I suggest testing it on a simulation with multiple nucleations before merging(need not be added to CI if a small case is not there, local testing to ensure receivers match should suffice). I don't have a testcase for it. I messaged @Aangniu for it requesting it.

davschneller:
Thanks for the review; and sure, I'll add some comments and hopefully also some simple tests.

And no—multiple simulations will not be removed by this PR; their implementation just changes a bit: instead of having multiple arrays per nucleation for the initial stress/pressure, we now pack them all into a single one (cf. the `count` operator).
Having a multi-nucleation test case would be great (actually, what would be stopping us just taking e.g. a tpv5-esque case with nucleation and add some extra nucleation "pulses" to it? No need to stop at 2 :) ).

src/Memory/Descriptor/DynamicRupture.h
davschneller: The last argument does the trick here—that'll allocate multiple copies per cell (but still delegates all indexing to us, the user).

(this mechanic was already present for several years in SeisSol but only used in one class to my knowledge—so now we use it here)

src/DynamicRupture/Output/OutputManager.h
davschneller: That's done automatically, unlike for e.g. `int`s (thanks to C++ being old and based on C etc.)—hence we're good with removing it here.

src/DynamicRupture/Output/OutputManager.h
davschneller: See comment below

src/DynamicRupture/Initializer/BaseDRInitializer.h
davschneller: Done (once I'll upload the review commit)

src/DynamicRupture/FrictionLaws/FrictionSolver.h
davschneller: We aren't doing that, see my comment below; but the implementation changes a bit (the nucleations are allocated a tiny bit differently).

src/DynamicRupture/Factory.h
davschneller: Wait... Actually, that one might be mis-named. The `DynamicRupture` object always was not an `LTSTree`, but the descriptor of what was allocated by the `LTSTree`.

(that said; the "LTS tree" was a generic datastructure—basically splitting various data with different types into "layers" which usually equal an LTS cluster and a halo type (Interior, Ghost, Copy). It had a superimposed tree structure because the data was e.g. initialized and used like that for history reasons—see e.g. how the cell local information is initialized in `LtsLayout`. ... However, we don't really need the tree structure for the most part; but rather iterating agnostically over all such layers is good enough for us, and contains less structure which makes it (usually) more flexible.)

src/Common/Templating.h
davschneller: Well, not really. :)  ... I'll remove it.

src/Common/Iterator.h
davschneller: We should be fine here (good catch though). If we are at the end already, we shouldn't even enter here. (if everything is well-defined; and/or delegate everything to the base iterator)

The current implementation is already consistent with the C++20 implementation (which should supersede my implementation once we switch); cf. https://en.cppreference.com/w/cpp/ranges/filter_view/iterator.html#increment .

docs/related-publications.rst
davschneller: Yes; that was decided to keep the publication lists from getting outdated too quickly.

src/Common/Iterator.h
vikaskurapati: Could you refer me to the part of the code that ensures that we don't enter here if we are at the end already?(asking only for my curiosity)

src/DynamicRupture/FrictionLaws/FrictionSolver.h
vikaskurapati: OK, the code bit here confused me a bit as we basically got rid of one dimension. May be some explanation of how it is handled now, here as comments or somewhere else may help.

src/Memory/Descriptor/DynamicRupture.h
vikaskurapati: OK. That clears up most of the comments.

vikaskurapati:
> Thanks for the review; and sure, I'll add some comments and hopefully also some simple tests.
> 

Thanks!

> And no—multiple simulations will not be removed by this PR; their implementation just changes a bit: instead of having multiple arrays per nucleation for the initial stress/pressure, we now pack them all into a single one (cf. the `count` operator). 

I hope you mean multiple nucleations. I could not find major changes which change multiple simulations in this. I understood it now. The storage layout and `LTStree.h` implementations are something I still need time to understand in full detail.

> Having a multi-nucleation test case would be great (actually, what would be stopping us just taking e.g. a tpv5-esque case with nucleation and add some extra nucleation "pulses" to it? No need to stop at 2 :) ).

I would say nothing, except that _I_ don't know what to expect in terms of results, to check if they make sense. I will ping @Aangniu once again to verify them for a TPV5-esque case and get back to you. 

Or maybe @Thomas-Ulrich knows something about it?

davschneller:
Yeah, I meant the multinuc (not the multisim). By the way (I forgot initially), we already have a tpv5-nuc case; that one might be easy to expand.

I've added some tests and some comments; but I guess that a _ton_ more can be done in the future at every part in SeisSol. However, it's best to bring the whole code into a better structure first (requiring at least #1411 and possibly #1421, plus a lot more refactoring down the road—to finally get to a useful code structure after the FORTRAN removal).
(and of course some code, especially w.r.t. the iterators, becomes partially obsolete once we start using C++20 ... However I'd only do that once either we can verify for ICX/ICC 2021 that it'll work, or SuperMUC/Frontera go EOL or get a software update)

src/Common/Iterator.h
davschneller: That's done in the loop; I've added an example comment and a test. (I did not really intend the filter to be used somewhere outside; hence its implementation is a little more crude than e.g. for `zip` or `enumerate`. If wanted, I think I can change that though—but C++20 should make all of that obsolete anyways)

davschneller:
I've added a test case with four nucleation triggers here https://github.com/SeisSol/precomputed-seissol/pull/20 . And can confirm that this branch passes the test.

src/Common/Iterator.h
vikaskurapati: I will leave the decision for yourself. If it would probably be obsolete in a few years, I would tend to not changing it much. I was just confused as to why the increment is before the check. Which loop are you referring to, though? 

Thanks for adding the test.

vikaskurapati:
> I've added a test case with four nucleation triggers here [SeisSol/precomputed-seissol#20](https://github.com/SeisSol/precomputed-seissol/pull/20) . And can confirm that this branch passes the test.

Thanks for the test. I tried comparing it with the master branch's results for the same test case. Surprisingly the results in absolute slip etc. vary by a very small amount(lower than eps, of course) for the outputs of the current master branch, and this branch. Do you have any suspicion as to why that could be the case? I am just curious if just the ordering of data structures could bring minor differences in 6th digit of receiver outputs already? Or may be it is slightly different LTS clustering across different runs?

davschneller:
> > I've added a test case with four nucleation triggers here [SeisSol/precomputed-seissol#20](https://github.com/SeisSol/precomputed-seissol/pull/20) . And can confirm that this branch passes the test.
> 
> Thanks for the test. I tried comparing it with the master branch's results for the same test case. Surprisingly the results in absolute slip etc. vary by a very small amount(lower than eps, of course) for the outputs of the current master branch, and this branch. Do you have any suspicion as to why that could be the case? I am just curious if just the ordering of data structures could bring minor differences in 6th digit of receiver outputs already? Or may be it is slightly different LTS clustering across different runs?

Thanks for the approval.

W.r.t. the small differences: I used a slightly older SeisSol version (from May 2025) for verifying that; so the difference might lie somewhere else as well. The errors are reproducible with the latest master, and also, for the e-16 errors, seem to vary between the runs. Interesting—might need more debugging.

src/DynamicRupture/FrictionLaws/FrictionSolver.h
davschneller: Ok; left a comment where the `nucleationCount` is actually passed/allocated to the storage system.

PR #1410, opened 2025-07-25
Fix plasticity kernels for multiple simulations

vikaskurapati:
The transposes were not necessary for multiple simulations in plasticity kernels. Now the results match in the receivers with TPV13.


src/Solver/TimeStepping/TimeCluster.cpp
```diff
@@ -823,6 +823,9 @@ void TimeCluster::finalize() {
 
 template <bool UsePlasticity>
 void TimeCluster::computeNeighboringIntegrationImplementation(double subTimeStart) {
+  // we cannot do an auto clusterData = this->clusterData, and an auto timeStep = timeStepSize()
```
davschneller: The comment may be a bit unclear; what did you want to achieve here?

To follow how I interpreted it: yes, you (probably) can do that—I understand it in the way that you'd like to just have local variables and reference that in the OpenMP loop as well (in the `shared` clause). Except that it should be `auto& clusterData`.

Are you certain that it's `clang-tidy` that complains here and not the compilers themselves? The reason is the `default(none)` at the clause right now which requires every variable in the loop to be referenced explicitly.
(which is due to some `clang-tidy` rule in fact; though it's disabled in our checks right now)

E.g. see https://github.com/SeisSol/SeisSol/blob/ad5e99008cc9c00f0a58657b8d101b3bac163363/src/Solver/TimeStepping/TimeCluster.cpp#L887-L906 for an example where I had added some new entries.

src/Initializer/InitProcedure/InitModel.cpp
```diff
@@ -384,6 +384,7 @@ void seissol::initializer::initprocedure::initModel(seissol::SeisSol& seissolIns
   logInfo() << "Order:" << ConvergenceOrder;
   logInfo() << "Precision:"
             << (Config::Precision == RealType::F32 ? "single (f32)" : "double (f64)");
+  logInfo() << "Number of fused simulations: " << Config::NumSimulations;
```
davschneller: Will display `1` for single simulations; i.e. not necessarily fused—how about something in the line of just "Number of [concurrent] simulations" ?

davschneller (changes requested):
Looks mostly good; however please squash the most recent commits which only concern formatting. IMO it's ok to keep such "formatting commits"—but use them sparingly; I try to do so as well. An exception are post-merge formatting/merge fix commits, as rebasing merges can be pretty cumbersome/error-prone.

(some projects like to squash all changes of a PR into a single commit; but personally I don't like that tbh. It would make it hard to debug or revert on a single-commit basis)

src/Solver/TimeStepping/TimeCluster.cpp
vikaskurapati: > The comment may be a bit unclear; what did you want to achieve here?
> 
> To follow how I interpreted it: yes, you (probably) can do that—I understand it in the way that you'd like to just have local variables and reference that in the OpenMP loop as well (in the `shared` clause). Except that it should be `auto& clusterData`.
> 
> Are you certain that it's `clang-tidy` that complains here and not the compilers themselves? The reason is the `default(none)` at the clause right now which requires every variable in the loop to be referenced explicitly. (which is due to some `clang-tidy` rule in fact; though it's disabled in our checks right now)
> 
> E.g. see
> 
> https://github.com/SeisSol/SeisSol/blob/ad5e99008cc9c00f0a58657b8d101b3bac163363/src/Solver/TimeStepping/TimeCluster.cpp#L887-L906
> 
> for an example where I had added some new entries.

You are right. It seems like I misunderstood the error before. I made the change, thank you.

vikaskurapati:
Thank you for your review. I addressed the comments and squashed the formatting commits.

PR #1411, opened 2025-08-05
Clean up the remaining (old) init procedure

davschneller:
Refactor all of the remaining parts of SeisSol (in effect a follow up to #1275 and #829 ). Also apply `clang-format` and `clang-tidy` to everything. Also, delete `LtsLayout`, and unify the code structure and usage even more.

As a result, we can shorten the code by a lot while keeping its functionality. And also modernize it—as well as paving the way to grander refinement of the initialization code.

Depends #1384 and #1408—which are now merged.
The code could still use a bit of polishing; but it should be good right now already.

More unit tests and more holistic refactoring may come in a subsequent PR.

### DIFF
--- a/src/Initializer/InitProcedure/Internal/Buckets.cpp
+++ b/src/Initializer/InitProcedure/Internal/Buckets.cpp
@@ -23,6 +23,7 @@
 #include <cstdint>
 #include <cstring>
 #include <utility>
+#include <utils/logger.h>
 #include <vector>
 #include <yateto/InitTools.h>
 
--- a/src/Solver/TimeStepping/TimeManager.cpp
+++ b/src/Solver/TimeStepping/TimeManager.cpp
@@ -30,6 +30,7 @@
 #include <Solver/TimeStepping/TimeCluster.h>
 #include <algorithm>
 #include <cassert>
+#include <cmath>
 #include <cstddef>
 #include <cstdint>
 #include <limits>


src/Initializer/LtsSetup.h
```diff
@@ -0,0 +1,82 @@
...
+
+    [[nodiscard]] constexpr auto neighborHasDerivatives(int face) const -> bool {
+        return test(face);
+    }
+
+    constexpr auto setNeighborGTS(int face, bool gts) -> LtsSetup& {
+        return set(face + 4, gts);
+    }
+
+    [[nodiscard]] constexpr auto neighborGTS(int face) const -> bool {
```
Thomas-Ulrich: why not a name with has or is?
neighborGTS is not very clear.
can you add a doctring to this function?

src/Initializer/LtsSetup.h
```diff
@@ -0,0 +1,82 @@
...
+
+    [[nodiscard]] constexpr auto hasDerivatives() const -> bool {
+        return test(9);
+    }
+
+    constexpr auto setCacheBuffers(bool val) -> LtsSetup& {
+        return set(10, val);
+    }
+
+    [[nodiscard]] constexpr auto cacheBuffers() const -> bool {
```
Thomas-Ulrich: hasCacheBuffers?

src/Initializer/TimeStepping/Common.h
```diff
@@ -85,26 +87,25 @@ namespace seissol::initializer::time_stepping {
  * @param faceNeighborIds face neighbor ids.
  * @param copy true if the cell is part of the copy layer (only required for correctness in dynamic rupture computations).
  **/
-static LtsSetup getLtsSetup(unsigned int localClusterId,
-                                  const unsigned int neighboringClusterIds[4],
-                                  const FaceType faceTypes[4],
-                                  const unsigned int faceNeighborIds[4], // TODO: Remove, outdated
+static LtsSetup getLtsSetup(const CellLocalInformation& ownPrimary,
```
Thomas-Ulrich: the docstring would need to be updated.

Thomas-Ulrich (approved):
LGTM as far as I looked, that is a few commits.
(and I won't be able to review the rest)

PR #1414, opened 2025-08-20
Fix "previous time DOFs" in Non-CK Situations

davschneller:
Currently, the receiver and energy kernels use the 0-th order derivatives to access the DOFs. Alas, that doesn't work for Legendre bases anymore; there it would be the time integral. Thus, we now apply the time kernel in this situation instead.

Might incur a slight performance regression in the CK regime; hence draft. (since we're now looking at the whole space-time derivatives; and not just the 0-th one)

Thus, use the time kernels with a dummy time instead (not super happy about that idea; but it seems to do its job).

Maybe also a TODO: use the (spatial)function of the computed step; instead of the previous step?


davschneller:
"Swallowed" into #1407.

PR #1419, opened 2025-08-22
Fix the non-volume writer executors getting stuck

davschneller:
Fix the simulation getting stuck after being already finished.

The stopwatches need to be triggered by all ranks right now (which caused an issue with the old fault and surface writers); thus we now do that.


vikaskurapati (approved):
Tested on Vista with multiple nodes. It solves the issue. Thank you.

PR #1422, opened 2025-08-28
Implementation of fused simulations with varying plasticity

vikaskurapati:
We separate the plasticity material parameters from materialfilename and put it separately in plasticityfilename. This allows us to varying plasticity coefficients across different simulations. Tested on CPUs, GPU changes ensured to not break serial simulations.

### DIFF
--- a/.github/workflows/build-seissol-cpu.yml
+++ b/.github/workflows/build-seissol-cpu.yml
@@ -10,7 +10,7 @@ on:
   - push
 
 env:
-  precomputed-hash: 57577066fddab59afc5670397549b1b3e6d6e7ac
+  precomputed-hash: fb41141ee0b38b48b10cc74c8c817860d4a745c5
 
 jobs:
   seissol-build-test:
@@ -324,6 +324,14 @@ jobs:
             volume: false
             surface: false
           - name: tpv13-fused
+            equation: elastic
+            multisim: 8
+            energies: true
+            receivers: true
+            fault: false # incomplete, hence disabled
+            volume: false
+            surface: false
+          - name: tpv13-fused-plasticity
             equation: elastic
             multisim: 8
             energies: true


src/Initializer/Parameters/ModelParameters.cpp
```diff
@@ -50,6 +54,21 @@ ModelParameters readModelParameters(ParameterReader* baseReader) {
   const auto boundaryFileName = reader->readPath("boundaryfileName");
   const std::string materialFileName =
       reader->readPathOrFail("materialfilename", "No material file given.");
+  const auto plasticityFileName = reader->readPath("plasticityfilename");
+
+  std::vector<std::string> plasticityFileNames(seissol::multisim::NumSimulations);
+
+  if (!plasticityFileName.value_or("").empty()) {
```
davschneller: Check for the model file name vs plasticity file name right here already. I.e. if plasticity file name empty, then use the model file name; otherwise use the dedicated file name.

src/Initializer/Parameters/ModelParameters.cpp
```diff
@@ -50,6 +54,21 @@ ModelParameters readModelParameters(ParameterReader* baseReader) {
...
+    plasticityFileNames[0] = plasticityFileName.value();
+  }
+
+  for (std::size_t i = 1; i < plasticityFileNames.size(); ++i) {
+    const auto fieldname = "plasticityfilename" + std::to_string(i);
+    if (reader->hasField(fieldname)) {
+      plasticityFileNames[i] = reader->read<std::string>(fieldname).value();
```
davschneller: Use `readPath` instead of `read<std::string>` ; also you can avoid the `if` entirely by instead using `readPath(fieldname).value_or(plasticityFileName);`.
To keep the default behavior from before.

src/Initializer/Parameters/ModelParameters.h
```diff
@@ -38,6 +38,8 @@ struct ModelParameters {
   double tv;
   std::string boundaryFileName;
   std::string materialFileName;
+  std::string plasticityFileName;
```
davschneller: Why is that one necessary besides the array? I'd unify them.

src/Initializer/InitProcedure/InitModel.cpp
```diff
@@ -106,11 +107,18 @@ void initializeCellMaterial(seissol::SeisSol& seissolInstance) {
       queryGen, seissolParams.model.materialFileName, meshReader.getElements().size());
 
   // plasticity (if needed)
-  std::vector<Plasticity> plasticityDB;
+  std::vector<std::vector<Plasticity>> plasticityDB;
   if (seissolParams.model.plasticity) {
     // plasticity information is only needed on all interior+copy cells.
-    plasticityDB = queryDB<Plasticity>(
-        queryGen, seissolParams.model.materialFileName, meshReader.getElements().size());
+    if (seissolParams.model.plasticityFileName.empty()) {
```
davschneller: Keep that check in the parameter reader if possible. (generally checking for an empty string is borderline antiquated practice; an optional might be better—or just always supplying a valid plasticity file name (even if it's the model file name))

generated-code/kernels/plasticity.py
```diff
@@ -170,17 +169,16 @@ def addKernels(generator, aderdg, matricesDir, PlasticityMethod, targets):
 
         if aderdg.multipleSimulations > 1:
             # for now, copy the tensors into here and rename them; until gemmforge/chainforge is deprecated
-            initialLoading = Tensor("initialLoadingM", (6,))
-            replicateInitialLoading = OptionalDimTensor(
-                "replicateInitialLoadingM",
+            initialLoadingM = OptionalDimTensor(
```
davschneller, line 172: Needs also to not be an `OptionalDimTensor` (won't change the results—but it's some unnecessary computations)

src/Model/Plasticity.h
```diff
@@ -10,31 +10,37 @@
...
-
+  real initialLoading[6 * seissol::multisim::NumSimulations];
+  real cohesionTimesCosAngularFriction[seissol::multisim::NumSimulations];
+  real sinAngularFriction[seissol::multisim::NumSimulations];
+  real mufactor; // Only dependent on mu which is to be constant for all simulations
+
+  PlasticityData(const std::array<Plasticity, seissol::multisim::NumSimulations>& plasticity,
+                 const Material* material) {
+    for (std::size_t i = 0; i < seissol::multisim::NumSimulations;
+         ++i) { // interleave these so that the kernel does not need any modifications
```
davschneller: Next line for the comment.

Maybe you can leave the comment out as that fact should follow from the tensor structure (and too many "rather immediate" comments may get old too quickly in the code)

Documentation/parameters.par
```diff
@@ -15,6 +15,8 @@ UseCellHomogenizedMaterial = 1
 !off-fault plasticity parameters (ignored if Plasticity=0)
 Plasticity=0
 Tv=0.05
+!yaml file defining spatial dependance of properties necessary for plasticity
+PlasticityFileName = 'plasticity.yaml'
```
davschneller: We usually set the default values here; so `PlasticityFileName = ''` . (and maybe mention that the model/material file name is used if it is not mentioned)

src/Initializer/InitProcedure/InitModel.cpp
```diff
@@ -106,11 +107,18 @@ void initializeCellMaterial(seissol::SeisSol& seissolInstance) {
       queryGen, seissolParams.model.materialFileName, meshReader.getElements().size());
 
   // plasticity (if needed)
-  std::vector<Plasticity> plasticityDB;
+  std::vector<std::vector<Plasticity>> plasticityDB;
```
davschneller: You can already resize the outer vector to `NumSimulations` (or reserve)

davschneller (changes requested):
Added some comments. Looks good to me in general—but make sure to keep the parameter reading and checking to the `Initializer/Parameters` folder where possible (it is here)—so we don't have to deal with that in other parts of the code when we can do it conveniently in there.

generated-code/kernels/plasticity.py
vikaskurapati: For multiple simulations, isn't it supposed to be an optionalDimTensor? I.e., it has all different simulations' initial loading stored in it? Or does it work differently for GPUs?

src/Model/Plasticity.h
vikaskurapati: I will prefer to keep this comment to remind myself why this is done this way. Mainly for further debugging purposes.

PR #1423, opened 2025-08-30
Some small DR Adjustments

davschneller:
We

* make the peak slip rate and rupture output more accurate (update it per DR timestep; not only per wavepropagation timestep)
* abstract the timestep count for the DR (that could even become a runtime variable in the future—in case accuracy is a concern)
* add a dummy point at the end of the timestep—since previously we'd just timestep _over_ the last quadpoint; treating it differently than the others. Downside: the DR becomes a little more expensive. Also, as tested with the precomputed tpv5; the energy values do change in the 5th digit already—i.e. probably a significant change. (I also tried, for fun, to double the number of timesteps in the DR; that changed the elastic energy _even_ more significantly)

Not 100%ly sure if the latter really solves the issue; or if maybe even Gauss Lobatto points would be better suited.

### DIFF
--- a/src/DynamicRupture/FrictionLaws/FrictionSolver.cpp
+++ b/src/DynamicRupture/FrictionLaws/FrictionSolver.cpp
@@ -17,18 +17,18 @@
 namespace seissol::dr::friction_law {
 
 FrictionSolver::FrictionTime FrictionSolver::computeDeltaT(const std::vector<double>& timePoints) {
-  std::vector<real> deltaT(ConvergenceOrder);
-  real sumDt = 0;
+  std::vector<double> deltaT(ConvergenceOrder);
 
-  deltaT[0] = timePoints[0];
-  sumDt = deltaT[0];
+  deltaT[0] = timePoints[0]; // - 0
   for (std::size_t timeIndex = 1; timeIndex < ConvergenceOrder; ++timeIndex) {
     deltaT[timeIndex] = timePoints[timeIndex] - timePoints[timeIndex - 1];
-    sumDt += deltaT[timeIndex];
   }
-  // to fill last segment of Gaussian integration
+
+  // include the last integration segment (from the last point to dt) into the last timestep
   deltaT[ConvergenceOrder - 1] = deltaT[ConvergenceOrder - 1] + deltaT[0];
-  sumDt += deltaT[0];
+
+  // use that time points are symmetric to compute dt
+  const auto sumDt = timePoints[ConvergenceOrder - 1] + timePoints[0];
 
   return {sumDt, deltaT};
 }
--- a/src/DynamicRupture/FrictionLaws/FrictionSolver.cpp
+++ b/src/DynamicRupture/FrictionLaws/FrictionSolver.cpp
@@ -17,18 +17,15 @@
 namespace seissol::dr::friction_law {
 
 FrictionSolver::FrictionTime FrictionSolver::computeDeltaT(const std::vector<double>& timePoints) {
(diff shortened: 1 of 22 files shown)


src/DynamicRupture/FrictionLaws/FrictionSolver.h
```diff
@@ -68,7 +68,7 @@ class FrictionSolver {
    * Adjust initial stress by adding nucleation stress * nucleation function
    * For reference, see: https://strike.scec.org/cvws/download/SCEC_validation_slip_law.pdf.
    */
-  real deltaT[ConvergenceOrder] = {};
+  real deltaT[misc::TimeSteps] = {};
```
Thomas-Ulrich, line 71: real here in double above intentionate?

src/DynamicRupture/FrictionLaws/FrictionSolver.cpp
```diff
@@ -7,28 +7,30 @@
...
+               << deltaT.size();
+  }
+
+  deltaT[0] = timePoints[0]; // - 0
+  for (std::size_t timeIndex = 1; timeIndex < misc::TimeSteps; ++timeIndex) {
     deltaT[timeIndex] = timePoints[timeIndex] - timePoints[timeIndex - 1];
-    sumDt += deltaT[timeIndex];
   }
-  // to fill last segment of Gaussian integration
-  deltaT[ConvergenceOrder - 1] = deltaT[ConvergenceOrder - 1] + deltaT[0];
```
Thomas-Ulrich, line 30: do I understand correctly that deltaT[ConvergenceOrder - 1]  will have a different value now than previously?

davschneller: Ah wait—I meant to "revert" that change; because I don't know what's better.

The `dt[last] = timestep - prevLastPoint` (instead of `dt[last] = lastPoint - prevLastPoint` or similar) doesn't feel right to me, however. ... We can now also try to e.g. use Lobatto points, or a higher convergence order (however, on my laptop GPU, the runtime is directly dependent on the number of timesteps).

src/DynamicRupture/FrictionLaws/FrictionSolver.h
davschneller: Yes, the `real` here is intentionate. The `double` in the `FrictionLawData` is only for "transport"; once we call the DR function, the data is copied to these `real` arrays. (as it was before)

From what it stands, only the `timeWeights` are kept `double` always right now—maybe that could be changed rather easily as well without much precision loss? (haven't tried it, to be honest—it could speed up things a tiny little bit on gaming GPUs however)

davschneller:
Just tested it again with our precomputed `tpv5` on my laptop (which has a pretty memory-bound-ish GPU); order 6 single precision.
No notable performance difference there—so I think we should be good.

Here's the receiver difference between master and this branch:
```
Relative L2 error of the different quantities at the different faultreceivers
  absolute slip friction coefficient peak sliprate traction rupture time  sliprate slip normal velocity dynstress time
1           0.0                  0.0      0.000326      0.0     0.001065  0.000001  0.0        0.000002       0.000174
2           0.0                  0.0      0.000005      0.0          0.0  0.000001  0.0        0.000002       0.000648
3           0.0                  0.0           0.0      0.0          0.0  0.000001  0.0        0.000002       0.002168
4           0.0                  0.0      0.000859      0.0      0.00294  0.000001  0.0        0.000002       0.002168
```
(though I'm not 100%ly sure if there might be something to attribute to the error being addressed in #1407 here)

PR #1427, opened 2025-09-06
Prepare for header cleanup

davschneller:
* Soften all backwards-breaking changes introduced by #625 (i.e. check for the old parameters if the new names aren't found).
* remove explicit LTO option; since CMake can do that by itself already
* Start tidying headers to switch to `"???.h"` eventually (`clang-tidy` will need to be applied to the header files then as well—which currently still partially fails)


CONTRIBUTING.md
```diff
@@ -171,6 +177,20 @@ you still want to show it to SeisSol maintainers. In this case, click on a green
...
+#### Step 4 (Alternative)
+
+You can also run most of the linters manually via the following few lines:
+
+```bash
+.ci/format.sh $(which clang-format) .
+python .ci/filename.py --fix src app
+python .ci/header.py --fix src app
+black codegen/kernels/**/*.py
+isort --profile black codegen/kernels
```
Thomas-Ulrich, line 189: is there a command to for  markdownlint?

Thomas-Ulrich (approved):
LGTM. 
Good idea to bring back backwards compatibility.

CONTRIBUTING.md
davschneller: Yes; I've added it.
Albeit, you'll need to ignore the errors found in the submodules for now (which there are).

PR #1434, opened 2025-10-03
Print eta damp parameter value

davschneller:
And rename from "eta hack" to "eta damp"


fabian-kutschera (approved):
LGTM!

AliceGabriel:
Stop, please let us not call this 'eta-hack' anywhere in the code base! Let's agree on a better naming convention!

davschneller:
Yes, replacing it is easy. Waiting for a new name now. :)

EDIT: renamed to `damp` for now

PR #1436, opened 2025-10-04
Fix location flag output for free surface

davschneller:
(resulted from a datatype mismatch; the XdmfWriter still wants an `unsigned int` there)


src/Memory/Tree/Backmap.h
```diff
@@ -107,8 +107,10 @@ class StorageBackmap {
                          const TRef* layerData,
                          std::size_t cell,
                          std::size_t index) {
-    const auto position =
-        static_cast<std::size_t>((layerData + index) - zeroPosition) / sizeof(TRef);
+    const auto layerPosition = (reinterpret_cast<std::intptr_t>(layerData) -
+                                reinterpret_cast<std::intptr_t>(zeroPosition)) /
+                               sizeof(TRef);
+    const auto position = layerPosition + index;
```
Thomas-Ulrich, line 113: just to double check: this is not equivalent to the initial code (intended right?).

davschneller: Yes, it's not equivalent and intended. (that part was broken due to being not used)

PR #1440, opened 2025-10-15
update supermucNG phase 1 documentation to use stack24.5

Thomas-Ulrich:
Using stack24.5, it is possible to install Seissol quite straightforwardly.

~~To merge, two issues need to be fixed:~~
- ~~The generated executable is much slower than with intel23 (issue to come).~~
- ~~The environment Python does not allow installing new modules via pip.~~

### DIFF
--- a/docs/supermuc.rst
+++ b/docs/supermuc.rst
@@ -30,8 +30,8 @@ Update your ``~/.bashrc`` file as follows:
 
     # Add the custom SeisSol module path
     module use /hppfs/work/pn49ha/di73yeq4/user_spack24.5/SNG1/modules
-    # To access the easi Python module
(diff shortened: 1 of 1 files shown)


docs/supermuc.rst
```diff
@@ -279,20 +234,42 @@ and ``~/.spack/modules.yaml``:
               precision=double: double
               cuda: cuda
               debug: debug
+          projections:
```
davschneller, line 242: Still include ICC here? (or not anymore? We absolutely don't need to :) )

Thomas-Ulrich: it might be that the projections are actually not needed.
icc get a projection to intel23 with stack 23.1 (but there were no projection defined for oneapi within 23.1 when I first tried to install tandem with oneapi25 on stack 23.1).

Thomas-Ulrich: just tested, without the projections, I get
`seissol/1.3.1-o4-elas-dunav-double-impi` vs  ` seissol/1.3.1-oneapi25-o4-elas-dunav-double-impi` with.
So it is needed for oneapi.
(my understanding is that lrz defined some projection for intel23 already).

PR #1446, opened 2025-10-29
update lumi installation

Thomas-Ulrich:
Here are some update on LUMI (based on rocm/6.0.3).
It compiles and runs.

Here is an example of performance on 2 nodes, with a probably too small mesh to make big conclusions.
(16TFlops)
[14058953.seissol.log.txt](https://github.com/user-attachments/files/23214625/14058953.seissol.log.txt)

### DIFF
--- a/docs/lumi.rst
+++ b/docs/lumi.rst
@@ -41,15 +41,12 @@ Run ``pwd`` and copy the path there. Run the following script there.
 Next, we load the necessary modules for our SeisSol build.
 We set the compilers to the cray compiler wrappers (which in our case use ``amdclang`` internally).
 
-**NOTE: these modules are outdated.**
-
 .. code-block:: bash
 
     module load LUMI/24.03 partition/G
     module load cpeAMD/24.03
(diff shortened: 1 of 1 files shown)


docs/lumi.rst
```diff
@@ -45,19 +45,16 @@ We set the compilers to the cray compiler wrappers (which in our case use ``amdc
...
-    module load amd/5.6.1
-
-    module load buildtools/23.09
-
-    module load Boost/1.82.0-cpeAMD-23.09
+    module load LUMI/24.03 partition/G
+    module load cpeAMD/24.03
+    module load rocm/6.0.3
+    module load amd/6.0.3
+    module load Boost/1.83.0-cpeAMD-24.03
```
davschneller: Should only be needed for an ACPP installation (which has become much more difficult starting with ROCm 6.0 and beyond; since libLLVM isn't bundled with ROCm anymore since then)

davschneller (approved):
LGTM; thanks for the update.

PR #1450, opened 2025-11-06
Apply clang-tidy to the headers; refactor the CMake setup

davschneller:
I.e. make `seissol-common-properties` non-imported. Also, change all local headers to use `"..."` and not `<...>`
Thus, clang-tidy will add all future internal headers in the form `#include "..."`; not `#include <...>` anymore.
However, submodules will stay at `#include <...>` for now (since they are, in effect, independent of SeisSol).

Furthermore, rename the `MPI` class to `Mpi`, to prevent conflicts with an Intel MPI object/class (which is also called `MPI`), and place the `tests` into the top-level folder—to adhere a more "canonical" CMake structure (towards one file per target). The CMake structure is also pushed further by this PR; namely moving most subdirectory logic into the respective subdirectories; though more will need to be done in that regard in a future PR.

Also, we enable more clang-tidy tests and remove a free-floating `using namespace seissol;`.

### DIFF
--- a/app/Main/Main.cpp
+++ b/app/Main/Main.cpp
@@ -44,7 +44,7 @@
 namespace {
 using namespace seissol;
 
-auto readYamlParams(const std::string& parameterFile) -> std::shared_ptr<YAML::Node> {
+std::shared_ptr<YAML::Node> readYamlParams(const std::string& parameterFile) {
   // Read parameter file input from file
   fty::Loader<fty::AsLowercase> loader{};
   std::shared_ptr<YAML::Node> inputParams = nullptr;
@@ -57,7 +57,7 @@ auto readYamlParams(const std::string& parameterFile) -> std::shared_ptr<YAML::N
 }
 } // namespace
 
-auto main(int argc, char* argv[]) -> int {
+int main(int argc, char* argv[]) {
   try {
 #ifdef ACL_DEVICE
     seissol::Mpi::mpi.bindAcceleratorDevice();
--- a/docs/tpv16.rst
+++ b/docs/tpv16.rst
@@ -120,7 +120,7 @@ due to the heterogenous stress ratio on the fault. Figure
 [fig:tpv16slip] shows the fault slip rate along strike-direction at
 T=5.5 s.
 
-.. figure:: ./figures/PSR_16.png
+.. figure:: figures/PSR_16.png
    :alt: Mapview of fault slip rate along strike-direction.
    :width: 11.00000cm
    :align: center
@@ -131,14 +131,14 @@ There are several receivers on the fault surface. Figure
 [fig:tpv16fault] shows slip rate along the strike- and downdip-direction
 on the fault at point (15 km, 0 km, -9 km).
 
-.. figure:: ./figures/sliprate_16.png
+.. figure:: figures/sliprate_16.png
    :alt: Fault slip along strike- (left)and downdip- (right) direction.
    :width: 11.00000cm
    :align: center
 
    Fault slip along strike- (left)and downdip- (right) direction.
 
-.. figure:: ./figures/tpv16_uvw1.png
+.. figure:: figures/tpv16_uvw1.png
    :alt: Velocity at two opposite stations across the fault surface.
    :width: 11.00000cm
    :align: center
--- a/src/DynamicRupture/FrictionLaws/CpuImpl/RateAndState.h
+++ b/src/DynamicRupture/FrictionLaws/CpuImpl/RateAndState.h
@@ -171,7 +171,7 @@ class RateAndStateBase : public BaseFrictionLaw<RateAndStateBase<Derived, TPMeth
                                                              this->deltaT[timeIndex],
                                                              localSlipRate[pointIndex]);
       }
-      this->tpMethod.calcFluidPressure(
+      tpMethod.calcFluidPressure(
           normalStress, this->mu, localSlipRate, this->deltaT[timeIndex], false, ltsFace);
 
       updateNormalStress(normalStress, faultStresses, timeIndex, ltsFace);
--- a/src/Equations/elastic/Model/ElasticSetup.h
+++ b/src/Equations/elastic/Model/ElasticSetup.h
@@ -84,62 +84,64 @@ struct MaterialSetup<ElasticMaterial> {
                                         Tneigh& qGodNeighbor) {
     qGodNeighbor.setZero();
 
+    // matR == eigenvector matrix
+
     // Eigenvectors are precomputed
-    Matrix99 mR = Matrix99::Zero();
+    Matrix99 matR = Matrix99::Zero();
 
     if (testIfAcoustic(local.mu)) {
-      mR(0, 0) = local.lambda;
-      mR(1, 0) = local.lambda;
-      mR(2, 0) = local.lambda;
-      mR(6, 0) = std::sqrt((local.lambda) / local.rho);
+      matR(0, 0) = local.lambda;
+      matR(1, 0) = local.lambda;
+      matR(2, 0) = local.lambda;
+      matR(6, 0) = std::sqrt((local.lambda) / local.rho);
 
-      // scale for better condition number of mR
-      mR(3, 1) = local.lambda;
-      mR(5, 2) = local.lambda;
+      // scale for better condition number of matR
+      matR(3, 1) = local.lambda;
+      matR(5, 2) = local.lambda;
     } else {
-      mR(0, 0) = local.lambda + 2 * local.mu;
-      mR(1, 0) = local.lambda;
-      mR(2, 0) = local.lambda;
-      mR(6, 0) = std::sqrt((local.lambda + 2 * local.mu) / local.rho);
+      matR(0, 0) = local.lambda + 2 * local.mu;
+      matR(1, 0) = local.lambda;
+      matR(2, 0) = local.lambda;
+      matR(6, 0) = std::sqrt((local.lambda + 2 * local.mu) / local.rho);
 
-      mR(3, 1) = local.mu;
-      mR(7, 1) = std::sqrt(local.mu / local.rho);
+      matR(3, 1) = local.mu;
+      matR(7, 1) = std::sqrt(local.mu / local.rho);
 
-      mR(5, 2) = local.mu;
-      mR(8, 2) = std::sqrt(local.mu / local.rho);
+      matR(5, 2) = local.mu;
+      matR(8, 2) = std::sqrt(local.mu / local.rho);
     }
 
-    // scale for better condition number of mR
-    mR(4, 3) = local.lambda + 2 * local.mu;
-    mR(1, 4) = local.lambda + 2 * local.mu;
-    mR(2, 5) = local.lambda + 2 * local.mu;
+    // scale for better condition number of matR
+    matR(4, 3) = local.lambda + 2 * local.mu;
+    matR(1, 4) = local.lambda + 2 * local.mu;
+    matR(2, 5) = local.lambda + 2 * local.mu;
 
     if (testIfAcoustic(neighbor.mu)) {
-      // scale for better condition number of mR
-      mR(7, 6) = neighbor.lambda;
-      mR(8, 7) = neighbor.lambda;
+      // scale for better condition number of matR
+      matR(7, 6) = neighbor.lambda;
+      matR(8, 7) = neighbor.lambda;
 
-      mR(0, 8) = neighbor.lambda;
-      mR(1, 8) = neighbor.lambda;
-      mR(2, 8) = neighbor.lambda;
-      mR(6, 8) = -std::sqrt((neighbor.lambda + 2 * neighbor.mu) / neighbor.rho);
+      matR(0, 8) = neighbor.lambda;
+      matR(1, 8) = neighbor.lambda;
+      matR(2, 8) = neighbor.lambda;
+      matR(6, 8) = -std::sqrt((neighbor.lambda + 2 * neighbor.mu) / neighbor.rho);
     } else {
-      mR(5, 6) = neighbor.mu;
-      mR(8, 6) = -std::sqrt(neighbor.mu / neighbor.rho);
+      matR(5, 6) = neighbor.mu;
+      matR(8, 6) = -std::sqrt(neighbor.mu / neighbor.rho);
 
-      mR(3, 7) = neighbor.mu;
-      mR(7, 7) = -std::sqrt(neighbor.mu / neighbor.rho);
+      matR(3, 7) = neighbor.mu;
+      matR(7, 7) = -std::sqrt(neighbor.mu / neighbor.rho);
 
-      mR(0, 8) = neighbor.lambda + 2 * neighbor.mu;
-      mR(1, 8) = neighbor.lambda;
-      mR(2, 8) = neighbor.lambda;
-      mR(6, 8) = -std::sqrt((neighbor.lambda + 2 * neighbor.mu) / neighbor.rho);
+      matR(0, 8) = neighbor.lambda + 2 * neighbor.mu;
+      matR(1, 8) = neighbor.lambda;
+      matR(2, 8) = neighbor.lambda;
+      matR(6, 8) = -std::sqrt((neighbor.lambda + 2 * neighbor.mu) / neighbor.rho);
     }
 
     if (faceType == FaceType::FreeSurface) {
       const MaterialType materialtype =
           testIfAcoustic(local.mu) ? MaterialType::Acoustic : MaterialType::Elastic;
-      getTransposedFreeSurfaceGodunovState(materialtype, qGodLocal, qGodNeighbor, mR);
+      getTransposedFreeSurfaceGodunovState(materialtype, qGodLocal, qGodNeighbor, matR);
     } else {
       Matrix99 chi = Matrix99::Zero();
       if (!testIfAcoustic(local.mu)) {
@@ -148,7 +150,7 @@ struct MaterialSetup<ElasticMaterial> {
       }
       chi(0, 0) = 1.0;
 
-      const auto godunov = ((mR * chi) * mR.inverse()).eval();
+      const auto godunov = ((matR * chi) * matR.inverse()).eval();
 
       // qGodLocal = I - qGodNeighbor
       for (unsigned i = 0; i < godunov.cols(); ++i) {
--- a/src/Equations/poroelastic/Model/PoroelasticSetup.h
+++ b/src/Equations/poroelastic/Model/PoroelasticSetup.h
@@ -305,21 +305,23 @@ struct MaterialSetup<PoroElasticMaterial> {
         chiPlus(i, i) = 1.0;
       }
     }
-    CMatrix mR = localEigenvectors * chiMinus + neighborEigenvectors * chiPlus;
+
+    // matR == eigenvector matrix
+    CMatrix matR = localEigenvectors * chiMinus + neighborEigenvectors * chiPlus;
     // set null space eigenvectors manually
-    mR(1, 4) = 1.0;
-    mR(2, 5) = 1.0;
-    mR(12, 6) = 1.0;
-    mR(11, 7) = 1.0;
-    mR(4, 8) = 1.0;
+    matR(1, 4) = 1.0;
+    matR(2, 5) = 1.0;
+    matR(12, 6) = 1.0;
+    matR(11, 7) = 1.0;
+    matR(4, 8) = 1.0;
     if (faceType == FaceType::FreeSurface) {
-      Matrix realR = mR.real();
+      Matrix realR = matR.real();
       getTransposedFreeSurfaceGodunovState(
           MaterialType::Poroelastic, qGodLocal, qGodNeighbor, realR);
     } else {
-      CMatrix invR = mR.inverse();
-      CMatrix godunovMinus = mR * chiMinus * invR;
-      CMatrix godunovPlus = mR * chiPlus * invR;
+      CMatrix invR = matR.inverse();
+      CMatrix godunovMinus = matR * chiMinus * invR;
+      CMatrix godunovPlus = matR * chiPlus * invR;
 
       for (unsigned i = 0; i < qGodLocal.shape(1); ++i) {
         for (unsigned j = 0; j < qGodLocal.shape(0); ++j) {
--- a/src/Model/Common.h
+++ b/src/Model/Common.h
@@ -35,8 +35,8 @@ constexpr bool testIfAcoustic(T mu) {
 }
 
 template <typename Tmaterial, typename Tmatrix>
-void getTransposedCoefficientMatrix(const Tmaterial& material, unsigned dim, Tmatrix& mM) {
-  MaterialSetup<Tmaterial>::getTransposedCoefficientMatrix(material, dim, mM);
+void getTransposedCoefficientMatrix(const Tmaterial& material, unsigned dim, Tmatrix& matM) {
+  MaterialSetup<Tmaterial>::getTransposedCoefficientMatrix(material, dim, matM);
 }
 
 template <typename Tmaterial, typename T>
@@ -63,7 +63,7 @@ template <typename T, typename Tmatrix>
 void getTransposedFreeSurfaceGodunovState(MaterialType materialtype,
                                           T& qGodLocal,
                                           T& qGodNeighbor,
-                                          Tmatrix& mR);
+                                          Tmatrix& matR);
 
 template <typename T>
 void getPlaneWaveOperator(const T& material,
@@ -130,9 +130,9 @@ void getElasticPlaneWaveOperator(
     const MaterialT& material,
     const double n[3],
     std::complex<double> mdata[MaterialT::NumQuantities * MaterialT::NumQuantities]) {
-  yateto::DenseTensorView<2, std::complex<double>> mM(
+  yateto::DenseTensorView<2, std::complex<double>> matM(
       mdata, {MaterialT::NumQuantities, MaterialT::NumQuantities});
-  mM.setZero();
+  matM.setZero();
 
   double data[MaterialT::NumQuantities * MaterialT::NumQuantities];
   yateto::DenseTensorView<2, double> coeff(data,
@@ -144,7 +144,7 @@ void getElasticPlaneWaveOperator(
 
     for (unsigned i = 0; i < MaterialT::NumQuantities; ++i) {
       for (unsigned j = 0; j < MaterialT::NumQuantities; ++j) {
-        mM(i, j) += n[d] * coeff(j, i);
+        matM(i, j) += n[d] * coeff(j, i);
       }
     }
   }
@@ -153,7 +153,7 @@ void getElasticPlaneWaveOperator(
 
   for (unsigned i = 0; i < MaterialT::NumQuantities; ++i) {
     for (unsigned j = 0; j < MaterialT::NumQuantities; ++j) {
-      mM(i, j) -= std::complex<double>(0.0, coeff(j, i));
+      matM(i, j) -= std::complex<double>(0.0, coeff(j, i));
     }
   }
 }
@@ -250,19 +250,21 @@ template <typename T, typename Tmatrix>
 void seissol::model::getTransposedFreeSurfaceGodunovState(MaterialType materialtype,
                                                           T& qGodLocal,
                                                           T& qGodNeighbor,
-                                                          Tmatrix& mR) {
+                                                          Tmatrix& matR) {
   for (size_t i = 0; i < seissol::model::MaterialT::NumElasticQuantities; i++) {
     for (size_t j = 0; j < seissol::model::MaterialT::NumElasticQuantities; j++) {
       qGodNeighbor(i, j) = std::numeric_limits<double>::signaling_NaN();
     }
   }
 
+  // matR == eigenvector matrix
+
   qGodLocal.setZero();
   switch (materialtype) {
   case MaterialType::Acoustic: {
     // Acoustic material only has one traction (=pressure) and one velocity comp.
     // relevant to the Riemann problem
-    qGodLocal(0, 6) = -1 * mR(6, 0) * 1 / mR(0, 0); // mS
+    qGodLocal(0, 6) = -1 * matR(6, 0) * 1 / matR(0, 0); // mS
     qGodLocal(6, 6) = 1.0;
     break;
   }
@@ -273,8 +275,8 @@ void seissol::model::getTransposedFreeSurfaceGodunovState(MaterialType materialt
     const std::array<int, 4> tractionIndices = {0, 3, 5, 9};
     const std::array<int, 6> velocityIndices = {6, 7, 8, 10, 11, 12};
     const std::array<int, 4> columnIndices = {0, 1, 2, 3};
-    const Matrix44 matR11 = mR(tractionIndices, columnIndices);
-    const Matrix64 matR21 = mR(velocityIndices, columnIndices);
+    const Matrix44 matR11 = matR(tractionIndices, columnIndices);
+    const Matrix64 matR21 = matR(velocityIndices, columnIndices);
     const Matrix64 matS = (-(matR21 * matR11.inverse())).eval();
     setBlocks(qGodLocal, matS, tractionIndices, velocityIndices);
     break;
@@ -283,8 +285,8 @@ void seissol::model::getTransposedFreeSurfaceGodunovState(MaterialType materialt
     const std::array<int, 3> tractionIndices = {0, 3, 5};
     const std::array<int, 3> velocityIndices = {6, 7, 8};
     using Matrix33 = Eigen::Matrix<double, 3, 3>;
-    const Matrix33 matR11 = mR(tractionIndices, {0, 1, 2});
-    const Matrix33 matR21 = mR(velocityIndices, {0, 1, 2});
+    const Matrix33 matR11 = matR(tractionIndices, {0, 1, 2});
+    const Matrix33 matR21 = matR(velocityIndices, {0, 1, 2});
     const Matrix33 matS = (-(matR21 * matR11.inverse())).eval();
     setBlocks(qGodLocal, matS, tractionIndices, velocityIndices);
     break;
--- a/app/Proxy/Main.cpp
+++ b/app/Proxy/Main.cpp
@@ -26,7 +26,7 @@
 
 using namespace seissol::proxy;
 
-auto main(int argc, char* argv[]) -> int {
+int main(int argc, char* argv[]) {
   std::stringstream kernelHelp;
   auto allowedKernels = Aux::getAllowedKernels();
   kernelHelp << "Kernels to benchmark. A comma-separated list of (those kernels will be run "
--- a/docs/tpv5.rst
+++ b/docs/tpv5.rst
@@ -20,7 +20,7 @@ fault surface. As it propagates away from the nucleation zone, it
 encounters two square patches with initial stress conditions that are
 different from the rest of the fault surface.
 
-.. figure:: ./figures/tpv5_mesh.png
+.. figure:: figures/tpv5_mesh.png
    :alt: Diagram of TPV5.
    :width: 9.00000cm
    :align: center
--- a/src/Equations/acoustic/Model/AcousticSetup.h
+++ b/src/Equations/acoustic/Model/AcousticSetup.h
@@ -26,25 +26,25 @@ template <>
 struct MaterialSetup<AcousticMaterial> {
   template <typename T>
   static void
-      getTransposedCoefficientMatrix(const AcousticMaterial& material, unsigned dim, T& mM) {
-    mM.setZero();
+      getTransposedCoefficientMatrix(const AcousticMaterial& material, unsigned dim, T& matM) {
+    matM.setZero();
 
     const double rhoInv = 1.0 / material.rho;
 
     switch (dim) {
     case 0:
-      mM(1, 0) = material.lambda;
-      mM(0, 1) = rhoInv;
+      matM(1, 0) = material.lambda;
+      matM(0, 1) = rhoInv;
       break;
 
     case 1:
-      mM(2, 0) = material.lambda;
-      mM(0, 2) = rhoInv;
+      matM(2, 0) = material.lambda;
+      matM(0, 2) = rhoInv;
       break;
 
     case 2:
-      mM(3, 0) = material.lambda;
-      mM(0, 3) = rhoInv;
+      matM(3, 0) = material.lambda;
+      matM(0, 3) = rhoInv;
       break;
 
     default:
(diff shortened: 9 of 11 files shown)


app/Main/Main.cpp
```diff
@@ -44,7 +42,9 @@
 #include "Version.h"
 
 namespace {
-std::shared_ptr<YAML::Node> readYamlParams(const std::string& parameterFile) {
+using namespace seissol;
+
+auto readYamlParams(const std::string& parameterFile) -> std::shared_ptr<YAML::Node> {
```
vikaskurapati: While I agree that this trailing type helps for complex return types to make them more readable, here for this simple return type, don't you think the traditional return suits okay? Or do you intend to make it consistent that this is the kind of return type which we are going to use across all of the software package?

src/DynamicRupture/FrictionLaws/CpuImpl/LinearSlipWeakening.cpp
```diff
@@ -6,11 +6,13 @@
 // SPDX-FileContributor: Author lists in /AUTHORS and /CITATION.cff
 
 #include "LinearSlipWeakening.h"
+
```
vikaskurapati: Why do we need an extra white space here? And have this out of the alphabetic order? Is it because this `.cpp` and `.h` are directly linked?

davschneller:
The NVHPC build runs currently fail in the container creation already (before anything SeisSol-specific could happen)—something might have changed on the Github Actions side recently probably. (the runs also fail on master for the very same reason)

vikaskurapati:
> The NVHPC build runs currently fail in the container creation already (before anything SeisSol-specific could happen)—something might have changed on the Github Actions side recently probably. (the runs also fail on master for the very same reason)

OK. I started looking at the PR. Is there someway we could reproduce the CI workflow on one of the available machines to check if there is an inherent issue?

davschneller:
Thanks for doing that (I know; it's file-wise a bit big).

For NVHPC, I did try to build it on my local workstation—it seemed to work.
I'm running it right now on the `davschneller/ci-merge-test` branch (which contains an updated CI; but still incomplete—it currently uses my local workstation for everything); that should test some other compilers as well.

EDIT: the tests seemed to pass as usual (besides the existing bugs); however there was a bug with the point source initialization. Fixed in the commit below.

app/Main/Main.cpp
vikaskurapati: Why are we using trailing returns for simple return types? Or are we now enforcing this for the entire repository now?

app/Main/Main.cpp
```diff
@@ -44,7 +42,9 @@
 #include "Version.h"
 
 namespace {
-std::shared_ptr<YAML::Node> readYamlParams(const std::string& parameterFile) {
+using namespace seissol;
```
vikaskurapati, line 45: I had the impression that we were getting rid of this namespace. Or did I understand it wrong?

docs/tpv16.rst
```diff
@@ -120,7 +120,7 @@ due to the heterogenous stress ratio on the fault. Figure
 [fig:tpv16slip] shows the fault slip rate along strike-direction at
 T=5.5 s.
 
-.. figure:: ./LatexFigures/PSR_16.png
+.. figure:: ./figures/PSR_16.png
```
vikaskurapati: Any particular reason why we have a ./ here in contrast to the above paths?

docs/tpv16.rst
```diff
@@ -131,14 +131,14 @@ There are several receivers on the fault surface. Figure
 [fig:tpv16fault] shows slip rate along the strike- and downdip-direction
 on the fault at point (15 km, 0 km, -9 km).
 
-.. figure:: ./LatexFigures/sliprate_16.png
+.. figure:: ./figures/sliprate_16.png
```
vikaskurapati: Any particular reason why we have a ./ here in contrast to the above paths?

docs/tpv16.rst
```diff
@@ -131,14 +131,14 @@ There are several receivers on the fault surface. Figure
...
    :alt: Fault slip along strike- (left)and downdip- (right) direction.
    :width: 11.00000cm
    :align: center
 
    Fault slip along strike- (left)and downdip- (right) direction.
 
-.. figure:: ./LatexFigures/tpv16_uvw1.png
+.. figure:: ./figures/tpv16_uvw1.png
```
vikaskurapati: Any particular reason why we have a ./ here in contrast to the above paths?

src/Common/Offset.h
```diff
@@ -12,11 +12,19 @@
 
 namespace seissol {
 
+// NOLINTBEGIN (bugprone-macro-parentheses)
+
+// IMPORTANT: these macros use member names directly; and thus cannot be safeguarded by parentheses.
+// (I'm sorry, clang-tidy)
```
vikaskurapati: You don't need to apologise. :) I would remove this comment.

src/DynamicRupture/Misc.h
```diff
@@ -66,19 +65,19 @@ static constexpr double TpMaxWaveNumber = 10.0;
 static constexpr unsigned int NumBoundaryGaussPoints =
     init::QInterpolated::Shape[multisim::BasisFunctionDimension];
 
-template <class TupleT, class F, std::size_t... I>
-constexpr F forEachImpl(TupleT&& tuple, F&& functor, std::index_sequence<I...> /*unused*/) {
-  return (void)std::initializer_list<int>{
-             (std::forward<F>(functor)(std::get<I>(std::forward<TupleT>(tuple)), I), 0)...},
-         functor;
+template <std::size_t I, typename F, typename TupleT>
+constexpr F forEachElement(F&& functor, TupleT&& tuple) {
```
vikaskurapati, line 69: Don't you think using a trailing return here is a good idea?

Also, this function is not clear on what it is doing just by looking at the code. It needs documentation on what it it is intended to be done, and also how it is to be used, along with test for it if possible.

src/DynamicRupture/Output/Builders/PickPointBuilder.h
```diff
@@ -83,31 +84,36 @@ class PickPointBuilder : public ReceiverBasedOutputBuilder {
...
-        receiver.elementIndex = element.localId;
-        receiver.elementGlobalIndex = element.globalId;
-
-        receiver.reference =
-            transformations::tetrahedronGlobalToReference(meshVertices[element.vertices[0]].coords,
-                                                          meshVertices[element.vertices[1]].coords,
-                                                          meshVertices[element.vertices[2]].coords,
-                                                          meshVertices[element.vertices[3]].coords,
-                                                          receiver.global.getAsEigen3LibVector());
+      try {
```
vikaskurapati, line 90: Do we still have the problem with the receivers not being found? Didn't you already fix it with a modified algorithm?

src/Equations/Datastructures.h
```diff
@@ -34,7 +33,7 @@
 
 // IWYU pragma: end_exports
 
-#include <Config.h>
+#include "Config.h"
```
vikaskurapati: Why is this in the end in contrast to the others?

src/Equations/elastic/Model/ElasticSetup.h
```diff
@@ -80,66 +80,66 @@ struct MaterialSetup<ElasticMaterial> {
...
-                                        Tloc& QgodLocal,
-                                        Tneigh& QgodNeighbor) {
-    QgodNeighbor.setZero();
+                                        Tloc& qGodLocal,
+                                        Tneigh& qGodNeighbor) {
+    qGodNeighbor.setZero();
 
     // Eigenvectors are precomputed
-    Matrix99 R = Matrix99::Zero();
+    Matrix99 mR = Matrix99::Zero();
```
vikaskurapati: Please use a better variable name.

src/Equations/poroelastic/Model/PoroelasticSetup.h
```diff
@@ -306,26 +305,26 @@ struct MaterialSetup<PoroElasticMaterial> {
         chiPlus(i, i) = 1.0;
       }
     }
-    CMatrix R = localEigenvectors * chiMinus + neighborEigenvectors * chiPlus;
+    CMatrix mR = localEigenvectors * chiMinus + neighborEigenvectors * chiPlus;
```
vikaskurapati: a better variable name is nice to have.

src/Geometry/CubeGenerator.cpp
```diff
@@ -1526,7 +1531,7 @@ void CubeGenerator::cubeGenerator(const std::array<std::size_t, 4> numCubes,
 
   // Get maximum number of neighbors (required to get collective MPI-IO right)
   const int maxNeighbors = bndSize;
-  // MPI_Allreduce(MPI_IN_PLACE, &maxNeighbors, 1, MPI_INT, MPI_MAX, seissol::MPI::mpi.comm());
+  // MPI_Allreduce(MPI_IN_PLACE, &maxNeighbors, 1, MPI_INT, MPI_MAX, seissol::Mpi::mpi.comm());
```
vikaskurapati: Do you still want to retain this comment for the future?

src/Geometry/PUMLReader.cpp
```diff
@@ -177,23 +174,22 @@ const std::array<std::int32_t, 4> FirstFaceVertex = {0, 0, 0, 1};
...
-                       const char* partitioningLib,
+PUMLReader::PUMLReader(const std::string& meshFile,
+                       const std::string& partitioningLib,
                        seissol::initializer::parameters::BoundaryFormat boundaryFormat,
                        seissol::initializer::parameters::TopologyFormat topologyFormat,
                        initializer::time_stepping::LtsWeights* ltsWeights,
                        double tpwgt)
-    : MeshReader(seissol::MPI::mpi.rank()), boundaryFormat(boundaryFormat),
-      topologyFormat(topologyFormat) {
+    : MeshReader(seissol::Mpi::mpi.rank()), boundaryFormat(boundaryFormat) {
```
vikaskurapati: Why remove only topologyFormat, and keep boundaryFormat here? I see that the topologyFormat is directly used from the function inputs here. These kinds of changes are quite confusing in the end, mainly when it is not clear just by looking at the local code, without any description.

src/Initializer/InitProcedure/InitLayout.cpp
```diff
@@ -247,7 +253,8 @@ void setupMemory(seissol::SeisSol& seissolInstance) {
             cellInformation[index].neighborConfigIds[face] = 0;
           } else {
             secondaryCellInformation[index].faceNeighbors[face] = StoragePosition::NullPosition;
-            cellInformation[index].neighborConfigIds[face] = -1;
+            cellInformation[index].neighborConfigIds[face] =
+                std::numeric_limits<std::uint32_t>::max();
```
vikaskurapati, line 256: I don't understand the reasoning behind changing this from -1 to max

src/Initializer/Parameters/LtsParameters.h
```diff
@@ -57,10 +56,7 @@ class LtsParameters {
   [[nodiscard]] bool isAutoMergeUsed() const;
   [[nodiscard]] double getAllowedPerformanceLossRatioAutoMerge() const;
   [[nodiscard]] AutoMergeCostBaseline getAutoMergeCostBaseline() const;
-  [[nodiscard]] double getWiggleFactor() const;
   [[nodiscard]] LtsWeightsTypes getLtsWeightsType() const;
-  void setWiggleFactor(double factor);
```
vikaskurapati, line 62: From what I understand you moved the logic of this to another function setTimeStepScale or similar. It probably worked out, and nothing changed, but I would still suggest to verify results for some simulation with a wiggleFactor to be absolutely sure.

src/Initializer/Parameters/ModelParameters.h
```diff
@@ -8,40 +8,40 @@
...
 struct ITMParameters {
-  bool itmEnabled;
-  double itmStartingTime;
-  double itmDuration;
-  double itmVelocityScalingFactor;
-  ReflectionType itmReflectionType;
+  bool itmEnabled{false};
+  double itmStartingTime{};
+  double itmDuration{};
+  double itmVelocityScalingFactor{};
```
vikaskurapati: A better default scaling factor is 1.0

src/Initializer/Parameters/ModelParameters.h
```diff
@@ -8,40 +8,40 @@
...
+  ReflectionType itmReflectionType{};
```
vikaskurapati: I'm not sure what default values are used for enum classes, but a better default would be ReflectionType::BothWaves here.

src/Initializer/Parameters/ModelParameters.h
```diff
@@ -8,40 +8,40 @@
...
-  bool useCellHomogenizedMaterial;
-  double freqCentral;
-  double freqRatio;
-  double gravitationalAcceleration;
-  double tv;
+  bool hasBoundaryFile{false};
+  bool plasticity{false};
+  bool useCellHomogenizedMaterial{true};
+  double freqCentral{};
+  double freqRatio{};
```
vikaskurapati: I think the default freqRatio is better set to 1.0 given that we do a log of it.

src/Memory/MemoryAllocator.cpp
```diff
@@ -11,117 +11,142 @@
...
   // presumably a clang-tidy bug; it'd want to make ptrBuffer const
   // NOLINTNEXTLINE(misc-const-correctness)
   void* ptrBuffer{nullptr};
   bool error = false;
 
-  /* handle zero allocation */
+  // handle zero-size allocations
   if (size == 0) {
-    // logWarning() << "allocation of size 0 requested, returning nullptr; (alignment: " <<
-    // alignment << ", memkind: " << memkind << ").";
```
vikaskurapati, line 36: Why is this warning not useful anymore? Or does this case not occur anymore?

src/Memory/MemoryAllocator.h
```diff
@@ -141,7 +147,7 @@ class ManagedAllocator {
    * @param  alignment alignment of the memory chunk in byte.
    * @return pointer, which points to the aligned memory of the given size.
    **/
-  void* allocateMemory(size_t size, size_t alignment = 1, enum Memkind memkind = Standard);
+  void* allocateMemory(size_t size, size_t alignment = 1, Memkind memkind = Memkind::Standard);
```
vikaskurapati, line 150: There is a parameter missing in the documentation above.

src/Model/Common.h
```diff
@@ -36,13 +34,13 @@ constexpr bool testIfAcoustic(T mu) {
 }
 
 template <typename Tmaterial, typename Tmatrix>
-void getTransposedCoefficientMatrix(const Tmaterial& material, unsigned dim, Tmatrix& M) {
-  MaterialSetup<Tmaterial>::getTransposedCoefficientMatrix(material, dim, M);
+void getTransposedCoefficientMatrix(const Tmaterial& material, unsigned dim, Tmatrix& mM) {
```
vikaskurapati: Can we have a better variable naming for matrix here?

src/Model/Common.h
```diff
@@ -153,21 +151,21 @@ void getElasticPlaneWaveOperator(
...
+      mM(i, j) -= std::complex<double>(0.0, coeff(j, i));
     }
   }
 }
 
 } // namespace seissol::model
 
 template <typename T, typename Tmatrix, typename Tarray1, typename Tarray2>
-void setBlocks(T qGodLocal, Tmatrix S, Tarray1 tractionIndices, Tarray2 velocityIndices) {
+void setBlocks(T qGodLocal, Tmatrix mS, Tarray1 tractionIndices, Tarray2 velocityIndices) {
```
vikaskurapati, line 164: A better name for mS, please.

src/Parallel/MPI.cpp
```diff
@@ -79,7 +81,7 @@ void seissol::MPI::printAcceleratorDeviceInfo() {
 #endif
 }
 
-void seissol::MPI::setDataTransferModeFromEnv() {
+void seissol::Mpi::setDataTransferModeFromEnv() {
   // TODO (Ravil, David): switch to reading this option from the parameter-file.
   // Waiting for David to finish his `no-fortran` PR
   const auto envVariable =
```
vikaskurapati, line 85: Is this not already done? Or does this comment refer to the fortran part of poroelasticity?

src/Solver/TimeStepping/TimeCluster.h
```diff
@@ -51,11 +49,11 @@ namespace seissol::time_stepping {
 class TimeCluster : public AbstractTimeCluster {
   private:
   // Last correction time of the neighboring cluster with higher dt
-  double lastSubTime;
+  double lastSubTime{0};
```
vikaskurapati, line 52: May be very nitpicky here, but in a lot of places, you use {} for default initialization, but here you specifically put a {0} to initialize this. Isn't a double initialized by a 0 by default anyway? Is there a particular reason this is done this way here?

src/Solver/TimeStepping/TimeCluster.h
```diff
@@ -51,11 +49,11 @@ namespace seissol::time_stepping {
...
 
   // The timestep of the largest neighbor. Not well-defined (and not used) for the largest local
   // timecluster.
-  double neighborTimestep;
+  double neighborTimestep{0};
```
vikaskurapati, line 56: May be very nitpicky here, but in a lot of places, you use {} for default initialization, but here you specifically put a {0} to initialize this. Isn't a double initialized by a 0 by default anyway? Is there a particular reason this is done this way here?

src/Solver/TimeStepping/TimeManager.cpp
```diff
@@ -185,12 +186,10 @@ void TimeManager::addClusters(const initializer::ClusterLayout& clusterLayout,
   }
 
   // Create ghost time clusters for MPI
-  const auto preferredDataTransferMode = MPI::mpi.getPreferredDataTransferMode();
+  const auto preferredDataTransferMode = Mpi::mpi.getPreferredDataTransferMode();
   const auto persistent = usePersistentMpi(seissolInstance.env());
   for (auto& layer : memoryManager.getLtsStorage().leaves(Ghost | Interior)) {
 
-    const auto clusterId = layer.getIdentifier().lts;
```
vikaskurapati, line 192: Why is it better here to call the function twice instead of calling it once and using it twice? This seems counterintuitive to me.

src/SourceTerm/Manager.cpp
```diff
@@ -99,6 +101,12 @@ void computeMInvJInvPhisAtSources(
   krnl.execute();
 }
 
+struct SourceFile {
+  std::vector<std::size_t> originalIndex;
+  std::vector<std::size_t> meshIds;
```
vikaskurapati, line 106: Just out of curiosity, why no default initialization braces for vectors?

tests/Geometry/VariableSubsampler.t.h
```diff
@@ -0,0 +1,82 @@
...
+
+    const seissol::refinement::DivideTetrahedronBy4<double> refineBy4;
+    const seissol::refinement::VariableSubsampler<double> subsampler(
+        1, refineBy4, ConvergenceOrder, Quantities, AlignedQuantities);
+
+    std::array<real, Quantities * SubTriangles> expectedDOFs{};
+
+    std::uniform_real_distribution<> rngdist(0.0, 1.0);
+
+    // For order 3 there are 108 DOFs (taking alignment into account)
```
vikaskurapati: This is for a particular architecture, or? Are the tests here for a particular architecture like this? Mentioning that here may help to keep track.

tests/Initializer/PointMapper.t.h
```diff
@@ -0,0 +1,48 @@
...
+#include "Geometry/MockReader.h"
+#include "Initializer/PointMapper.h"
+
+#include <Eigen/Dense>
+#include <random>
+
+namespace seissol::unit_test {
+
+TEST_CASE("Point mapper") {
+  // We do all tests in double precision
```
vikaskurapati, line 17: Do they fail in single precision? Or is there some other reason for it?

tests/TestHelper.h
```diff
@@ -122,6 +137,18 @@ ostream& operator<<(ostream& stream, const std::vector<T>& vec) {
   stream << "}";
   return stream;
 }
+
+template <typename T, std::size_t N>
+ostream& operator<<(ostream& stream, const std::array<T, N>& vec) {
+  stream << "{";
+  for (const auto& item : vec) {
+    stream << item << ",";
```
vikaskurapati: I personally prefer putting spaces around the comma, making it easier to read on the console for vectors. However, if you prefer it this way for some other reason, without spaces, I won't complain.

app/Main/Main.cpp
davschneller: Not yet :)
Though we could now switch to it if everyone feels comfortable with that change.

app/Main/Main.cpp
davschneller: Ah, it was about getting rid of it in the header files; sorry should've clarified. :)

(though some style guides want to get rid of it _absolutely_ everywhere; but IMO having them increases the readability to have it in the source files there ... On the other hand, thinking about it now—it should be possible to place the anonymous namespaces inside the `seissol` namespace; thus eliminating the need for the `using` directives here; maybe that's preferable. I'll try it out and report back—though it shouldn't make any difference in the end.)

docs/tpv16.rst
davschneller: I guess it's a "historic" reason if there is one; I can't think of one ad hoc.

src/DynamicRupture/Misc.h
davschneller: Since `F` is also used in the arguments, I don't think it'll help us much. ... However, I'll try.

As a side note, after thinking about it tbh, the underlying design in this place feels a bit overkill to me; though I'd personally need to look a bit deeper to see if there's an easier (i.e. less C++-demanding) replacement for it.

src/DynamicRupture/Output/Builders/PickPointBuilder.h
davschneller: That problem should be fixed, yes. The extra try-catch was due to some obscure other errors (i.e. the linter complained about that happening).

src/Equations/Datastructures.h
davschneller: Ah, that's probably because the `IWYU pragma` lines prevent uniting the include blocks. I'll move it up manually.

src/Equations/elastic/Model/ElasticSetup.h
davschneller: Any good idea on how to name matrix variables?
I did use `matR` in other cases; but it all doesn't seem like a "proper" solution to me tbh.

src/Geometry/CubeGenerator.cpp
davschneller: We can get rid of it.
(or just remove the whole cube generator for now—though having an "inline" mesh generator is sometimes probably also nice to have I guess)

src/Geometry/PUMLReader.cpp
davschneller: It might be more useful to also "localize" the `boundaryFormat` probably—if that helps.

src/Initializer/InitProcedure/InitLayout.cpp
davschneller: The `-1` works the same way (to get the `0xffffffff` value), but throws a linter warning due to the implicit sign conversion.
It _is_ possible to silence the conversion probably but using the explicit max limit feels more "proper".

src/Initializer/Parameters/LtsParameters.h
davschneller: Yes, it got moved there, because previously, the wiggle factor was written into the SeisSol parameters and then read from it again. However, ideally, the parameters should be a constant input (which they are now with this PR).

I'll see if I can get a test for it. (or maybe we should also add one to the CI?)

src/Memory/MemoryAllocator.cpp
davschneller: Zero-length allocations should be allowed; they are allowed to happen. :) Or at least, we'll need to catch zero-length allocations somewhere the latest; so here seems like a good place for that.
The warning has been disabled (but kept as a comment) already since late 2017.

src/Parallel/MPI.cpp
davschneller: Yeah, it does refer to #829 . Though I think this one is better as an environment variable, since it's more system-dependent and not simulation-dependent (i.e. it doesn't make a difference w.r.t. the simulation results).

src/Solver/TimeStepping/TimeCluster.h
davschneller: Yeah, I guess it doesn't make a semantic difference. I used it mostly because here we consciously _know_ that the `0` is what we want, and not just an initializer.
But it's probably all dependent on "taste" after all.

src/Solver/TimeStepping/TimeManager.cpp
davschneller: The calls happen only inside asserts which are disabled during non-debug time; hence the linter complained.
Also, the calls should be really cheap; in essence the `getIdentifier()` is just a member accessor.

src/SourceTerm/Manager.cpp
davschneller: They have a default constructor—so we don't need that for them.
That's in contrast to the standard number datatypes which stay un-initialized unless we force C++ to initialize them; same for C-style arrays and `std::array`s which work on those types. I guess that's some C baggage in the end.

Cf. https://en.cppreference.com/w/cpp/language/default_initialization.html or https://herbsutter.com/2024/08/07/reader-qa-what-does-it-mean-to-initialize-an-int/ (seemingly, that behavior gets adjusted/warned for in C++26).

tests/Geometry/VariableSubsampler.t.h
davschneller: Ah I see—I set the alignment to `4` in the `constexpr`s above. Let me clarify it.

tests/Initializer/PointMapper.t.h
davschneller: The coordinate calculations currently always happen in double precision, since that is only done during the initialization and won't impact the simulation runtime.

It sometimes did make a difference, IIRC the tpv16 single-precision test case was affected by that.

tests/TestHelper.h
davschneller: You're right, spaces to look better there. Let me add them.

src/DynamicRupture/Misc.h
davschneller: Looking at it again—there are already tests sort of testing it. (`tests/DynamicRupture/Output/Variables.t.h`)

davschneller:
All reviews should be complete; except the renaming of the matrices—I may need some input on how to either generically rename them (e.g. make `A` -> `matA` or the likes), or get specific names for each of them.

I added one additional `clang-tidy` to rule (cf. https://github.com/SeisSol/SeisSol/pull/1450/commits/1f077aacc8ef20e728d5425d6ec5bdab6acb8c95 ; that one actually helped to find some typo bugs).

app/Main/Main.cpp
vikaskurapati: Personally, I don't really like trailing types for simple return types like in this scenario, and I could not find a reason online why they are supposed to be used unless the return type is too complex w.r.t some extra template deduction issues, or dependency on the type of the parameters etc. For example, in this scenario, I would have stayed with a normal return type. This is not a major issue, though. So, _if everyone_ else wants it this way, I would not protest.

app/Main/Main.cpp
vikaskurapati: It may not be relevant here, but using such `using namespace ...` commands make me kinda scared thinking that the same methods may possibly be existing in different namespaces and I am not sure which one exactly is being called in this instance. If this is a readability issue, this takes precedence for sure, but otherwise, we should avoid using these as much as possible, IMO.

docs/tpv16.rst
vikaskurapati: Okay. Can we remove it, and does it still work? If so, can we remove it for consistency?

src/DynamicRupture/Misc.h
vikaskurapati: I am sorry, I am confused, what is being tested in `tests/DynamicRupture/Output/Variables.t.h` w.r.t this method?

src/Equations/elastic/Model/ElasticSetup.h
vikaskurapati: > Any good idea on how to name matrix variables? I did use `matR` in other cases; but it all doesn't seem like a "proper" solution to me tbh.

I think `matR` is better than `mR` definitely. I was thinking something on the lines of `matEV` but now that I think more about it, it clashes with eigenvalues too in English. As there is already a comment before the variable initialization, it may help to come back and identify these are eigenvectors. May be having such a small comment before unclear variable initializations solves the problem. What do you think?

src/Geometry/CubeGenerator.cpp
vikaskurapati: I'd remove just the comment

src/DynamicRupture/FrictionLaws/CpuImpl/RateAndState.h
```diff
@@ -171,13 +171,8 @@ class RateAndStateBase : public BaseFrictionLaw<RateAndStateBase<Derived, TPMeth
...
       }
-      this->tpMethod.calcFluidPressure(normalStress,
-                                       this->mu,
-                                       localSlipRate,
-                                       this->deltaT[timeIndex],
-                                       false,
-                                       timeIndex,
-                                       ltsFace);
+      this->tpMethod.calcFluidPressure(
+          normalStress, this->mu, localSlipRate, this->deltaT[timeIndex], false, ltsFace);
```
vikaskurapati, line 175: Is the `tpMethod` in line 64 same as the `tpMethod` here? i.e., they both have the same instance of the object? If yes, can we remove the `this->` here safely?

src/Initializer/InitProcedure/InitLayout.cpp
vikaskurapati: It would probably never happen to us, but in case an LTS cluster has enough elements to overflow this, would we then shift it to `uint64_t`?

src/Parallel/MPI.cpp
vikaskurapati: > Yeah, it does refer to #829 . Though I think this one is better as an environment variable, since it's more system-dependent and not simulation-dependent (i.e. it doesn't make a difference w.r.t. the simulation results).

I agree with the environment variable part. Thanks!

src/SourceTerm/Manager.cpp
vikaskurapati: But does it hurt to have braces here? i.e., any unknown behavior, any linter complaints etc.?

vikaskurapati:
> All reviews should be complete; except the renaming of the matrices—I may need some input on how to either generically rename them (e.g. make `A` -> `matA` or the likes), or get specific names for each of them.
> 
> I added one additional `clang-tidy` to rule (cf. [1f077aa](https://github.com/SeisSol/SeisSol/commit/1f077aacc8ef20e728d5425d6ec5bdab6acb8c95) ; that one actually helped to find some typo bugs).

I did another sweep, and it looks good to me. For matrices, `matA` format sounds better than `mA`, and if a matrix's role in that method is not clear, a small comment near it's initialization may help, but is not necessary.

app/Main/Main.cpp
davschneller: Agreed in general; though it can make things more readable—but then you can also import each one of them. In essence, a `using namespace XYZ` is equivalent to saying `from XYZ import *` in Python which can also be a bit bad.

I just checked, if there's ambiguity, there will be an error. E.g.
```
namespace A {
    void test() {}
}

namespace B {
    void test() {}
}

int main() {
    using namespace A;
    using namespace B;
    test();
    return 0;
}
```
should fail to compile.

app/Main/Main.cpp
davschneller: Ok; I'm gonna revert it for now—applying the trailing return types is done in a matter of minutes. Maybe something to discuss in a larger audience; also w.r.t. which programming style guide to eventually adopt.

I think https://blog.petrzemek.net/2017/01/17/pros-and-cons-of-alternative-function-syntax-in-cpp/ is the best overview I've seen yet.

To add one more benefit:

* if you list multiple functions with trailing types, their names will start at the same position, e.g.
```
auto fn1() -> type1;
auto fn2() -> verylongtype2;
```
and so on.
* it's also a bit easier with formatting when you change the return type.

In total, I'm tending more towards the trailing return types—we'll get used to them eventually, and Python or Rust also use the same syntax in that case.

docs/tpv16.rst
davschneller: I'll try.

src/DynamicRupture/Misc.h
davschneller: The `forEach` is the only way `forEachElement` is supposed to be called—and that's tested in there.

src/Geometry/CubeGenerator.cpp
davschneller: Should be done

src/DynamicRupture/FrictionLaws/CpuImpl/RateAndState.h
davschneller: Yes, I think we can. I'll do it.

src/SourceTerm/Manager.cpp
davschneller: There is a rule for that as well, namely redundant initialization (`readability-redundant-member-init`).

src/Initializer/InitProcedure/InitLayout.cpp
davschneller: In those cases, we would need to switch the respective affected `uint32_t`s to `uint64_t`s, yes. But given that we'd need about over 4 bln elements on a single rank in GTS, I'm not sure if we'll reach it at least under the current systems.
(but who knows the future then)

app/Main/Main.cpp
vikaskurapati: > Ok; I'm gonna revert it for now—applying the trailing return types is done in a matter of minutes. Maybe something to discuss in a larger audience; also w.r.t. which programming style guide to eventually adopt.
> 
> I think https://blog.petrzemek.net/2017/01/17/pros-and-cons-of-alternative-function-syntax-in-cpp/ is the best overview I've seen yet.

The end of this blog also mentions why it is not widely used yet. But as you mentioned, this is something for a larger audience to discuss. :)

> 
> To add one more benefit:
> 
> * if you list multiple functions with trailing types, their names will start at the same position, e.g.
> 
> ```
> auto fn1() -> type1;
> auto fn2() -> verylongtype2;
> ```
> 
> and so on.
> 
> * it's also a bit easier with formatting when you change the return type.
> 
> In total, I'm tending more towards the trailing return types—we'll get used to them eventually, and Python or Rust also use the same syntax in that case.

In Python, it makes sense as the types are not declared otherwise. But for simpler types, I prefer the normal return types because they just make the code look cleaner and easier to write. In total, I tend to keep it simple, rather than enforcing this.

app/Main/Main.cpp
vikaskurapati: > Agreed in general; though it can make things more readable—but then you can also import each one of them. In essence, a `using namespace XYZ` is equivalent to saying `from XYZ import *` in Python which can also be a bit bad.
> 
> I just checked, if there's ambiguity, there will be an error. E.g.
> 
> ```
> namespace A {
>     void test() {}
> }
> 
> namespace B {
>     void test() {}
> }
> 
> int main() {
>     using namespace A;
>     using namespace B;
>     test();
>     return 0;
> }
> ```
> 
> should fail to compile.

What I meant was for something like this: https://godbolt.org/z/MGj93M146. But again, I may be overthinking it, as it is really unlikely we have common method names as the `std`.

src/Initializer/Parameters/LtsParameters.h
vikaskurapati: > Yes, it got moved there, because previously, the wiggle factor was written into the SeisSol parameters and then read from it again. However, ideally, the parameters should be a constant input (which they are now with this PR).
> 
> I'll see if I can get a test for it. (or maybe we should also add one to the CI?)

Yes, please. If there is a test that tests functionality, it is nice to have it in the CI, such that further changes can be tested for this too.

src/Memory/MemoryAllocator.cpp
vikaskurapati: Can we safely remove the comment if its not relevant anymore?

src/DynamicRupture/FrictionLaws/CpuImpl/RateAndState.h
vikaskurapati: > Yes, I think we can. I'll do it.

Sorry for bringing this up again, but can we also remove `this->` in `mu` and `deltaT[timeIndex]` as well? Or is there something else with the exact variables?

src/Memory/MemoryAllocator.cpp
davschneller: Yes, it's gone now. :)

src/DynamicRupture/FrictionLaws/CpuImpl/RateAndState.h
davschneller: Just tried it—and it's unfortunately not possible there (gives a compile error). I guess due to them being part of a superclass and this class having a template or something like that.

(as a general side comment: we currently basically maintain _two_ DR implementations with equal functionality which, at some point, should/could be simplified to a single one. One for CPU, one for GPU. Current show stopper is the SIMD code generation with OpenMP—that will need an "OpenMP SIMD barrier" which doesn't seem to be an option in OpenMP right now. The alternative, which will probably be the way to go I'm afraid, is to handle everything with the CPU/GPU iterators as in the `FrictionSolverCommon` ... The GPU variant keeps all data pointers in an extra struct instead of as class members, thus allowing for a more consistent syntax probably)

app/Main/Main.cpp
davschneller: That could happen sometime still. :) Also, `std` grows over the C++ versions or with not-yet-but-suddenly-used headers—who knows.

I get what you mean, e.g. that one variant "shadows" the other and/or `std` suddenly takes precedence in cases where it should not.

A tiny bit nitpickity w.r.t the code: I'm afraid unfortunately `test::sqrt` isn't called in that case (exchange it by `return -std::sqrt(num);` to see that the code still returns positive values). Since `std::sqrt` is templated, it'll give you a function specialization which has an `int` input arg which takes precedent over the `test::sqrt` accepting only `float`. You can add a `sqrt(4.f)` to enforce a `float` input arg; but then you'll get an ambiguity error again. Or alternatively replace `float test::sqrt(float num)` by `template<typename T> T test::sqrt(T num)`; then you'll get an ambiguity error also for `sqrt(4)` and `sqrt(4.0)`.

The internet also didn't provide a straight-forward example that quickly; so the errors might be even more obscure maybe in the end. (which is even worse IMO)

src/Initializer/Parameters/LtsParameters.h
davschneller: Ok; I got one for the CI—but no explicit "unit test" for the wiggle factor part yet (i.e. have a _tiny_ sample mesh and compute the wiggle factor, and compare it against a pre-computed/analytical solution from the interface). I.e. we can see if we get a similar timestep only via the fault receiver output, I'm afraid.

Results look good for this PR; and I'll throw the update of precomputed-seissol (once the PR is reviewed and accepted) into #1455 .

PR #1453, opened 2025-11-13
Move setDevice to serial region

uphoffc:
setDevice is currently called in a parallel region and this tripped up the debugger. I dunno whether this is intentional, but if it is the setDevice function needs to be thread-safe.

### DIFF
--- a/src/Initializer/InitProcedure/Internal/Buckets.cpp
+++ b/src/Initializer/InitProcedure/Internal/Buckets.cpp
@@ -239,8 +239,8 @@ void setupBuckets(LTS::Layer& layer, std::vector<solver::RemoteCluster>& comm) {
   const auto derivativeSize = yateto::computeFamilySize<tensor::dQ>();
 
 #ifdef ACL_DEVICE
-    device::DeviceInstance& device = device::DeviceInstance::getInstance();
-    device.api->setDevice(0);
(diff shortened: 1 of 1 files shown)


davschneller (commented):
Hi, it _is_ intentional for HIP/CUDA (they handle the device assignment per thread—cf. e.g. https://developer.nvidia.com/blog/cuda-pro-tip-always-set-current-device-avoid-multithreading-bugs/ ). Not sure about e.g. AdaptiveCpp-PCUDA in that regard if we were to support it.

... that reminds me, if copy initialization is enough, we could probably also completely avoid the GPU memory touching here (in the `initBucketItem` function); since that'll happen a bit later in the initialization.

(P.S.: to prevent any problems with extra formatting commits—there's a `pre-commit` config now to do that automatically)

Thomas-Ulrich:
Do I understand correctly that this PR should be closed?

davschneller:
It's best to confirm first that https://github.com/SeisSol/Device/pull/53 works and works well. Then yes, we can.

davschneller:
Or maybe see #1457 as an even more useful alternative.

( https://github.com/SeisSol/Device/pull/53 can be then ignored for now)

davschneller:
This PR has been resolved via #1457 which removes the part in question altogether. Thanks again for your contribution; I'm closing here now.

Issue #1442, opened 2025-10-21
runninng seissol on NG phase 2


Thomas-Ulrich:
I'm running a model with DR on NG phase2, with this batch file:
[seissol_batch_file.txt](https://github.com/user-attachments/files/23024392/seissol_batch_file.txt)

and I get the following error:
[seissol_log_ng2.txt](https://github.com/user-attachments/files/23024393/seissol_log_ng2.txt)

installed with
```
spack install -j 40 seissol@master convergence_order=4 dr_quad_rule=dunavant equations=elastic precision=single +intel_gpu intel_gpu_arch=pvc +sycl_gemm sycl_backend=oneapi  %oneapi@=2025.1.0  ^easi +python ^unzip %gcc ^intel-oneapi-compilers@2025.1.0 
```

(master is at this commit 4414cb043248a0af921fbaa75ec47fd3d171438d)

and this change to spack:

https://github.com/SeisSol/seissol-spack-aid/compare/main...NG_stack24.5

```
(venv-stack24.5-3.10) di73yeq4@login21:~/seissol-spack-aid> module list
Currently Loaded Modulefiles:
 1) stack/24.5.0{arch=auto}   2) intel/2025.1.0   3) user_spack/24.5.0   4) python/3.10.12-extended   5) intel-mpi/2021.15.0   6) seissol/master-oneapi25-o4-elas-dunav-single-gpu-impi  

Key:
auto-loaded  default-version  {variant=value}  
```

I can provide the model if necessary.

uphoffc:
Thomas have a look at https://github.com/SeisSol/Device/pull/52

Thomas-Ulrich:
Thanks. This seems to work, but I get into another issue then:

```
2025-11-13 11:45:38.559 info 0 : Setting up DR... 
2025-11-13 11:45:38.564 info 0 : Storage "dr" ; variables: 32.9328 MiB 
2025-11-13 11:45:38.585 info 0 : Storage "dr" ; scratchpads: 5.8090 MiB 
2025-11-13 11:45:38.587 info 0 : Setting up data exchange and face displacements (buckets)... 
2025-11-13 11:45:38.594 info 0 : Storage "cluster" ; buckets: 83.8768 MiB 
[...]

PR #1454, opened 2025-11-13
Add ghost-layer asserts for device buckets

davschneller:
Cf. #1442 for the assert triggering while it should've already triggered in the CPU part then.


Thomas-Ulrich (approved):
fixed the issue

PR #1455, opened 2025-11-18
Add running the proxy to `ctest`; test module system

davschneller:
* add the proxy to the testing system. I.e. if the proxy doesn't run, the tests will fail
* also add the proxy to the coverage (though not "pure" unit tests; maybe we might want two coverages in the long run)
* test the `Modules` folder and fix some associated bugs


cmake/CodeCoverage.cmake
```diff
@@ -231,10 +231,11 @@ function(setup_target_for_coverage_lcov)
     set(multiValueArgs EXCLUDE EXECUTABLE EXECUTABLE_ARGS DEPENDENCIES LCOV_ARGS GENHTML_ARGS)
     cmake_parse_arguments(Coverage "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
-    get_property(crosscompiling_emulator
-            TARGET ${Coverage_EXECUTABLE}
-            PROPERTY CROSSCOMPILING_EMULATOR
-            )
+    # if you supply seissol-serial-test or similar:
```
vikaskurapati, line 234: I don't understand why we don't need it anymore.

davschneller:
Ok; confirmed by some samples.

Side note, we might "officially" add the precomputed-seissol end-to-end tests to the coverage (though we might even wanna report two coverages in the end; one "total" coverage and an extra unit test coverage); though I'm not 100%ly sure yet how to do it smartly (e.g. with fetch contents)—the repository is a bit big and the testing takes longer than instantly.
(maybe we'll need to add kernel-level comparisons at some point)

PR #1459, opened 2025-11-24
Disable NVHPC tests

davschneller:
Disable NVHPC tests; since the container seems to fail now. Still a TODO: rebuild the container.

(more CI updates are prepared on the `davschneller/ci-merge` branch; however I fear that'll still take a bit)

### DIFF
--- a/.github/workflows/build-seissol-cpu.yml
+++ b/.github/workflows/build-seissol-cpu.yml
@@ -63,10 +63,10 @@ jobs:
             cxx: icpx
             fc: ifx
             container: seissol/gha-gpu-intel:davschneller-gpu-image
-          - cc: nvc
-            cxx: nvc++
-            fc: nvfortran
-            container: seissol/gha-gpu-nvhpc:davschneller-gpu-image
+          #- cc: nvc
+          #  cxx: nvc++
+          #  fc: nvfortran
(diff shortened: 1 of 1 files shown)


Thomas-Ulrich (approved):

vikaskurapati:
Do we really want to disable tests because they fail? I would rather be inclined to wait until the fixes are there instead of disabling them .

davschneller:
Normally I'd agree; but in this case, we'll need to update the Github Actions CI scripts anyways in all cases that I'm aware of—unless they were to suddenly fix the regression again (which I doubt).
Again, it fails _before_ we do anything SeisSol-related already. I.e. when we setup the container itself. With SeisSol and NVHPC, nothing is wrong to my knowledge (yet).
As an alternative, we could already switch the container to `self-hosted` now already. It'll probably be a bit slower as of speaking, but it might work. It can be pretty unstable, so I wouldn't want to definitely rely on it for now.

davschneller:
However, I'd do two things then:

* disable the tests for NVHPC (those will take longer; and we'll need to adjust specifically for the self-hosted runner)
* maybe we should also disable most build tests during development in general (at least until #1421 is merged and it's sped up properly w.r.t. compile time). Just to reduce pressure on the CI pipelines. We can then re-enable all of them when we plan to merge. And maybe run the full tests already before merging automatically.

PR #1461, opened 2025-11-27
GPU L2 Compression Support

davschneller:
Currently guarded behind `SEISSOL_L2_COMPRESS` which is disabled by default; and only has an effect for NVIDIA cards right now. (maybe at sometime also Intel according to Level Zero)
Could/will help with some memory-bound kernels and situations (it did have an effect on my RTX 4070 mobile in the proxy).

### DIFF
--- a/src/Memory/Descriptor/Boundary.h
+++ b/src/Memory/Descriptor/Boundary.h
@@ -22,9 +22,9 @@ inline auto allocationModeBoundary() {
   if constexpr (!isDeviceOn()) {
     return AllocationMode::HostOnly;
   } else {
-    return useUSM() ? AllocationMode::HostDeviceUnified
-                    : (useDeviceL2Compress() ? AllocationMode::HostDeviceCompress
-                                             : AllocationMode::HostDeviceSplit);
+    const auto modeMaybeCompress = useDeviceL2Compress() ? AllocationMode::HostDeviceCompress
+                                                         : AllocationMode::HostDeviceSplit;
+    return useUSM() ? AllocationMode::HostDeviceUnified : modeMaybeCompress;
   }
 }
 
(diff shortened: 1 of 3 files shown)


uphoffc:
Memory compression controls are exposed here: https://oneapi-src.github.io/level-zero-spec/level-zero/latest/core/api.html#ze-memory-compression-hints-ext-desc-t

On the Arc B580 compression is enabled by default. (You need to force uncompressed if you want uncompressed.) Compression could be useful for material parameters, for example if they are constant across the domain.

davschneller:
Ok, thanks for the infos! Any other hint structs to add to the L0 allocator calls besides the "main" one? (referring to a simplified form of https://github.com/intel/llvm/blob/c757480800a9e0224b81e509bf88d2b77067da69/unified-runtime/source/adapters/level_zero/usm.cpp#L183-L204 )

uphoffc:
I'm not aware of any other then the relaxed allocation limit you linked to and the compression hints.

davschneller:
Ok thanks for the info. I'll add that into the Device module.

Other than that, this PR only takes care of the forwarding; so it can be merged (and it'll work for NV). Once the Device module is updated, it'll then also work for Intel.

src/Memory/Descriptor/Boundary.h
```diff
@@ -22,7 +22,9 @@ inline auto allocationModeBoundary() {
   if constexpr (!isDeviceOn()) {
     return AllocationMode::HostOnly;
   } else {
-    return useUSM() ? AllocationMode::HostDeviceUnified : AllocationMode::HostDeviceSplit;
+    return useUSM() ? AllocationMode::HostDeviceUnified
+                    : (useDeviceL2Compress() ? AllocationMode::HostDeviceCompress
```
vikaskurapati: Sorry to be nitpicky here, but how about `HostDevicCompressed`?

davschneller:
It's implemented for CUDA; I have a patch open for Level Zero, but I'll still need to test it. However, once we add the respective Device module patch and update the device module here, it's available automatically within SeisSol; no further update needed. (presumably)

PR #1462, opened 2025-11-27
Rate+State Loop in FP32

davschneller:
Switch the R+S loop to single precision when compiling for single precision, by eliminating too large intermediate values.

It turns out, the issue why the R+S loop diverged for FVW and AL/SL in FP32 was seemingly not actually a precision, but a range issue. In particular: we compute `asinh(x * exp(c))`. Now, `exp(c)` may be really large for moderately-big `c`, but `asinh(exp(c))` is small; growing only linearly in comparison (`asinh` effectively equals `ln(2x)` for large x). I.e. if we modify the calculation of `asinh(exp(c))` a tiny bit to only call `exp(-2c)` in these cases, we can avoid using double precision entirely.

### DIFF
--- a/src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
+++ b/src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
@@ -46,7 +46,7 @@ struct Settings {
   Computes asinh(x * exp(c)). Reason is: exp(c) can grow really large (too large for float);
   but actually asinh(exp(c)) \approx c for large c.
 
-  Hence, we compute instead
+  Hence, we compute instead (x > 0)
   asinh(x * exp(c))
   = asinh((x * exp(c)) + sqrt((x * exp(c))**2 + 1))
   = asinh(exp(c) * (x + sqrt(x**2 + exp(-2c))))
@@ -59,9 +59,28 @@ struct Settings {
 #pragma omp declare simd
 template <typename T>
 SEISSOL_HOSTDEVICE constexpr T arsinhexp(T x, T expLog, T exp) {
-  if (expLog > 0) {
-    return expLog + std::log(x + std::sqrt(x * x + exp));
+  // Switch is empirically chosen; to prevent issues with
+  // or replacement formula not being accurate enough if x * exp(c) is small
+  constexpr T Switch = 10;
+  constexpr T Threshold = 50;
+  constexpr T Log2 = 0.69314718055994530943;
+  int xexp{};
+  (void)std::frexp(x, &xexp);
+
+  // make sure to invert the constant we'd use otherwise (if the exponent is too big/small)
+
+  // use the new code path only if we really need to
+  if (expLog + std::max(xexp, 0) * Log2 > Switch || expLog >= Threshold) {
+    if (expLog <= 0) {
+      exp = 1 / exp;
+    }
+    const T xa = std::abs(x);
+    const T xs = x >= 0 ? 1 : -1;
+    return xs * (expLog + std::log(xa + std::sqrt(xa * xa + exp * exp)));
   } else {
+    if (expLog > 0) {
+      exp = 1 / exp;
+    }
     const auto v = exp * x;
     return std::asinh(v);
   }
@@ -76,7 +95,7 @@ template <typename T>
 SEISSOL_HOSTDEVICE constexpr T computeCExp(T cExpLog) {
   T cExp{};
   if (cExpLog > 0) {
-    cExp = std::exp(-2 * cExpLog);
+    cExp = std::exp(-cExpLog);
   } else {
     cExp = std::exp(cExpLog);
   }
@@ -89,9 +108,23 @@ SEISSOL_HOSTDEVICE constexpr T computeCExp(T cExpLog) {
 #pragma omp declare simd
 template <typename T>
 SEISSOL_HOSTDEVICE constexpr T arsinhexpDerivative(T x, T expLog, T exp) {
-  if (expLog > 0) {
-    return 1 / std::sqrt(x * x + exp);
+  constexpr T Switch = 10;
+  constexpr T Threshold = 50;
+  constexpr T Log2 = 0.69314718055994530943;
+  int xexp{};
+  (void)std::frexp(x, &xexp);
+
+  // make sure to invert the constant we'd use otherwise (if the exponent is too big/small)
+
+  if (expLog + std::max(xexp, 0) * Log2 > Switch || expLog >= Threshold) {
+    if (expLog <= 0) {
+      exp = 1 / exp;
+    }
+    return 1 / std::sqrt(x * x + exp * exp);
   } else {
+    if (expLog > 0) {
+      exp = 1 / exp;
+    }
     const auto v = exp * x;
     return exp / std::sqrt(1 + v * v);
   }
@@ -113,7 +146,7 @@ SEISSOL_HOSTDEVICE constexpr T arsinhexpDerivative(T x, T expLog, T exp) {
 #pragma omp declare simd
 template <typename T>
 SEISSOL_HOSTDEVICE constexpr T logsinh(T x, T c) {
-  const T sign = c > 0 ? 1 : -1;
+  const T sign = c >= 0 ? 1 : -1;
   const T absC = std::abs(c);
   return absC + std::log(x / 2 * -sign * std::expm1(-2 * absC));
 }


src/DynamicRupture/FrictionLaws/GpuImpl/FastVelocityWeakeningLaw.h
```diff
@@ -28,59 +29,64 @@ class FastVelocityWeakeningLaw
...
+    real cExp{};
+    real acLin{};
   };
 
   SEISSOL_DEVICE static MuDetails getMuDetails(FrictionLawContext& ctx, double localStateVariable) {
-    const double localA = ctx.data->a[ctx.ltsFace][ctx.pointIndex];
-    const double c = 0.5 / ctx.data->drParameters.rsSr0 * std::exp(localStateVariable / localA);
-    return MuDetails{localA, c, localA * c};
+    const real localA = ctx.data->a[ctx.ltsFace][ctx.pointIndex];
+    const real cLin = 0.5 / ctx.data->drParameters.rsSr0;
```
Thomas-Ulrich, line 72: what means Lin in cLin?

src/DynamicRupture/FrictionLaws/GpuImpl/SlowVelocityWeakeningLaw.h
```diff
@@ -32,33 +32,38 @@ class SlowVelocityWeakeningLaw
...
-    const auto localF0 = ctx.data->f0[ctx.ltsFace][ctx.pointIndex];
-    const auto localB = ctx.data->b[ctx.ltsFace][ctx.pointIndex];
-    const double c =
-        0.5 / ctx.data->drParameters.rsSr0 * std::exp((localF0 + localB * log1) / localA);
-    return MuDetails{localA, c, localA * c};
+    const double localF0 = ctx.data->f0[ctx.ltsFace][ctx.pointIndex];
+    const double localB = ctx.data->b[ctx.ltsFace][ctx.pointIndex];
+
+    const real cLin = 0.5 / ctx.data->drParameters.rsSr0;
+    const real cExpLog = (localF0 + localB * log1) / localA;
```
Thomas-Ulrich, line 50: That's not the subject of the PR, but: wouldn't it make more sense to store one_over_localA and one_over_rsSr0? (or maybe getMuDetails is only called once?)

src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
```diff
@@ -38,6 +40,79 @@ struct Settings {
...
+    return std::asinh(v);
+  }
+}
+
+/**
+  Helper function to arsinhexp. Since for asinh(x * exp(c)),
+  we can assume c to be constant, we can pre-compute exp(c) or exp(-2c).
+ */
+template <typename T>
+SEISSOL_DEVICE constexpr T computeCExp(T cExpLog) {
```
Thomas-Ulrich: wouldn't it be more readable to have computeExpC and computeExpMinusTwoC?

src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
```diff
@@ -38,6 +40,79 @@ struct Settings {
...
+    cExp = std::exp(cExpLog);
+  }
+  return cExp;
+}
+
+/**
+  Derivative to arsinhexp.
+ */
+template <typename T>
+SEISSOL_DEVICE constexpr T darsinhexp(T x, T expLog, T exp) {
```
Thomas-Ulrich: arsinhexp_derivative?

src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
```diff
@@ -38,6 +40,79 @@ struct Settings {
...
+    return exp / std::sqrt(1 + v * v);
+  }
+}
+
+/**
+  Compute log(x * sinh(c)).
+  if c > 0, then
+  log(x * (e(c) - e(-c)) / 2)
+  = log(e(c) * x / 2 * (1 - e(-2c)))
+  = c * log(x / 2 * -expm1(-2c))
```
Thomas-Ulrich: c + log(x / 2 * -expm1(-2c))

src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
```diff
@@ -38,6 +40,79 @@ struct Settings {
...
+
+  if c < 0, then
+  log(x * sinh(c)) = -c * log(x / 2 * expm1(2c))
```
Thomas-Ulrich: +

src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
```diff
@@ -38,6 +40,79 @@ struct Settings {
...
+
+  In total,
+  log(x * sinh(c)) = -c * log(x / 2 * -sign(c) * expm1(-2|c|))
```
Thomas-Ulrich: |c| +

src/DynamicRupture/FrictionLaws/GpuImpl/SlowVelocityWeakeningLaw.h
```diff
@@ -32,33 +32,38 @@ class SlowVelocityWeakeningLaw
...
+    const real localA = ctx.data->a[ctx.ltsFace][ctx.pointIndex];
     const double localSl0 = ctx.data->sl0[ctx.ltsFace][ctx.pointIndex];
     const double log1 = std::log(ctx.data->drParameters.rsSr0 * localStateVariable / localSl0);
-    const auto localF0 = ctx.data->f0[ctx.ltsFace][ctx.pointIndex];
-    const auto localB = ctx.data->b[ctx.ltsFace][ctx.pointIndex];
-    const double c =
-        0.5 / ctx.data->drParameters.rsSr0 * std::exp((localF0 + localB * log1) / localA);
-    return MuDetails{localA, c, localA * c};
+    const double localF0 = ctx.data->f0[ctx.ltsFace][ctx.pointIndex];
+    const double localB = ctx.data->b[ctx.ltsFace][ctx.pointIndex];
```
Thomas-Ulrich: doubles on purpose?

Thomas-Ulrich (changes requested):
Great idea!

src/DynamicRupture/FrictionLaws/GpuImpl/SlowVelocityWeakeningLaw.h
davschneller: Yes, that's only run once per Newton iteration. Since the state variable stays constant throughout a Newton iteration, we can pre-compute everything that only depends on the state variable and not on the slip rate to solve for.

src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
davschneller: Ok, done

src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
davschneller: Ah thanks—it's adjusted

src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
davschneller: Done

src/DynamicRupture/FrictionLaws/RateAndStateCommon.h
davschneller: Done

src/DynamicRupture/FrictionLaws/GpuImpl/SlowVelocityWeakeningLaw.h
davschneller: No, missed that one. :) Changed now.

src/DynamicRupture/FrictionLaws/GpuImpl/FastVelocityWeakeningLaw.h
davschneller: The `Lin` should mean linear in that context. The idea was to decompose `c` into `c = cLin * exp(cExp)` (and avoid the `exp` call here since it might go out of range)

davschneller:
Thanks for the review. Just to note—there's currently still a possible numerical inaccuracy with computing `asinh(x * exp(c))` if `x` is close to zero (e.g. `x=1e-7` for FP32); i.e. that case will require better treatment, maybe by an adjusted `asinh` implementation.

davschneller:
Ok, updated it to use `asinh(x * exp(c))` still directly for `log(x) + c` (approximately) small, and `c` not too big.

Though there can probably more be done; like reformulating it all to `c + log(x) + log1p(sqrt(1 + exp(-2c) / x**2))` if `x` is also small; or the extreme cases where we can write `c + log(x) + log(2)` (both aren't implemented in this PR).

Issue #1052, opened 2024-02-14
Very long initialization during texascale run


NicoSchlw:
**Describe the bug**
Initialization during texascale run takes 40 min

**Expected behavior**
Faster Initialization

**To Reproduce**
1. Which version do you use? Provide branch and commit id. 
 nico/2nuc-latest-master; v1.1.3-109-g10ae92ee
2. Which build settings do you use? Which compiler version do you use? 
SeisSol_Release_dskx_6_viscoelastic2; intel-19.1.1
3. On which machine does your problem occur? If on a cluster: Which modules are loaded?
Frontera
Currently Loaded Modules:
  1) intel/19.1.1
  2) impi/19.0.9
  3) git/2.24.1
  4) autotools/1.2
  5) cmake/3.24.2
  6) pmix/3.1.4
  7) hwloc/1.11.12
  8) xalt/2.10.34
  9) TACC
 10) python3/3.9.2
 11) intel-mpi/2019.0.9-intel-19.1.1.217-plmutsc
 12) pkgconf/1.9.5-intel-19.1.1.217-orhsldj
 13) zlib-ng/2.1.4-intel-19.1.1.217-xwszawp
 14) hdf5/1.12.2-intel-19.1.1.217-sji7bdx
 15) netcdf-c/4.7.4-intel-19.1.1.217-guy2c4r
 16) numactl/2.0.14-intel-19.1.1.217-fbfn6qe
 17) asagi/1.0.1-intel-19.1.1.217-6oes3fl
 18) cxxtest/develop-intel-19.1.1.217-yavleqi
 19) impalajit/main-intel-19.1.1.217-w45mc4n
 20) curl/7.29.0-intel-19.1.1.217-kiwu5vb
 21) ncurses/6.4-intel-19.1.1.217-jkmcgrd
 22) readline/8.2-intel-19.1.1.217-rdugitv
 23) unzip/6.0-intel-19.1.1.217-s2gzz22
 24) lua/5.3.2-intel-19.1.1.217-sqpjecx
 25) yaml-cpp/0.6.2-intel-19.1.1.217-p36vrxa
 26) easi/1.2.0-intel-19.1.1.217-lwzthnz
 27) eigen/3.4.0-intel-19.1.1.217-wni5u35
 28) libxsmm/1.17-intel-19.1.1.217-knnzdf4
 29) memkind/1.13.0-intel-19.1.1.217-pvz3fkl
[...]

PR #1468, opened 2025-12-10
Quadpoint-wise initial plasticity loading

### LINKED ISSUE #760: Initialize plasticity properties and associated initial stress quadrature points wise
Plasticity properties and associated initial stress are initialized element-wise.
It would make sense to initialize them (at least the stress) quadrature point-wise given that plasticity is evaluated at the quadrature points).
(I'm just pinning that here to remember. I realized they are initialized element-wise when reading the homogenization PR (#644).

davschneller:
Add an option to sample the quadrature points per point, not barycentrically. (the latter method stays as well)

Addresses and resolves #760 .

### DIFF
--- a/docs/parameters.par
+++ b/docs/parameters.par
@@ -12,8 +12,10 @@
 MaterialFileName = '33_layered_constant.yaml'
 !1: Compute average materials for each cell, 0: sample material values at element barycenters
 UseCellHomogenizedMaterial = 1
-!off-fault plasticity parameters (ignored if Plasticity=0)
+!off-fault plasticity parameters (disabled iff Plasticity=0)
 Plasticity=0
+!comma-separated list of groups for which plasticity is disabled, even if we enable it globally
+plasticitydisabledgroups=''
 Tv=0.05
 !yaml file defining spatial dependance of properties necessary for plasticity. MaterialFileName is taken for plasticity if this is not provided
 PlasticityFileName = ''
@@ -58,6 +60,7 @@ avoidic = 0    ! 1 iff use the IC only for analytical boundary conditions
 /
 
 &DynamicRupture
+enabled = 1                  ! explicitly enable/disable the DR (different to setting FL=0; disabling it converts all DR faces to default inner faces, and disables any DR-related I/O)
 FL = 16                      ! Friction law
 !0: none, 16:LSW, 103: RS with strong velocity weakening
 !yaml file defining spatial dependance of fault properties
--- a/src/Geometry/MeshReader.cpp
+++ b/src/Geometry/MeshReader.cpp
@@ -86,6 +86,16 @@ void MeshReader::scaleMesh(const Eigen::Matrix3d& scalingMatrix) {
   }
 }
 
+void MeshReader::disableDR() {
+  for (auto& elem : m_elements) {
+    for (std::size_t j = 0; j < Cell::NumFaces; ++j) {
+      if (elem.boundaries[j] == 3) {
+        elem.boundaries[j] = 0;
+      }
+    }
+  }
+}
+
 /**
  * Reconstruct the fault information from the boundary conditions
  */
--- a/src/Geometry/MeshReader.h
+++ b/src/Geometry/MeshReader.h
@@ -117,6 +117,11 @@ class MeshReader {
   void exchangeGhostlayerMetadata();
 
   /**
+   * Disable the DR by converting all DR faces (BC = 3) to regular faces (BC = 0).
+   */
+  void disableDR();
+
+  /**
     Create a linearized ghost layer view.
     Currently, the ghost layer arrays copy each cell per rank-boundary face.
     Meaning that a cell may appear multiple times remotely.
--- a/src/Initializer/BatchRecorders/PlasticityRecorder.cpp
+++ b/src/Initializer/BatchRecorders/PlasticityRecorder.cpp
@@ -25,7 +25,6 @@ using namespace seissol::recording;
 void PlasticityRecorder::record(LTS::Layer& layer) {
   setUpContext(layer);
 
-  auto* pstrains = currentLayer->var<LTS::PStrain>(AllocationPlace::Device);
   size_t nodalStressTensorCounter = 0;
   real* scratchMem =
       static_cast<real*>(currentLayer->var<LTS::IntegratedDofsScratch>(AllocationPlace::Device));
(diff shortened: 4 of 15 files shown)


Thomas-Ulrich:
Thank you!
I think this could be important for studies looking at near-surface plasticity strain output, or for studies with a relatively coarse fault mesh (e.g. Sumatra or Myanmar). On the other hand, I agree that if because of ASAGI this increases a lot the initialization time of seissol then that would be problematic. 
Do you think it would be possible to have a switch similar as
`UseCellHomogenizedMaterial = 1` to disable the feature?

davschneller:
Happy to help here. W.r.t. normalization: yes, that should be absolutely doable, I've implemented it onto the current setup (it will really evaluate only the barycenter points). The quadpoint-wise (non-barycentric) initialization is currently the default.

W.r.t. ASAGI also note this comment here: https://github.com/SeisSol/SeisSol/issues/1052#issuecomment-3067284287 . I haven't tested it yet, but setting those two options might let ASAGI be much faster.
If not, I can look into the issue once more. (there seem to be _really_ many knobs for ASAGI)

src/Model/Plasticity.h
```diff
@@ -17,40 +20,55 @@
...
-      initialLoading[static_cast<std::size_t>(4 * seissol::multisim::NumSimulations) + i] =
-          plasticity[i].sYZ;
-      initialLoading[static_cast<std::size_t>(5 * seissol::multisim::NumSimulations) + i] =
-          plasticity[i].sXZ;
-
-      const double angularFriction = std::atan(plasticity[i].bulkFriction);
-
-      cohesionTimesCosAngularFriction[i] = plasticity[i].plastCo * std::cos(angularFriction);
-      sinAngularFriction[i] = std::sin(angularFriction);
+  PlasticityData(const std::array<const Plasticity*, seissol::multisim::NumSimulations>& plasticity,
```
Thomas-Ulrich, line 36: const std::array<const Plasticity* is passed as input but not used?
Same name for the 2 arguments except the lowercase?

src/Model/Plasticity.h
```diff
@@ -17,40 +20,55 @@
...
+                 const Material* material,
+                 bool pointwise) {
+    auto initialLoadingV = init::initialLoading::view::create(initialLoading);
```
Thomas-Ulrich, line 39: initialLoading is declared in one of the header I guess?

src/Initializer/ParameterDB.cpp
```diff
@@ -357,14 +410,18 @@ void MaterialParameterDB<T>::evaluateModel(const std::string& fileName,
...
-  const unsigned numPoints = query.numPoints();
+  const auto numPoints = query.numPoints();
 
   std::vector<T> materialsFromQuery(numPoints);
   easi::ArrayOfStructsAdapter<T> adapter(materialsFromQuery.data());
   MaterialParameterDB<T>().addBindingPoints(adapter);
 
   easiEvalSafe(model, query, adapter, "volume material");
 
+  if (m_materials->size() < numPoints) {
```
Thomas-Ulrich, line 421: in which case is it necessary?

codegen/matrices/plasticity-ip-matrices-2.json
```diff
@@ -0,0 +1,10 @@
...
...16, 1, "0.3470037660383519" ], [ 16, 2, "0.05784760393614262" ], [ 16, 3, "0.5280743882734471" ], [ 17, 1, "0.3470037660383519" ], [ 17, 2, "0.05784760393614262" ], [ 17, 3, "0.2975743150127527" ], [ 18, 1, "0.3470037660383519" ], [ 18, 2, "0.05784760393614262" ], [ 18, 3, "0.06707424175205831" ], [ 19, 1, "0.0729940240731497" ], [ 19, 2, "0.7301650280476316" ], [ 19, 3, "0.17465664523839902" ], [ 20, 1, "0.0729940240731497" ], [ 20, 2, "0.7301650280476316" ], [ 20, 3, "0.09842047393960934" ], [ 21, 1, "0.0729940240731497" ], [ 21, 2, "0.7301650280476316" ], [ 21, 3, "0.02218430264081966" ], [ 22, 1, "0.0729940240731497" ], [ 22, 2, "0.3795782302805906" ], [ 22, 3, "0.4857317270371134" ], [ 23, 1, "0.0729940240731497" ], [ 23, 2, "0.3795782302805906" ], [ 23, 3, "0.27371387282312987" ], [ 24, 1, "0.0729940240731497" ], [ 24, 2, "0.3795782302805906" ], [ 24, 3, "0.06169601860914631" ], [ 25, 1, "0.0729940240731497" ], [ 25, 2, "0.08212156786344241" ], [ 25, 3, "0.7496645282216933" ], [ 26, 1, "0.0729940240731497" ], [ 26, 2, "0.08212156786344241" ], [ 26, 3, "0.42244220403170396" ], [ 27, 1, "0.0729940240731497" ], [ 27, 2, "0.08212156786344241" ], [ 27, 3, "0.09521987984171464" ] ]
```
Thomas-Ulrich, line 2: I wonder if it would be useful to add a description saying what is each array?
(I mean vNodes, v, vD and Vinv) or change their names to something more clear.

src/Model/Plasticity.h
davschneller: Ah, you mean the array right? The whole `const std::array<const Plasticity*, seissol::multisim::NumSimulations>&` is a single type; effectively giving us a const reference to an array of const pointers to `Plasticity` data structures.

That's useful if we have a whole section of plasticity datastructures we want to work with (as we do here, if we use quadpoint-wise initialization); and alas we're not at C++20 yet; then we could use `std::span` instead.

src/Model/Plasticity.h
davschneller: Yes; it's declared in the Yateto tensor collection (that's best for abstracting away anything like multisim or different tensor extensions, and to avoid differences in tensor layout vs. contiguous).

src/Initializer/ParameterDB.cpp
davschneller: We don't really check the input size of the `m_materials` array; i.e. you could even pass a vector of size 0 (I know, the interface doesn't seem too ideal yet). So just to prevent any strange override bugs to happen, I've added this condition here which enforces that we always have enough space available.

codegen/matrices/plasticity-ip-matrices-2.json
davschneller: In principle, yes, possible; though in practice probably it's not that good to have it in this file. The plasticity files are generated by our matrix generator (SeisSol/Matrices, formerly the Matlab scripts).
We can of course rename them. `v` is supposed to stand for "Vandermonde matrix" (i.e. just the basis evaluated at the input points); `vD` for the derivative of it (hence it's a tensor), and `Vinv` the (pseudo-)inverse. `vNodes` are the points we build the matrix out of. All of those are used in the plasticity and the energy (there, we always use the more exact basis).

I'm wondering if it would be best to do the renaming and documenting in one big sweep for all matrices and files.

davschneller:
Thanks for the review. For the Github changes: I'm afraid it's not really a bug; the reason are the matrix files. For the XML matrix files (the "older" ones), we have basically one line per matrix entry. While for the JSON files, we have one line per whole tensor. Thus, we remove a lot of lines and add much less.

codegen/matrices/plasticity-ip-matrices-2.json
Thomas-Ulrich: yes, we can do that in another PR.

PR #1469, opened 2025-12-11
Nodal Plasticity Storage (and merge more kernels)

### LINKED ISSUE #725: Negative plastic moment
**Describe the bug**
The plastic moment output is negative for a few time steps in the energy.csv file

**Expected behavior**
The plastic moment should increase monotonically 

**To Reproduce**
1. Which version do you use? Provide branch and commit id.
fix-707-2-nuc branch
2. Which build settings do you use? Which compiler version do you use? 
double precision, viscoelastic2, skx, order 5
Frontera, modules are given in the job script
Ridgecrest paper setup: https://zenodo.org/record/6842773#.Y2kizL7MLUQ

If you suspect a problem in the numerics/physics add a screenshot of your output.

If you encounter any errors/warnings/... during execution please provide the console output.

**Additional context**
[RC-paper-energy.csv](https://github.com/SeisSol/SeisSol/files/9952990/RC-paper-energy.csv)
[4973897.out.txt](https://github.com/SeisSol/SeisSol/files/9952992/4973897.out.txt)

davschneller:
Convert all plasticity storage to nodal.
As a result, we can combine several more kernels and remove some matrix multiplications.
For the output, we still convert back to modal and determine the output from that right now—but maybe we can also e.g. interpolate differently instead.
(though it might already make some difference that we don't lose the extra info between steps)

Once done, it should resolve #725 .

### DIFF
--- a/src/Kernels/DeviceAux/sycl/PlasticityAux.cpp
+++ b/src/Kernels/DeviceAux/sycl/PlasticityAux.cpp
@@ -33,20 +33,20 @@ auto getrange(std::size_t size, std::size_t numElements) {
   }
 }
 
-void adjustDeviatoricTensors(real** __restrict nodalStressTensors,
-                             real** __restrict pstrainPtr,
-                             unsigned* __restrict isAdjustableVector,
-                             std::size_t* __restrict yieldCounter,
-                             const seissol::model::PlasticityData* __restrict plasticity,
-                             double oneMinusIntegratingFactor,
-                             double tV,
-                             double timeStepWidth,
-                             const size_t numElements,
-                             void* queuePtr) {
+void plasticityNonlinear(real** __restrict nodalStressTensors,
+                         real** __restrict pstrainPtr,
+                         unsigned* __restrict isAdjustableVector,
+                         std::size_t* __restrict yieldCounter,
+                         const seissol::model::PlasticityData* __restrict plasticity,
+                         double oneMinusIntegratingFactor,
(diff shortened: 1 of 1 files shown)


src/ResultWriter/EnergyOutput.cpp
```diff
@@ -598,7 +602,25 @@ void EnergyOutput::computeVolumeEnergies() {
...
+          // integrating over all collocation points suffices
+          const real* __restrict qEta = &pstrainCell[tensor::QStressNodal::size()];
+
+          alignas(Alignment) real qEtaQuad[tensor::QEtaNodalProject::size()]{};
+
+          kernel::plProject krnl;
+          set_evalAtQP(krnl, global->evalAtQPMatrix);
+          set_vInv(krnl, global->vandermondeMatrixInverse);
+          krnl.QEtaNodal = qEta;
+          krnl.QEtaNodalProject = qEtaQuad;
```
Thomas-Ulrich, line 615: I don't understand why you need to project qEta to another nodal basis?

davschneller: That only happens if we use the `nb` basis for plasticity—as the energy calculation uses `ip` ; otherwise it becomes a no-op. ... Or is there a better way when using the `nb` basis to directly obtain the integrated data?

Thomas-Ulrich: I see! Thx.

PR #1471, opened 2025-12-12
Begin auto-determining architectures if not specified explicitly

davschneller:
In case it helps—determine the host arch and device arch mostly automatically. _If_ you run the configuration once on a compute node (the auto-search will then be disabled for subsequent calls).

For GPUs, you'll still need to set `DEVICE_BACKEND` right now.

Still a little experimental (e.g. it still would need to store the `-march=native` flags somehow); otherwise, it should work well already now. I've tested it with NVIDIA and AMD cards, both work.


gabriellemhobson:
Hi, I've been testing this branch on Vista, compiling by submitting jobs to the gh-dev queue. I have tested it for both CPU and GPU; when I did not specify `-DDEVICE_BACKEND=cuda`, it compiled for CPU, and when I did specify `-DDEVICE_BACKEND=cuda`, it compiled for CPU. I specified `-DDEVICE_ARCH=auto` and `-DHOST_ARCH=auto`. When I use it as-is, the code correctly identifies the architecture, based on this output:

```
-- Determining host arch supported on the current system
-- Performing Test HAS_SVE128
-- Performing Test HAS_SVE128 - Success
-- Performing Test HAS_SVE256
-- Performing Test HAS_SVE256 - Failed
-- Performing Test HAS_SVE512
-- Performing Test HAS_SVE512 - Failed
```

And sve128 is what I typically use on Vista. However, the installation later fails after `make -j` (after ~11 min) with this error message:

```
File "/work/10752/ghobson/vista/seissol_gpu_wz/SeisSol-GPU-branch/codegen/generate.py", line 330, in <module>
    main()
  File "/work/10752/ghobson/vista/seissol_gpu_wz/SeisSol-GPU-branch/codegen/generate.py", line 95, in main
    arch = deriveArchitecture(host_arch, device_arch)
  File "/work/10752/ghobson/vista/seissol_gpu_wz/SeisSol-GPU-branch/codegen/yateto/arch.py", line 235, in deriveArchitecture
    raise NotImplementedError(f'The architecture {host_def.archname} is unknown to Yateto, and no custom alignment was given')
NotImplementedError: The architecture sve-128 is unknown to Yateto, and no custom alignment was given
```

Looking at the changes, I wondered whether line 117-118 of `cmake/auto_arch.cmake` is causing the issue. It reads

```
   elseif(HAS_SVE128)
            set(myarch "sve-128")
   endif()
```

When I changed `set(myarch "sve-128")` to `set(myarch "sve128")`, the installation completed with no issue. Could someone please take a look and see whether this is a bug? Or let me know if there is an additional compiler flag etc. I should specify to avoid the error (perhaps related to the custom alignment the error message asks for)? 

Thank you! I'm happy to test further, just let me know.

davschneller:
Thanks for testing—that's indeed a bug (affecting the detection of SVE dummy archs and the RISC-V V dummy archs). I've fixed it.

PR #1472, opened 2025-12-12
Fix/restrict alignment padding

davschneller:
Fix a mismatch between transfer pointer taking and aligning data to the `Alignment` parameter—that causes issues if `Vectorsize < Alignment`.


Thomas-Ulrich (approved):
LGTM

PR #1473, opened 2025-12-13
Disable plasticity per group if wanted

davschneller:
Allow to disable plasticity per material group, removing the need to specify a too high `plastCo` to emulate that.

Currently, we don't split the clusters by plasticity or not (i.e. there _might_ be tiny imbalances at least in the CPU case)—but that would also be doable with #1411 merged.

### DIFF
--- a/docs/parameters.par
+++ b/docs/parameters.par
@@ -15,7 +15,7 @@ UseCellHomogenizedMaterial = 1
 !off-fault plasticity parameters (disabled iff Plasticity=0)
 Plasticity=0
 !comma-separated list of groups for which plasticity is disabled, even if we enable it globally
-plasticitydisable=''
+plasticitydisabledgroups=''
 Tv=0.05
 !yaml file defining spatial dependance of properties necessary for plasticity. MaterialFileName is taken for plasticity if this is not provided
 PlasticityFileName = ''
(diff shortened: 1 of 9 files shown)


docs/parameters.par
```diff
@@ -12,8 +12,10 @@
 MaterialFileName = '33_layered_constant.yaml'
 !1: Compute average materials for each cell, 0: sample material values at element barycenters
 UseCellHomogenizedMaterial = 1
-!off-fault plasticity parameters (ignored if Plasticity=0)
+!off-fault plasticity parameters (disabled iff Plasticity=0)
 Plasticity=0
+!comma-separated list of groups for which plasticity is disabled, even if we enable it globally
+plasticitydisable=''
```
Thomas-Ulrich: plasticityDisabledGroups or plasticityDisabledRegions?

src/Solver/TimeStepping/TimeCluster.h
```diff
@@ -130,6 +130,8 @@ class TimeCluster : public AbstractTimeCluster {
 
   seissol::memory::MemkindArray<std::size_t> yieldCells;
 
+  std::size_t plasticityCells{0};
```
Thomas-Ulrich: numPlasticCells?
(avoid confusion that it can be a pointer to the cells?)

src/Initializer/CellLocalInformation.h
```diff
@@ -31,6 +31,9 @@ struct CellLocalInformation {
 
   // LTS setup
   LtsSetup ltsSetup;
+
+  // plasticity enabled
+  bool plasticity{false};
```
Thomas-Ulrich: plasticityEnabled?

PR #1474, opened 2025-12-15
Add an option to definitively disable the DR

davschneller:
Add an option that converts all DR faces to "regular" inner faces; and disable any DR-related input/output.
For comparison, `FL=0` still uses a dummy friction law.

### DIFF
--- a/tests/Geometry/MeshReader.t.h
+++ b/tests/Geometry/MeshReader.t.h
@@ -1,0 +1,48 @@
+// SPDX-FileCopyrightText: 2020 SeisSol Group
+//
+// SPDX-License-Identifier: BSD-3-Clause
+// SPDX-LicenseComments: Full text under /LICENSE and /LICENSES/
+//
+// SPDX-FileContributor: Author lists in /AUTHORS and /CITATION.cff
+
+#include <doctest.h>
+
+#include "Common/Constants.h"
+#include "Geometry/MeshReader.h"
+#include "TestHelper.h"
+
+#include <Eigen/Dense>
+#include <array>
+
+namespace seissol::unit_test {
+class MockReader2 : public seissol::geometry::MeshReader {
+  public:
+  explicit MockReader2(int boundaryType) : seissol::geometry::MeshReader(0) {
+    m_vertices.resize(4);
+
+    m_elements.resize(1);
+    m_elements.at(0).vertices[0] = 0;
+    m_elements.at(0).vertices[1] = 1;
+    m_elements.at(0).vertices[2] = 2;
+    m_elements.at(0).vertices[3] = 3;
+    m_elements.at(0).boundaries[0] = boundaryType;
+    m_elements.at(0).boundaries[1] = boundaryType;
+    m_elements.at(0).boundaries[2] = boundaryType;
+    m_elements.at(0).boundaries[3] = boundaryType;
+  }
+};
+
+TEST_CASE("MeshReader") {
+  SUBCASE("No DR") {
+    MockReader2 rdr(3);
+
+    rdr.disableDR();
+
+    REQUIRE(rdr.getElements()[0].boundaries[0] == 0);
+    REQUIRE(rdr.getElements()[0].boundaries[1] == 0);
+    REQUIRE(rdr.getElements()[0].boundaries[2] == 0);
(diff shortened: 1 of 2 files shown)


src/Geometry/MeshReader.h
```diff
@@ -116,6 +116,8 @@ class MeshReader {
 
   void exchangeGhostlayerMetadata();
 
+  void disableDR();
```
vikaskurapati, line 122: Please add a docstring/documentation.

src/Initializer/InitProcedure/InitMesh.cpp
```diff
@@ -72,12 +72,17 @@ void postMeshread(seissol::geometry::MeshReader& meshReader,
...
 
-  logInfo() << "Extracting fault information.";
   const auto& drParameters = seissolInstance.getSeisSolParameters().drParameters;
   const VrtxCoords center{drParameters.referencePoint[0],
                           drParameters.referencePoint[1],
                           drParameters.referencePoint[2]};
-  meshReader.extractFaultInformation(center, drParameters.refPointMethod);
+  if (true) {
+    logInfo() << "The Dynamic Rupture component has been disabled for this simulation.";
+    meshReader.disableDR();
```
vikaskurapati, line 81: Does it not mean that it always goes into this branch and disables DR always? Should we not take the DR enable flag from the parameters?

davschneller: Yes, we should. :D Thanks, forgot that one from the quick build-up of the branch.

docs/parameters.par
```diff
@@ -58,6 +58,7 @@ avoidic = 0    ! 1 iff use the IC only for analytical boundary conditions
 /
 
 &DynamicRupture
+enabled = 1                  ! explicitly enable/disable the DR (different to setting FL=0; it'll remove all DR faces)
```
Thomas-Ulrich: it'll remove all DR faces
->
it'll convert all DR faces to "regular" inner faces and disable any DR-related input/output.

?
(or maybe that's too long)

vikaskurapati: I like the longer version. Helps to understand better if someone is unfamiliar with what is happening. How about "converts all DR faces to "regular" inner faces and disables any DR-related I/O" ?

davschneller: Ok; I've added/updated it.

davschneller:
Ok, I've added a short test as well (though I guess for the _whole_ mesh reading system, we'll still need a _lot_ more tests coming up).

vikaskurapati:
> Ok, I've added a short test as well (though I guess for the _whole_ mesh reading system, we'll still need a _lot_ more tests coming up).

Agreed, one strategy is to keep adding tests with modifications in different PRs. That would be slowly building up the test suite.

davschneller:
Yes, that sounds good (though I'm not sure if it'll be feasible for absolutely everything right off the bat).

But we might also want to look into unit testing and verifying all the rest of SeisSol that already exists and is not changed—but separately from introducing new features (to not over-burden the PRs that add something new).
E.g. to verify the whole kernel dispatchers, point sources, DR, and so on. And everything written manually. Though I'd prefer it to be a bit more lenient when it comes to MPI and/or GPUs since we'll still need to provide the test infrastructure for that (but once the structure is there, we can always demand new tests).

E.g. I've started adding some tests for the module folder in #1455 .

vikaskurapati:
> Yes, that sounds good (though I'm not sure if it'll be feasible for absolutely everything right off the bat).
> 

Of course. But wherever possible, we should try to add them already. 

> But we might also want to look into unit testing and verifying all the rest of SeisSol that already exists and is not changed—but separately from introducing new features (to not over-burden the PRs that add something new). E.g. to verify the whole kernel dispatchers, point sources, DR, and so on. And everything written manually. Though I'd prefer it to be a bit more lenient when it comes to MPI and/or GPUs since we'll still need to provide the test infrastructure for that (but once the structure is there, we can always demand new tests).
> 

Regarding MPI and/or GPUs, since there is no infrastructure in place to demand new tests for CI/unit testing, developers should test a couple of cases that they believe would likely be affected by changes, and ensure that they are working as intended. Tracking these continuously would help in the long run, imo. 

> E.g. I've started adding some tests for the module folder in #1455 .

PR #1479, opened 2025-12-30
Some Docs Updates

davschneller:
* Update the docs a bit—to make them a bit more coherent (including some docs changes from #1331 )
* Switch from the RTD theme to the Book theme (looks better; at least IMO—and it seems to be pretty "trendy" right now)
* Normalize all image file names (in `docs/figures`) to kebab-case.

### DIFF
--- a/docs/build-parameters.rst
+++ b/docs/build-parameters.rst
@@ -42,6 +42,9 @@ You may explicitly compile and install multiple of these configurations at the s
     * ``single``: use single precision. Recommended in general for faster simulations. But especially for consumer GPUs (i.e. Nvidia GeForce, AMD Radeon, Intel ARC etc.), since these have usually a high performance difference between single and double precision.
     * ``double``: use double precision. Recommended, if your simulation fails with Inf/NaN errors in single precision builds. See also https://github.com/SeisSol/SeisSol/issues/200 .
 
+- ``NUMBER_OF_FUSED_SIMULATIONS``: the number of simulations run simultaneously.
+- ``NEW_BINARY_NAMING``: use an updated binary naming scheme which uses lower-case letters and a shorter notation
+
 Besides these, the host or, if enabled, the device architecture and backend are also encoded in the name of the executable.
 
 Generic parameters
@@ -61,10 +64,12 @@ Generic parameters
 - ``LTO``: enable link-time optimization
 - ``GRAPH_PARTITIONING_LIBS``: compile for the given graph partitioning libraries. Allowed options are:
 
-    * ``parmetis``: Repository: https://github.com/KarypisLab/ParMETIS
-    * ``PTScotch``: Repository:
-    * ``ParHIP``: Repository: https://github.com/KaHIP/KaHIP
-    * ``none``: Do not require a graph partitioning library. Recommended only for single-node/proxy builds.
+    * ``parmetis``: repository: https://github.com/KarypisLab/ParMETIS
+    * ``PTScotch``: repository: https://www.labri.fr/perso/pelegrin/scotch/
+    * ``ParHIP``: repository: https://github.com/KaHIP/KaHIP
+    * ``none``: do not require a graph partitioning library. Recommended only for single-node/proxy builds.
+- ``BUILD_DOCS``: build the (sphinx) documentation as HTML
+- ``BUILD_DOXYGEN``: build the doxygen docs for the SeisSol code
 
 CPU-specific parameters
 -----------------------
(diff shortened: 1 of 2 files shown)


docs/basic-code-structure.rst
```diff
@@ -9,32 +9,43 @@
...
-Solver        Time-stepping and code executed during a simulation.
-SourceTerm    Everything related to kinematic rupture models.
-============= =============
+Keep all folders in ``PascalCase`` here.
+
+=============== =============
+Folder          Description
+=============== =============
+Common          Routines which are used in several parts of the code, or are generic enough to not fit anywhere else.
+Dynamic Rupture Contains the Dynamic Rupture implementation (i.e. friction laws, input, output).
```
Thomas-Ulrich: no space (DynamicRupture)

docs/build-archs.rst
```diff
@@ -57,30 +57,30 @@ CPU architectures
      - Intel Knight's Corner (Xeon Phi coprocessor)
      - Generates Knight's Corner-specific instructions.
      - Intel Xeon Phi coprocessor
-     - (not known anymore)
+     - (none known anymore)
```
Thomas-Ulrich, line 63: Legacy ?
(Old, but still functional)

docs/build-archs.rst
```diff
@@ -57,30 +57,30 @@ CPU architectures
...
    * - ``rome``
      - AMD Zen 2
-     - Generates AVX2 instructions. For the libxsmm kernel generator, it is deemed equivalent to ``hsw``.
+     - Generates AVX2 instructions. For the kernel generators, it is deemed equivalent to ``hsw``.
      - Ryzen 3??? series, 7?2?, 8?2? series
      - LUMI (CPU partition)
    * - ``milan``
      - AMD Zen 3
-     - Generates AVX2 instructions. For the libxsmm kernel generator, it is deemed equivalent to ``hsw``.
+     - Generates AVX2 instructions. For the kernel generators, it is deemed equivalent to ``hsw``.
```
Thomas-Ulrich: but it still matters in terms of hardcoded cache size?

docs/build-dependencies.rst
```diff
@@ -190,20 +190,20 @@ You can find the installation instructions for it `under this link <https://easy
...
+
 Code generators for CPUs (optional, recommended)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 We support the following CPU code generators:
 
-- libxsmm (``libxsmm\_gemm\_generator``) will give reasonable performance on most ``x86`` machines. Its JIT variant also supports ARM CPUs.
-- PSpaMM (``pspamm-generator``): can handle some special cases faster; recommended mostly on AVX512-capable machines in conjunction with LIBXSMM. Otherwise slightly slower than LIBXSMM.
+- libxsmm (``libxsmm\_gemm\_generator``) will give reasonable performance on most ``x86`` machines. Its JIT variant also supports ARM and RISC-V CPUs.
+- PSpaMM (``pspamm-generator``): is the default for handling sparse matrix multiplications; and the LIBXSMM generator can come over as a bit dated in that area. It usually compares well against LIBXSMM; but is a tiny bit slower than it for dense-dense matrix multiplications.
```
Thomas-Ulrich, line 207: Can you give examples of equations solved with sparse matrix multiplications?  and with dense?

docs/build-run.rst
```diff
@@ -104,12 +105,30 @@ For the IO, the following environment variables are recommended to be set, when
...
 
-For GPUs, SeisSol should be best launched with one rank per GPU. To select, SeisSol will automatically pick the first visible GPU to it.
-However, some systems make all GPUs on a node visible to all processes running on it—potentially resulting in all SeisSol processes
-selecting the same GPU. To avoid that, SeisSol provides a launch script given as ``shared/seissol-launch`` which selects the GPU according to the node-local SLURM rank.
+For GPUs, SeisSol will utilize one GPU per rank.
+If multiple GPUs are visible to SeisSol, it will automatically pick the first visible GPU when ordering.
+
+Thus, care needs to be taken, as some systems make all GPUs on a node visible to all processes running on it—potentially resulting in all SeisSol processes
+selecting the same GPU.
+To avoid that, SeisSol provides a launch script given as ``shared/seissol-launch`` which selects the GPU according to the node-local SLURM rank.
```
Thomas-Ulrich, line 123: works only for nvidia GPUS?

docs/build-archs.rst
davschneller: If you mean the cacheline size and alignment: that's the same here. For the cache size/arch-specific settings for the compiler: yes those differ—though IIRC that won't really affect LIBXSMM and/or PSpaMM directly. I.e. the codegen is the same—but the compiler options are different.

E.g. for `skx` vs. `bergamo`: `skx` sometimes generates 256-bit vector instructions; whereas `bergamo` will generate "proper" 512-bit vector instructions in that place. E.g. for copy-scale-add operations.

docs/build-dependencies.rst
davschneller: Right now that's basically almost all of them. :) At least since #1243 : now we compute the whole ADER space-time evolution as sparse-dense and dense-sparse matrix products; similar for the flux computations with sparse-dense. (at least on the CPU; the sparse-dense principle is harder to generalize towards the GPU; but probably still somewhat possible) In all of these cases, we use PSpaMM even with LIBXSMM enabled.
Before that, we had the "classic" (GPU-portable) version that only the Astar and source matrices are sparse; and everything else dense. And PSpaMM was used only in these cases.

So there are almost no dense-dense matrix multiplications left.
Except the DR projection and back.

docs/build-run.rst
davschneller: That should also work for AMD—at least I had used it in the past that way IIRC. (or did you have a different experience on LUMI? The script in the LUMI docs, however, should in essence do the same as the `seissol-launch` script)
In fact, the switch to checking `ROCR_VISIBLE_DEVICES` instead of `HIP_VISIBLE_DEVICES` is because the former also works if you subdivide your AMD GPU (i.e. you can split an MI300A into 3 or 6 sub-GPUs if you want—I know that splitting it in 4 is enough to run 3×tpv5 on a single MI300A).

Thomas-Ulrich: Good to know (then we can update the LUMI doc).
By the way there is a typo in the Usage line (should be seissol-launch).

```
(local) ulrichth@uan01:/project/project_465002391/ulrich/seissol_base/seissol/build> cat seissol-launch 
#!/bin/bash
# SPDX-FileCopyrightText: 2022 SeisSol Group
#
# SPDX-License-Identifier: BSD-3-Clause
# SPDX-LicenseComments: Full text under /LICENSE and /LICENSES/
#
# SPDX-FileContributor: Author lists in /AUTHORS and /CITATION.cff

# This is a sample gpu launch file.
# you can make a copy of the generated one
# and adjust it based on your needs

# Usage: mpirun -n <num_proc> ./launch ./SeisSol* ./parameters.par

if [[ ! -z "${OMPI_COMM_WORLD_LOCAL_RANK}" ]]; then
  LOCAL="${OMPI_COMM_WORLD_LOCAL_RANK}"

elif [[ ! -z "${MV2_COMM_WORLD_LOCAL_RANK}" ]]; then
  LOCAL="${MV2_COMM_WORLD_LOCAL_RANK}"

elif [[ ! -z "${PMI_RANK}" ]]; then
  LOCAL="${PMI_RANK}"

elif [[ ! -z "${SLURM_LOCALID}" ]]; then
  LOCAL="${SLURM_LOCALID}"

else
  LOCAL=0
fi

export HIP_VISIBLE_DEVICES="${LOCAL}"
"$@"
```

docs/build-dependencies.rst
Thomas-Ulrich: Nice! Would it make sense to add these info in the docs too?

docs/build-archs.rst
Thomas-Ulrich: ok!
Then I would suggest adding somewhere:

"Note that it may still be useful to provide the exact build architecture (e.g., bergamo) instead of a generic one (e.g., skx), as this allows the compiler to use customized cache sizes and architecture-specific settings."

docs/build-archs.rst
davschneller: Ah, the entry for `(none known anymore)` means if there are any products/clusters which still use it (which I don't really know anymore). IIRC GCC was already thinking about removing at least knc, if not knl as well.

I don't think we have any hardware anymore to test them, nor that it would make much sense probably
(I think I have a KNC card in my office—but I doubt it would make sense to re-activate it really ;) ).

docs/build-archs.rst
davschneller: Ok, sounds good to me. The generic architectures are, by now, `avx10-512` etc.; but you'd need to set `mtune` and so on manually.

There's also #1471 which could help with finding the architecture automatically (but still lacks automatically setting `-march=native -mcpu=native -mtune=native`).

docs/build-dependencies.rst
davschneller: Yes, I can—you're right, right now it sounds like a "choose one" kind of option while it actually isn't.

docs/build-run.rst
davschneller: Ok, noted and adjusted (in the upcoming commit).

PR #1481, opened 2026-01-03
Even more cleanup

davschneller:
* Harmonize private/protected member names: remove `m_` prefix, and suffix all private/protected members with `_`.
* Apply clang-tidy on the device-only files a bit
* Split the CMake file list into subdirectories; each producing an `OBJECT` library (to clean up the central files-to-build list and prevent potential merge problems with it).
* Allow static device kernel libraries.

### DIFF
--- a/src/Parallel/Runtime/Stream.h
+++ b/src/Parallel/Runtime/Stream.h
@@ -107,7 +107,13 @@ class StreamRuntime {
 
   template <typename F>
   void enqueueHost(F&& handler) {
-    device().api->streamHostFunction(streamPtr_->get(), std::forward<F>(handler));
+    if (Backend != DeviceBackend::Hip) {
+      device().api->streamHostFunction(streamPtr_->get(), std::forward<F>(handler));
+    } else {
+      // if the stream host function call isn't implemented or slow, we'll need to synchronize
+      wait();
+      std::invoke(handler);
+    }
   }
 
   template <typename F>


src/Geometry/Refinement/VariableSubSampler.h
```diff
@@ -25,24 +25,24 @@ namespace seissol::refinement {
 template <class T>
 class VariableSubsampler {
   private:
-  std::vector<basisFunction::SampledBasisFunctions<T>> m_BasisFunctions;
+  std::vector<basisFunction::SampledBasisFunctions<T>> BasisFunctions_;
```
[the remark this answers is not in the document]
vikaskurapati: Do you agree with this? Or `BasisFunctions_` named this way for some reason I am missing?

src/Geometry/Refinement/VariableSubSampler.h
```diff
@@ -25,24 +25,24 @@ namespace seissol::refinement {
...
 
   /** The original number of cells (without refinement) */
-  unsigned int mNumCells;
+  unsigned int mNumCells_;
```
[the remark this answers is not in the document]
vikaskurapati: I agree with this. Do you also agree?

src/Geometry/Refinement/VariableSubSampler.h
```diff
@@ -25,24 +25,24 @@ namespace seissol::refinement {
...
 
-  std::size_t kSubCellsPerCell;
-  std::size_t kNumVariables;
-  std::size_t kNumAlignedDOF;
+  std::size_t kSubCellsPerCell_;
+  std::size_t kNumVariables_;
+  std::size_t kNumAlignedDOF_;
```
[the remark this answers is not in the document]
vikaskurapati: Do you agree with this or is `k` left as a prefix for some other reason?

src/Geometry/MeshReader.h
```diff
@@ -49,38 +49,34 @@ struct LinearGhostCell {
 
 class MeshReader {
   protected:
-  int mRank{0};
+  int rank_{0};
```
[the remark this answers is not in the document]
vikaskurapati: Do you think this makes sense? I understand this did not presumably cause an issue till now, but do you foresee any confusing behavior in the future if we let it stay this way?

src/Initializer/InitProcedure/Init.cpp
```diff
@@ -25,6 +25,7 @@
 #include "Monitoring/Unit.h"
 #include "Numerical/Statistics.h"
 
+#include <Device/device.h>
 #include <ostream>
 #include <sstream>
 #endif
```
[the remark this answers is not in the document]
davschneller: That is unfortunately incorrect. It _is_ guarded by an `ACL_DEVICE`.

src/Geometry/Refinement/VariableSubSampler.h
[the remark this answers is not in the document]
davschneller: Yes, I agree. Might be an oversight. :) Somehow `clang-tidy` didn't catch it.

(tbf these file might soon be obsolete anyways; as soon #1180 lands)

src/Geometry/Refinement/VariableSubSampler.h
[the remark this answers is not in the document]
davschneller: Same. I've changed it.

src/Initializer/InitProcedure/Init.cpp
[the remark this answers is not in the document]
vikaskurapati: Exactly. This is proof that AI can make mistakes. ;)

src/Geometry/MeshReader.h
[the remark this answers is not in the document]
davschneller: It _might_ have been the reason why the cube generator code didn't work correctly after all.
The `rank_` variable for the `MeshReader` was always set correctly; just not for the `CubeGenerator` (if someone tried to use that in a multi-rank setting).
I've fixed it.

Though it shadowing the `rank()` method should not be true.

src/Geometry/Refinement/VariableSubSampler.h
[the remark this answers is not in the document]
davschneller: Yes; agreed. I've adjusted it.

src/Common/Iterator.h
```diff
@@ -58,69 +58,70 @@ class Zip {
     Iterator(bool lenient,
              const std::tuple<IteratorTs...>& iterators,
              const std::tuple<IteratorTs...>& iteratorEnds)
-        : iterators(iterators), iteratorEnds(iteratorEnds),
-          ended(tupleAnyEqual(iterators, iteratorEnds)), lenient(lenient) {
-      if (!lenient && ended && iterators != iteratorEnds) {
+        : iterators_(iterators), iteratorEnds_(iteratorEnds),
+          ended_(tupleAnyEqual(iterators, iteratorEnds)), lenient_(lenient) {
+      if (!lenient && ended_ && iterators != iteratorEnds) {
```
vikaskurapati, line 63: It should probably not make much difference, but why are we using `ended_`, i.e., the class' private variable only for `ended` and not the other two here, i.e., for `lenient` and `iterators`?

davschneller: That looks like a `clang-tidy` oversight tbh; I'll adjust it

src/DynamicRupture/FrictionLaws/CpuImpl/AgingLaw.h
```diff
@@ -41,7 +41,7 @@ class AgingLaw : public SlowVelocityWeakeningLaw<AgingLaw<TPMethod>, TPMethod> {
                                            double stateVarReference,
                                            double timeIncrement,
                                            double localSlipRate) const {
-    const double localSl0 = this->sl0[faceIndex][pointIndex];
+    const double localSl0 = this->sl0_[faceIndex][pointIndex];
```
vikaskurapati, line 44: Could we also get rid of the `this->` here, or is it clearer with it?

src/DynamicRupture/FrictionLaws/CpuImpl/BaseFrictionLaw.h
```diff
@@ -50,30 +50,30 @@ class BaseFrictionLaw : public FrictionSolver {
...
       return;
     }
 
     if constexpr (model::MaterialT::SupportsDR) {
 
       SCOREP_USER_REGION_DEFINE(myRegionHandle)
-      std::copy_n(frictionTime.deltaT.begin(), frictionTime.deltaT.size(), this->deltaT);
-      this->mFullUpdateTime = fullUpdateTime;
+      std::copy_n(frictionTime.deltaT.begin(), frictionTime.deltaT.size(), this->deltaT_);
+      this->mFullUpdateTime_ = fullUpdateTime;
```
vikaskurapati: Should we get rid of the `m` suffix here?

davschneller: Yes, I think we should (that's also one of the `m_` relics; one that stayed a bit longer)

src/DynamicRupture/FrictionLaws/CpuImpl/BaseFrictionLaw.h
```diff
@@ -157,26 +157,26 @@ class BaseFrictionLaw : public FrictionSolver {
...
+                                        impAndEta_[ltsFace],
                                         timeWeights,
-                                        spaceWeights,
-                                        godunovData[ltsFace],
-                                        slipRateMagnitude[ltsFace],
-                                        this->drParameters->energiesFromAcrossFaultVelocities);
+                                        spaceWeights_,
+                                        godunovData_[ltsFace],
+                                        slipRateMagnitude_[ltsFace],
+                                        this->drParameters_->energiesFromAcrossFaultVelocities);
```
vikaskurapati, line 181: This kinda confuses me. All internal data structures don't have ' this -> ' except `drParameters_`. Is there a reason for this, or is it more of a legacy artifact? Should we get rid of it, too?

src/DynamicRupture/FrictionLaws/CpuImpl/AgingLaw.h
davschneller: _Might_ be harder, probably since the class is templated (that sometimes needs us to explicitly state the `this` for some reasons... ... Though, arguably, it reminds me of some languages like Rust or JavaScript that enforce using `this` or `self` in all cases for clarity).

src/DynamicRupture/FrictionLaws/CpuImpl/FastVelocityWeakeningLaw.h
vikaskurapati, line 1: The same question with `this->` here. I am avoiding writing the same comment in the other files from here to avoid polluting the PR with it too many times.

src/DynamicRupture/FrictionLaws/CpuImpl/BaseFrictionLaw.h
davschneller: Could be due to templates and inheritance. IIRC it didn't really work without that there.

We could try to go the GPU-esque way and have an extra datastructure which carries the pointers (or mid-term try to use our memory storage struct pointers instead). ... Or for now use `this` everywhere there. Or keep it as it is.

src/DynamicRupture/FrictionLaws/CpuImpl/AgingLaw.h
vikaskurapati: Okay. I generally lean towards maintaining it consistently, i.e., either all use `this->` or none do. A bit more leaning towards not having it. But if it is explicitly necessary, I guess there is not much we could do.

src/DynamicRupture/FrictionLaws/CpuImpl/ImposedSlipRates.h
```diff
@@ -31,38 +31,40 @@ class ImposedSlipRates : public BaseFrictionLaw<ImposedSlipRates<STF>> {
                              std::array<real, misc::NumPaddedPoints>& /*strengthBuffer*/,
                              std::size_t ltsFace,
                              uint32_t timeIndex) {
-    const real timeIncrement = this->deltaT[timeIndex];
-    real currentTime = this->mFullUpdateTime;
+    const real timeIncrement = this->deltaT_[timeIndex];
+    real currentTime = this->mFullUpdateTime_;
```
vikaskurapati: m here

src/DynamicRupture/FrictionLaws/CpuImpl/LinearSlipWeakening.h
```diff
@@ -150,58 +150,58 @@ class LinearSlipWeakeningLaw : public BaseFrictionLaw<LinearSlipWeakeningLaw<Spe
...
 
   void calcStateVariableHook(std::array<real, misc::NumPaddedPoints>& stateVariable,
                              uint32_t timeIndex,
                              std::size_t ltsFace) {
     alignas(Alignment) real resampledSlipRate[misc::NumPaddedPoints]{};
-    specialization.resampleSlipRate(resampledSlipRate, this->slipRateMagnitude[ltsFace]);
+    specialization_.resampleSlipRate(resampledSlipRate, this->slipRateMagnitude_[ltsFace]);
 
-    real time = this->mFullUpdateTime;
+    real time = this->mFullUpdateTime_;
```
vikaskurapati: m here

src/DynamicRupture/FrictionLaws/GpuImpl/BaseFrictionSolver.h
```diff
@@ -63,7 +63,7 @@ struct FrictionLawContext {
 };
 
 #ifdef __CUDACC__
-SEISSOL_DEVICE inline void deviceBarrier(FrictionLawContext& ctx) { __syncthreads(); }
+SEISSOL_DEVICE inline void deviceBarrier(FrictionLawContext& /*ctx*/) { __syncthreads(); }
 #elif defined(__HIP__)
 SEISSOL_DEVICE inline void deviceBarrier(FrictionLawContext& ctx) { __syncthreads(); }
```
vikaskurapati: Why is `ctx` not commented out here in this function and in the else branch? From what I understand, only the SYCL barrier uses `ctx`.

src/DynamicRupture/FrictionLaws/GpuImpl/BaseFrictionSolverCudaHip.cpp
```diff
@@ -77,20 +79,20 @@ void BaseFrictionSolver<T>::evaluateKernel(seissol::parallel::runtime::StreamRun
...
-  args.resampleMatrix = resampleMatrix;
-  args.tpInverseFourierCoefficients = devTpInverseFourierCoefficients;
-  args.tpGridPoints = devTpGridPoints;
-  args.heatSource = devHeatSource;
+  args.data = this->data_;
+  args.spaceWeights = this->devSpaceWeights_;
+  args.resampleMatrix = this->resampleMatrix_;
+  args.tpInverseFourierCoefficients = this->devTpInverseFourierCoefficients_;
+  args.tpGridPoints = this->devTpGridPoints_;
+  args.heatSource = this->devHeatSource_;
```
vikaskurapati, line 98: Just out of curiosity, it worked without `this->` before, but when we append an `_`, `this->` is needed. Do we know why this is the case?

src/DynamicRupture/FrictionLaws/GpuImpl/BaseFrictionSolverSycl.cpp
```diff
@@ -32,27 +32,28 @@ void BaseFrictionSolver<T>::evaluateKernel(seissol::parallel::runtime::StreamRun
...
   args.fullUpdateTime = fullUpdateTime;
 
-  sycl::nd_range rng{{this->currLayerSize * misc::NumPaddedPoints}, {misc::NumPaddedPoints}};
+  sycl::nd_range rng{{this->currLayerSize_ * misc::NumPaddedPoints}, {misc::NumPaddedPoints}};
   queue->submit([&](sycl::handler& cgh) {
     // NOLINTNEXTLINE
     sycl::local_accessor<real> sharedMemory(misc::NumPaddedPoints, cgh);
 
     cgh.parallel_for(rng, [=](sycl::nd_item<1> item) {
+      FrictionLawArgs argsLocal = args;
```
vikaskurapati: Just out of curiosity, why was this extra argument copy required?

src/Geometry/CubeGenerator.h
```diff
@@ -22,13 +22,11 @@
 namespace seissol::geometry {
 
 class CubeGenerator : public seissol::geometry::MeshReader {
-  int rank;
-  int nProcs;
+  int rank_;
+  int nProcs_;
```
vikaskurapati, line 26: Could these be made consts? If I understand correctly, they won't change throughout the program.

src/Geometry/MeshReader.cpp
```diff
@@ -113,7 +113,7 @@ void MeshReader::extractFaultInformation(
 
       // DR boundary
 
-      if (i.neighborRanks[j] == mRank) {
+      if (i.neighborRanks[j] == seissol::Mpi::mpi.rank()) {
```
vikaskurapati: It probably is not a big deal given that this function call is not really expensive, but you may be able to call `mpi.rank()` once at the beginning of the function, and use it here.

src/Geometry/MeshReader.cpp
```diff
@@ -137,8 +137,8 @@ void MeshReader::extractFaultInformation(
       // Does not work for all meshes
       //                VrtxCoords elementCenter;
       //                VrtxCoords faceCenter;
-      //                MeshTools::center(*i, m_vertices, elementCenter);
-      //                MeshTools::center(*i, j, m_vertices, faceCenter);
+      //                MeshTools::center(*i, vertices_, elementCenter);
+      //                MeshTools::center(*i, j, vertices_, faceCenter);
       //
       //                bool isPlus = (MeshTools::distance(elementCenter, center)
       //                        < MeshTools::distance(faceCenter, center));
```
vikaskurapati: Is the comment still relevant?

src/Geometry/MeshReader.cpp
```diff
@@ -205,14 +205,14 @@ void MeshReader::extractFaultInformation(
       // Compute second vector in the plane, orthogonal to the normal and tangent 1 vectors
       MeshTools::cross(f.normal, f.tangent1, f.tangent2);
 
-      auto remoteNeighbor = i.neighbors[j] == static_cast<int>(m_elements.size());
+      auto remoteNeighbor = i.neighbors[j] == static_cast<int>(elements_.size());
```
vikaskurapati: Const, may be?

src/Geometry/MeshReader.cpp
```diff
@@ -232,19 +232,19 @@ void MeshReader::extractFaultInformation(
...
       if (isPlus || neighborIndex >= 0) {
-        m_hasPlusFault = true;
+        hasPlusFault_ = true;
       }
     }
   }
 
   // Sort fault neighbor lists and update MPI fault indices
-  for (auto& i : m_MPIFaultNeighbors) {
+  for (auto& i : mpiFaultNeighbors_) {
```
vikaskurapati: Can we use a better variable name instead of just `i`? `neighbor` or something like that?

src/Geometry/MeshReader.cpp
davschneller: Yes, can do for verbosity reasons.

The `mpi.rank()` should _not_ call MPI internally; that's only done once at the beginning of the program.

src/Geometry/MeshReader.cpp
davschneller: I think not (it's probably super old, and there were not yet complaints about the current method); I can remove it.

src/Geometry/MeshReader.cpp
davschneller: Yes, definitely. IIRC that was a missed refactor anyways. Same for the `: elements_)` loop before that.

src/Geometry/CubeGenerator.h
davschneller: `const` member variables are alas not such a good idea; as that'll unintendedly hinder a lot of other things w.r.t. the class behavior. There's IIRC a clang-tidy rule against that currently. (cf. also https://old.reddit.com/r/cpp/comments/8wbeom/coding_guideline_avoid_const_member_variables/ for some opinions)

Yes; by that extension, the whole `seissolInstance_`s might also not work well (as mentioned in that post); they should in principle also be scheduled for removal at some time. (they _are_ better than just having a globally-static SeisSol object like before—but it's still not ideal)

src/Initializer/BatchRecorders/DataTypes/Table.h
```diff
@@ -28,50 +26,22 @@ class GenericTableEntry {
...
-  }
-
-  GenericTableEntry& operator=(const GenericTableEntry& other) = delete;
-
-  virtual ~GenericTableEntry() {
-    if (deviceDataPtr != nullptr) {
-      device.api->freeGlobMem(deviceDataPtr);
-      deviceDataPtr = nullptr;
-    }
-  }
```
vikaskurapati, line 59: I don't understand why these methods are no longer useful. Is the remaining part of the code modified such that these methods become obsolete?

src/DynamicRupture/FrictionLaws/GpuImpl/BaseFrictionSolver.h
davschneller: That's probably a refactor miss (IIRC I ran clang-tidy once with CUDA enabled; but not with HIP)

src/Initializer/BatchRecorders/DataTypes/Table.h
davschneller: That's now covered somewhere else; via the `MemkindArray` type. There, we now take care of the manual allocation/deallocation, so we don't need to deal with it here. Helps avoiding common mistakes when implementing the same pattern over and over again. (cf. using a `vector` to using bare pointers with `malloc` and `free` or `new` and `delete`)

src/DynamicRupture/FrictionLaws/GpuImpl/BaseFrictionSolverCudaHip.cpp
davschneller: I think I added the `this` for consistency in that case—again, since it "feels" more complete that way.

... though that does indeed duplicate with the trailing `_`; I can see if I can remove them again.

src/DynamicRupture/FrictionLaws/GpuImpl/BaseFrictionSolverSycl.cpp
davschneller: I wasn't 100%ly certain if the `args` was really captured by value in all cases, since we only use a reference on it. So I added the explicit copy that might (hopefully) be elided. ... I'll try checking again; maybe `std::move` works here as well.

src/Initializer/Parameters/ParameterReader.h
```diff
@@ -146,26 +146,27 @@ class ParameterReader {
...
     }
   }
 
-  YAML::Node node; // apparently the YAML nodes use a reference semantic. Hence, we do it like this.
-  std::string rootPath;
-  bool empty;
-  std::unordered_set<std::string> visited;
-  std::unordered_map<std::string, std::shared_ptr<ParameterReader>> subreaders;
+  YAML::Node
+      node_; // apparently the YAML nodes use a reference semantic. Hence, we do it like this.
```
vikaskurapati: Can we skip linting here? Pushing this to the next line is weird formatting, imo.

src/Initializer/MemoryManager.cpp
```diff
@@ -331,54 +332,60 @@ bool requiresNodalFlux(FaceType f) {
...
   auto product = factory->produce();
-  m_dynRup = std::move(product.storage);
-  m_DRInitializer = std::move(product.initializer);
-  m_FrictionLaw = std::move(product.frictionLaw);
-  m_FrictionLawDevice = std::move(product.frictionLawDevice);
-  m_faultOutputManager = std::move(product.output);
+  dynRup_ = std::move(product.storage);
+  DRInitializer_ = std::move(product.initializer);
+  FrictionLaw_ = std::move(product.frictionLaw);
+  FrictionLawDevice_ = std::move(product.frictionLawDevice);
```
vikaskurapati: This is probably too picky, but why do we have capitalization differences for three variables here? Was that intentional?

davschneller: It isn't really; that's also a `clang-tidy` miss. One of more out there it seems.

src/Initializer/Parameters/ParameterReader.h
davschneller: In that case, maybe it's best to use the line before that for the comment

src/Solver/TimeStepping/TimeCluster.cpp
```diff
@@ -276,39 +284,40 @@ void TimeCluster::computeDynamicRuptureDevice(SEISSOL_GPU_PARAM DynamicRupture::
...
-    streamRuntime.runGraph(
-        computeGraphKey, layerData, [&](seissol::parallel::runtime::StreamRuntime& streamRuntime) {
-          dynamicRuptureKernel.batchedSpaceTimeInterpolation(
-              table, pointsCollocate.data(), streamRuntime);
-        });
-    device.api->popLastProfilingMark();
+    streamRuntime_.runGraph(computeGraphKey,
+                            layerData,
+                            [&](seissol::parallel::runtime::StreamRuntime& /*streamRuntime*/) {
+                              dynamicRuptureKernel_.batchedSpaceTimeInterpolation(
```
vikaskurapati, line 290: If `streamRuntime` is not required anymore, can't we remove it from the lambda function arguments?

src/DynamicRupture/FrictionLaws/CpuImpl/BaseFrictionLaw.h
vikaskurapati: Fine with either. I just wanted to make sure it was intentional.

src/DynamicRupture/FrictionLaws/GpuImpl/BaseFrictionSolver.h
vikaskurapati: It may be a good idea to run it once with each of the backends we have. If not all, at least with CUDA and HIP.

src/Geometry/MeshReader.cpp
vikaskurapati: Okay. If it is not really expensive, I don't have a problem. Verbosity isn't a big deal here either. We can leave it as is.

src/Geometry/MeshReader.cpp
```diff
@@ -101,22 +101,22 @@ void MeshReader::disableDR() {
  */
 void MeshReader::extractFaultInformation(
     const VrtxCoords& refPoint, seissol::initializer::parameters::RefPointMethod refPointMethod) {
-  for (auto& i : elements_) {
+  for (auto& cell : elements_) {
```
vikaskurapati, line 104: At the risk of being nitpicky, or annoying, can I ask to name this `element` in stead of `cell` just because the iterator is named `elements_`? I am approving it in either cases because this is absolutely me being nitpicky.

PR #1482, opened 2026-01-03
Print units for the terminal energy output

davschneller:
* always use the scientific notation for the energies/moments etc. (however, please double-check the units if possible)
* output the total number of cells and the distribution over the ranks (and use decimal separators for displaying the total count here)
* add an additional, shorter pinning display (the "long" display stays next to it, though); akin to the free CPU masks

### DIFF
--- a/src/Monitoring/Unit.cpp
+++ b/src/Monitoring/Unit.cpp
@@ -9,6 +9,7 @@
 
 #include <array>
 #include <cmath>
+#include <cstddef>
 #include <cstdint>
 #include <iomanip>
 #include <ios>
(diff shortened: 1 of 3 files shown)


davschneller:
Sure; here you go:

Data from my laptop; very small tpv5-esque mesh w/o mini SeisSol.

```
2026-01-05 11:22:54.489 info 0 : OpenMP worker affinity (this process): "01------89|------" ( 0-1,8-9 ) 
2026-01-05 11:22:54.489 info 0 : OpenMP worker affinity (this node)   : "012345--89|0123--" ( 0-5,8-13 )
```
(the brackets at the end are new)

```
2026-01-05 11:22:54.688 info 0 : Cell count: 19'183 (per rank: 6394.33 ± 1496.77 ; range: [ 4280 ; 7539 ]) 
```

Thomas-Ulrich:
Ok, now I understand, nice!

PR #1487, opened 2026-01-21
Combine the local flux kernels

davschneller:
Combine all four local flux kernels on the GPU, as they are called in all cases but the DR. I.e. there are not many exceptions; too little to really take care of them (and they disturb the GPU control flow). So we just set the respective flux matrices to 0 and execute the respective operations anyways.

Makes the four kernels in sequence more compute-bound, as we accumulate onto the same data (`Q`) using the same data (`I`) for all four faces.

(on my local laptop (a pretty "memory-bound GPU"), the proxy absolutely loves the change; though it did also benefit from L2 compression in that area—one probably mostly cancels the effects of the other. And this one here is better)

This change is compatible with all three current code generation frameworks (as it can be translated into a single fused GEMM node in Yateto).


codegen/kernels/aderdg.py
```diff
@@ -66,6 +66,10 @@ def __init__(self, order, multipleSimulations, matricesDir):
 
         Aplusminus_spp = self.flux_solver_spp()
         self.AplusT = Tensor("AplusT", Aplusminus_spp.shape, spp=Aplusminus_spp)
+        self.AplusTz = [
+            Tensor(f"AplusTz({i})", Aplusminus_spp.shape, spp=Aplusminus_spp)
+            for i in range(4)
+        ]
```
vikaskurapati, line 72: Can we please have a better name for Tz instead? At first glance, it is not clear what this Tz tensor refers to.

codegen/kernels/aderdg.py
```diff
@@ -406,6 +410,19 @@ def addLocal(self, generator, targets):
...
+                    plusFluxMatrixAccessor(i) * self.I["lq"] * self.AplusTz[i]["qp"]
+                    for i in range(4)
+                ],
+                start=self.Q["kp"],
+            )
+            generator.add(
+                "gpu_localFluxAll",
+                localFluxAll,
+                target="gpu",
+            )
```
vikaskurapati, line 424: Do we need both `localFlux` and `localFluxAll` for GPUs now if we are doing this kernel merging?

src/Kernels/LinearCK/Local.h
```diff
@@ -73,6 +73,7 @@ class Local : public LocalKernel {
 #ifdef ACL_DEVICE
   kernel::gpu_volume deviceVolumeKernelPrototype;
   kernel::gpu_localFlux deviceLocalFluxKernelPrototype;
+  kernel::gpu_localFluxAll deviceLocalFluxAllKernelPrototype;
```
vikaskurapati, line 76: Same question as in the code generation: Is there a reason we want to keep both the local fluxes for GPUs now?

vikaskurapati:
> Combine all four local flux kernels on the GPU, as they are called in all cases but the DR. I.e. there are not many exceptions; too little to really take care of them (and they disturb the GPU control flow). So we just set the respective flux matrices to 0 and execute the respective operations anyways.
> 
> Makes the four kernels in sequence more compute-bound, as we accumulate onto the same data (`Q`) using the same data (`I`) for all four faces.
> 
> (on my local laptop (a pretty "memory-bound GPU"), the proxy absolutely loves the change; though it did also benefit from L2 compression in that area—one probably mostly cancels the effects of the other. And this one here is better)

So, if I understand correctly, we do more FLOP/s now to keep a simpler control flow, which, according to your experiments, is better? Can you quickly test it on one of the clusters in use, say Vista with Grace-Hopper, with some benchmark like TPV13 for comparison?

davschneller:
Sure, here you go. Here's the data for 0.5s tpv5 on a single MI300A with the latest `dev` tensorforge for order 4 with FP32.

* w/o this patch: 16.6322 s (= 10.4498 HW-TFLOP/s)
* with this patch: 15.0442 s (= 11.6024 HW-TFLOP/s; with about 0.43% more work)

And you can see it really in the proxy here, with about 17 TFLOP/s vs. 14-15 TFLOP/s (for with vs. without). The proxy also has all faces as inner faces in all cases for that benchmark.

For order 6 / FP32 on the same hardware, the relative difference is currently much lower (but there might still be a little optimization potential I think; especially in conjunction with #1490 mid-term—though it's currently expected to see less speedup the more compute bound (i.e. higher arithmetic intensity) our kernels are, relatively speaking):

* w/o this patch: 57.6414 s (= 17.3421 HW-TFLOP/s)
* with this patch: 56.2379 s (= 17.8377 HW-TFLOP/s; with about 0.35% more work)

The difference in the proxy is much less as well.

In total, it'll help in most current cases. Also; I forgot to mention: it's also compatible with all code generation frameworks without any more Yateto changes.

I think it's best to keep the code for the old kernels in for now (to resolve the questions you had—otherwise I agree; it'd be nice to remove those as well)—just in case there's a mesh where you have an _extremely_ high ratio of DR faces compared to "normal" faces. Though, ad hoc, I can't really think of one.

(there are currently two ideas to refine this kernel a bit more; but both of them will require more Yateto and codegen work)

vikaskurapati:
> Sure, here you go. Here's the data for 0.5s tpv5 on a single MI300A with the latest `dev` tensorforge for order 4 with FP32.
> 
> * w/o this patch: 16.6322 s (= 10.4498 HW-TFLOP/s)
> * with this patch: 15.0442 s (= 11.6024 HW-TFLOP/s; with about 0.43% more work)
> 
> And you can see it really in the proxy here, with about 17 TFLOP/s vs. 14-15 TFLOP/s (for with vs. without). The proxy also has all faces as inner faces in all cases for that benchmark.
> 

That is around 15% improvement, nice! :) 

> For order 6 / FP32 on the same hardware, the relative difference is currently much lower (but there might still be a little optimization potential I think; especially in conjunction with #1490 mid-term—though it's currently expected to see less speedup the more compute bound (i.e. higher arithmetic intensity) our kernels are, relatively speaking):
> 
> * w/o this patch: 57.6414 s (= 17.3421 HW-TFLOP/s)
> * with this patch: 56.2379 s (= 17.8377 HW-TFLOP/s; with about 0.35% more work)
> 
> The difference in the proxy is much less as well.

So, would you say that the effects are similar to what we saw during the work with fused-simulations? i.e., that the speedup is higher for lower-orders due to the computations now being pushed towards compute-bound regime given that they are memory-bound? 

> 
> In total, it'll help in most current cases. Also; I forgot to mention: it's also compatible with all code generation frameworks without any more Yateto changes.
> 
> I think it's best to keep the code for the old kernels in for now (to resolve the questions you had—otherwise I agree; it'd be nice to remove those as well)—just in case there's a mesh where you have an _extremely_ high ratio of DR faces compared to "normal" faces. Though, ad hoc, I can't really think of one.
> 

Okay, but correct me if I am wrong: with this PR, merging will be automatic, i.e., we won't be using the old kernels during computations anymore, right? If we want to keep the code for the old kernels in for such cases, there is a high ratio of DR faces compared to "normal" faces. Should we not have an option to run the code with older kernels?

I think the Turkey mesh has a relatively higher ratio of DR faces than TPV5, but it is a big mesh. If you want to compare performance across with a big case, it may be a good candidate. Even Ridgecrest may be a good candidate, but there R+S may also play a role.

> (there are currently two ideas to refine this kernel a bit more; but both of them will require more Yateto and codegen work)

davschneller:
Thanks! :)
Ok; I should've addressed everything now (I hope); to answer the remaining questions:

* I've kept in the "old" kernels for now; they can be enabled at compile time. If it is really better in some cases, then it might be better to make the switch happen at runtime in the mid-term—possibly automatically and per cluster. 
* For fused simulations: yes, I'd expect the uplift to be a bit higher in general, since we're more on the memory-bound side with them in any case—assuming full sparsity usage (and maybe there's an uplift even when not).

src/Kernels/LinearCK/Local.cpp
```diff
@@ -282,12 +285,32 @@ void Local::computeBatchedIntegral(
...
+    for (std::size_t face = 0; face < Cell::NumFaces; ++face) {
+      localFluxKrnl.AplusTAll(face) = const_cast<const real**>(
+          entry.get(inner_keys::Wp::Id::LocalIntegrationData)->getDeviceDataPtr());
+      localFluxKrnl.extraOffset_AplusTAll(face) =
+          SEISSOL_ARRAY_OFFSET(LocalIntegrationData, nApNm1, face);
+    }
+    localFluxKrnl.linearAllocator.initialize(tmpMem.get());
+    localFluxKrnl.streamPtr = runtime.stream();
+    localFluxKrnl.execute();
+#endif
```
vikaskurapati, line 311: I am slightly confused about the scope of this `ifdef` here. Is it supposed to be in the for loop for all the star matrices? Please confirm that this is what you intended before merging(preferably with some testing)

src/Kernels/LinearCK/Local.cpp
```diff
@@ -282,12 +285,32 @@ void Local::computeBatchedIntegral(
...
+    localFluxKrnl.numElements = entry.get(inner_keys::Wp::Id::Dofs)->getSize();
+    localFluxKrnl.Q = (entry.get(inner_keys::Wp::Id::Dofs))->getDeviceDataPtr();
+    localFluxKrnl.I =
+        const_cast<const real**>((entry.get(inner_keys::Wp::Id::Idofs))->getDeviceDataPtr());
+
+    for (std::size_t face = 0; face < Cell::NumFaces; ++face) {
+      localFluxKrnl.AplusTAll(face) = const_cast<const real**>(
+          entry.get(inner_keys::Wp::Id::LocalIntegrationData)->getDeviceDataPtr());
+      localFluxKrnl.extraOffset_AplusTAll(face) =
+          SEISSOL_ARRAY_OFFSET(LocalIntegrationData, nApNm1, face);
```
vikaskurapati, line 306: As we are offsetting the pointer here, can we check if `LocalIntegrationData` is consistently padded before this? i.e., it may be useful to assert that this data structure is indeed aligned with this offset.

vikaskurapati (approved):
LGTM apart from minor comments.

src/Kernels/LinearCK/Local.cpp
davschneller: Yes, it is for all the star matrices; similarly to what the `gpu_volume` struct does a bit move above.
Since we now apply all four local flux contributions in the same kernel, we need all four of them at hand.

src/Kernels/LinearCK/Local.cpp
davschneller: Good point—that's still a bit open tbh (normally, the alignment should work—or at least I can't think of a situation where it shouldn't right now; but I agree, it's a bit flaky). I can see if I can add a macro to check that. (yes, unfortunately that one's still a macro right now, since `offsetof` is)
Though, to be sure we're talking about the same thing—the `SEISSOL_[ARRAY_]OFFSET` only concerns the offset inside the `LocalIntegrationData` struct; i.e. external padding will not influence this one.

Mid-term it might be, in fact, best to switch to a byte-level offset (currently, the offset is in `real`); that'll include Yateto changes and the likes.

PR #1488, opened 2026-01-22
Fix some bugs and tests

davschneller:
Fix some segfaults in some configurations; mostly as introduced by #1468 .

* remove struct member alignment for the `Plasticity` data structures (which caused segfaults in some unspecified cases when using GCC)
* fix a segfault when looking up `vNodes` (Yateto currently requires that all calls are inside the storage; so we filter out lookups outside of it)
* fix the material vector not being allocated in the `anisotropic` parameter evaluator by default
* fix an error in the mesh comparison script with the `global-id` ordering; also now complain if the mesh ordered by global IDs doesn't match

### DIFF
--- a/postprocessing/validation/meshcompare.py
+++ b/postprocessing/validation/meshcompare.py
@@ -22,17 +22,25 @@ def compare(file, file_ref, epsilon):
     geom_ref = mesh_ref.ReadGeometry()
     connect_ref = mesh_ref.ReadConnect()
 
+    matching = False
+    global_id_correct = None
+
     # read reference mesh
     if "global-id" in mesh_ref.ReadAvailableDataFields():
         # rely on stably-sorting the refined cells here
         preIds = mesh.Read1dData("global-id", mesh.nElements, isInt=True)
         ids = np.argsort(preIds)
-        preIds_ref = mesh_ref.Read1dData("global-id", mesh.nElements, isInt=True)
-        ids_ref = np.argsort(preIds)
-    else:
+        preIds_ref = mesh_ref.Read1dData("global-id", mesh_ref.nElements, isInt=True)
+        ids_ref = np.argsort(preIds_ref)
+
+        global_id_correct = np.all(np.abs(geom[connect[ids]] - geom_ref[connect_ref[ids_ref]]) < 1e-10)
+        matching = global_id_correct
+        print(f'Global IDs present; conformant: {global_id_correct}')
+
+    if not matching:
         # if the reference solution has no global IDs, heuristically sort by barycenter, weighed by 3D cube position
         print(
-            "No global-id field found in the reference. Order the cells by their barycenter in a 3D array."
+            "No or incorrect global-id field found in the reference. Order the cells by their barycenter in a 3D array."
         )
         rngs = [(np.max(geom[:, i]), np.min(geom[:, i])) for i in range(3)]
         offset = np.array([rng[1] for rng in rngs])
@@ -136,7 +144,15 @@ def compare(file, file_ref, epsilon):
             print(f"{q:3}: {relative_error} [rel.]")
             errors[i] = relative_error
 
+    failure = False
+    if global_id_correct == False:
+        print('Global IDs present, but did not match.')
+        failure = True
+
     if np.any(errors > epsilon):
         print(f"Relative/absolute error {epsilon} exceeded for quantities")
         print([quantity_names[i] for i in np.where(errors > epsilon)[0]])
+        failure = True
+
+    if failure:
         sys.exit(1)


src/Model/Plasticity.h
```diff
@@ -26,9 +26,9 @@ struct PlasticityData {
   static constexpr auto PointCount = tensor::vNodes::Shape[0];
 
   // initial loading (stress tensor)
-  alignas(Alignment) real initialLoading[tensor::initialLoading::size()]{};
-  alignas(Alignment) real cohesionTimesCosAngularFriction[tensor::meanStress::size()]{};
-  alignas(Alignment) real sinAngularFriction[tensor::meanStress::size()]{};
+  real initialLoading[tensor::initialLoading::size()]{};
+  real cohesionTimesCosAngularFriction[tensor::meanStress::size()]{};
+  real sinAngularFriction[tensor::meanStress::size()]{};
```
vikaskurapati, line 31: Why did you remove the alignment here? No performance gain anymore? Or was it crashing due to LIBXSMM? Or was there some other reason?

src/Initializer/ParameterDB.cpp
```diff
@@ -563,6 +565,10 @@ void MaterialParameterDB<AnisotropicMaterial>::evaluateModel(const std::string&
   auto suppliedParameters = model->suppliedParameters();
   // TODO(Sebastian): inhomogeneous materials, where in some parts only mu and lambda are given
   //                  and in other parts the full elastic tensor is given
+  const auto numPoints = query.numPoints();
+  if (m_materials->size() < numPoints) {
+    m_materials->resize(numPoints);
+  }
```
vikaskurapati, line 571: Are we creating extra entries in `m_materials`  by doing this? I don't understand the logic/reasoning behind why we are doing this.

vikaskurapati:
> Fix some segfaults in some configurations.

It would be better to have more details about these configurations, what exactly the problem was, and how they are being fixed.

src/Model/Plasticity.h
davschneller: Somehow that gave me a segfault with GCC (look also at the current tests); and IIRC we don't need the alignment anywhere right now (i.e. it's not used as an `A` or `C` matrix input anywhere).

src/Initializer/ParameterDB.cpp
davschneller: We don't; the reason is that I moved the `m_materials` allocation into that place; but forgot the extra parameter initializer for the anisotropic material.

davschneller:
Ok, done

src/Model/Plasticity.h
vikaskurapati: > Somehow that gave me a segfault with GCC (look also at the current tests); and IIRC we don't need the alignment anywhere right now (i.e., it's not used as an `A` or `C` matrix input anywhere).

I am sorry, I am still confused. I used it with GCC(on Vista, for example), and it did not cause a segfault in these regions. Is it a recent thing? If so, do we know what induced this extra seg fault? 

That is okay that we don't need the Alignment, but just this Alignment causing seg faults is something that I find fascinating.

src/Initializer/ParameterDB.cpp
vikaskurapati: > We don't; the reason is that I moved the `m_materials` allocation into that place; but forgot the extra parameter initializer for the anisotropic material.

I am sorry, which place are you referring to? So this change is not related to a seg fault in the PR? So, was this an issue with anisotropic simulations in the earlier commits?

vikaskurapati:
> Fix some segfaults in some configurations; mostly as introduced by #1468 .
> 
> * remove alignment for the `Plasticity` data structures (which caused segfaults in some cases)
> * fix a segfault when looking up `vNodes` (Yateto currently requires that all calls are inside the storage; so we filter out lookups outside of it)
> * fix the material vector not being allocated in the `anisotropic` parameter evaluator by default
> * fix an error in the mesh comparison script with the `global-id` ordering; also now complain if the mesh ordered by global IDs doesn't match

Sorry to nitpick, but it would be nice to know in which cases a segfault occurred in plasticity. The other details seem enough, imo.

src/Initializer/ParameterDB.cpp
davschneller: Before, the `m_materials` was already allocated in the `InitModel.cpp`; but I've moved it now here in to the `ParameterDB.cpp` (via #1468 ; i.e. only since yesterday).

src/Model/Plasticity.h
davschneller: I mean—in this case it's alignment of member variables (which should be possible as well; but maybe buggy/not that well yet implemented? Or I've mis-understood something; who knows); so I've removed it again here. Also only the case since #1468 , i.e. yesterday (maybe we would've had some bug reports in otherwise already).

davschneller:
I'm sorry; I gotta pass here. :) I only know that they occur(ed) in some tests I've run (and didn't quite check which ones)—removing the alignment for the members in the `struct` seemed to help here.
However, those were only due to #1468; I can assure you that.

davschneller:
If specifying that it happens only with GCC is enough (cf. the CI for tpv13 ... Side note: we might need to re-record tpv13 at some time; the comparison exceeded the current limit already a bit earlier, due to more exact fault output—and the quadpoint-wise initialization (which can be switched off) seems to have tripped all the rest.), then so it is done. I can also reproduce with a newer GCC than in the CI on my local machine.
(the error doesn't happen with Clang or ICX)

vikaskurapati:
Okay, my comment was more about the reasoning behind it, as I found it interesting that just aligning something can cause a segfault in a few cases. But if you say it won't affect it in any way, I am fine with removing it. I just wanted to understand why it was happening. Did you also test the anisotropy case with your changes? I am approving it now because the changes don't seem problematic, but I would request you to do/confirm with an anisotropic case.

davschneller:
I've let run the anisotropy case locally; and could confirm both that the error existed and that this PR solves it.

davschneller:
Found the error why the alignment didn't work—see #1529 for an explanation.

PR #1489, opened 2026-01-22
update lumi install

Thomas-Ulrich:
Update LUMI install documentation to the new software stack.


Thomas-Ulrich:
I've addressed your review (I removed the mpich.pc as I think LUMI have fixed their mpich installation in the meanwhile).
Let me know if we can merge.

davschneller:
Thanks for the changes—I think we can.

PR #1532, opened 2026-03-07
MPI/Filesystem refactor

davschneller:
* add convenience methods for some more MPI calls. Thus, we don't need that many free-standing MPI calls.
* add methods to wait until a folder/file is visible to all ranks (as IIRC there were errors with that happening with many ranks on a large system)
* add a split between output directory and file prefix
* rework the MPI class to just encapsule a communicator with operations on said communicator only (which might differ from `MPI_COMM_WORLD`), and begin moving away from it being a singleton class. In particular, move the "generic" setup and printing methods to different files.


app/Main/Main.cpp
```diff
@@ -60,8 +61,22 @@ std::shared_ptr<YAML::Node> readYamlParams(const std::string& parameterFile) {
...
   try {
 #ifdef ACL_DEVICE
-    seissol::Mpi::mpi.bindAcceleratorDevice();
     device::DeviceInstance& device = device::DeviceInstance::getInstance();
+    /**
+     * @brief Inits Device(s).
+     *
+     * Some MPI implementations create a so-called context between GPUs and OS Processes inside of
+     * MPI_Init(...). It results in allocating some memory buffers in memory attached to the nearest
+     * NUMA domain of a core where a process is running. In case of somebody wants to bind a
```
vikaskurapati, line 70: Grammar? In case 'if'? Or, just write 'if somebody wants to..' or 'in case somebody'....

docs/parameters.par
```diff
@@ -137,7 +137,11 @@ LtsAutoMergeCostBaseline = 'bestWiggleFactor' ! Baseline used for auto merging c
 /
 
 &Output
-OutputFile = '../output_tpv33/tpv33'
+OutputFile = '../output_tpv33/tpv33' ! deprecated. Equals the path "outdir/outname" .
```
vikaskurapati, line 141: Is this planned to be removed forever? If yes, we will need to update all the examples, pre-computed seissol repository, and all the workflows everyone is using. 

If there is no very particular reason for doing this, I would suggest to provide both the options. i.e., your outdir, outname, etc. options, and this one too. 

Also, please add to the comment that it would be ignored if the parameters underneath are given.

src/Parallel/ParallelFS.cpp
vikaskurapati, line 1: FS for filesystem?

src/Parallel/ParallelFS.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+void waitExistence(const filesystem::path& path) {
+  std::size_t counter = 0;
+  auto entry = filesystem::directory_entry(path);
+  while (!entryExistsGlobally(entry)) {
+    using namespace std::chrono_literals;
+    std::this_thread::sleep_for(10ms);
+    entry.refresh();
+    ++counter;
+    if (counter % 10 == 0) {
```
vikaskurapati: Would it make sense to do this only for DEBUG builds? Is this something you anticipate to happen often?

Mainly in clusters with slower filesystems like Frontera, may be?

src/Parallel/SystemInfo.cpp
```diff
@@ -0,0 +1,60 @@
...
+  hostNames_ = Mpi::mpi.collectContainer(hostName);
+
+  logInfo() << "Running on (rank=0):" << hostNames_.front();
+}
+
+void SystemInfo::loadCPUInfo() {
+  // do nothing. Yet.
+}
+
+void SystemInfo::loadGPUInfo() {
```
vikaskurapati, line 42: Is this not printing GPU Info more than loading it?

src/Parallel/SystemInfo.cpp
```diff
@@ -0,0 +1,60 @@
...
+    utils::StringUtils::rtrim(hostName);
+    hostName.pop_back();
+  }
+  hostNames_ = Mpi::mpi.collectContainer(hostName);
+
+  logInfo() << "Running on (rank=0):" << hostNames_.front();
+}
+
+void SystemInfo::loadCPUInfo() {
+  // do nothing. Yet.
```
vikaskurapati: If this is not doing anything yet, can we remove it? What is the purpose of it?

src/Parallel/SystemInfo.h
```diff
@@ -0,0 +1,44 @@
...
+   * @return hostnames for all ranks in the communicator of the application
+   */
+  [[nodiscard]] const auto& getHostNames() const { return hostNames_; }
+
+  [[nodiscard]] const auto& getPCIAddresses() const { return pcis_; }
+
+  private:
+  void loadHostInfo();
+  void loadCPUInfo();
+  void loadGPUInfo();
```
vikaskurapati, line 40: Docstrings for these new functions will be nice. 

What is the intended difference between HostInfo and CPUInfo is something I would not understand just by looking at the function signatures.

vikaskurapati (commented):
Minor comments about the code. I will test it next. There are a few comments from copilot which may be worth addressing

vikaskurapati:
The upcoming comments are from Gemini -- if something is wrong, you could ignore them.

src/Parallel/MPI.h
```diff
@@ -223,6 +297,49 @@ class Mpi : public MpiBasic {
     MPI_Bcast(const_cast<InternalType*>(container.data()), size, mpiType, root, comm.value());
   }
 
+  template <typename T>
+  T allreduce(const T& value, MPI_Op op, std::optional<MPI_Comm> comm = {}) const {
```
vikaskurapati, line 301: Gemini comment: Performance: Using std::optional<MPI_Comm> and checking .has_value() inside frequently called reduction wrappers adds a small but non-zero overhead in the hot path of the solver. Given this is a high-performance simulation code, consider overloading the function instead of using std::optional.

But given that these are called only once towards the end of correction kernels, I think they should not be that big an overhead, no?

src/Parallel/MPI.h
```diff
@@ -97,6 +151,26 @@ class Mpi : public MpiBasic {
     }
   }
 
+  template <typename T>
+  void registerMpiType(MPI_Datatype type, bool commit = false) {
```
vikaskurapati, line 155: Safety: If commit is true, the class takes "ownership" of the type's lifecycle. However, if the user passes a predefined type (like MPI_DOUBLE), calling MPI_Type_free on it during finalize() (Line 368) is undefined behavior or an error in many MPI implementations. Add a check to ensure we only free user-defined types.

src/Parallel/ParallelFS.cpp
```diff
@@ -0,0 +1,83 @@
...
+  const auto total = Mpi::mpi.allreduce(local, MPI_MIN);
+  return total == 1;
+}
+
+void waitExistence(const filesystem::path& path) {
+  std::size_t counter = 0;
+  auto entry = filesystem::directory_entry(path);
+  while (!entryExistsGlobally(entry)) {
+    using namespace std::chrono_literals;
+    std::this_thread::sleep_for(10ms);
```
vikaskurapati, line 37: Risk: This is a "busy-wait" across the network. While 10ms is reasonable, on large-scale parallel filesystems (Lustre/GPFS), file metadata propagation can be inconsistent. If one rank sees the file and another doesn't, this loop could hang indefinitely if the underlying filesystem has a stale cache. A barrier is safer than a sleep-loop for synchronizing FS state.

src/Parallel/ParallelFS.cpp
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+void moveEntryGlobally(const filesystem::path& entry, const filesystem::path& newentry) {
+  logInfo() << "Rename" << entry << "to" << newentry;
+  if (Mpi::mpi.rank() == 0) {
+    filesystem::rename(entry, newentry);
+  }
+
+  waitExistence(newentry);
```
vikaskurapati, line 52: Race Condition: On some systems, rename is not atomic across all nodes immediately. If Rank 0 renames a file and Rank 1 immediately calls waitExistence, the metadata might not have hit the MDS (Metadata Server) yet. You should include an MPI_Barrier before and after filesystem operations to ensure Rank 0 has finished the syscall before others start looking for the result.

src/Parallel/SystemInfo.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+void SystemInfo::loadHostInfo() {
+  std::string hostName(256, ' ');
+  if (gethostname(hostName.data(), 256) != 0) {
+    hostName = "unknown-host";
+  } else {
+    utils::StringUtils::rtrim(hostName);
+    hostName.pop_back();
+  }
+  hostNames_ = Mpi::mpi.collectContainer(hostName);
```
vikaskurapati, line 37: Is only rank 0 calling this, or all ranks?

vikaskurapati:
I ran a test with Ridgecrest, and I can confirm that nothing of importance seems to break or diverge. We can merge this once the comments are addressed.

davschneller:
Thanks for the review; I've not yet addressed absolutely everything—I'll do the rest soon.

src/Parallel/SystemInfo.cpp
davschneller: Every rank needs to call it—though the result is only returned to rank 0 (by `collectContainer`)

src/Parallel/ParallelFS.cpp
davschneller: Question back to the LLM: filesystem-level barrier or MPI barrier? (the latter should already be triggered due to the allreduce maybe)

src/Parallel/MPI.h
davschneller: That's right; it shouldn't matter. That said—the more "proper" solution mid-term would be to eliminate that argument altogether, by having separate MPI wrappers per communicator.

src/Parallel/SystemInfo.h
davschneller: Ok, done. The info print would be:

* HostInfo: prints info about the OS, setup etc.
* CPUInfo: should have printed the CPU model etc. in symmetry to the GPU info printing. But NYI.

src/Parallel/SystemInfo.cpp
davschneller: Removed it for now; for the purpose see the comment below.

src/Parallel/SystemInfo.cpp
davschneller: It is kind of also loading the info; since it's collecting the PCIe addresses over all ranks (for debugging/checking pinning purposes)

src/Parallel/ParallelFS.cpp
davschneller: We should be good. I've lowered the frequency of the prints to 1s; and the prints should only happen from rank 0 (that's configured inside the logging module; seems to be opaque for the LLMs).

src/Parallel/ParallelFS.cpp
davschneller: Yes.

src/SeisSol.h
```diff
@@ -208,6 +213,8 @@ class SeisSol {
 
   utils::Env& env() { return env_; }
 
+  const SystemInfo& systemInfo() { return systemInfo_; }
+
```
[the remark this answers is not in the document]
davschneller: Done

src/Parallel/SystemInfo.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+  logInfo() << "Running on (rank=0):" << hostNames_.front();
+}
```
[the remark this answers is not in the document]
davschneller: Not needed; that's implied.

src/Initializer/Parameters/OutputParameters.cpp
```diff
@@ -269,8 +270,25 @@ OutputParameters readOutputParameters(ParameterReader* baseReader) {
...
+
+  std::string outputDir;
+  std::string outputName;
+
+  if (outputDirR.has_value() && outputNameR.has_value()) {
+    outputDir = outputDirR.value();
+    outputName = outputNameR.value();
+  } else {
+    const auto prefix = reader->readOrFail<std::string>(
+        "outputfile", "Output directory, output name, nor output file prefix not defined.");
```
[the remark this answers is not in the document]
davschneller: Done.

src/Parallel/ParallelFS.cpp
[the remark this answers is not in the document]
davschneller: Not needed; implied.

src/Parallel/ParallelFS.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  // file should already exist for a much longer time
+  if (entryExistsGlobally(entry)) {
+    const auto actualTimeStamp = timeStamp.value_or(
+        []() { return utils::TimeUtils::timeAsString("%Y-%m-%dT%H-%M-%S", time(nullptr)); }());
+    std::stringstream backupFileName;
+    backupFileName << fileName << ".bak_" << actualTimeStamp << '.' << fileExtension;
+    const auto copyPath = seissol::filesystem::path(backupFileName.str());
+    moveEntryGlobally(seissol::filesystem::directory_entry(path),
+                      seissol::filesystem::directory_entry(copyPath));
```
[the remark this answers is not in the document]
davschneller: Done.

src/Parallel/Pin.cpp
```diff
@@ -339,12 +339,7 @@ CpuMask Pinning::getNodeMask() {
     workerMaskArray[cpu] = CPU_ISSET(cpu, &workerMask);
   }
 
-  MPI_Allreduce(MPI_IN_PLACE,
-                workerMaskArray.data(),
-                workerMaskArray.size(),
-                MPI_CHAR,
-                MPI_BOR,
-                Mpi::mpi.sharedMemComm());
+  Mpi::mpi.allreduceContainer(workerMaskArray, MPI_BOR);
```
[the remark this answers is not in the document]
davschneller: Done.

src/Parallel/ParallelFS.cpp
vikaskurapati: Gemini's response: 

The allreduce inside entryExistsGlobally provides the necessary MPI synchronization, but the loop is still required to handle the asynchronous nature of distributed metadata. An MPI_Barrier alone is insufficient because it doesn't force a cache refresh on the non-root nodes. The current 'busy-wait' is the standard (albeit slightly ugly) way to implement a global filesystem sync in HPC.

src/Parallel/SystemInfo.cpp
vikaskurapati: Okay, in that case, this was it's comment:

" Memory/Scalability: On 100,000+ ranks, every single rank will now store a vector of 100,000 strings. This is a significant memory footprint (~25-50 MB per rank just for hostnames). You should only collect these on Rank 0 or store them more efficiently (e.g., a map of unique hostnames to rank ranges). "

But given that we don't go to so many ranks yet, we could safely ignore the comment, imo. What do you think?

vikaskurapati:
> Thanks for the review; I've not yet addressed absolutely everything—I'll do the rest soon.

OK. Please send me a reminder when done, so that I can check again and run a simulation again.

docs/parameters.par
davschneller: As long as we keep the FORTRAN parameter file format, it'll stay—that's just our current unwritten quasi-policy more or less. Everything that was once supported (since probably about the introduction of easi) should still work out of the box. More or less, at least I think. (e.g. no Netcdf meshes)

I'll add the comment to the options below.

At some point, we might wanna switch to a new parameter format—then I'd remove the `outputfile` option entirely—but we should/will provide an update script that's being checked by the CI *all the time* and e.g. splits that parameter into the two new ones. I.e., in general: we update the parameters in SeisSol -> the modernization script will also need to update to the current standard.
That way, we can move out all the parameter reading logic from the C++ core except the very basic one and add new structures and features (e.g. more scripting, less hard-coded ICs (but scripting), more output writers of one kind with different intervals, different initial data for different simulations that run at the same time etc.) without the backlog of old ones. And also more easily support pre-easi configurations if ever needed. Plus, also auto-convert every easi script to some new scripting format (e.g. currently Lua) automatically; so you also don't need to install that one anymore.

app/Main/Main.cpp
davschneller: Guess so. :) (I just copied that comment from where it was before—IIRC written sometime earlier 2023 not by me)

PR #1538, opened 2026-03-13
Add some (mostly "vibe-coded") unit tests

davschneller:
* Add simple unit tests to several areas of SeisSol
* Add first unit tests with MPI
* Fix some bugs in relation to those tests (mainly ports from #1180)
* Use mostly `CHECK` instead of `REQUIRE` (i.e. don't fail and abort the unit tests in most cases; only fail and continue)

Seems like currently the coverage is up by +8% for both functions and code lines.

Disclaimer: most of the unit tests were generated using AI (despite some a bit "nonsensical" tests and generally a "random" distribution, I'm positively surprised about their quality). The model used is Claude Opus 4.6.


davschneller:
Can be looked at. In fact, it might be best to also trigger the Copilot review here for certainty.

PR #1541, opened 2026-03-18
A bit more cleanup; make `INTEGRATE_QUANTITIES` work once more

davschneller:
* add literals for sized ints and `size_t` (backport from C++23)
* remove some unused kernels
* refactor the point-to-cell mapping
* re-enable `INTEGRATE_QUANTITIES`, add it to the default build, and make it GPU-ready (add a `SimulationSettings` class to encapsule settings like "is plasticity on" or "do we need to integrate over the DOFs"). Also it now supports the high-order output in space. However it still uses a lower-order time integration scheme internally. ... The parameter is called `integrationMask` in the `output` section of the parameter file—it operates like `ioutputmask`.


docs/wave-field-output.rst
```diff
@@ -48,19 +48,15 @@ The 6 first digits controls the components of the off-fault plastic
...
-    (int_sigma_xx, int_sigma_yy, int_sigma_zz, int_sigma_xy, int_sigma_yz, and int_sigma_xz),
-    and the 3 last digits the displacement components (displacement_x, displacement_y, displacement_z).
-    Note that this output is associated with the prefix-low.xdmf file, and can only output
-    the cell average quantities.
+IntegrationMask
+---------------
+
+IntegrationMask allows switching on and off the writing of time integrated SeisSol unknowns.
+The 6 first digits control the components of the time integrated stress tensor
+(int_sigma_xx, int_sigma_yy, int_sigma_zz, int_sigma_xy, int_sigma_yz, and int_sigma_xz),
```
vikaskurapati, line 56: It is okay to have them, but I am unaware of where we would use integrated stress quantities. Interesting, though.

vikaskurapati (commented):
LGTM code-wise. I will need to build a simulation case to test if the integration is done correctly. I will let you know when I am done with that

vikaskurapati (approved):
Checked `INTEGRATE_QUANTITIES` with a small coarse planar wave. Seems to be doing fine.

PR #1546, opened 2026-03-25
More more former parameters as "deprecated" instead of "unknown"

davschneller:
Cf. the warnings thrown by the scenario used in #1539 .


vikaskurapati (approved):
LGTM

PR #1555, opened 2026-04-09
Output a completion percentage value (w.r.t. the total simulation time) on sync point

davschneller:
Cf. title. Just a little QoL improvement.


vikaskurapati (approved):

Thomas-Ulrich (approved):
Nice. What about printing an "estimated run time remaining" too?
(might be that I never pressed submit review)

davschneller:
That might be something useful; yes. ... Since we're now already merging, let me file a new PR for it

PR #1557, opened 2026-04-14
Refactor the face type input

davschneller:
* Unify face types `Regular` (0) and `Periodic` (6) internally (they're treated same way now)
* Classify the faces a bit better, and generalize the classification a bit. We now have internal and external faces; with external faces currently being factored into some with a fake neighbor (i.e. the same cell—that's only required if they call the neighboring kernel; which none of our current BCs actually does) and some with no neighbor. Also, there are currently no fake neighbor BCs. I've left it in for now—but we can also remove the bits of code for that case (mostly concerning the `LtsSetup.cpp` and `Buckets.cpp`).
* Clean up the face type usage internally; especially for viscoelastic (also fix a flop counting issue there)
* Add external face type/tag input. Allows a YAML dictionary for input; including open ranges (not that YAML-y, but it gives us backward compatibility to the current assignment). With maybe the whole input files being converted to YAML at some point in the future.
* Reduce data transfers for the boundary conditions—i.e. make all boundary conditions behave like outflow everywhere. That the BCs effectively behave like outflow in the kernels (but not the datastructures) was already so before even v1.0.0 (e.g. the visco2 kernels still had remnants of code in there that the `FreeSurface` BC should be run in the `Neighbor.cpp` kernel; but that's long been changed to the `Local.cpp` for elastic _and_ visco2); so we're now just concluding the changes and cleaning up.
* includes unit tests for the new features

### DIFF
--- a/docs/puml-mesh-format.rst
+++ b/docs/puml-mesh-format.rst
@@ -115,7 +115,7 @@ The following convention for defining a face ID is used:
 
 You may supply a separate face type configuration as a YAML file. The file consists of a sole YAML dictionary
 with the tags ``regular``, ``freeSurface``, ``freeSurfaceGravity``, ``dynamicRupture``, ``dirichlet``,
-``outflow``, or ``analytic``.
+``outflow``, or ``analytical``.
 
 Each entry has a list with the boundary condition IDs / face tags that are associated to it.
 You may also add a range of the format e.g. ``10-23`` or ``10-`` if you want all larger tags
@@ -139,7 +139,7 @@ As an example, the following would replicate the default SeisSol behavior.
       - 4
    outflow:
       - 5
-   analytic:
+   analytical:
       - 7
 
 Topological Connectivity (e.g. for periodic or stitched domains)
--- a/docs/puml-mesh-format.rst
+++ b/docs/puml-mesh-format.rst
@@ -101,7 +101,7 @@ SeisSol indexes the boundary conditions as follows:
 
 - 5: absorbing. Boundary condition; no neighbor.
 
-- 6: identified. Similar to a regular face (0) between two cells. Formerly known as "periodic".
+- 6: regular. Like face type 0, a regular face between two cells. Formerly known as "periodic" or "identified".
 
 - 7: analytical. Boundary condition given by
 
@@ -112,6 +112,35 @@ The following convention for defining a face ID is used:
 .. code-block:: python
 
    s_vert[0,:] = [0,2,1];   s_vert[1,:] = [0,1,3];    s_vert[2,:] = [1,2,3]; s_vert[3,:] = [0,3,2];
+
+You may supply a separate face type configuration as a YAML file. The file consists of a sole YAML dictionary
+with the tags ``regular``, ``freeSurface``, ``freeSurfaceGravity``, ``dynamicRupture``, ``dirichlet``,
+``outflow``, or ``analytic``.
+
+Each entry has a list with the boundary condition IDs / face tags that are associated to it.
+You may also add a range of the format e.g. ``10-23`` or ``10-`` if you want all larger tags
+to be matched to this boundary condition. Ranges are inclusive on both sides.
+
+As an example, the following would replicate the default SeisSol behavior.
+
+.. code-block:: yaml
+
+   regular:
+      - 0
+      - 6
+   freeSurface:
+      - 1
+   freeSurfaceGravity:
+      - 2
+   dynamicRupture:
+      - 3
+      - 65- # note the open range
+   dirichlet:
+      - 4
+   outflow:
+      - 5
+   analytic:
+      - 7
 
 Topological Connectivity (e.g. for periodic or stitched domains)
 ----------------------------------------------------------------
--- a/src/Common/SegmentMap.h
+++ b/src/Common/SegmentMap.h
@@ -1,0 +1,105 @@
+// SPDX-FileCopyrightText: 2026 SeisSol Group
+//
+// SPDX-License-Identifier: BSD-3-Clause
+// SPDX-LicenseComments: Full text under /LICENSE and /LICENSES/
+//
+// SPDX-FileContributor: Author lists in /AUTHORS and /CITATION.cff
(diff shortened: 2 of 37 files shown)


docs/puml-mesh-format.rst
```diff
@@ -101,7 +101,7 @@ SeisSol indexes the boundary conditions as follows:
 
 - 5: absorbing. Boundary condition; no neighbor.
 
-- 6: identified. Similar to a regular face (0) between two cells. Formerly known as "periodic".
+- 6: regular. Like face type 0, a regular face between two cells. Formerly known as "periodic" or "identified".
```
vikaskurapati, line 104: I am confused. How does one mark in the tags that a face is periodic?

src/Common/SegmentMap.h
```diff
@@ -0,0 +1,105 @@
...
+    const auto upper = ranges_.upper_bound(index);
+    const auto lower = ranges_.lower_bound(index);
+
+    if (exact != ranges_.end()) {
+      return exact->second;
+    } else if (upper->second == lower->second) {
+      return lower->second;
+    } else {
+      // error (we should not end up here)
+      logError() << "Map error: internal issue.";
```
vikaskurapati, line 95: Do we know the condition under which we will end up here? If yes, can we be a bit verbose in that scenario?

src/Geometry/CubeGenerator.cpp
```diff
@@ -588,20 +594,20 @@ void CubeGenerator::cubeGenerator(const std::array<std::size_t, 4> numCubes,
...
             //                                                                            * 20 + 11]
             //                                                                            =
-            //                                                                            boundaryMaxz;
+            //                                                                            faceMap.at(boundaryMaxz).value();
             //                                                                            elemBoundaries[(((numCubesPerPart[2]-1)*numCubesPerPart[1]+yy)*numCubesPerPart[0]+xx)
             //                                                                            * 20 + 15]
             //                                                                            =
-            //                                                                            boundaryMaxz;
+            //                                                                            faceMap.at(boundaryMaxz).value();
             //                                                                    }
```
vikaskurapati, line 611: It may be a good idea to remove these comments if they don't serve a purpose.

src/Initializer/BasicTypedefs.h
```diff
@@ -13,50 +13,75 @@ namespace seissol {
...
   Dirichlet = 4,
 
-  // absorbing/outflow boundary
+  // absorbing/outflow boundary (boundary, linear)
   Outflow = 5,
 
-  // periodic boundary
-  Periodic = 6,
+  // periodic boundary (obsolete; now equivalent to Regular)
+  // Periodic = 6,
```
vikaskurapati, line 43: I might have missed something, but I am unable to understand how periodic meshes are supposed to work now..

docs/puml-mesh-format.rst
davschneller: Right now, you mark the nodes of that face to be identical (but position) to the nodes of another cell, cf. https://seissol.readthedocs.io/en/latest/puml-mesh-format.html#topological-connectivity-e-g-for-periodic-or-stitched-domains or the GMSH documentation for a really similar approach.
The advantage is that the periodicity is already visible in the mesh partitioning.

The `periodic` face has had no real or special meaning in the mainline code before to my knowledge (is was the same as `regular` in all cases). Maybe at some ancient point there was something—but at least not since v1.0.0.

vikaskurapati (commented):
LGTM mostly. I just want to test a case with periodic BCs before we merge this. Do you have an example or a test I could run somewhere with periodic BCs? Something without using the cube generator, ideally, as it is planned to be removed soon-ish?

@fabian-kutschera, could you please test an elastic-acoustic scenario with this branch if you have time to ensure nothing breaks?

src/Common/SegmentMap.h
davschneller: That's if a range is "half complete". I.e. right now the map basically is a contiguous list of ranges. E.g.

0 - 3: none

4 - 10: 4

11-12: none

13-infty: 10

Then we mark only the boundaries—i.e. 0, 3, 11, 12 with none; 4 and 10 with 4, and 13 and infty (or int max) with 10; those values in between do not get added to the internal map.
Still, and that's what shouldn't happen here, is that we query some interior of a range (e.g. 5,6,7,8, or 9), and their lower and upper boundary. Those now need to match (in the example, both would be 4). If they didn't, we had some internal setup problems.

davschneller:
For the periodic BCs, you can check out the pumlcube tool in the Meshing repository. It'll generate you some with boundary condition 6 (because that is hard-coded for there to generate you periodicity).

Or, though I'm not sure if that's 100%ly stable, create a GMSH file with periodicity and use pumgen.

docs/puml-mesh-format.rst
vikaskurapati: OK, thanks! I have not yet generated periodic meshes this way. I used the cube generator for my small convergence tests. I will try to do one, and see if it works the way we intend to.

vikaskurapati:
> For the periodic BCs, you can check out the pumlcube tool in the Meshing repository. It'll generate you some with boundary condition 6 (because that is hard-coded for there to generate you periodicity).
> 
> Or, though I'm not sure if that's 100%ly stable, create a GMSH file with periodicity and use pumgen.

The latter sounds like a more viable option to do. Let me see if I can do something like without big troubles. If it works, it is probably an Example to be added.

vikaskurapati:
> > For the periodic BCs, you can check out the pumlcube tool in the Meshing repository. It'll generate you some with boundary condition 6 (because that is hard-coded for there to generate you periodicity).
> > Or, though I'm not sure if that's 100%ly stable, create a GMSH file with periodicity and use pumgen.
> 
> The latter sounds like a more viable option to do. Let me see if I can do something like without big troubles. If it works, it is probably an Example to be added.

Tested with a traveling wave. Even though the simulation was losing energy, I suspect it is not due to periodicity, but how traveling wave initial conditions are imposed. So, LGTM.

vikaskurapati (approved):

Issue #1553, opened 2026-04-08
SeisSol Order 3 fails with a seg fault with the Ridgecrest simulation


vikaskurapati:
### Describe the bug

Order 3 binary crashes with a seg fault, but order 4 binary starts the simulation. The error is:

2026-04-08 10:34:39.487 info 0 : Start simulation epoch. (from 0 s to 20 s) 
srun: error: i01r01c04s05: task 6: Segmentation fault (core dumped)
srun: Terminating StepId=5528235.0
srun: error: i01r01c04s11: task 12: Segmentation fault (core dumped)
srun: error: i01r01c04s06: task 7: Segmentation fault (core dumped)
srun: error: i01r01c03s11: task 0: Segmentation fault (core dumped)
srun: error: i01r01c04s12: task 13: Segmentation fault (core dumped)
[2026-04-08T10:34:40.411] error: --task-epilog failed status=15
[2026-04-08T10:34:39.473] error: --task-epilog failed status=15
[2026-04-08T10:34:40.412] error: --task-epilog failed status=15
[2026-04-08T10:34:40.068] error: --task-epilog failed status=15
[2026-04-08T10:34:39.683] error: --task-epilog failed status=15
[2026-04-08T10:34:39.683] error: --task-epilog failed status=15
srun: error: i01r01c03s12: task 1: Segmentation fault (core dumped)
srun: error: i01r01c04s01: task 2: Segmentation fault (core dumped)
[2026-04-08T10:34:40.413] error: --task-epilog failed status=15
srun: error: i01r01c04s04: task 5: Segmentation fault (core dumped)
srun: error: i01r01c04s02: task 3: Segmentation fault (core dumped)
srun: error: i01r01c04s10: task 11: Segmentation fault (core dumped)
srun: error: i01r01c04s07: task 8: Segmentation fault (core dumped)
[2026-04-08T10:34:39.476] error: --task-epilog failed status=15
[2026-04-08T10:34:40.071] error: --task-epilog failed status=15
[2026-04-08T10:34:39.476] error: --task-epilog failed status=15
srun: error: i01r01c04s08: task 9: Segmentation fault (core dumped)
srun: error: i01r01c04s03: task 4: Segmentation fault (core dumped)
srun: error: i01r01c05s02: task 15: Segmentation fault (core dumped)
srun: error: i01r01c04s09: task 10: Segmentation fault (core dumped)
srun: error: i01r01c05s01: task 14: Segmentation fault (core dumped)
[...]

PR #1558, opened 2026-04-15
Fix Alignment for Fast Velocity Weakening Law.

vikaskurapati:
Solves #1553 . 

Checked on SuperMUC.


davschneller (approved):

davschneller:
Probably the first one wouldn't even need the alignment (the `B` matrix right now basically never needs it; as it's broadcasted); but it doesn't hurt to have it anyways.

Issue #1559, opened 2026-04-16
significant slow-down on LUMI-G when using (a few) on- and off-fault receivers


Thomas-Ulrich:
### Describe the bug

I'm investigating why a given setup was slower with the same mesh but different parameter on LUMI-G.
For the record, I need ~60s for simulating 0.5s in the slow setup (B, based on FL16 and with more outputs), while I need only 20s in the fast setup (A, based on FL34).
First I thought it is related to the friction law. But it turned out that if I switch FL to 16 in A, it remains fast.

Then I compared each setup, and realized that adding 4 receivers in A, with:

```
ReceiverOutput = 1
RFileName = 'tmp/seissol_station_5.txt'
pickdt = 0.05
ReceiverOutputInterval = 5
```
leads to an increase from ~20s to 35s for simulating 0.5s.

Switching from energy output every 0.5s to every 0.25s further increase to 40s (not related to volume output).

fast run: [17560596.fl33.fast.out.txt](https://github.com/user-attachments/files/26779815/17560596.fl33.fast.out.txt)

slower run (energy output every 0.25s, and 4 receivers added): [17562418.fl33.slow.out.txt](https://github.com/user-attachments/files/26779816/17562418.fl33.slow.out.txt)

EDIT: adding the on-fault receiver yield the final 10s increase.
```
! parameterize ascii fault file outputs
&Pickpoint
printtimeinterval = 30       ! Index of printed info at timesteps
OutputMask = 1 1 1 1 1 1 1 1 1 1 1 1  ! turn on and off fault outputs
PPFileName = 'tmp/fault_receiver.txt'
outputinterval = 1.0 ! Write-out interval for the on-fault receivers.
/
```

I also try updating the stripe on the newly created folder with
```
lfs setstripe -E 1M -L mdt -E -1 -c 8 output_fl33_test/
```
but this did not help.

### To Reproduce

1. Which version do you use? Provide master efb9a7c5d29113dd5d06dd52c9ec6c3ded957090
2. Which build settings do you use? Which compiler version do you use?
as in https://seissol.readthedocs.io/en/latest/lumi.html
SeisSol_Release_sgfx90a_hip_5_elastic
3. On which machine does your problem occur? If on a cluster: Which modules are loaded?
LUMI-G
```
(local) ulrichth@uan02:/scratch/project_465002391/ulrich/2025-03-28_Mw7.7_Burma_Myanmar_Earthq_us7000pn9s_1_m11_large> module list

[...]

PR #1560, opened 2026-04-16
Avoid host function calls with HIP

davschneller:
Normally, the receiver (both on and off fault) computation should be offloaded to the CPU—that's currently done via extra streams which launch CPU tasks.

However, there seems to be a problem with at least the HIP internal implementation or SeisSol—so let's disable the host function calls for HIP until further evaluation.

Might help with #1559

### DIFF
--- a/src/Parallel/Runtime/Stream.h
+++ b/src/Parallel/Runtime/Stream.h
@@ -106,7 +106,13 @@ class StreamRuntime {
 
   template <typename F>
   void enqueueHost(F&& handler) {
-    device().api->streamHostFunction(streamPtr->get(), std::forward<F>(handler));
+    if (Backend != DeviceBackend::Hip) {
+      device().api->streamHostFunction(streamPtr->get(), std::forward<F>(handler));
+    } else {
+      // if the stream host function call isn't implemented or slow, we'll need to synchronize
+      wait();
+      std::invoke(handler);
+    }
   }
 
   template <typename F>


Thomas-Ulrich:
it seems the code is then stuck at:
```
2026-04-16 19:01:37.488 info 0 : Writing energy output at time 0 
2026-04-16 19:01:37.932 info 0 :    Total momentum (X, Y, Z): 0.000000e+00 Ns , 0.000000e+00 Ns , 0.000000e+00 Ns 
2026-04-16 19:01:37.932 info 0 : Writing energy output at time 0 Done. 
2026-04-16 19:01:37.933 info 0 : Time spent for initial IO: 8.7380 s (per rank: (8.7380e+00 ± 4.1075e-05) s ; range: [ 8.7380e+00 s , 8.7381e+00 s ]) 
2026-04-16 19:01:37.933 info 0 :  
2026-04-16 19:01:37.933 info 0 : Start simulation epoch. (from 0 s to 0.5 s) 
```

(here for more than 5min).
tested twice.

davschneller:
Ok; I'll implement/try some other approach.

davschneller:
You can try again—I've switched this branch to a different strategy (hopefully this one works better...). Otherwise, I'll need to look into a more serious rewrite of how the receivers are handled right now.

Thomas-Ulrich:
now works https://github.com/SeisSol/SeisSol/issues/1559#issuecomment-4279110970, thanks!

Thomas-Ulrich (approved):

PR #1561, opened 2026-04-20
Remove the cube generator (for now)

davschneller:
Remove the cube generator code.
Alas the code is not in a good (code quality) shape as it's built from the old Netcdf reader and `cube_c`; and generally a bit hard to maintain, also it only works on a single rank. In short: it'll need a full rewrite.

We can either keep the code until something better comes along (and bother with maintaining it), or we can just let the users rely on external meshes in all cases for the time being—as what the cube generator does can also be replicated by external tools; plus, we need a parameter input file in any way (i.e. there _will_ be input files, so one small additional mesh file shouldn't hurt). Any opinions?

I'm not sure if it's good to fully remove the whole interface (being able to use "ad hoc" meshes might be a nice feature; especially for testing); however it's thinkable—then we could also remove the parameter section and the respective docs.
Since we have something like `pumlcube` now already, maybe in the future we can again offer direct input; but via PUML directly.

Besides, we remove a mesh comparison program (written in C++), as that's been replaced by the python mesh comparison scripts by now.


vikaskurapati:
> Remove the cube generator code. Alas the code is not in a good (code quality) shape as it's built from the old Netcdf reader and `cube_c`; and generally a bit hard to maintain, also it only works on a single rank. In short: it'll need a full rewrite.
> 

I don't mind this, tbh. I don't know of anyone who is actively using this. 

> We can either keep the code until something better comes along (and bother with maintaining it), or we can just let the users rely on external meshes in all cases for the time being—as what the cube generator does can also be replicated by external tools; plus, we need a parameter input file in any way (i.e. there _will_ be input files, so one small additional mesh file shouldn't hurt). Any opinions?
> 

I think I got too detached from this. The only thing that is in my mind is how would we run periodic meshes now. That is the one example we can provide while removing this. 

> I'm not sure if it's good to fully remove the whole interface (being able to use "ad hoc" meshes might be a nice feature; especially for testing); however it's thinkable—then we could also remove the parameter section and the respective docs. Since we have something like `pumlcube` now already, maybe in the future we can again offer direct input; but via PUML directly.
> 
> Besides, we remove a mesh comparison program (written in C++), as that's been replaced by the python mesh comparison scripts by now.

I think it is better for @Thomas-Ulrich to comment on this from a long-term perspective, then merge it.

Thomas-Ulrich (approved):
LGTM.

PR #1531, opened 2026-02-28
Allow more outer RSF iterations

davschneller:
* currently the outer RSF iterations were fixed to two; but it seemed to have some effect to sometimes run it more. (inner RSF iterations: Newton-Raphson for the slip rate—but fixed state and stress; outer RSF: varying state and stress, but regularized fixed-point iteration for the slip rate ... Combining the two iterations might be possible if we also add a state derivative; and also a derivative for the whole TP procedure ... Maybe it is worth it, but maybe not; I can't really tell yet) The increase in runtime seems to be pretty negligible; or sometimes there might even be a slight decrease when we only need one iteration
* re-wire the non-convergence variables for RSF. Currently, the non-convergence is checked similarly as for inf/nan; i.e. if we encounter it during evaluation.
* slightly rework the DR GPU output copies
* clearly separate the DR parameters as read in the parameter file, and the friction law constants, as used in the computation (difference: the former allows more types; could be serialized some time in the future; and is used in more places—whereas the latter is only used in the actual computation)
* allow adjusting the DR RSF loop parameters from the parameter file

PR #1563, opened 2026-04-21
Fix a computation "swap" in the CPU DR

davschneller:
Fix the order between computing the state variable and computing helper variables for `mu` at the end. Only affects/affected the CPU DR.

(should be also quasi-obsolete via #1531 )


vikaskurapati:
image

The results are not changing much with this change, and the misfits relevant to UQ studies change by only the 4th significant digit.

vikaskurapati (approved):

PR #1564, opened 2026-04-21
Accept MEMORY_LAYOUT filename or rewritten absolute path; add fallback test

vikaskurapati:
We add handling of `MEMORY_LAYOUT` a bit more automatic instead of always needing to use an absolute path when we use a different layout than prescribed like `dense.xml`. 

EDIT: partially vibe-coded to get the CMake working properly.


davschneller:
As a question; you could already give e.g. `cpu/dense.xml` if you wanted right now (originally, it was just `dense.xml` ; but then the split to `cpu` and `gpu` happened). ... So e.g. what works more now?

vikaskurapati:
> As a question; you could already give e.g. `cpu/dense.xml` if you wanted right now (originally, it was just `dense.xml` ; but then the split to `cpu` and `gpu` happened). ... So e.g. what works more now?

Just `dense.xml`, and it picks cpu or gpu automatically depending on the target. If not useful, we can close it. :smile:

davschneller:
Hmm yes, that is useful; thanks for changing that. Let me walk over the code and give you feedback.

Though ad hoc one first ad hoc comment: ideally denote for any commit that you have what AI tools you've used (if any). We don't really have a strict policy yet; but it might be good to be able to keep track where and what models influenced us. Or at least, that's what I've been doing so far (cf. #1538 ).

And another thought on the test infrastructure (not necessarily required for this PR): the CMake test looks good; though maybe the codegen could at some point use a pytest infrastructure (and probably Yateto as well).

codegen/generate.py
```diff
@@ -192,15 +192,38 @@ def generate_equation(subfolders, equation, order):
...
+            config_dir = os.path.join(script_dir, "config", subfolder)
+
+            if os.path.isabs(cmdLineArgs.memLayout):
+                if os.path.isfile(cmdLineArgs.memLayout):
+                    mem_layout = cmdLineArgs.memLayout
+                else:
+                    # CMake FILEPATH cache values can become absolute paths in the build tree.
+                    # Recover by using the basename in the active target config folder.
+                    mem_layout = os.path.join(
+                        config_dir, os.path.basename(cmdLineArgs.memLayout)
```
davschneller: What do you mean by this case here? E.g. something like `/dense.xml`?
But what would happen if someone were to input `/somewhere-else/dense.xml`?

tests/Codegen/TestMemoryLayoutFallback.cmake
```diff
@@ -0,0 +1,72 @@
...
+
+if(NOT SEISSOL_SOURCE_DIR)
+  message(FATAL_ERROR "SEISSOL_SOURCE_DIR is required")
+endif()
+
+if(NOT SEISSOL_BINARY_DIR)
+  message(FATAL_ERROR "SEISSOL_BINARY_DIR is required")
+endif()
+
+set(CodegenSourceDir "${SEISSOL_SOURCE_DIR}/codegen")
```
davschneller: Not quite sure yet, but maybe it's worth moving this test into a `codegen/tests` folder.

codegen/generate.py
```diff
@@ -192,15 +192,38 @@ def generate_equation(subfolders, equation, order):
...
-            script_dir = os.path.dirname(os.path.abspath(__file__))
-            mem_layout = os.path.join(script_dir, "config", cmdLineArgs.memLayout)
         else:
-            print(f"Using the memory layout config file {cmdLineArgs.memLayout}")
-            mem_layout = cmdLineArgs.memLayout
+            script_dir = os.path.dirname(os.path.abspath(__file__))
+            subfolder = "gpu" if "gpu" in targets else "cpu"
+            config_dir = os.path.join(script_dir, "config", subfolder)
+
+            if os.path.isabs(cmdLineArgs.memLayout):
```
davschneller: consider putting the whole path resolution in a separate method if possible (or maybe even the `kernel` folder; maybe `memlayout.py`?)

tests/Codegen/TestMemoryLayoutFallback.cmake
```diff
@@ -0,0 +1,72 @@
...
+set(OutputDir "${SEISSOL_BINARY_DIR}/tests/GeneratedCodeMemoryLayoutFallback")
+set(NonExistingAbsDense "/tmp/seissol-test-not-existing/dense.xml")
+
+file(REMOVE_RECURSE "${OutputDir}")
+file(MAKE_DIRECTORY "${OutputDir}")
+
+execute_process(
```
davschneller: In all honesty—running the whole codegen for just testing if the file name resolving works seems like a little overkill.

* you test much more than just the memory layout selection (e.g. if something breaks further downsteam etc. this test will fail as well)
* it takes much longer (one more Yateto codegen—including the whole optimization procedure which takes its time)

Strongly consider factoring out the behavior into a method of its own and replace it by a pytest test which you can feed either only this unit test or many, many more path comparisons. As that's in essence what you check with the finding tests below.

docs/build-parameters.rst
```diff
@@ -77,7 +77,10 @@ CPU-specific parameters
 - ``HOST_ARCH``: the parameter to tune the architecture for. See build-archs for an overview. Applies both compiler options and GEMM generation. If you get an illegal instruction error, check this variable first. Since there are many options, we have an own page for them. In short, use ``hsw`` when using a Linux/Windows PC or an older Mac. Use ``apple-m1`` or ``apple-m2`` on the latest Mac computers.
 - ``MEMKIND``: enables HBM support.
 - ``NUMA_AWARE_PINNING``: pin the free CPUs (those used for the communication and IO threads) according to the given NUMA domains.
-- ``MEMORY_LAYOUT``: the sparsity patterns to apply. If not given, it will be chosen by the CPU architecture.
+- ``MEMORY_LAYOUT``: the sparsity patterns to apply. If set to ``auto``, it is chosen from the
+  target-specific config folder (``codegen/config/cpu`` or ``codegen/config/gpu``). If you pass
```
davschneller: `you` feels a bit off here; but I can't pin down why. Maybe "Passing" instead of "If you pass"?

docs/build-parameters.rst
```diff
@@ -77,7 +77,10 @@ CPU-specific parameters
 - ``HOST_ARCH``: the parameter to tune the architecture for. See build-archs for an overview. Applies both compiler options and GEMM generation. If you get an illegal instruction error, check this variable first. Since there are many options, we have an own page for them. In short, use ``hsw`` when using a Linux/Windows PC or an older Mac. Use ``apple-m1`` or ``apple-m2`` on the latest Mac computers.
 - ``MEMKIND``: enables HBM support.
 - ``NUMA_AWARE_PINNING``: pin the free CPUs (those used for the communication and IO threads) according to the given NUMA domains.
-- ``MEMORY_LAYOUT``: the sparsity patterns to apply. If not given, it will be chosen by the CPU architecture.
+- ``MEMORY_LAYOUT``: the sparsity patterns to apply. If set to ``auto``, it is chosen from the
```
davschneller: Maybe make active: "the code generation chooses from [...]"

davschneller (commented):
Generally a good idea to add it that way. But, w.r.t. the test structure: it might be good to add a pytest unit test instead though—going via CMake with an integration test covers too much and too little at the same time:

* too much because you test much more than just your changes (and it _does_ take its time—which is to be avoided if possible)
* too little because you test only one path configuration—while you'd actually need to test multiple in the most ideal case.

Might be good to wait for #1567 to land and extract the pathfinding into a method on its own and move it into the `kernels` directory maybe even. (i.e. to make the `generate.py` a bit slimmer) Then you can add direct unit tests for many different configurations.

codegen/generate.py
vikaskurapati: I intended for it to be just `dense.xml` and automatically configure depending on if it is GPU, or CPU build. But if someone were to input `/somewhere-else/dense.xml` like `cpu/dense.xml` or some absolute path, they should also work currently.

tests/Codegen/TestMemoryLayoutFallback.cmake
vikaskurapati: I just removed the tests currently. We can add a py-test based test once you finish #1567

davschneller (approved):

PR #1567, opened 2026-04-22
Add some Python unit tests

davschneller:
Add pytest-based Python tests for several components and add them to the CI. Besides, fix some smaller bugs in the Python scripts.

Adds the codegen pytest tests also to ctest.

Not yet 100%ly sure nor happy about how the folder structure in the `tests` folder evolves. Maybe it makes sense to move all C++ tests into a `tests/src` folder instead.
Also the runtime of the "golden file" tests is a bit concerning; but I'm not sure if there's a good way around it.

The bulk of the tests was (AI) generated using Opus 4.7.


.github/workflows/pytest-python.yml
```diff
@@ -0,0 +1,63 @@
...
+#
+# SPDX-License-Identifier: BSD-3-Clause
+# SPDX-LicenseComments: Full text under /LICENSE and /LICENSES/
+#
+# SPDX-FileContributor: Author lists in /AUTHORS and /CITATION.cff
+
+name: pytest-python
+
+on:
+  push:
```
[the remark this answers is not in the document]
vikaskurapati: Ehh... I don't think the comment makes much sense. If it is being tested on every push anyway, we don't need to add an additional pull_request trigger. What do you think?

.github/workflows/pytest-python.yml
```diff
@@ -0,0 +1,63 @@
...
+          python -m pip install --upgrade pip
+          # Runtime deps for the tested code
+          pip install numpy pandas scipy setuptools lxml seissolxdmf h5py
+          # Test framework
+          pip install pytest pytest-cov hypothesis
+
+          # (not needed and disabled; kept for completeness)
+          # Install yateto from the checked-out submodule so codegen
+          # integration tests can actually construct ADERDG instances.
+          # pip install -e ./submodules/yateto
```
vikaskurapati, line 45: I am confused. So, how does YATeTo get installed in the Docker container in this scenario? Or does it mean that YATeTo is no longer required as a Python package to run these tests?

cmake/process_users_input.cmake
```diff
@@ -31,12 +31,22 @@ option(LIKWID "Link with the likwid marker interface for proxy" OFF)
 option(INTEGRATE_QUANTITIES "Compute cell-averaged integrated velocity and stress components (currently breaks compilation)" OFF)
 option(ADDRESS_SANITIZER_DEBUG "Use address sanitzer in debug mode" OFF)
 
+# General switch for unit testing. By default enables the C++ unit tests.
 option(TESTING "Compile unit tests and collect them for execution" OFF)
```
vikaskurapati, line 35: The default is `OFF`, but the comment says it is enabled. Or, do you mean that the C++ unit tests are used when `OFF` is used and something when `ON`?

.github/workflows/pytest-python.yml
davschneller: It seems the symlink confuses both Copilot and Opus (I'll still need to answer to the other comments here—but several of them speak to this issue); they think they need to re-install it separately. Though they should never have to.

cmake/process_users_input.cmake
davschneller: The "By default" means (i.e. should mean): if you set `TESTING=ON`, then the C++ unit tests will be included.

davschneller: Ah, maybe it's "enables" vs. "enable"? I might've ellipsed an "it" here.

tests/python/codegen/test_aderdg_math.py
```diff
@@ -0,0 +1,415 @@
...
+    closed-form formulas. These are called everywhere the codegen allocates
+    tensors, so off-by-one here means wrong memory allocation everywhere.
+    """
+
+    @pytest.fixture
+    def elastic(self):
+        from kernels.equations.elastic import ElasticADERDG
+
+        # Orders in the actual CMake build matrix: 2..8
+        return lambda order: _mk(ElasticADERDG, order=order)
```
vikaskurapati, line 63: Could you replace 'Orders' with 'Convergence Orders'? I, for one, always got confused about where we use polynomial order and where convergence order (in different softwares). So mentioning that explicitly may be helpful.

tests/python/codegen/test_aderdg_math.py
```diff
@@ -0,0 +1,415 @@
...
+        ],
+    )
+    def test_numberOf3DBasisFunctions(self, elastic, order, expected):
+        adg = elastic(order)
+        assert adg.numberOf3DBasisFunctions() == expected
+
+    @pytest.mark.parametrize("order", [1, 2, 3, 4, 5, 6, 7, 8])
+    def test_numberOf3DQuadraturePoints_is_cubic(self, elastic, order):
+        adg = elastic(order)
+        assert adg.numberOf3DQuadraturePoints() == (order + 1) ** 3
```
vikaskurapati, line 102: Is the order here still convergence order? Why are we moving from 1...8 here? Afaik, we don't support convergence order 1, right?

davschneller: We don't—though I guess it wouldn't hurt to have it in that test.

And yes, it's still the convergence order (other occurrences in the code follow a similar pattern).

tests/python/codegen/test_aderdg_math.py
davschneller: Will do.

... Maybe that's mid-term something to adjust more generally (i.e. `order` -> `convergence_order`/`convergenceOrder` in the codegen); or at least have some page with the common conventions somewhere well-visibly.

tests/python/codegen/test_aderdg_math.py
```diff
@@ -0,0 +1,415 @@
...
+        """A DoF can't be both a velocity and a traction component."""
+        v_indices = np.nonzero(adg.extractVelocities().any(axis=0))[0]
+        t_indices = np.nonzero(adg.extractTractions().any(axis=0))[0]
+        assert not set(v_indices) & set(
+            t_indices
+        ), "Velocity and traction index sets must be disjoint"
+
+
+class TestAcousticDofIndices:
+    """AcousticADERDG: 4 quantities = 1 pressure + 3 velocities.
```
vikaskurapati: This is nit-picky, but can we write negative pressure here?

davschneller: Ok, can do.

tests/python/codegen/test_generate_goldenfile.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def test_acoustic_generates(self, tmp_path):
+        result = _invoke_generate(tmp_path, equation="acoustic", order=3)
+        assert result.returncode == 0, (
+            f"generate.py acoustic failed:\n"
+            f"stdout:\n{result.stdout[-1000:]}\n"
+            f"stderr:\n{result.stderr[-1000:]}"
+        )
+        assert (tmp_path / "equation-acoustic-3-double").is_dir()
+
```
vikaskurapati, line 229: Can we also add something with poroelastic here to check that quantities for poroelastic here?

tests/python/validation/test_meshcompare.py
```diff
@@ -0,0 +1,321 @@
...
+                "v1": np.array([1.0, 2.0]),
+                "clustering": np.array([7.0, 8.0]),
+            },
+            "int_fields": {"global-id": np.array([0, 1])},
+        }
+        meshcompare.compare("sim.xdmf", "ref.xdmf", epsilon=0.01)
+
+
+class TestMeshCompareDSBugWorkaround:
+    """meshcompare has an inline comment and workaround for the DS field:
```
vikaskurapati, line 276: What is DS here referring to?

.github/workflows/pytest-python.yml
vikaskurapati: So, basically it is already installed elsewhere, and you don't need to install it here again, or am I understanding something wrong?

cmake/process_users_input.cmake
vikaskurapati: The `default` is what confused me. You could say the default is `OFF`; `ON` enables C++ unit tests. And then the other unit tests are already documented accordingly. Or, if you want to make it more verbose, that is also OK.

tests/python/codegen/test_aderdg_math.py
vikaskurapati: Both are good ideas and I don't have any priority.

tests/python/codegen/test_aderdg_math.py
vikaskurapati: Sure, it does not hurt, but it is not necessary, imo. Unless, you plan to add the FV implementation in YATeTo sometime. I tried it in the beginning once, and ended up not putting much effort into it.

.github/workflows/pytest-python.yml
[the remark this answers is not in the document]
davschneller: Yes, that's also why I avoided the `pull_request` tags for now. IIRC it really made PR commits double trigger for us.

The `pull_request` tag is mostly relevant for PRs that come from outside our team (and outside the repository)—currently you have to trigger them manually. I.e. clone the branch to this repository (the SeisSol main one). I.e. maybe in a future PR, we can enable the `pull_request` trigger, if it doesn't stem from a branch which is hosted locally, i.e. here.

.github/workflows/pytest-python.yml
davschneller: It should work via the symlink. Yateto is normally not installed for SeisSol; but packaged with it as a submodule. We just use a symlink in the `codegen` directory to make it appear like a local Python module.

I think that might be a major point of confusion for the LLMs (also for Claude Opus 4.7 btw).

tests/python/codegen/test_generate_goldenfile.py
davschneller: Yeah, let me add it.

tests/python/validation/test_meshcompare.py
davschneller: This special case here: https://github.com/SeisSol/SeisSol/blob/52fdea131d058edd6a8ea9e8fab121db134f52c7/postprocessing/validation/meshcompare.py#L129-L134
(fault output quantity `DS`)

I'm not even sure if that bug is still there.

vikaskurapati (approved):
LGTM, thanks!

PR #1569, opened 2026-04-25
Move the struct-to-parameter maps to the material structs themselves

davschneller:
* Like the title says, move the struct-to-parameter maps from the `ParameterDB.cpp` file to the material structs
* Canonicalize the material type downcasting for the time being; i.e. if there's a constructor in a material for either elastic or acoustic, and we only give isotropic elastic/acoustic data, then use that. (not super happy with hard-coding elastic/acoustic here; but we don't have any other—otherwise, we'd need a type list per material—doable as well)

A future PR will move the material averagers into the material repositories.


src/Equations/anisotropic/Model/Datastructures.cpp
```diff
@@ -33,10 +34,13 @@ double AnisotropicMaterial::getMuBar() const { return (c44 + c55 + c66) / 3.0; }
...
-      c23(m.lambda), c33(m.lambda + 2 * m.mu), c44(m.mu), c55(m.mu), c66(m.mu) {
-  rho = m.rho;
-}
+    : Material(m.rho), c11(m.lambda + 2 * m.mu), c12(m.lambda), c13(m.lambda),
+      c22(m.lambda + 2 * m.mu), c23(m.lambda), c33(m.lambda + 2 * m.mu), c44(m.mu), c55(m.mu),
+      c66(m.mu) {}
+
+AnisotropicMaterial::AnisotropicMaterial(const AcousticMaterial& m)
+    : Material(m.rho), c11(m.lambda), c12(m.lambda), c13(m.lambda), c22(m.lambda), c23(m.lambda),
+      c33(m.lambda) {}
```
vikaskurapati: With anisotropic materials, it is okay to do the isotropic way to initiate the material parameters with lambda, mu, rho, and the numerical inversion would work correctly. But if we do it with acoustic materials, the anisotropic material's implementation does not have `mu=0` acoustic handling that isotropic elastic material has. And we have not yet solved and implemented the Riemann problem for it. The matrices become singular, and I suspect the solutions will not be correct. I believe that we should, for now at least, throw an error when something like this happens.

Another question I have is, why are we initializing anisotropic materials with acoustic materials? What is the use case here?

vikaskurapati (commented):
LGTM from the overview. I will need to test at least one acoustic, one elastic, and one poro-elastic simulation to check that the results did not break with these changes. I will post my results once I get the allocation for it.

P.S.: I did not check Copilot comments. Please consider them if they make sense.

vikaskurapati:
```
2026-05-06 03:32:39.212 info 0 : Initialize cell material parameters. 
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
[c642-071:1907474] *** Process received signal ***
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
[c642-071:1907474] *** Process received signal ***
[c642-071:1907474] Signal: Segmentation fault (11)
[c642-071:1907474] Signal code: Address not mapped (1)
[c642-071:1907474] Failing at address: 0x1d1b51
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
[c642-062:2211657] *** Process received signal ***
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
[c642-072:640548] *** Process received signal ***
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
[c642-072:640548] *** Process received signal ***
[c642-072:640548] Signal: Segmentation fault (11)
[c642-072:640548] Signal code: Address not mapped (1)
[c642-072:640548] Failing at address: 0x9c661
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
terminate called recursively
--------------------------------------------------------------------------
prterun noticed that process rank 2 with PID 1907474 on node c642-071 exited on
signal 6 (Aborted).
--------------------------------------------------------------------------
TACC:  MPI job exited with code: 134 
TACC:  Shutdown complete. Exiting.
```
I get this error on vista's gh nodes when I run Ridgecrest simulation with this branch.

davschneller:
Problem fixed (already a few commits back); it works now.

PR #1570, opened 2026-04-25
Speed up the volume energy computation

### LINKED ISSUE #1556: Inconsitency for the elastic energy computation in code vs. documentation
### Describe the bug

The formula for the elastic energy computation as written in the code seems to differ from the one shown in the documentation.
The difference exists since #527 .

### Expected behavior

The formulas (and results) match.

### Additional Comments

The formula om tje dpcs is given as 
```
1 / (2*mu) * integrate( sigma_ij * sigma_ij + (lambda / (3*lambda + 2*mu)) * sigma_kk^2 )
```
Even though I'd interpret the `sigma_kk` rather as `sigma_ij * delta_ij` (where `delta` denotes the Kronecker delta) and not as an extra loop over `k`, it still differs from the code:
Internally, we compute (using the same notation as above)
```
1 / (2*mu) * integrate( sigma_ij * sigma_ij - (lambda / (3*lambda + 2*mu)) * sigma_kk * sigma_ll)
```
This is because we have https://github.com/SeisSol/SeisSol/blob/74335ae64e2c813ad36e2d42768510280b07eca3/src/ResultWriter/EnergyOutput.cpp#L464-L469 ; and the `sumUnaxialStresses` basically adds a `sigma_ll` (and not `sigma_kk` or `sigma_ij * delta_ij`).

(also, side note; we could probably optimize the volume energy computation by a lot by using pre-computed mass matrices ... Since we don't have any mainlined high-order material or curvilinearity yet)

davschneller:
Compute the scalar products for the energy computation out of the (diagonal) mass matrix and the dofs directly; not via quadrature points.

Furthermore, optimize the energy computations in the fused regime; by switching the loop ordering to avoid additional matrix multiplications.

Should work for all current energies (and also those of e.g. poroelastic and anisotropic if the AI chatbots told me the correct formulas). Will still fail for e.g. #1297; maybe we'll need to re-introduce the "old" codepath for that one; also curvilinearity (if ever implemented) will require some additional quirks.

Also abstract the energy computation interface and the energies concerned—and move the computation code to the equation systems themselves.

Furthermore, begin adding anisotropic volume elastic and poroelastic volume energy support (possibly also better viscoelastic volume energy support). The docs were mostly constructed using AI; via Opus 4.6 and Opus 4.7.


davschneller:
(https://github.com/SeisSol/yateto/pull/105 will fix the remaining failing tests)

docs/energy-output.rst
```diff
@@ -9,67 +9,514 @@
...
-- Plastic moment
-    :math:`\int_{\Omega_e} \mu \eta  \,\mathbf{dx}`, with :math:`\mu` the second Lame coefficient and \eta a scalar quantity measuring the accumulated material damage.
-
-Currently, the output is only supported for the elastic wave equation.
+Energy in the water layer
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The water layer is modelled as an acoustic medium over the domain :math:`\Omega_a`. The constitutive behaviour is governed by the relation :math:`p = -K \nabla \cdot \mathbf{u}`, with bulk modulus :math:`K`, pressure :math:`p`, and displacement :math:`\mathbf{u}`.
+
+**Gravitational energy** is associated with the deformed sea surface :math:`\Gamma_\mathrm{free}` (the top boundary of :math:`\Omega_a`):
```
fabian-kutschera: In Lotto & Dunham (2015), this is called "gravitational potential energy"

docs/energy-output.rst
```diff
@@ -9,67 +9,514 @@
...
+
+with :math:`\rho` the density, :math:`g` the gravitational acceleration and :math:`\eta` the sea-surface elevation. This formulation follows from the linearised free-surface boundary condition with gravitational restoring force [LottoDunham2015]_.
+
+**Acoustic energy:**
+
+.. math::
+
+   W_\mathrm{ac} = \int_{\Omega_a} \frac{1}{2K}\, p^2 \,\mathrm{d}\mathbf{x}
+
+with :math:`p` the acoustic pressure and :math:`K` the bulk modulus (compressibility).
```
fabian-kutschera: K is the bulk modulus (not the compressibility, thus, remove it?)

docs/energy-output.rst
```diff
@@ -9,67 +9,514 @@
...
+
+**Gravitational energy** is associated with the deformed sea surface :math:`\Gamma_\mathrm{free}` (the top boundary of :math:`\Omega_a`):
+
+.. math::
+
+   W_\mathrm{grav} = \int_{\Gamma_\mathrm{free}} \frac{1}{2} \rho\, g\, \eta^2 \,\mathrm{d}S
+
+with :math:`\rho` the density, :math:`g` the gravitational acceleration and :math:`\eta` the sea-surface elevation. This formulation follows from the linearised free-surface boundary condition with gravitational restoring force [LottoDunham2015]_.
+
+**Acoustic energy:**
```
fabian-kutschera: From Eric: "acoustic energy" is typically used to refer to the sum of kinetic and potential energies associated with acoustic waves. here, you are referring to the "acoustic strain energy" contribution only

davschneller: [reporting back from discussing Claude] Isn't "potential" here more accurate here than "strain"? (as in: referring to the quantity from which we get this energy) For "elastic energy" -> "elastic strain energy" I'm aboard.

PR #1571, opened 2026-04-27
Add support for anisotropic ITM

vikaskurapati:
A research group in Helsinki is working with ITM for anisotropic materials. This feature is introduced for this purpose. I cleaned up the original ITM with some preliminary tests for this purpose. Alas, I still don't have a small test case for this to add to the pre-computed repository. 

The tests were generated with the help of AI. Model -- Codex 5.3 within the Copilot CLI.


src/Physics/InstantaneousTimeMirrorManager.cpp
```diff
@@ -94,73 +124,103 @@ void InstantaneousTimeMirrorManager::updateVelocities() {
...
+  }
 
-    for (auto& cluster : clusters_) {
-      cluster->setClusterTimes(cluster->getClusterTimes() * timeStepScalingFactor_);
-      auto* neighborClusters = cluster->getNeighborClusters();
-      for (auto& neighborCluster : *neighborClusters) {
-        neighborCluster.ct.setTimeStepSize(neighborCluster.ct.getTimeStepSize() *
-                                           timeStepScalingFactor_);
-      }
+  for (auto& cluster : clusters_) {
```
davschneller, line 195: Make an extra function that scales the cluster times; to avoid code duplication between the two paths.

src/Physics/InstantaneousTimeMirrorManager.cpp
```diff
@@ -94,73 +124,103 @@ void InstantaneousTimeMirrorManager::updateVelocities() {
...
+    auto* neighborClusters = cluster->getNeighborClusters();
+    for (auto& neighborCluster : *neighborClusters) {
+      neighborCluster.ct.setTimeStepSize(neighborCluster.ct.getTimeStepSize() /
+                                         velocityScalingFactor_);
     }
   }
 }
 
+void InstantaneousTimeMirrorManager::updateVelocities() {
+  updateVelocitiesForMaterialType<seissol::model::MaterialT>();
```
davschneller, line 205: If feasible, it might be best to move the template parameter a bit "down", i.e. inside the loop per cell.

So that the only real "different" part is the lambda as in the (isotropic) elastic loop.

src/Physics/InstantaneousTimeMirrorManager.cpp
```diff
@@ -75,7 +98,14 @@ void InstantaneousTimeMirrorManager::syncPoint(double currentTime) {
   isEnabled_ = false;
 }
 
-void InstantaneousTimeMirrorManager::updateVelocities() {
+template <typename MaterialType>
+void InstantaneousTimeMirrorManager::updateVelocitiesForMaterialType() {
+  logError() << "ITM material update is not implemented for this material type.";
```
davschneller: It might be best to show that error message a little earlier; at best during setup as well.

_If_ it's activated.

davschneller (commented):
Hi,

sorry for taking a bit; here's for a first review round with two stylistic comments to avoid code duplication.

In general, the ITM alas has a bit of a feel like the initial conditions even more like that—several different configuration options which only work for a subset of equation systems etc.

Meaning, since it's conceptually simple, we will want to move the different reflection methods into the scripting facility mid-term probably. I.e. that you can specify a "material transform" as follows (written in "pseudo easi"):

```
in: [rho, lambda, mu, scalingfactor]
out: [rho, lambda, mu]
function:
  rho: rho * scalingfactor
  lambda: lambda * lambda / mu * scalingfactor
  mu: 123
```

That way, we can encode more methods without needing to update or recompile SeisSol – that'll make it much more convenient for the users to play around with it—and also move the maintenance burden away from you.
And the material transforms should be fast enough when just executed once or twice a simulation.

The necessary infrastructure for the change should be paved via #1573 .


The existing options via the parameter file (like in this PR) will continue to function nonetheless even if we (you or me) introduce that—until we switch parameter formats with a migration script.

vikaskurapati:
> Hi,
> 
> sorry for taking a bit; here's for a first review round with two stylistic comments to avoid code duplication.
> 
> In general, the ITM alas has a bit of a feel like the initial conditions even more like that—several different configuration options which only work for a subset of equation systems etc.
> 

I don't know if I understand your comment correctly, but if you want to formulate an ITM like an initial condition, you will need to phrase it into three simulation phases, phase 1, WP with one particular material parameters, phase 2, scaled up, down etc., phase 3, with scaled back. So end of phase 1 will be IC for phase 2, yes, but if you want to use the easi initial conditions, you will need to do it in two separate simulations. I would avoid doing it two separate simulations as post-processing in one particular simulation is easier. 


> Meaning, since it's conceptually simple, we will want to move the different reflection methods into the scripting facility mid-term probably. I.e. that you can specify a "material transform" as follows (written in "pseudo easi"):
> 
> ```
> in: [rho, lambda, mu, scalingfactor]
> out: [rho, lambda, mu]
> function:
>   rho: rho * scalingfactor
>   lambda: lambda * lambda / mu * scalingfactor
>   mu: 123
> ```
> 

The material transform you mentioned, needs to be triggered in time. So, I am particularly keen on doing it that way. 

> That way, we can encode more methods without needing to update or recompile SeisSol – that'll make it much more convenient for the users to play around with it—and also move the maintenance burden away from you. And the material transforms should be fast enough when just executed once or twice a simulation.

SeisSol needs to be compiled only once, depending on which material you are using. Like how we are doing right now anyway. We don't need to recompile SeisSol unless, there are more modes of ITM being implemented. Currently, for elastic materials, the one which is already there works, and for anisotropy, this one should work. 

For playing with ITM parameters, you need to work with the parameters file anyway. So, that should not be a major problem. This, imo, is a simpler way of doing it.

> 
> The necessary infrastructure for the change should be paved via #1573 .
> 
> The existing options via the parameter file (like in this PR) will continue to function nonetheless even if we (you or me) introduce that—until we switch parameter formats with a migration script.

src/Physics/InstantaneousTimeMirrorManager.cpp
vikaskurapati: I am not sure if I understand what you mean. The only real "different" part is both lambda and mu in isotropic elastic loop. I did it this way to avoid the if branching for every cell.

vikaskurapati:
I am unsure if I understood the comments correctly, but I made the changes as per what I understood.

davschneller:
Ok, I'll take a look again.


W.r.t. the simulations—absolutely. It will in all cases stay a single simulation, _even_ with the scripting as I suggested. In fact, the scripting should concern the material transform.

What will happen is:

* you start the simulation as you'd normally do
* we hit the first inversion point: instead of C++ code, we trigger a script for the material adjustment (as you call `setDensity`/`setLameParameters` right now); then update the materials as normal. Really, the _only_ thing that changes is that we now call Lua/easi instead of C++. Afterwards, we update the materials as we're used to.
* the second point works the same way: trigger a script as before via Lua/easi, then update the matrices as it's done right now.

And instead of having x,y,z as easi inputs, you take the existing material parameters and adjust them (that's possible with easi 1.5 upwards). We call easi/Lua _at_ the sync points instead of before. I.e. full flexibility and a single simulation. And you could even run acoustic and/or poroelastic ITMs without adjusting SeisSol. And all of that with really simple code changes.

Ideally, at some point, we'd want SeisSol to just be available on a cluster and no more need to recompile (especially in conjunction with e.g. #1421 ). Even though it's right now the straight-forward way to just hard-code everything, eventually we wanna get rid of that. Too many options for everything (again, the same is true for the initial conditions—which you can already also input via easi now for most cases).


But that said—I'll approve this PR after another review round. What I just described above might still be a bit ahead of us.

src/Physics/InstantaneousTimeMirrorManager.cpp
davschneller: Yeah; all good. :) I'll take it like that.

davschneller (approved):
LGTM.

PR #1572, opened 2026-04-30
Fix ITM scheduling and output alignment

vikaskurapati:
As part of research with the Helsinki team, they encountered a bug when running ITM-acoustic simulations. The two bugs that this PR is supposed to fix are as follows:

1. Alignment issues with LIBXSMM when using higher-order output.
2. ITM startup synchronization bug (0s -> 0s). ITM modules registered on `SynchronizationPoint` but were not getting their initial sync schedule at simulation start. So, the next sync time stayed at 0, and the simulator aborted with "did not advance in time". This fix also initializes the simulation start time for synchronization-point modules, ensuring that ITM's first trigger is scheduled correctly.

### DIFF
--- a/src/Initializer/InitProcedure/InitIO.cpp
+++ b/src/Initializer/InitProcedure/InitIO.cpp
@@ -7,6 +7,7 @@
 
 #include "InitIO.h"
 
+#include "Alignment.h"
 #include "Common/Constants.h"
 #include "Common/Filesystem.h"
 #include "Equations/Datastructures.h"
@@ -282,6 +283,8 @@ void setupOutput(seissol::SeisSol& seissolInstance) {
       for (std::size_t quantity = 0; quantity < seissol::model::MaterialT::Quantities.size();
            ++quantity) {
         if (seissolParams.output.waveFieldParameters.outputMask[quantity]) {
+          constexpr std::size_t MaxVtk3dPoints = tensor::vtk3d::Shape
+              [(sizeof(tensor::vtk3d::Shape) / sizeof(tensor::vtk3d::Shape[0])) - 1][1];
           writer.addPointData<real>(
               namewrap(seissol::model::MaterialT::Quantities[quantity], sim),
               {},
@@ -291,13 +294,15 @@ void setupOutput(seissol::SeisSol& seissolInstance) {
                 const auto* dofsSingleQuantity = dofsAllQuantities + QDofSizePadded * quantity;
                 kernel::projectBasisToVtkVolume vtkproj{};
                 memory::AlignedArray<real, multisim::NumSimulations> simselect{};
+                alignas(Alignment) std::array<real, MaxVtk3dPoints> alignedTarget{};
                 simselect[sim] = 1;
                 vtkproj.simselect = simselect.data();
                 vtkproj.qb = dofsSingleQuantity;
-                vtkproj.xv(order) = target;
+                vtkproj.xv(order) = alignedTarget.data();
                 vtkproj.collvv(ConvergenceOrder, order) =
                     init::collvv::Values[ConvergenceOrder + (ConvergenceOrder + 1) * order];
                 vtkproj.execute(order);
+                std::copy_n(alignedTarget.data(), tensor::vtk3d::Shape[order][1], target);
               });
         }
       }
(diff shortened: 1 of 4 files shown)


vikaskurapati:
I don't know why, but when I request a review from both Copilot and @davschneller, the request is automatically removed from @davschneller. I am tagging you again to review it. It was not intentionally removed.

src/Modules/Modules.cpp
```diff
@@ -82,9 +82,17 @@ double Modules::_callSyncHook(double currentTime, double timeTolerance, bool for
 void Modules::_callSimulationStartHook(std::optional<double> checkpointTime) {
   assert(static_cast<int>(nextHook_) <= static_cast<int>(ModuleHook::SynchronizationPoint));
 
+  const auto startTime = checkpointTime.value_or(0);
+
   for (auto& [_, module] : hooks_[static_cast<size_t>(ModuleHook::SimulationStart)]) {
     module->simulationStart(checkpointTime);
-    module->setSimulationStartTime(checkpointTime.value_or(0));
+    module->setSimulationStartTime(startTime);
```
davschneller: Remove this line (it's superfluous now), and adjust the comment below. The actual bug was that the code assumed that all modules with `SynchronizationPoint` also implement `SimulationStart`.

The `setSimulationStartTime` is only needed for those who call the sync point hook.

davschneller (commented):
One short comment; otherwise we're good.

(and yes; I'm aware that the high-order output, while nice and short to adjust compared to the old one, is still a bit messy to handle at the moment—working on it. Cf. also #1180 )

src/Initializer/InitProcedure/InitIO.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include "InitIO.h"
 
+#include "Alignment.h"
```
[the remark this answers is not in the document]
davschneller: The include is required by clang-tidy (and IWYU).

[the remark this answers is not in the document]
vikaskurapati: Yes, I think Copilot missed the Clang-Tidy requirements. Problems of still having the free-tier account, I guess.

vikaskurapati:
We also added GPU synchronization (thanks @davschneller), and I fixed the S-wave reflection parameters (thanks to student work), which were in #1571 but moved here too. If you have more ideas for unit tests, please let me know.

src/Physics/InstantaneousTimeMirrorManager.cpp
```diff
@@ -89,13 +96,15 @@ void InstantaneousTimeMirrorManager::updateVelocities() {
...
       // Refocusing only S-waves
 
-      // (preserved comment from before the refactor)
-      // lambda = -2.0 * velocityScalingFactor * mu + (lambda + 2.0 * mu) / velocityScalingFactor;
-      // mu *= velocityScalingFactor;
-      // rho *= velocityScalingFactor;
+      const auto newLambda =
+          -2.0 * velocityScalingFactor_ * mu + (lambda + 2.0 * mu) / velocityScalingFactor_;
+
+      if (newLambda < 0.0) {
```
davschneller, line 102: ((un)intended) slight overlap with #1571 ?

tests/Modules/Module.t.h
```diff
@@ -72,9 +72,24 @@ class TestModule : public seissol::Module {
...
+  void syncPoint(double time) override {
+    state = ModuleHook::SynchronizationPoint;
+    this->time = time;
+  }
+};
+
 TEST_CASE("Module runthrough") {
   TestModule mod1(true, 2);
   TestModule mod2(false, 3);
+  SyncOnlyModule syncOnly(1000);
```
davschneller, line 92: Might be good to test two of this kind. One with `1000`, and the other one with e.g. `2`. Just to see that syncpoint-only modules work and actually jump into the sync point state.

davschneller (approved):
Two more comments—otherwise LGTM.

src/Physics/InstantaneousTimeMirrorManager.cpp
vikaskurapati: Intended overlap, actually. I wanted this fix done so that local testing for all reflections works on this branch. If it is a problem, I can remove it from here or from #1571.

PR #1577, opened 2026-05-20
Fix rconv compilation

vikaskurapati:
Temporary fix for training. If it is removed later, we can move it elsewhere.


davschneller (approved):
LGTM.

davschneller:
To be honest, it might be worth rewriting rconv in Python mid-term (sounds like a job for an LLM tbh). Or just introduce a new format altogether to unite 42 and 50 (NRF/SRF and FSRM).

PR #1578, opened 2026-05-20
Update runners to Ubuntu 26.04

davschneller:
Switch all CI jobs to use Ubuntu 26.04 (as it's available now).

The Docker images in the `.ci` folder will be updated in a separate PR (there are already more modern versions of them on some branch).

Requires switching from ParMETIS to SCOTCH as the former isn't shipped with Ubuntu 26.04 anymore (luckily SeisSol supports both).


Thomas-Ulrich (approved):
LGTM

PR #1580, opened 2026-06-01
Fix Poro DR (after some recent changes)

davschneller:
Fix the poro DR from diverging. It broke, since the initialization of the imposed states to zero got moved.


Thomas-Ulrich (approved):
LGTM

PR #1583, opened 2026-06-11
More DR single precision synchronization

davschneller:
Fix in essence the last differences between the CPU and GPU DR implementation, and make the single precision usage even more consequential.

### DIFF
--- a/src/DynamicRupture/FrictionLaws/GpuImpl/RateAndState.h
+++ b/src/DynamicRupture/FrictionLaws/GpuImpl/RateAndState.h
@@ -74,45 +74,27 @@ class RateAndStateBase : public BaseFrictionSolver<RateAndStateBase<Derived, TPM
    */
   SEISSOL_DEVICE static void calcInitialVariables(FrictionLawContext& __restrict ctx,
                                                   uint32_t timeIndex) {
-    auto& devStateVariableBuffer{ctx.stateVariableBuffer};
-    auto& devFaultStresses{ctx.faultStresses};
-    auto* devSlipRateMagnitude{ctx.data->slipRateMagnitude};
-    auto* devSlipRate1{ctx.data->slipRate1};
-    auto* devSlipRate2{ctx.data->slipRate2};
-    auto* devInitialStressInFaultCS{ctx.data->initialStressInFaultCS};
-
-    auto& devAbsoluteShearTraction{ctx.initialVariables.absoluteShearTraction};
-    auto& devLocalSlipRate{ctx.initialVariables.localSlipRate};
-    auto& devStateVarReference{ctx.initialVariables.stateVarReference};
-
     updateNormalStress(ctx, timeIndex);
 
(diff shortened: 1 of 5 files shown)


src/DynamicRupture/FrictionLaws/GpuImpl/RateAndState.h
```diff
@@ -164,27 +146,24 @@ class RateAndStateBase : public BaseFrictionSolver<RateAndStateBase<Derived, TPM
...
     // the only mu calculation left, outside of the fixed-point loop
-    auto details = Derived::getMuDetails(ctx, localStateVariable);
-    ctx.data->mu[ctx.ltsFace][ctx.pointIndex] = Derived::updateMu(ctx, slipRateMagnitude, details);
+    const auto details = Derived::getMuDetails(ctx, localStateVariable);
+    const auto mu = Derived::updateMu(ctx, slipRateMagnitude, details);
+
+    ctx.data->mu[ctx.ltsFace][ctx.pointIndex] = mu;
 
-    const real strength = -ctx.data->mu[ctx.ltsFace][ctx.pointIndex] * devNormalStress;
+    const real strength = -mu * ctx.initialVariables.normalStress;
```
Thomas-Ulrich, line 163: It is quite strange that it is called initialVariables, since it should also include the transient contribution to the normal stress.

src/DynamicRupture/FrictionLaws/GpuImpl/RateAndState.h
```diff
@@ -144,9 +127,8 @@ class RateAndStateBase : public BaseFrictionSolver<RateAndStateBase<Derived, TPM
 
       hasConvergedInner &= hasConvergedLocal;
 
-      devLocalSlipRate = 0.5 * (localSlipRateMagnitude + slipRateTest);
```
Thomas-Ulrich, line 147: isn't the multiplication faster?

Thomas-Ulrich (approved):

src/DynamicRupture/FrictionLaws/GpuImpl/RateAndState.h
davschneller: Hmm; the `initialVariables` feels like "initially computed before the RSF iteration".

Thomas-Ulrich: ok now I understand. then can be left as it is.

src/DynamicRupture/FrictionLaws/GpuImpl/RateAndState.h
davschneller: In this case, the compiler (once optimizing, i.e. no pure Debug builds) will transform the `/ 2` to a `* 0.5[f]` . The `/ 2` is IMO a bit easier to read than a `static_cast<real>(0.5) * ` in that case; that's why I adjusted it.

PR #1585, opened 2026-06-18
Make the new binary naming scheme the default

davschneller:
Approaching a new release, let's switch the naming scheme which already exists for quite some time and is easier to type. It'll go from 

`SeisSol_Release_dskx_6_elastic`

to

`seissol-cpu-elastic-o6-f64`

.

In particular, we'll remove the Release/Debug distinction (especially helpful when debugging something—also you can use the Ninja multibuilds if you really want to have separate release/debug builds) and the architecture.

Each new number entry (like e.g. `o6`) should have one letter and the number itself; thus the order of the elements should in principle be interchangeable.

Moreover, `SeisSol_proxy` becomes `proxyseissol`; so that you can easier auto-complete SeisSol vs SeisSol proxy.

~Open questions are:~
* ~keep `cpu` / `gpu`?~
* ~any info missing in the file name?~
* ~once #1421 lands, the scheme will simplify to a simple `seissol-cpu` / `seissol-gpu` executable. (or just `seissol`) I.e. should we wait for the latter to land? If we decide so, I'll close this PR again.~ not yet considered; we switch here.

If you want to keep using the old option, just set `NEW_BINARY_NAMING=OFF` in CMake (e.g. `cmake .. -DNEW_BINARY_NAMING=OFF` ).


Thomas-Ulrich:
keep cpu/gpu? Maybe not, as this information should be found in the log, and if you are on a gpu machine you will not compile the cpu version.
Actually, it would be nice if this kind of info can be printed with seissol -h for example (Xinju told me -h does not work any more).
Even f64 could be removed then.
Any info  missing? I don't think so.
Should we wait for the latter to land? you choose.

davschneller:
Ok, I can add these infos to the `-h` option or similar again (`-h` still works; but doesn't show any info). EDIT: see #1586 .

...

or maybe make it a different option / script?

The build infos (precision, equation system etc.) should also be printed when running SeisSol, i.e. in the log.


Against removing `f64` would be, alas, if you'd install a single-precision and a double-precision version in the same folders. That's why we still have these args in the first place right now—or at least that's why I'd keep them still.

davschneller:
W.r.t. cpu vs gpu: ok, I'll remove it.
(probably too much "developer thinking" from my side; that you'd want on a home computer both versions)

Or, one more idea: suffixes via another CMake option. In case we forgot something.

Some last points w.r.t. the naming:

* should the plasticity method be in the file name
* same for the DR quadrature rule
* ...or maybe make both always available in runtime for all variants?

davschneller:
One more comment (thanks @fabian-kutschera for that one, and also the solution suggestions):

`p6` might be a little confusing, as we don't mean "polynomial degree" 6, but 5 here (but convergence order 6). I.e. (at least) three options:

* keep it `p6`
* switch to the degree: `p6` -> `p5`
* switch the letter: e.g. `p6` -> `o6` (<-- that's the currently adopted one)

(meanwhile, the `cpu`/`gpu` label is gone; and there's now an extra custom slot)

NicoSchlw:
I think o6 is a good choice

Thomas-Ulrich (approved):
LGTM!
For GPU/CPU.
Maybe worth (in the future, not necessarily with this PR), to throw a warning if there are GPUs on the nodes and we using a CPU build.
(and to throw an error if we try to run a GPU compiled seissol on a node without GPU, or incompatible GPU. but maybe there is already that?).

davschneller:
> LGTM! For GPU/CPU. Maybe worth (in the future, not necessarily with this PR), to throw a warning if there are GPUs on the nodes and we using a CPU build. (and to throw an error if we try to run a GPU compiled seissol on a node without GPU, or incompatible GPU. but maybe there is already that?).

Hmm; that might be difficult to detect I fear. ... I can try.
Though maybe that's something more for a wrapper script of some sort. You know, some script that also automatically handles any pinning etc. — would something like that be useful?
(I've already let Claude create such one once some time ago, but never tried it out ^^)

PR #1587, opened 2026-06-26
Improve the checkpointing docs

davschneller:
Describe the current format (again, some changes might still come to it—hence it's all written with some care).

Still TODO is to add maybe an example output.

Also, we now allow "soft failures" for checkpoints; i.e. if variables or storage structures aren't present, we don't fail but just use the default values. ... Might be replaced by a willing opt out instead.


docs/checkpointing.rst
```diff
@@ -18,6 +18,8 @@ The checkpointing mechanism enables the storage and restart of a simulation to a
 
 The checkpointing mechanism will automatically adapt to any given element partition. In particular, it is independent of the number of nodes.
 
+Note that, currently, we cannot give any guarantee that the checkpoints are portable between SeisSol versions; use at your own descretion.
```
Thomas-Ulrich, line 21: typo: descretion

docs/checkpointing.rst
```diff
@@ -64,3 +66,35 @@ Note however, that the following restrictions apply.
...
+Last date of update: 2026-06-26.
+
+Note: we do not recommend this procedure, as the format is still subject to change and might be incompatible with future versions of SeisSol. Use with care.
+
+General format outline:
+
+* the checkpoint is a Hdf5 file, stored in the ``checkpoint`` group.
+* the group contains two attributes:
+
+  * ``__order``, which is the order that checkpoint was recorded in
```
Thomas-Ulrich, line 83: precise: accuracy order (= polynomial degree+1)?

docs/checkpointing.rst
```diff
@@ -64,3 +66,35 @@ Note however, that the following restrictions apply.
...
+  * ``__time``, denoting the simulation time that the checkpoint was recorded at
```
Thomas-Ulrich, line 84: could also include __padding, __friction_law, __quadrature_rule?

davschneller: That might be a good idea; I'll add them.

PR #1594, opened 2026-07-01
Correctly name the parameters for strain rate output

davschneller:
Previously called just "strain" output (but it was always strain rate—since we didn't integrate over the whole displacement. Cf. also how the energies are computed), for the off-fault receivers.


NicoSchlw (approved):
Thanks, LGTM

PR #1595, opened 2026-07-03
Rename `-p` for acoustic

davschneller:
Currently, the negative pressure output is denoted by `-p` for acoustic (i.e. true acoustic; not the elastic-emulated one). If used in some software, that might cause issues with the naming or at least require additional quotation marks (due to the `-` in the name).

Hence, we replace it here by `pprime`.


fabian-kutschera:
If I'm not mistaken, SeisSol outputs the Eulerian pressure perturbation (Pa) in the acoustic medium. This is often referred to as p′ in the literature. Thus, my suggestions would be to call it `pprime`, which is also the way how we refer to it in the TTPV benchmarks.

davschneller:
Ok; changed it.

fabian-kutschera (approved):
LGTM.

PR #1597, opened 2026-07-07
Split Yateto kernel builds per Yateto call

davschneller:
Split up the build of Yateto kernels per Yateto call.
(mostly relevant for #1421 ; but also to a little extent now already)

Each Yateto call generates its own set of files—now we forward it to the build system; and we do not require any more "aggregator files" for the CPP ones; only the header files.

The code generation still happens in a single sweep (due to the subroutines; we might need to pickle them to get around that).

Small tidbit of the current setup: we will now need to run Python during configure, as we run the generation script now twice (once during configure to obtain all files to compile, and once during build to actually generate the files).

### DIFF
--- a/codegen/CMakeLists.txt
+++ b/codegen/CMakeLists.txt
@@ -10,15 +10,11 @@ add_custom_target(build-time-make-directory ALL
     COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_CURRENT_BINARY_DIR}/GeneratedCode")
 
 set(GENERATED_FILES_FOR_SEISSOL "${CMAKE_CURRENT_BINARY_DIR}/GeneratedCode/subroutine.h"
-                                "${CMAKE_CURRENT_BINARY_DIR}/GeneratedCode/tensor.cpp"
                                 "${CMAKE_CURRENT_BINARY_DIR}/GeneratedCode/subroutine.cpp"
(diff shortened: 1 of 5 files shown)


codegen/CMakeLists.txt
```diff
@@ -56,7 +51,59 @@ add_custom_command(
...
+math(EXPR TARGETCOUNT "${TARGETCOUNT} - 1")
+
+set(SEISSOL_CODEGEN_KERNELS)
+set(SEISSOL_CODEGEN_TESTS)
+set(SEISSOL_CODEGEN_HEADERS)
+
+foreach(IDX RANGE ${TARGETCOUNT})
+  string(JSON TARGETNAME MEMBER ${SEISSOL_CODEGEN_TARGETS} ${IDX})
+
+  string(JSON TARGET_COUNT LENGTH ${SEISSOL_CODEGEN_TARGETS} "${TARGETNAME}" "kernels")
```
Thomas-Ulrich, line 72: Does it works if kernels is empty?
TARGET_COUNT will be -1.

Thomas-Ulrich (approved):

codegen/CMakeLists.txt
davschneller: That's intended here—since the runs for `0 <= IDX <= TARGETCOUNT`. So the -1 is needed (not 0) to explicitly cancel the loop.

Issue #1601, opened 2026-07-08
RS non-convergence of TPV105 (only near t=0s)


Thomas-Ulrich:
### Describe the bug

I'm running tpv105 in higher resolution (o5), with an improved mesh and with the latest seissol.
I'm using this branch for the new mesh (but anyway the mesh is not the problem).
https://github.com/SeisSol/Examples/tree/thomas/tpv105_improve/tpv105
The code runs fine (large throughput ~2TFLOP/s on 16 nodes on NG), but there are convergence problems at t=0s. After t=0s, the results look correct (although I would need to double check comparing on the SCEC website).

The log shows many non-convergence near t= dt.
```
2026-07-08 14:00:31.907 warn 6 : A rate and state cell failed to converge at the given settings; at the cell around [0, 0, -3000] at simulation time 0.000273388 s. PointIDs of failure (inner, outer loop failures): [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24] []
```
also, latter on, it also show non-convergence near t=0 (which is strange, but maybe the fact that t=0 appears after t=dt is only a buffered output issue).

```
2026-07-08 14:02:35.285 warn 14 : A rate and state cell failed to converge at the given settings; at the cell around [20016.6, 0, -17618.7] at simulation time 0 s. PointIDs of failure (inner, outer loop failures): [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]
```

[5712954.tpv105.out.txt](https://github.com/user-attachments/files/29800446/5712954.tpv105.out.txt)

At t=0s, the slip rate is large in the top 3km (velocity strengthing region)
<img width="1475" height="695" alt="Image" src="https://github.com/user-attachments/assets/41b573da-f32c-4448-b4ea-ce5a44305503" />

at t=0.5s, the slip rate look as expected

<img width="1463" height="658" alt="Image" src="https://github.com/user-attachments/assets/62d61d36-5285-4c96-9a7f-870d37eae92e" />

note also the large SRs on the first fault receiver at t=0s
````
(venv-stack24.5-3.10) di73yeq4@login03:/hppfs/scratch/0A/di73yeq4/Examples/tpv105> head output/tpv105_3d-faultreceiver-00001.dat -n 15
[...]

PR #1604, opened 2026-07-09
fix arsinhexp argument order (RS initializer)

Thomas-Ulrich:
Fixes the non-zero SR issue identified in #1601.


davschneller (approved):

davschneller:
(for the record; that issue was introduced in #1531 —and now fixed here)

PR #1608, opened 2026-07-21
Improve Newton-Raphson Loop

davschneller:
To address #1601 (i.e. convergence floors don't really get found); switch to rtsafe.

Meaning: we use Newton-Raphson as before, but also bisection (b/c our functions in use are monotonous), and choose the faster of the two in the end.
Furthermore, we use the geometric variant for bisection and run the Newton-Raphson step in the logarithmic regime. (for faster convergence in both cases)

(there may be better versions by now; but these here work well enough it seems)

Code generated with and inspired by Opus 4.8 and Opus 5.


davschneller:
Should address #1601 ; not quite there yet though. Or potentially, the tolerance is now interpreted even stricter with this branch.

Generally, the solutions (CI) don't differ much in F64. Comparing the F32 solutions to master F64, this one is a little closer to it than master F32. (side note; the tpv101-slip and tpv105 currently seem to be the most unstable CI test cases in these regards)
Errors for F32 vs. a master of when the comments were written lie at around e-3 for the receivers, and below e-4 (or even e-8) for the fault output.

PR #1611, opened 2026-07-24
Add global cell/face ID to the receiver header

davschneller:
Add a reference to the cell or face on which we've found the receiver; mainly for debugging purposes—but mainly it'll also have other uses.

Happy to discuss the naming (`face/cell-global-id`), and/or the layout.

### DIFF
--- a/src/DynamicRupture/Output/OutputManager.cpp
+++ b/src/DynamicRupture/Output/OutputManager.cpp
@@ -418,6 +418,8 @@ void OutputManager::initPickpointOutput() {
               file << "# x1\t" << makeFormatted(point[0]) << '\n';
               file << "# x2\t" << makeFormatted(point[1]) << '\n';
               file << "# x3\t" << makeFormatted(point[2]) << '\n';
+              file << "# face-global-id\t"
+                   << receiver.elementGlobalIndex * 4 + receiver.localFaceSideId << '\n';
             }
 
             // stress info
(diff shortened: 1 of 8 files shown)


src/DynamicRupture/Output/OutputManager.cpp
```diff
@@ -418,6 +418,8 @@ void OutputManager::initPickpointOutput() {
               file << "# x1\t" << makeFormatted(point[0]) << '\n';
               file << "# x2\t" << makeFormatted(point[1]) << '\n';
               file << "# x3\t" << makeFormatted(point[2]) << '\n';
+              file << "# face-global-id\t"
+                   << receiver.elementGlobalIndex * 4 + receiver.localFaceSideId << '\n';
```
Thomas-Ulrich: I would suggest writing 2 lines with receiver.elementGlobalIndex and receiver.localFaceSideId.

davschneller: Maybe it's worth outputting all three values — the global ID is what we currently also output as unique identifiers for the fault. But it might not be the case that we're keeping them this way — e.g. maybe those could be compressed in the future (to just enumerate the fault, and nothing else); then they won't match `globalCellId * 4 + faceSide` anymore. And just in general to have a match between fault elementwise and fault pickpoint.

Also, then just the plus global index or also minus? (both should be doable I think) Same for the side.

Thomas-Ulrich: up to you. Add it if you think it can help for debugging purposes.

davschneller: Ok, added it.

PR #1612, opened 2026-07-24
Output Kernel Memory Throughput and Refactor Flop Counter

davschneller:
Refactor the flop counter, to make it much more generic. I.e. no more extra functions for incrementing this or that—and we also use the `PerformanceEstimate` struct from the proxy everywhere now.

Also, we propagate the kernel memory throughput metric — which outputs all read and written non-constant data combined over all kernels and elements. Thus providing an over-estimation of the underlying achieved memory throughput/bandwidth; though on the GPU (due to kernel == element loop) we should roughly get something realistic between L2 and main memory behavior of the kernel.


davschneller:
~Needs https://github.com/SeisSol/yateto/pull/114 for a better kernel bandwidth estimate.~ done

PR #1624, opened 2026-08-19
Fix some issues with ICC

davschneller:
(ICC has sometimes problems with `if constexpr`, so we make everything where not necessary dynamic — won't hurt the performance there)


fabian-kutschera (approved):
LGTM (even though outside my expertise; but pm with David)

PR #1626, opened 2026-08-24
Make ICC compile the size functions

davschneller:
Due to seemingly a bug with SFINAE and `using` in templates.


fabian-kutschera (approved):
Was able to compile successfully.

PR #1627, opened 2026-08-26
Fix receiver strain rate parameter name

davschneller:
(always needs to be all lower case)
(... maybe it's good to catch that in software at some point)


fabian-kutschera (approved):
LGTM. 
Should fix:
"receivercomputestrainrate" in "output" was given in the parameter file, but is unknown to SeisSol.

davschneller:
(the current GPU CI errors there can be ignored)

PR #1630, opened 2026-08-30
From 48 to 16 (again); but this time also from 11 to 8

davschneller:
Reduce the amount of neighboring flux kernels by relabeling vertices, as it once already was considered (thanks Claude for bringing up this idea — again after about 10 years ago it seems). And combine with the flux matrix splitting.

Reasons being:

* the GPU implementation currently uses the non-split flux matrices anyways. And there, having at most 16 instead of 48 kernel launches pays off. Same for the DR going from 16 matrices down to 8. (which can be, except for Stroud, be probably further reduced to 2-4 permutation plus 4 collocation matrices)
* combine the move from 48 to 16 matrices with the flux matrix splitting into the 11 matrices (`rDivMT(0)-(3)`, `fP(0)-(2)`, `rT(0)-(3)`) on top, bringing them down to 8 matrices there: we are able to combine `fP` and `rT` , in particular we always use `fP(0)` now thanks to the relabel. The nonzero flop slightly _increase_ though, as we cannot use `fP(1)` anymore. (while, for the proxy o6, the hardware flops go down to 75-80% of before) Thus: instead of having 3 `fP` and 4 `rT` matrices, we now have 4 `fPrT` matrices. (still, the CPU speedup is only marginal)

~I'm still a bit sceptical with the current implementation because the energies are off a bit too much for my likening in the CI cases (it's just 4-6th digit; but it's visible), but maybe it's just a "fluke".~ Tested again; it's inside the norm. In fact, if you just re-arrange vertices, you can get similar differences in the energy value.

Note that the positive-orientation check for the tetrahedra becomes a mere internal confirmation check by this PR. Reason being, we now already ensure that all tetrahedra are reordered properly.

### DIFF
--- a/codegen/config/cpu/pspamm.xml
+++ b/codegen/config/cpu/pspamm.xml
@@ -13,6 +13,11 @@
 <matrix name="fMrT(2)" sparse="aligned"/>
 <matrix name="fMrT(3)" sparse="aligned"/>
 
+<matrix name="fPrT(0)" sparse="aligned"/>
+<matrix name="fPrT(1)" sparse="aligned"/>
+<matrix name="fPrT(2)" sparse="aligned"/>
+<matrix name="fPrT(3)" sparse="aligned"/>
+
 <matrix name="rDivM(0)" sparse="aligned"/>
 <matrix name="rDivM(1)" sparse="aligned"/>
 <matrix name="rDivM(2)" sparse="aligned"/>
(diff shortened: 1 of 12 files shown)


davschneller:
Tested again; it's inside the norm. In fact, if you just re-arrange vertices (e.g. run the permutation applied here already externally; or permute them randomly), you can get similar differences in the energy value.

PR #1632, opened 2026-08-31
Use quadweights for the potency computation

davschneller:
Use the given quadrature rule for the potency (and seismic moment) computation in the energy; not the average.

### DIFF
--- a/docs/breaking-changes-backward-compatibility.rst
+++ b/docs/breaking-changes-backward-compatibility.rst
@@ -118,3 +118,13 @@ Name of the Strain Rate Output for Off-Fault Receivers
 The strain rate output was named just "strain" output for the off-fault receivers.
 The corresponding option was likewise called :code:`ReceiverComputeStrain`,
 not :code:`ReceiverComputeStrainRate`.
+
+Potency and Seismic Moment Quadrature
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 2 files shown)


NicoSchlw (changes requested):
Hi David, I think this PR makes sense. But it should be quadweights instead of quadpoints, right?

Also, this is a breaking change, as an old setup will not yield the same seismic moment anymore. This should be documented.

davschneller:
Thanks — that's a big miss on my side. ... In any way; it's changed and now also documented.

For reference, here's the effect on the 20k cells TPV5 mesh after 1s (order 6; and we're underresolved anyways — so the effect here will be bigger than for a "production mesh" in all likelihood):

* without this change: `Seismic moment (without plasticity): 5.189359e+17 Nm , Mw: 5.74008`
* with it: `Seismic moment (without plasticity): 5.172700e+17 Nm , Mw: 5.73914`

NicoSchlw (approved):
Thanks for the updates. It looks good to me now. As a clarification for me, you removed the factor of 0.5 because the reference element has a surface area of 0.5? Is this worth a comment in the code?

davschneller:
It's actually something that's done like that in the friction energy computation as well — there, we also don't have the `0.5`; it's been asked before for that one: #849 and https://github.com/SeisSol/SeisSol/commit/1e1351c3b60c0b68ce88aa31270259adb9eace4f .

The TL;DR being, the quadrature points work on the reference triangle (0,0), (1,0), (0,1) which has area 1/2. Thus, to get to the area A of our triangle, we need to use the quadrature result, and multiply it by 2A.

PR #1635, opened 2026-09-01
add #include <cstdint>

NicoSchlw:
Add #include <cstdint> for compilation with GNU 15.2


davschneller (approved):

davschneller:
One comment: that `cstdint` needs to be placed a few lines above (so that they're in alphabetic order); the clang-format checks will fail right now

NicoSchlw:
Ok thanks, will do so

PR #1637, opened 2026-09-07
Make ruptureFrontThreshold a parameter file variable

NicoSchlw:
`ruptureFrontThreshold` dictates when the rupture time is picked, from which the rupture speed is calculated. Increasing this threshold can help to suppress negligible slip fronts and to pick the main rupture front instead. The default value is set to the original hardcoded value (0.001) to ensure backward compatibility.


davschneller (approved):
LGTM.

Would be good to also add it to `docs/parameters.par` , so that others can see it.

